PR #47, opened 2016-08-08
Enable fill coloring for choropleths

mcnuttandrew:
Responds to https://github.com/uber/deck.gl/pull/46, allows users to specify a rgb vector a property of each geojson layer being passed in. Example

screen shot 2016-08-08 at 11 56 13 am


src/layers/choropleth-layer/choropleth-layer.js
```diff
@@ -137,9 +137,9 @@ export default class ChoroplethLayer extends Layer {
   }
 
   calculateColors(attribute) {
-    const colors = this.state.groupedVertices.map(
+    const colors = this.state.groupedVerticesColors.map(
       vertices => vertices.map(
```
gnavvy: shall we change the vertices to colors?

PR #62, opened 2016-09-02
Added accessors

ibgreen:
@gnavvy -  All positions now use array format and all layers use accessors that application can override. 

- Apps no longer have to transform data to a specific format to use deck.gl.
- Also apps can now pass any iterable objects (e.g. Immutable Lists) as deck.gl data props.

This is a breaking change that will be merged to dev branch and initially published in a 3.0.0 beta release.

### DIFF
--- a/src/layers/arc-layer/arc-layer.js
+++ b/src/layers/arc-layer/arc-layer.js
@@ -22,11 +22,10 @@ import BaseLayer from '../base-layer';
 import {Model, Program, Geometry} from 'luma.gl';
 const glslify = require('glslify');
 
-const RED = [255, 0, 0];
-const BLUE = [0, 0, 255];
+const DEFAULT_COLOR = [0, 0, 255];
 
-const defaultGetPosition0 = x => x.position0;
-const defaultGetPosition1 = x => x.position1;
+const defaultGetSourcePosition = x => x.sourcePosition;
+const defaultGetTargetPosition = x => x.targetPosition;
 const defaultGetColor = x => x.color;
 
 export default class ArcLayer extends BaseLayer {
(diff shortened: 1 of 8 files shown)


ibgreen:
@gnavvy - Some work still remains on Choropleth and Grid Layers.

src/layers/choropleth-layer/choropleth-layer.js
```diff
@@ -51,11 +51,11 @@ export default class ChoroplethLayer extends BaseLayer {
     attributeManager.addDynamic({
       // Primtive attributes
       indices: {size: 1, update: this.calculateIndices, isIndexed: true},
-      positions: {size: 3, update: this.calculatePositions},
-      colors: {size: 3, update: this.calculateColors},
+      positions: {size: 3, update: this.calculateInstancePositions},
+      colors: {size: 3, update: this.calculateInstanceColors},
```
gnavvy: for choropleth layer, the positions and colors are not instanced.
(trying to pull these functions to the base-layer?)

src/layers/arc-layer/arc-layer.js
```diff
@@ -25,25 +25,35 @@ const glslify = require('glslify');
 const RED = [255, 0, 0];
 const BLUE = [0, 0, 255];
 
+const defaultGetPosition0 = x => x.position0;
+const defaultGetPosition1 = x => x.position1;
```
gnavvy: since we are doing the refactoring now, do you think it is more intuitive to name these two: sourcePosition and targetPosition, instead of position0 and position1, as we only have two vertices for the arc and the line layer?

ibgreen:
Addressed @gnavvy's review feedback

Issue #59, opened 2016-08-19
empty data array causes cascading errors


contra:
<img width="696" alt="screen shot 2016-08-19 at 4 01 27 pm" src="https://cloud.githubusercontent.com/assets/425716/17822659/39ec119a-6626-11e6-8bdf-fff5466de424.png">

tested on arc and scatterplot layers

ibgreen:
@contra This does not look good. Would like to fix but can't reproduce the cascade. 

Does "empty data array" mean `[]` or `null`? From the log it looks like the layer itself is null but that could the be a result of the error cascade...

To help pin this down I just opened a PR #66 :
- I've added a couple of regression test cases that instantiate the layers you mention with empty arrays, these work as expected. 
- I have also tried to change the ArcLayer's data array to `[]` in the example and that didn't cause a crash.
- I changed the ArcLayer's data to null in the example, that generated a single clean error, not a cascade.

Will attempt to land and publish asap so that you can try again with these changes in place.

ibgreen:
Landed as v2.4.6

PR #66, opened 2016-09-06
Add layer construction test cases, including empty and null data. 

ibgreen:
More robust layerName debug helper.


ibgreen:
@gnavvy @contra These changes are intended to mitigate and help debug problems reported in issue #59 - empty data array causes cascading errors.

PR #69, opened 2016-09-08
Update readme to specify dependency on a base node version

tangential:


README.md
```diff
@@ -32,6 +32,10 @@ Features:
 ```
 npm install --save deck.gl
 ```
+Note: deck.gl has a dependency on node version 0.12 or higher. If you use an older version, use a separate shell to install and build deck.gl
```
ibgreen: @tangential there are different node version managers (n, nvm, maybe more) so maybe just suggest nvm as one option

If you use an older version, you can install a node version manager like nvm and open a separate shell ...

tangential: Good point. Will edit inline.

PR #76, opened 2016-09-16
Fix picking

### LINKED ISSUE #73: click events not working?
index always comes back as -1 after updating to the latest (2.4.8)

ibgreen:
Fixes #73

@gnavvy @abmai @contra

Using the new luma logging it was easy to spot that the picking colors attribute was all zeroes. Fixed by a one line change to second parameter in `Layer.calculateInstancePickingColors`.

### DIFF
--- a/src/layers/base-layer.js
+++ b/src/layers/base-layer.js
@@ -396,7 +396,7 @@ export default class BaseLayer {
     this.willUnmount();
   }
 
-  calculateInstancePickingColors(attribute, numInstances) {
+  calculateInstancePickingColors(attribute, {numInstances}) {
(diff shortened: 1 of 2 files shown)


src/layers/base-layer.js
```diff
@@ -396,7 +396,7 @@ export default class BaseLayer {
     this.willUnmount();
   }
 
-  calculateInstancePickingColors(attribute, numInstances) {
+  calculateInstancePickingColors(attribute, {numInstances}) {
```
gnavvy: could you add jsdoc to this function on what is the second param and why we need to unpack numInstances out of it?

PR #78, opened 2016-09-20
Update deck.gl dev branch to use uber eslint definitions. 

ibgreen:
- Move eslint config to package.json
- Update source to conform
- Matches changes 

Stronger lint rules for deck.gl, matching luma.gl's new lint rules to ensure a consistent coding standard across the two repos.
@gnavvy @shaojingli


package.json
```diff
@@ -95,12 +97,17 @@
...
     "start": "budo example/main.js:example/bundle.js --live --open --port 3000 --css example/main.css --title 'deck.gl' --watch-glob '**/*.{html,css,js,glsl}' -- -t babelify -t envify"
   },
-  "standard": {
-    "ignore": [
-      "**/build/**",
-      "**/dist/**",
-      "**/examples/**",
-      "**/Documents/**"
-    ]
+  "eslintConfig": {
```
chrisirhc: Wondering why you went with `eslintConfig` vs `.eslintrc`, is there an advantage of going with this approach?

src/test/viewport-spec.js
```diff
@@ -91,14 +90,13 @@ test('Viewport projection matrix', t => {
   t.end();
 });
 
-
-test('Viewport.project#2D', t => {
+test('Viewport.projectFlat', t => {
```
chrisirhc: I'm guessing this is to resolve a regression?

package.json
ibgreen: Not strong reason, except that I like a clean root directory (without a lot of little files), and now the eslint config is so compact that it became reasonable to put it inside package.json. Easy to revert in case of any issues.

src/test/viewport-spec.js
ibgreen: Not really, there were some tricky lint warnings here so I had to rewrite.
We have added new projectFlat functions that ignore perspective, to match the old project ViewportMercatorProject functions so it seemed like the right thing to do.
Looks like it actually caused a regression though, this test case is failing now.

src/layers/core/choropleth-layer/choropleth-layer.js
```diff
@@ -168,15 +168,15 @@ export default class ChoroplethLayer extends BaseLayer {
   }
 
   calculateColors(attribute) {
-    const colors = this.state.groupedVerticesColors.map(
+    const nestedColors = this.state.groupedVerticesColors.map(
       colors => colors.map(
         color => this.props.drawContour ?
           // TODO - why destructure and rebuild array?
```
gnavvy: - why destruct and rebuild array?
the input color array could be vec4, instead of vec3

package.json
gnavvy: personally, I prefer to keep them in the dot files so that you can copy them over different projects.
I can easily forget I need to copy over part of the package.json file when scaffolding another project.

PR #82, opened 2016-09-24
Add performance example and docs

ibgreen:
@gnavvy @shaojingli Adds 1 million and 10 million point ScatterPlotLayers to the example.
@Pessimistress Adds initial performance doc

screen shot 2016-09-24 at 12 00 57 pm

### DIFF
--- a/docs/performance.md
+++ b/docs/performance.md
@@ -1,0 +1,99 @@
+# deck.gl Performance Notes
+
+
+## General Expectations
+
+While performance is obviously hardware dependent here are some guidelines:
+
+On recent MacBook Pros with dual graphics cards, we see most basic layers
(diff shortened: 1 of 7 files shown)


docs/performance.md
```diff
@@ -0,0 +1,99 @@
...
+
+deck.gl is built on luma.gl which has extensive debugging and instrumentation
+support.
+
+TBA
+
+
+## Implementation notes:
+
+Possible performance improvements:
```
contra: would be great if these were open as issue and tagged as `help wanted`

ibgreen: @contra Good idea: https://github.com/uber/deck.gl/issues/83

PR #87, opened 2016-09-27
Added more math functions to fp64 support. Added unit tests, perf test, etc...

shaojingli:
Another thousand line diff for rewiew.... @gnavvy @ibgreen

1) Implemented more "correct" math functions for fp64. Now, addition, subtraction, multiplication, division, square root, exponential and log are available
2) Implemented a workaround for NVIDIA to prevent its compiler from optimizing necessary operations away
3) Implemented a few performance test layers to test the performance of fp64 layers
4) Implemented unit tests for fp64 math functions
5) Refactored shaders and eliminated quite a few duplicate functions

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,7 +93,8 @@
     "test-headless-src": "babel-node --only=deck.gl/test,deck.gl/src src/test/headless.js",
     "test-browser": "budo src/test/browser.js:build/test-bundle.js --dir test --live --open --port 3001 --watch-glob '**/*.{html,css,scss,js,glsl}' -- -t babelify -t glslify",
     "test-electron": "browserify dist/test/electron.js | testron | faucet",
-    "start": "budo example/main.js:example/bundle.js --live --open --port 3000 --css example/main.css --title 'deck.gl' --watch-glob '**/*.{html,css,js,glsl}' -- -t babelify -t envify"
+    "start": "budo example/main.js:example/bundle.js --live --open --port 3000 --css example/main.css --title 'deck.gl' --watch-glob '**/*.{html,css,js,glsl}' -- -t babelify -t envify",
+    "test-shader": "budo src/test/shader.js:build/test-bundle.js --dir test --live --open --port 3001 --watch-glob '**/*.{html,css,scss,js,glsl}' -- -t babelify -t glslify"
   },
   "standard": {
     "ignore": [
--- a/src/test/shader.js
+++ b/src/test/shader.js
@@ -53,11 +53,11 @@ function addSpan(contents, div) {
 }
 
 function addDiv(contents) {
-    var testInfoDiv = document.createElement("div");
-    document.body.appendChild(testInfoDiv);
-    testInfoDiv.setAttribute("class", "testInfo");
+  var testInfoDiv = document.createElement("div");
+  document.body.appendChild(testInfoDiv);
+  testInfoDiv.setAttribute("class", "testInfo");
 
-    return testInfoDiv;
+  return testInfoDiv;
 }
 
 function logToConsole(msg) {
--- a/src/test/shader.js
+++ b/src/test/shader.js
@@ -41,15 +41,15 @@ function glEnumToString(gl, value) {
 };
 
 function addSpan(contents, div) {
-    if (div == undefined) {
-      var divs = document.body.getElementsByClassName("testInfo");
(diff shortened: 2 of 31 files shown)


package.json
```diff
@@ -93,7 +93,8 @@
     "test-headless-src": "babel-node --only=deck.gl/test,deck.gl/src src/test/headless.js",
     "test-browser": "budo src/test/browser.js:build/test-bundle.js --dir test --live --open --port 3001 --watch-glob '**/*.{html,css,scss,js,glsl}' -- -t babelify -t glslify",
     "test-electron": "browserify dist/test/electron.js | testron | faucet",
-    "start": "budo example/main.js:example/bundle.js --live --open --port 3000 --css example/main.css --title 'deck.gl' --watch-glob '**/*.{html,css,js,glsl}' -- -t babelify -t envify"
+    "start": "budo example/main.js:example/bundle.js --live --open --port 3000 --css example/main.css --title 'deck.gl' --watch-glob '**/*.{html,css,js,glsl}' -- -t babelify -t envify",
+    "test-shader": "budo src/test/shader.js:build/test-bundle.js --dir test --live --open --port 3001 --watch-glob '**/*.{html,css,scss,js,glsl}' -- -t babelify -t glslify"
```
ibgreen: Nit: Keep the 'test-*' targets together

src/test/shader.js
```diff
@@ -0,0 +1,640 @@
...
+/* eslint-disable no-var, max-statements */
+import 'babel-polyfill';
+import {document, window} from 'global';
+import {Buffer, createGLContext, Program, glGetDebugInfo} from 'luma.gl';
+
+const glslify = require('glslify');
+
+// Utilities functions that to be moved to a common place for future tests
+
+function glEnumToString(gl, value) {
```
ibgreen: Similar functions exist in luma.gl/src/context.js etc. We should consolidate.

src/test/shader.js
```diff
@@ -0,0 +1,640 @@
...
+  if (ndx < 0) {
+    var msg = "getError expected" + ((glErrors.length > 1) ? " one of: " : ": ");
+    testFailed(msg + expected +  ". Was " + glEnumToString(gl, err) + " : " + opt_msg);
+  } else {
+    //var msg = "getError was " + ((glErrors.length > 1) ? "one of: " : "expected value: ");
+    //testPassed(msg + expected + " : " + opt_msg);
+  }
+};
+
+// Special utility functions for df64 tests
```
ibgreen: These shouldn't be in here, right? They should be in a general utility file for 64 bit floats.

src/test/shader.js
```diff
@@ -0,0 +1,640 @@
...
+  }
+  for (var p in gl) {
+    if (gl[p] == value) {
+      return p;
+    }
+  }
+  return "0x" + value.toString(16);
+};
+
+function addSpan(contents, div) {
```
ibgreen: While it is cool to have an HTML interface, it should be simpler/less maintenance to just use `tape` to run the tests and have `tap-browser-color` HTML-ify the resulting TAP results. This is already set up in the CI tests. We could just have a folder for tests that should only run under the browser, if that is an issue.

src/test/shader.js
```diff
@@ -0,0 +1,640 @@
...
+  if (!fp_texture_support)
+  {
+    console.error("no floating point texture support!");
+  }
+  return gl;
+}
+
+function initializeTexTarget(gl)
+{
+  var framebuffer = gl.createFramebuffer();
```
ibgreen: It would be really neat to do this with the luma Framebuffer/Texture classes. We should be able to write this with very few lines of code, and have it as an example of how to write shader tests. 
I'd be happy to look at that if you think it makes sense.

package.json
shaojingli: Fixed.

src/test/shader.js
shaojingli: Agree. Will consolidate later.

src/test/shader.js
shaojingli: These will be moved to a general fp64 library later.

shaojingli:
@ibgreen I agree with your comments. Let's do incremental changes later. We need to refactor all fp64 related JS utility as well as shaders into a new library.

PR #89, opened 2016-09-29
Additional fine tuning of fp64 functions and layers.

shaojingli:
Fixed a bug in the Nvidia workaround path in twoSub, twoSum and twoProd functions. This bug shows itself through the sin/cos/tan function testings.

Made project-fp64() using the recently implemented log_fp64() and tan_fp64() so the mercator projection by GPU is truly high precision.

Other than that, mostly coding style and other miscellaneous changes addressing previous review comments.
1) Fixed a few lines of code left from debugging
2) Fixed the copy right notice in shader files
3) Constant name changes according to conventions

Please review @gnavvy @ibgreen


shaojingli:
@ibgreen  not really... judging from naked eyes... 

scatterplot64 layer with 1M points are still acceprable, as before. but we need some data to be sure though.

PR #90, opened 2016-10-06
Better geojson support in ChoroplethLayer

Pessimistress:
- renders MultiPolygon and Polygon with holes
- add strokeWidth prop for drawContour: true mode
- add getColor accessor (default to use color property of the feature)
- use Uint32Array for indices when available to support larger feature sets
- throws error when vertex count exceeds UNSIGNED_SHORT limit

@ibgreen @gnavvy

### DIFF
--- a/src/layers/core/choropleth-layer/choropleth-layer.js
+++ b/src/layers/core/choropleth-layer/choropleth-layer.js
@@ -158,7 +158,7 @@ export default class ChoroplethLayer extends BaseLayer {
       const numVertices = polygon.length;
 
       // use vertex pairs for gl.LINES => [0, 1, 1, 2, 2, ..., n-1, n-1, 0]
-      let indices = [...acc, offset];
+      const indices = [...acc, offset];
       for (let i = 1; i < numVertices - 1; i++) {
         indices.push(i + offset, i + offset);
       }
--- a/src/layers/core/choropleth-layer/choropleth-layer.js
+++ b/src/layers/core/choropleth-layer/choropleth-layer.js
@@ -128,14 +128,29 @@ export default class ChoroplethLayer extends BaseLayer {
     normalizedGeojson.features.map((feature, featureIndex) => {
       const {properties, geometry} = feature;
       const {coordinates, type} = geometry;
-      if (type === 'MultiPolygon') {
+      switch(type) {
+      case 'MultiPolygon': {
         const choropleths = coordinates.map(coords => ({
           coordinates: coords,
           featureIndex
         }));
         this.state.choropleths.push(...choropleths);
-      } else if (type === 'Polygon') {
+        break;
+        }
+      case 'Polygon':
         this.state.choropleths.push({coordinates, featureIndex});
+        break;
+      case 'LineString':
+        this.state.choropleths.push({coordinates: [coordinates], featureIndex});
+        break;
+      case 'MultiLineString': {
+        const choropleths = coordinates.map(coords => ({
+          coordinates: [coords],
+          featureIndex
+        }));
+        this.state.choropleths.push(...choropleths);
+        break;
+        }
       }
     });
 
@@ -143,7 +158,7 @@ export default class ChoroplethLayer extends BaseLayer {
       choropleth => {
         return choropleth.coordinates.map(
           polygon => polygon.map(
-            coordinate => [coordinate[0], coordinate[1], 0]
+            coordinate => [coordinate[0], coordinate[1], coordinate[2] || 0]
           )
         );
       }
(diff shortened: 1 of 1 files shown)


src/layers/core/choropleth-layer/choropleth-layer.js
```diff
@@ -58,10 +67,14 @@ export default class ChoroplethLayer extends BaseLayer {
         {size: 3, update: this.calculatePickingColors, noAlloc: true}
     });
 
+    const indexType = gl.getExtension('OES_element_index_uint') ?
```
ibgreen: Should we call it `IndexType` (capitalized) since it holds a type and you will call `new IndexType`?

src/layers/core/choropleth-layer/choropleth-layer.js
```diff
@@ -94,70 +108,107 @@ export default class ChoroplethLayer extends BaseLayer {
         drawMode: this.props.drawContour ? 'LINES' : 'TRIANGLES'
       }),
       vertexCount: 0,
-      isIndexed: true
+      isIndexed: true,
+      onBeforeRender() {
+        this.userData.oldStrokeWidth = gl.getParameter(gl.LINE_WIDTH);
+        this.program.gl.lineWidth(this.userData.strokeWidth || 1);
```
ibgreen: Move the `|| 0` to where we assign strokeWidth? Principle: validate data on the edge of the component, not "inside" it.

contra:
Does this support multipolygons with z indexes specified in the coordinates array?

Pessimistress:
@contra I have not encountered a geojson with z index in the coordinates. Can you point me to the specs for this?

contra:
@Pessimistress http://geojson.org/geojson-spec.html#positions

Positions in GeoJSON are [ x, y, z ] (Z is optional, but present frequently).

Couple of examples:
- Flight path (LineString)
- Airspace (MultiPolygon shapes in 3D)

Pessimistress:
@contra Thanks for suggesting this. With some minor tweaks this layer definitely works with 3D LineStrings in its `drawContour` mode. 

For mesh polygons and multipolygons, we are using `earcut` for triangulation - seems like it only considers x and y. Which is broken when 1)  not all points in the polygon are coplanar 2) the polygon is perpendicular to the ground. The latter case can be easily work around; first one is more tricky. In your experience, how are these 3d geoJSON generated? Could you point me to some real world samples for testing purpose?

src/layers/core/choropleth-layer/choropleth-layer.js
```diff
@@ -210,3 +207,89 @@ export default class ChoroplethLayer extends BaseLayer {
...
+  let choropleths;
+
+  switch(type) {
+  case 'MultiPolygon':
+    choropleths = coordinates;
+    break;
+  case 'Polygon':
+    choropleths = [coordinates];
+    break;
+  case 'LineString':
```
ibgreen: OK for now, but long-term, should line strings be handled here? They are not closed paths and should not be filled, so I assume this will not always give expected results? 

- It's almost like we should have one PolygonLayer, one LineStringLayer and one PointLayer.
- Even better, we should change deck.gl so that one layer can have multiple primitives. That would be a great addition for deck.gl v3.

src/layers/core/choropleth-layer/choropleth-layer.js
```diff
@@ -210,3 +207,89 @@ export default class ChoroplethLayer extends BaseLayer {
   }
 
 }
+
+// convert geojson to choropleths
```
ibgreen: Thanks for adding these. 

The wording 'to choropleths' is a little vague. How about this function converts list of features from a GeoJSON object to a list of GeoJSON polygons (i.e. array of polygons, first is outline, remaining are holes).

Maybe the return type can be specified differently?
  @returns {[Number,Number][][]}

JSDoc requires using "double" star comments to be recognized by tools such as linters and doc generators, i.e
`/** ... */` instead of `//`

src/layers/core/choropleth-layer/choropleth-layer.js
contra: @ibgreen I think LineStrings are out of scope for this component - for now should def do a layer for each primitive. Would love a robust and generic GeoJSON layer that can render anything in the spec - basically trying to hack one together right now, can contribute code from that if I can split it out

ibgreen: > Would love a robust and generic GeoJSON layer that can render anything in the spec - basically trying to hack one together right now, can contribute code from that if I can split it out

@contra - One problem I see is that a GeoJSON layer might need three separate primitives or models (GL.TRIANGLES for the filled polygon areas, GL.LINES for polygon outlines and LineStrings and and perhaps a GL.POINTS primitive for the Points.)

This could either be handled by having a Layer that uses multiple primitives (on the roadmap but not yet supported by deck.gl) or allow composite layers (also on the roadmap but not yet supported). 

An alternative is to "extrude" all one and zero dimensional objects (i.e. lines and points) into triangles. This would only require one primitive and generate the best visuals. There is an example of how to do this in the dev branch (layers/experimental-layers/extruded-choropleth-layer) but the extrusion used there doesn't currently handle steep angles gracefully.

You are welcome to put up a PR if you make some progress so that we can compare our approaches.

PR #91, opened 2016-10-08
Workaround for the inaccurate math functions of Intel GPUs

shaojingli:
Use the fp64 version of log() function for Intel GPUs. I added a #ifdef to the custom shaders of all existing layers. Please review, test and verify that it fixes the offset issue.

@ibgreen @Pessimistress @abmai @twojtasz


shaojingli:
The headless test failed...because glGetDebugInfo is returning nil. Is this an error or it's actually a limitation of our tests? @ibgreen

PR #94, opened 2016-10-10
Dev choropleth

contra:


contra:
Wow, no idea why github decided to do this. Clicked a button on my fork and it filed this against upstream - sorry about that!

PR #99, opened 2016-10-13
Add docs. Reorg files/directories. Demote HexagonLayer to sample layer.

ibgreen:
@Pessimistress Some additional docs. More in the works.


ibgreen:
@contra - Good catch, thanks!

For your reference since you are reviewing: the document update will be a step by step process. At this point I was mainly moving the existing documentation out of the repository README into separate doc files that can be linked to from the new web page. The various pieces certainly need another round of editing and review.

contra:
@ibgreen I can help out w/ docs if you want - I think it would be especially useful to document and have examples for the data parameters.

ibgreen:
@contra - that would be fantastic. We need simple docs of the parameters for each of the 5 core layers. If you want to go through a couple of those that would be great. 

For now I will focus on completing the articles on coordinate system handling, layer subclassing etc.

PR #103, opened 2016-10-14
Fixed the 64bit Arc Layer on Windows with Nvidia, Intel GPUs

shaojingli:
Fixed flickering issue of 64-bit arc layer
Fixed the 64-bit tan() function returning incorrect value on Intel GPU on Windows
Other improvements of 64-bit shader library

### DIFF
--- a/shaderlib/fp64/cos-fp64.glsl
+++ b/shaderlib/fp64/cos-fp64.glsl
@@ -91,6 +91,21 @@ vec2 cos_fp64(vec2 a) {
     vec2 u = vec2(0.0, 0.0);
     vec2 v = vec2(0.0, 0.0);
 
+#ifdef NVIDIA_WORKAROUND
+    if (abs(float(abs_k) - 1.0) < 0.5) {
+        u = COS_TABLE_0;
+        v = SIN_TABLE_0;
+    } else if (abs(float(abs_k) - 2.0) < 0.5) {
+        u = COS_TABLE_1;
+        v = SIN_TABLE_1;
+    } else if (abs(float(abs_k) - 3.0) < 0.5) {
+        u = COS_TABLE_2;
+        v = SIN_TABLE_2;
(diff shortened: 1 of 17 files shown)


shaderlib/fp64/cos-fp64.glsl
```diff
@@ -91,6 +91,21 @@ vec2 cos_fp64(vec2 a) {
     vec2 u = vec2(0.0, 0.0);
     vec2 v = vec2(0.0, 0.0);
 
+#ifdef NVIDIA_WORKAROUND
```
ibgreen: I think we should either just call this flag `NVIDIA` - or be more specific, `NVIDIA_COMPARISON_WORKAROUND`

src/layers/fp64/arc-layer/arc-layer-vertex.glsl
```diff
@@ -20,6 +20,8 @@
 
 #define SHADER_NAME arc-layer-vertex-shader
 
+uniform float ONE; // nvidia workaround
```
ibgreen: Consider integrating this the new shader assembly system and just do this in JS when loading the shader
```
  vs: assembleShader(gl, {
    fp64: true,
    vs: glslify()
  })
```

src/layers/fp64/scatterplot-layer/scatterplot-layer.js
```diff
@@ -100,7 +100,7 @@ export default class ScatterplotLayer extends BaseLayer {
     const debugInfo = glGetDebugInfo(gl);
 
     if (checkRendererVendor(debugInfo, 'nvidia')) {
-      nvidiaDef += '#define INTEL_WORKAROUND 1\n';
```
ibgreen: Again either `INTEL` or `INTEL_PRECISION_WORKAROUND` or similar.

PR #106, opened 2016-10-17
Changed how platform/GPU specific workaround code are added to shaders

shaojingli:
1) changed how platform specific defines are added to shaders
2) lumped the GPU/platform detection together
3) more intel specific workaround to solve unresolved corner case in fp64 maths
4) cleaned up fp64 shaders


src/layers/core/arc-layer/arc-layer.js
```diff
@@ -82,7 +84,8 @@ export default class ArcLayer extends BaseLayer {
 
     return new Model({
       program: new Program(gl, {
-        vs: assembleShader(gl, {vs: glslify('./arc-layer-vertex.glsl')}),
+        vs: assembleShader(gl, {vs: glslify('./arc-layer-vertex.glsl')},
+          getPlatformShaderDefines(gl)),
```
ibgreen: The idea here was that `assembleShader` would automatically add the platform defines to the output. I.e. it should not be necessary to call `getPlatformShaderDefines` when using `assembleShader`

src/lib/utils/get-platform-shader-defines.js
```diff
@@ -0,0 +1,41 @@
+import {glGetDebugInfo} from 'luma.gl';
```
ibgreen: Isn't there a `shader-utils` directory now?

src/lib/utils/get-platform-shader-defines.js
```diff
@@ -0,0 +1,41 @@
...
+}
+
+export function getPlatformShaderDefines(gl) {
+  let platformDefines = '';
+  const debugInfo = glGetDebugInfo(gl);
+
+  if (checkRendererVendor(debugInfo, 'nvidia')) {
+    platformDefines += '#define NVIDIA_GPU \n' +
+    '#define NVIDIA_FP64_WA 1\n' +
+    '#define NVIDIA_EQUATION_WA 1';
```
ibgreen: - There should be more documentation (in comments) about these flags, either here or at point of use, or both. 
- Also naming is better now, but still confusing (i.e. what is an EQUATION workaround?)
- I think `WA` could be spelled out as `WORKAROUND` - doesn't seem like an obvious abbreviation...

src/lib/utils/get-platform-shader-defines.js
```diff
@@ -0,0 +1,41 @@
...
+  } else if (checkRendererVendor(debugInfo, 'intel')) {
+    platformDefines += '#define INTEL_GPU \n #define INTEL_FP64_WA 1\n #define INTEL_LOG_WA 1';
+  } else if (checkRendererVendor(debugInfo, 'amd')) {
+    platformDefines += '#define AMD_GPU \n';
+  } else {
+    platformDefines += '#define DEFAULT_GPU \n';
+  }
```
ibgreen: I would assume that a number of defines are defined by the drivers - do we have any information about that?

Maybe something like 
```
#ifdef INTEL_...
#  define INTEL_WORKAROUND
#endif
```
Might be an alternative to driver name scraping?

src/lib/utils/get-platform-shader-defines.js
```diff
@@ -0,0 +1,41 @@
...
+export function getPlatformShaderDefines(gl) {
+  let platformDefines = '';
+  const debugInfo = glGetDebugInfo(gl);
+
+  if (checkRendererVendor(debugInfo, 'nvidia')) {
+    platformDefines += '#define NVIDIA_GPU \n' +
+    '#define NVIDIA_FP64_WA 1\n' +
+    '#define NVIDIA_EQUATION_WA 1';
+  } else if (checkRendererVendor(debugInfo, 'intel')) {
+    platformDefines += '#define INTEL_GPU \n #define INTEL_FP64_WA 1\n #define INTEL_LOG_WA 1';
```
ibgreen: Multiline strings?

```
platformDefines += `\
#define INTEL_GPU
#define INTEL_FP64_WA 1
#define INTEL_LOG_WA 1
`;
```
I'd also avoid spaces before the hash unless intentionally indenting in an `#ifdef`

PR #109, opened 2016-10-18
Voronoi

gnavvy:
The Voronoi layer re-uses 95+% of the code from the scatterplot layer.
The major difference is the primitive geometry: a circle as in the scatterplot layer vs. a cone in the Voronoi layer.
Currently, the cone primitive has a fixed but relative large radius that defines the maximum responsive area for a Voronoi site.
screen shot 2016-10-19 at 12 41 10 am
To obtain a full-screen Voronoi Diagram, increase the radius to a larger number that covers the diagonal length of the screen.

TODO (some in later PRs):
- [ ] Add a separate projection function in the shader, or add layer specific camera settings. Ideally, the cones in the Voronoi layer should always orthogonal to the viewport that does not transform as we tilt the camera;
- [ ] Add documentation;
- [ ] Add an interactive example with two layers: a regular scatterplot layer + a transparent Voronoi layer for proximate picking;
- [ ] Add a performance test example with 10K moving points as sites to demonstrate the efficiency;
- [ ] Extends the Voronoi layer (point-distance-field) to more general (line-/curve-/polygon-distance-field) layers.


src/layers/samples/voronoi-layer/voronoi-layer.js
```diff
@@ -0,0 +1,145 @@
...
+    }
+  }
+
+  calculateInstanceColors(attribute) {
+    const {data, getColor} = this.props;
+    const {value, size} = attribute;
+    let i = 0;
+    for (const point of data) {
+      const color = getColor(point);
+      // use random colors for demostration
```
kriskowal: It would be possible and desirable to assign one of [four colors][4C] to each veronoi cell. It would involve a dynamic programming solution, traversing the adjacency matrix of these cells, assigning the color not currently used by an adjacent cell.
[4C]: https://en.wikipedia.org/wiki/Four_color_theorem

ibgreen: >It would be possible and desirable to assign one of four colors to each veronoi cell

@kriskowal - That would be really cool. If the application was able to calculate such a coloring, it could certainly use the `getColor` accessor to render it.

Now the beauty of this particular implementation of Voronoi is that it doesn't need to calculate an adjacency matrix, it just consults the WebGL depth buffer for each pixel to find the closest cone generated by the center points. 

So if we implement this, the four color calculation should probably be kept separate as an option on the JavaScript side. 

I took a quick look at npm but to my surprise couldn't find any graph coloring modules...

kriskowal: > fourcolors will be yours. Oh yes, fourcolors will be yours.
> — https://www.npmjs.com/package/fourcolors

PR #120, opened 2016-10-19
Extend the far plane to show models with negative z coord

shaojingli:
Extend the far plane for creating the perspective projection matrix so that models with negative height are also shown

@Pessimistress @ibgreen


ibgreen:
@Pessimistress  Published as beta22

PR #121, opened 2016-10-20
Lifecycle updates. Composite Layers. GeoJsonLayer skeleton. Doc updates.

ibgreen:


ibgreen:
@shaojingli - This broke the FP64 layers should be easy to fix.
@contra - the initial GeoJson layer is now on dev.

Also likely breaks support for lineWidth, strokeWidth for some layers, will be restored when settings support lands in luma.gl.

PR #126, opened 2016-10-21
Extruded choropleth layer 64-bit

shaojingli:
64-bit extruded choropleth layer. Good for displaying 3D buildings
64-bit line layer.


src/layers/fp64/scatterplot-layer/scatterplot-layer.js
```diff
@@ -65,20 +65,6 @@ export default class ScatterplotLayer extends Layer {
...
-    });
-  }
-
-  didMount() {
-    this.updateUniforms();
-  }
-
-  willReceiveProps(oldProps, newProps) {
-    super.willReceiveProps(oldProps, newProps);
-    this.updateUniforms();
```
shaojingli: These are no longer get called. Uniform ONE is set in the base layer

src/layers/fp64/arc-layer/arc-layer.js
```diff
@@ -65,7 +65,6 @@ export default class ArcLayer extends Layer {
     this.setState({model});
 
     attributeManager.addInstanced({
-      instancePositions: {size: 4, update: this.calculateInstancePositions},
```
shaojingli: instancePositions are not used in 64bit arc layer. it's replaced by instanceTargetPositionFP64 and instanceSourcePositionFP64

src/layers/fp64/choropleth-layer/choropleth-layer.js
```diff
@@ -75,7 +75,6 @@ export default class ChoroplethLayer64 extends Layer {
     const IndexType = gl.getExtension('OES_element_index_uint') ?
       Uint32Array : Uint16Array;
 
-    this.setUniforms({opacity: this.props.opacity});
```
shaojingli: Same... opacity is a base layer attribute, and the corresponding uniform is set in base layer

example/main.js
```diff
@@ -45,6 +45,7 @@ const MAPBOX_ACCESS_TOKEN = process.env.MAPBOX_ACCESS_TOKEN ||
   'Set MAPBOX_ACCESS_TOKEN environment variable or put your token here.';
 
 const CHOROPLETHS_FILE = './example/data/sf.zip.geo.json';
+const EXTRUDED_CHOROPLETHS_FILE = './example/data/sfo.buildings.zoom13.json';
```
ibgreen: I don't think putting this file in the repo is right. We'll need to find/make some other small test file instead (or load from a service), we should probably filter-branch out this file to keep the repo from getting too big.

src/layers/fp64/choropleth-layer/choropleth-layer.js
ibgreen: Agreed. In addition I think the base layer applies gamma to the opacity, having it here defeats that.

src/layers/fp64/extruded-choropleth-layer/extruded-choropleth-layer-vertex.glsl
```diff
@@ -0,0 +1,91 @@
...
+  return ambient + uPointLightAttenuation * diffuse;
+}
+
+void main(void) {
+  vec2 projected_xy[2];
+  project_position_fp64(positions, projected_xy);
+  vec2 scaled_height = mul_fp64(heights, vec2(projectionPixelsPerUnit.x * elevation, 0.0));
+
+  vec2 vertex_pos_modelspace[4];
+  vertex_pos_modelspace[0] = projected_xy[0];
```
ibgreen: Longer term - Feels like there should be a function for this (`make_vec4_fp64`? I wonder if we used a struct instead of an array if we could define a type `vec4_64` to encapsulate this?)

src/layers/fp64/extruded-choropleth-layer/extruded-choropleth-layer-vertex.glsl
```diff
@@ -0,0 +1,91 @@
...
+  vertex_pos_modelspace[1] = projected_xy[1];
+  vertex_pos_modelspace[2] = sum_fp64(scaled_height, vec2(1.0, 0.0));
+  vertex_pos_modelspace[3] = vec2(1.0, 0.0);
+
+  vec2 vertex_pos_clipspace[4];
+  project_to_clipspace_fp64(vertex_pos_modelspace, vertex_pos_clipspace);
+
+  gl_Position = vec4(
```
ibgreen: Would it not be reasonable for `project_to_clipspace` to pick the x-coordinates and return a normal vec4? Is it likely that we will want to do 64 bit math in clip space?

src/layers/fp64/extruded-choropleth-layer/extruded-choropleth-layer-vertex.glsl
```diff
@@ -0,0 +1,91 @@
...
+  	vertex_pos_clipspace[0].x,
+  	vertex_pos_clipspace[1].x,
+  	vertex_pos_clipspace[2].x,
+  	vertex_pos_clipspace[3].x
+  );
+
+  vec3 color = applyLighting(
+  	vec3(
```
ibgreen: Again I'd love a function (`to_vec3_32`?)

src/layers/fp64/extruded-choropleth-layer/extruded-choropleth-layer.js
```diff
@@ -0,0 +1,449 @@
...
+import {fp64ify} from '../../../lib/utils/fp64';
+
+const glslify = require('glslify');
+
+import {vec3} from 'gl-matrix';
+
+import earcut from 'earcut';
+import flattenDeep from 'lodash.flattendeep';
+
+const ATTRIBUTES = {
```
ibgreen: We've removed the ATTRIBUTES from the other layer and places these directly in attributeManager.add()

src/layers/fp64/extruded-choropleth-layer/extruded-choropleth-layer.js
```diff
@@ -0,0 +1,449 @@
...
+      normals: {update: this.calculateNormals}
+    });
+
+    this.setState({
+      numInstances: 0,
+      model: this.getModel(gl)
+    });
+
+    this.setUniforms({
+      elevation: Number.isFinite(elevation) ? elevation : 1,
```
ibgreen: I think these should be passed in draw()

src/layers/fp64/extruded-choropleth-layer/extruded-choropleth-layer.js
```diff
@@ -0,0 +1,449 @@
...
+    return new Model({
+      id: this.props.id,
+      program: new Program(gl, assembleShaders(gl, {
+        vs: glslify('./extruded-choropleth-layer-vertex.glsl'),
+        fs: glslify('./extruded-choropleth-layer-fragment.glsl'),
+        fp64: true,
+        project64: true
+      })),
+      geometry: new Geometry({
+        id: this.props.id,
```
ibgreen: I think model sets the id on its geometry. Not needed.

src/layers/fp64/extruded-choropleth-layer/extruded-choropleth-layer.js
```diff
@@ -0,0 +1,449 @@
...
+  draw({uniforms}) {
+    this.state.model.render({
+      ...uniforms
+    });
+  }
+
+  // each top vertex is on 3 surfaces
+  // each bottom vertex is on 2 surfaces
+  calculatePositions(attribute) {
+
```
ibgreen: empty line - kill?

src/layers/fp64/extruded-choropleth-layer/extruded-choropleth-layer.js
```diff
@@ -0,0 +1,449 @@
...
+      }),
+      vertexCount: 0,
+      isIndexed: true
+    });
+  }
+
+  draw({uniforms}) {
+    this.state.model.render({
+      ...uniforms
+    });
```
ibgreen: Here just add the attributes from setUniforms

src/layers/fp64/line-layer/line-layer.js
```diff
@@ -0,0 +1,145 @@
...
+  willReceiveProps(oldProps, nextProps) {
+    super.willReceiveProps(oldProps, nextProps);
+    this.state.model.userData.strokeWidth = nextProps.strokeWidth;
+  }
+
+  draw({uniforms}) {
+    this.state.model.render(
+      uniforms,
+      {
+        lineWidth: this.props.strokeWidth
```
ibgreen: This doesn't work yet (pending luma update). 

But this should

```
    const {gl} = this.context;
    const oldStrokeWidth = gl.getParameter(GL.LINE_WIDTH);
    gl.lineWidth(this.props.strokeWidth || 1);

    this.state.model.render(uniforms);

    gl.lineWidth(oldStrokeWidth || 1);
``

File: src/layers/fp64/line-layer/line-layer.js [hunk 14]
```diff
@@ -0,0 +1,145 @@
...
+        fs: glslify('./line-layer-fragment.glsl'),
+        fp64: true,
+        project64: true
+      })),
+      geometry: new Geometry({
+        drawMode: 'LINE_STRIP',
+        positions: new Float32Array(positions)
+      }),
+      isInstanced: true,
+      onBeforeRender() {
```
ibgreen: If using above technique, kill this.

src/layers/fp64/scatterplot-layer/scatterplot-layer.js
ibgreen: Yep. more precisely, the new mechanism is that is being returned by viewport.getUniforms() which is what is passed in to draw()

example/main.js
shaojingli: This file is just several hundred kilo... similar to sf.zip.geo.json file. I guess we could get away with it for now. Or, I could generate a smaller file.

src/layers/fp64/extruded-choropleth-layer/extruded-choropleth-layer-vertex.glsl
shaojingli: yep.. that will happen when we move fp64 lib to luma.. not this week or next week though

src/layers/fp64/extruded-choropleth-layer/extruded-choropleth-layer-vertex.glsl
shaojingli: I agree. Will change this

src/layers/fp64/extruded-choropleth-layer/extruded-choropleth-layer.js
shaojingli: This uniform never changed... so I feel it's better to put them here...ideally they should just be set once, but right now we still set them every time we draw..

ibgreen: > This uniform never changed... so I feel it's better to put them here...ideally they should just be set once, but right now we still set them every time we draw..

Are you thinking about eliminating the WebGL calls that actually set the uniforms on every render? I am not sure what is the best way to do that, would love to get your thoughts offline. WebGL 2 has uniform buffers but haven't looked much at them.

If not, then I don't think there is much of an advantage to set these up front, as they will just be added to the render call by the model class or the program class instead.

shaojingli:
Get mixed with 32-bit and 64-bit lineLayer and arcLayer linewidth settings... will correct it in next revision

Issue #55, opened 2016-08-15
scatterplot - outline


contra:
would be great to be able to specify an outline color + width on scatterplot points

temporary workaround is to render another layer with the same data, but different colors and a radius 1 unit higher at a z of 0, with your actual data points at a z of 1.

gnavvy:
had the same issue for the choropleth layer. in deck.gl we tend to render one primitive for one layer, so the two-layer workaround is preferred for the moment.

mcnuttandrew:
The above ^ PR doesn't quite address the letter of this issue, but provides an enhancement for the two layer approach

mcnuttandrew:
Hey @contra, with the merge of #130 I think this issue is resolved, so i'm going to close it.

contra:
@mcnuttandrew The scatterplot outline mode does only the outline, right? This ticket was about doing a dot with an outline (or shadow) for visibility reasons.

PR #130, opened 2016-10-25
Add outline to scatterplot

mcnuttandrew:
screen shot 2016-10-24 at 4 50 23 pm

Add outline to scatterplot in response to #55

### DIFF
--- a/example/layer-examples.js
+++ b/example/layer-examples.js
@@ -68,6 +68,8 @@ const ScatterplotLayerExample = props =>
   new ScatterplotLayer({
     data: props.points,
     opacity: 0.5,
+    outlineMode: true,
+    outlineStrokeWidth: 2,
     pickable: true,
     onHover: props.onScatterplotHovered,
     onClick: props.onScatterplotClicked
@@ -89,6 +91,7 @@ const ScatterplotLayerMeterExample = props =>
       {position: [1000, 1000]}
     ],
     opacity: 0.5,
+    outlineMode: true,
     pickable: true,
     onHover: props.onScatterplotHovered,
     onClick: props.onScatterplotClicked
--- a/example/layer-examples.js
+++ b/example/layer-examples.js
@@ -91,7 +91,6 @@ const ScatterplotLayerMeterExample = props =>
       {position: [1000, 1000]}
     ],
     opacity: 0.5,
-    outlineMode: true,
     pickable: true,
     onHover: props.onScatterplotHovered,
     onClick: props.onScatterplotClicked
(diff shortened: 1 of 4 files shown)


src/layers/core/scatterplot-layer/scatterplot-layer.js
```diff
@@ -98,10 +110,17 @@ export default class ScatterplotLayer extends Layer {
         })
       ),
       geometry: new Geometry({
-        drawMode: 'TRIANGLE_FAN',
+        drawMode: drawMode || 'TRIANGLE_FAN',
         positions: new Float32Array(positions)
       }),
-      isInstanced: true
+      isInstanced: true,
+      onBeforeRender() {
```
ibgreen: In deck.gl v3 it is possible to override Layer.draw(). That enables cleaner handling of gl parameters compared to setting state and using callbacks on the model
 See [ArcLayer](https://github.com/uber/deck.gl/blob/dev/src/layers/core/arc-layer/arc-layer.js#L70).

example/layer-examples.js
```diff
@@ -68,6 +68,8 @@ const ScatterplotLayerExample = props =>
   new ScatterplotLayer({
     data: props.points,
     opacity: 0.5,
+    outlineMode: true,
+    outlineStrokeWidth: 2,
```
ibgreen: I believe that all the other layers that support line width prop just call it `strokeWidth`, so calling it that would be a little more consistent.

src/layers/core/scatterplot-layer/scatterplot-layer.js
```diff
@@ -65,6 +67,16 @@ export default class ScatterplotLayer extends Layer {
   }
 
   willReceiveProps(oldProps, newProps) {
+    if (oldProps.outlineMode !== newProps.outlineMode) {
+      const {gl} = this.state;
+      const model = this._getModel(gl, newProps.outlineMode ? 'LINE_LOOP' : 'TRIANGLE_FAN');
```
ibgreen: I'd prefer to use GL constants here. `GL.LINE_LOOP` and `GL.TRIANGLE_FAN`. `GL` can be imported from luma, see other layers on dev (if examples are needed). 

The strings will be looked up and resolved to those constants, but no need to make the reader guess what is happening.

src/layers/core/scatterplot-layer/scatterplot-layer.js
```diff
@@ -65,6 +67,16 @@ export default class ScatterplotLayer extends Layer {
...
+
+      this.state.model = model;
+    }
+    if (newProps.outlineMode) {
```
ibgreen: This could be cut if going with `draw`

example/layer-examples.js
contra: I think to make it consistent with choropleth (`drawContour`) the `outlineMode` option should be `drawOutline`

PR #132, opened 2016-10-26
unbreak screen grid layer

Pessimistress:
- Fix broken ScreenGridLayer on the deck.gl 3
- Add custom color ramp
- Add position and weight accessors

### DIFF
--- a/src/layers/core/screen-grid-layer/screen-grid-layer-vertex.glsl
+++ b/src/layers/core/screen-grid-layer/screen-grid-layer-vertex.glsl
@@ -41,10 +41,10 @@ uniform vec3 selectedPickingColor;
 varying vec4 vColor;
 
 void main(void) {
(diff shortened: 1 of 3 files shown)


src/layers/core/screen-grid-layer/screen-grid-layer-vertex.glsl
```diff
@@ -27,20 +27,24 @@
...
+uniform vec4 minColor;
+uniform vec4 maxColor;
 uniform float renderPickingBuffer;
 uniform vec3 cellScale;
 uniform vec3 selectedPickingColor;
 
 varying vec4 vColor;
 
 void main(void) {
+  vec4 instanceColor = mix(minColor, maxColor, instanceCount / maxCount);
```
ibgreen: This temporary variable`instanceColor`, made me think it was an attribute, I had to read the code two times. Maybe just `color` would be simple?

PR #138, opened 2016-10-31
Split the viewport into GL and non-GL related classes.

ibgreen:
@shaojingli @gnavvy - Step 1 towards moving core viewport to viewport-mercator-project.

### DIFF
--- a/src/viewport/index.js
+++ b/src/viewport/index.js
@@ -1,6 +1,10 @@
+// Note: The base viewport class will be moved to a separate module
+// viewport-mercator-project
 export * from './viewport';
 export {default as BaseViewport} from './viewport';
+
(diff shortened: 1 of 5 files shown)


src/viewport/index.js
```diff
@@ -1,2 +1,6 @@
-export {default as default} from './viewport';
-export {default as Viewport} from './viewport';
+export * from './viewport';
+export {default as BaseViewport} from './viewport';
+export * from './webgl-viewport';
+export {default as WebGLViewport} from './webgl-viewport';
+export {default as Viewport} from './webgl-viewport';
```
gnavvy: could you add a comment that Viewport is for legacy support (and will be removed in future releases)?

PR #139, opened 2016-10-31
Update layer docs

Apercu:
Standardized parameters types case among layer docs, and moved the default values (if any) down the heading, to provide more control over the markup with its value and possible description.

Had a little doubt about keeping the additional notes in the *Choropleth* one, since it's in the `data` object so I kept it for now.

### DIFF
--- a/docs/layer.md
+++ b/docs/layer.md
@@ -9,11 +9,9 @@ The layer class also defines the lifecycle methods of the layers, as well as
 a a couple of support methods which the lifecycle methods can use. These are
 important only when subclassing layers and are therefore described separetely.
 
-
 ## Layer Properties
 
-
-### `id` (string, optional): layer id
+##### `id` (String, optional)
 
 The id must be unique among all your layers. The layer's id defaults to the
 `Layer` class name. If you have more than one instance of the same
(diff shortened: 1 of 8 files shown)


docs/layer.md
```diff
@@ -25,53 +23,56 @@ which helps avoid id collisions in this case.
 
 E.g. assuming a composite GeoJsonLayer layer that renders two sublayers,
 choropleths and lines, with those ids:
-```
```
ibgreen: If we remove the triple quotes, the examples will not be formatted as code. Is this intentional?

docs/layer.md
```diff
@@ -25,53 +23,56 @@ which helps avoid id collisions in this case.
...
+    if (showMyLayer) {
+      new MyLayer({visible: showMyLayer});
+    }
+
 In the second example (conditional rendering) the layer state will
 be destroyed and regenerated every time the showMyLayer flag changes.
 
-
-### `opacity` (number, required)
+##### `opacity` (Number, required)
```
ibgreen: I think the number of hashes `###` is important as it determines the sizing of the header. We should be consistent across files. Suggest using 3 hashes unless visuals are better with 5.

docs/layers/arc-layer.md
```diff
@@ -1,9 +1,42 @@
...
 
-**Layer-specific Parameters**
+Inherits from all [Base Layer properties](/docs/layer.md).
 
-* `data` (array, required) array of objects: [{ position: {x0, y0, x1, y1},
-color }, ...]
+## Layer-specific Properties
+
+##### `strokeWidth` (Number, optional)
+
```
ibgreen: There was no empty line in layer.md. Let's be consistent, whatever way looks best.

docs/layer.md
Apercu: If you indent your code with at least 4 spaces it will be code-formatted, this way you don't have to add extra markup to do it, and it removes the extra left-padding. We could keep the triple backtick, but would need to remove the extra spaces on each line.

docs/layer.md
Apercu: I found an `<h3>` (1.25em as per github css) for the props was quite aggressive visually, and that the `<h5>` font size was slightly better for this kind of thing, that way we better differentiate our API from the headings.

If you think otherwise I'm very open to this, since it's pretty personal and a secondary issue.

docs/layer.md
ibgreen: OK thanks, I didn't know that. We can go with the indent.

docs/layer.md
ibgreen: No problem, 5 is OK, as long as we are consistent.

PR #140, opened 2016-10-31
Fix arc layer flickering

shaojingli:
Attempt to fix the arc layer flickering.

The flickering of the segment 0 of arc layer seems due to negative results returned by the paraboloid() and paraboloid_fp64() function.


ibgreen:
Aha, so `sqrt` is failing? Do we know what happens if GLSL functions fail on invalid range? Do they generate floating point NaN?

shaojingli:
@ibgreen they are "undefined" per GLSL specs... a lot of stuff in GLSL are undefined :( on some platforms it returns nan, but on others it returns inf...

PR #149, opened 2016-11-01
Fixes to lineWidth warning, context.viewport = null. opacity prop.

ibgreen:
Also enables prop diff tracing.


Apercu:
Since it modifies default props, docs should be updated to reflect this.

PR #151, opened 2016-11-01
Allow url rewrites in docs and fix routes

Apercu:


demo/src/javascripts/components/markdown-page.js
```diff
@@ -3,25 +3,35 @@ import marked from 'marked';
 import {highlight} from 'highlight.js';
 import React, {Component, PropTypes} from 'react';
 
+const urlRewrites = {
```
ibgreen: I think we should add a comment here explaining why we do this. Otherwise this can easily be broken in a future update.

PR #158, opened 2016-11-02
Add fps counter to demo page

Apercu:
While we figure out if we want a live demo or video, this might come handy if we take the first solution.

### DIFF
--- a/demo/src/javascripts/components/home.js
+++ b/demo/src/javascripts/components/home.js
@@ -1,4 +1,5 @@
 import 'babel-polyfill';
+import Stats from 'stats.js';
 import React, {Component} from 'react';
 import {connect} from 'react-redux';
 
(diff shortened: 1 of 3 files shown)


demo/src/javascripts/components/home.js
```diff
@@ -35,6 +36,18 @@ class Home extends Component {
     this._onScroll();
     this._resizeMap();
 
+    this._stats = new Stats();
+    this._stats.showPanel(0);
+    this.refs.fps.appendChild(this._stats.dom);
+
+    const calcFPS = () => {
+      this._stats.begin();
```
ibgreen: @apercu - yes, good first step. To get really correct results, we may need to hook it into the DeckGL rendering loop, but maybe this is good enough for now. Let's merge it so I can test.

PR #159, opened 2016-11-02
doc update

gnavvy:
- added link from the get started button to the get started page;
- simplified the text in the introduction page
- updated the getting-started page
* updated eslint to the latest version (2.* is not compatible with my editor, will revert if this causes any issue)

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,7 @@
     "discify": "^1.2.0",
     "electron-prebuilt": "^0.37.2",
     "envify": "^3.4.0",
-    "eslint": "^2.0.0",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -63,7 +63,7 @@
     "discify": "^1.2.0",
     "electron-prebuilt": "^0.37.2",
     "envify": "^3.4.0",
-    "eslint": "^2.0.0",
+    "eslint": "^3.9.1",
```
ibgreen: This is always a little risky, but as long as CI passes, I am fine with this.

Issue #156, opened 2016-11-02
All 64-bit layers now wobbling...


shaojingli:
All 64-bit layers are wobbling as of beta31. I need to do some regression tests tomorrow.
It's highly likely due to recent viewport changes.
Assigning to myself.

contra:
Still getting some pretty major wobbling on FP64 layers even on the latest dev branch

shaojingli:
@contra Saddly, the 64-bit layers are not supported on 2013 MBP with NVIDIA GPUs from the beginning...because the machine is not available to buy and we could secure one to test... there are some unique peculiarity in mac version of nvidia driver. 

I tested MBP with AMD and Intel GPUs, Linux and Windows with NVIDIA GPUs, the wobbling is not there as of today.

contra:
@shaojingli Any way I can help out with that? You can use my machine remotely to debug if you want.

PR #160, opened 2016-11-02
Fix the fp64 projection matrix

shaojingli:
Make the 64-bit layers work again... by providing correct projection matrix. Solved issue #156

### DIFF
--- a/src/viewport/webgl-viewport.js
+++ b/src/viewport/webgl-viewport.js
@@ -100,7 +100,7 @@ export default class WebGLViewport extends Viewport {
         [
           this.glProjectionMatrixFP64[(i * 4 + j) * 2],
           this.glProjectionMatrixFP64[(i * 4 + j) * 2 + 1]
-        ] = fp64ify(this.glProjectionMatrix[j * 4 + i]);
+        ] = fp64ify(this.viewProjectionMatrix[j * 4 + i]);
       }
     }
   }


src/viewport/webgl-viewport.js
```diff
@@ -100,7 +100,7 @@ export default class WebGLViewport extends Viewport {
         [
           this.glProjectionMatrixFP64[(i * 4 + j) * 2],
           this.glProjectionMatrixFP64[(i * 4 + j) * 2 + 1]
-        ] = fp64ify(this.glProjectionMatrix[j * 4 + i]);
+        ] = fp64ify(this.viewProjectionMatrix[j * 4 + i]);
```
ibgreen: About the comment above. I am a little surprised that we need to transpose... One of the main features of the `glMatrix` is supposedly that matrices are already aligned in the order webgl expects (column major memory layout).

src/viewport/webgl-viewport.js
```diff
@@ -100,7 +100,7 @@ export default class WebGLViewport extends Viewport {
         [
           this.glProjectionMatrixFP64[(i * 4 + j) * 2],
           this.glProjectionMatrixFP64[(i * 4 + j) * 2 + 1]
-        ] = fp64ify(this.glProjectionMatrix[j * 4 + i]);
```
ibgreen: I see, we ended up putting the truncated 32 bit values in the 64 bit matrix...

src/viewport/webgl-viewport.js
shaojingli: ummm... you might be right...let me change this later...

PR #163, opened 2016-11-02
Update some docs, and allow to remove links on website.

Apercu:


demo/src/javascripts/components/markdown-page.js
```diff
@@ -4,7 +4,8 @@ import {highlight} from 'highlight.js';
 import React, {Component, PropTypes} from 'react';
 
 const urlRewrites = {
-  '/docs/layers/base-layer.md': '#/layers/base-layer/layer'
+  '/docs/layers/base-layer.md': '#/layers/base-layer/layer',
+  '/docs/custom-layers.md#layerencodepickingcolorindex--number': false
```
ibgreen: Again, I think we need some comments on this urlRewrite thing. It is hard to understand what it for someone new.

docs/custom-layers.md
```diff
@@ -101,16 +95,11 @@ built-in picking while adding layer specific selection information, override
...
-its normal color calculation.
-
-
-See `Layer.encodePickingColor
-
+The layer is rendered with a special uniform set to 1, which causes it to
+render into a off-screen framebuffer using `pickingColor` or
+`instancePickingColor` attributes instead of using its normal color calculation.
 
+See [`Layer.encodePickingColor`](/docs/custom-layers.md#layerencodepickingcolorindex--number).
```
ibgreen: This link looks very fragile, I am sure it will break soon as we edit.

docs/tips-and-tricks.md
```diff
@@ -39,18 +38,16 @@ It is recommended, but not required, to use immutable data (containers AND
...
 
-
 ## Notes on picking
 
 **Note**: Because DeckGL layers are designed to take any type of iterable
 collection as data (which may not support "random access" array style
 references of its elements), the picking calculates and index but the
 actual object.
 
-FEATURE IDEA: The base layer could take an optional getObject(index) accessor
```
ibgreen: Let's delete this. It looks unfinished.

docs/custom-layers.md
Apercu: I fell like it's better so someone browsing the docs on Github may be able to click be directed to the corresponding method. If the heading changes and we forgot to change this, the link won't work, but there won't be any critical error. It will be transformed to a simple text on the website anyway, since we're not in html5 mode.

But I agree it may be additional work, so it's up to you if you want to remove it.

PR #169, opened 2016-11-03
Add 64 bit Layer docs

Apercu:


docs/layers/arc-layer-64.md
```diff
@@ -0,0 +1,9 @@
+# Arc Layer 64
```
ibgreen: I don't think we should split up the name of the layer. It should be ArcLayer64

docs/layers/arc-layer-64.md
```diff
@@ -0,0 +1,9 @@
...
+
+The Arc Layer takes in paired latitude and longitude coordinated points and
+render them as arcs that links the starting and ending points.
+
+    import {ArcLayer64} from 'deck.gl';
+
+Inherits from all [Base Layer](/docs/layers/base-layer.md) and
+[Arc Layer](/docs/layers/arc-layer.md) properties.
```
ibgreen: This makes it look like this is a subclass of ArcLayer, which it currently is not. Instead of `inherits`, say that `this layer has the same properties as`

docs/layers/arc-layer-64.md
```diff
@@ -0,0 +1,9 @@
+# Arc Layer 64
+
+The Arc Layer takes in paired latitude and longitude coordinated points and
```
ibgreen: We should just say that this is a 64 bit version of the standard ArcLayer. And refer the reader there, and maybe to the article about 64 bits.

PR #174, opened 2016-11-04
demo page revision

gnavvy:
another round of edit


demo/src/javascripts/components/home.js
```diff
@@ -125,42 +125,37 @@ class Home extends Component {
...
 
-              <h3>A Layered Approach to Visualization</h3>
+              <h3>
+                A layered approach with <a href="/#/layers/catalog/overview">growing catalog</a>
+              </h3>
               <p>
-              Based on the notion of layers and instancing deck.gl is
-              suitable to reason about and display data, perform
-              on-the-fly aggregations and visually explore large datasets.
+              Based on the notion of layering and instancing, deck.gl is suitable to display,
```
ibgreen: Skip instancing. I'd prefer something like this:

deck.gl allows complex visualizations to be constructed by reusing and composing layers , and provides an architecture enabling users to easily create and share new layers.  We already offer a catalog of proven layers and we have many more in the works.

demo/src/javascripts/components/home.js
```diff
@@ -125,42 +125,37 @@ class Home extends Component {
...
               <p>
-              deck.gl defines an architecture for packaging high performance
-              WebGL visualizations as reusable JavaScript layers. We already
-              offer a catalog of proven layers and we have many more in the
-              works.
+              By emulating float64 computations in the GPU, deck.gl supports rendering datasets
+              with unparalleled accuracy and performance.
               </p>
 
               <h3>Interoperability with Mapbox</h3>
```
ibgreen: "React and Mapbox Integrations"?

PR #177, opened 2016-11-04
Add thumbs on layer docs & lineLayer image

Apercu:


docs/layers/arc-layer-64.md
```diff
@@ -1,3 +1,7 @@
+<div align="center">
+  <img height="300" src="https://github.com/uber/deck.gl/raw/dev/demo/src/static/images/demo-thumb-02.jpg" />
```
ibgreen: - Should we name the image files after the layers to make it a little easier to maintain? Right now they just have numbers - `demo-thumb-02.jpg`

- Is it possible to load relative files rather than hard link to a branch?

- Also the image-06 might be a little large?

I'll need to see how it looks in the final version, wasn't able to view it here.

Apercu: - Definitely, wanted to keep it consistent at first.
- If we want to support both Github & website, it might be easier this way, without possibly having to rewrite image src on the website
- Strange, I'm sure I took a 800x800 picture, dunno why the size has been doubled

PR #178, opened 2016-11-04
Fixed the wobbling in meters projection mode

shaojingli:
Fixe the wobbling of meters projection mode by separating the transformation of origin and offset. 

 **mercatorProj(originPosition) * projectionMatrix** + offsetPosition * projectionMatrix

The **bold** part is done on CPU and the rest on GPU

Please verify before we merge @ibgreen

### DIFF
--- a/shaderlib/project/project.glsl
+++ b/shaderlib/project/project.glsl
@@ -8,7 +8,7 @@ const float PROJECT_MERCATOR_OFFSETS = 2.;
 
 uniform float projectionMode;
 uniform float projectionScale;
-uniform vec2 projectionCenter;
+uniform vec4 projectionCenter;
 uniform vec3 projectionPixelsPerUnit;
 uniform mat4 projectionMatrix;
 uniform mat4 projectionMatrixUncentered;
@@ -209,7 +209,7 @@ vec2 project_position(vec2 position) {
     return (position + vec2(TILE_SIZE / 2.0)) * projectionScale;
   }
   if (projectionMode == PROJECT_MERCATOR_OFFSETS) {
-    return project_scale(position) + projectionCenter;
+    return project_scale(position);
   }
   // Covers projectionMode == PROJECT_MERCATOR
   return project_mercator_(position) * WORLD_SCALE * projectionScale;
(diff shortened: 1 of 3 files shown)


src/layers/core/scatterplot-layer/scatterplot-layer-vertex.glsl
```diff
@@ -34,9 +34,7 @@ varying vec4 vColor;
 void main(void) {
   vec3 center = project_position(instancePositions.xyz);
   vec3 vertex = positions * project_scale(radius * instancePositions.w);
-  gl_Position = project(vec4(center, 1.0)) +
-                project(vec4(vertex, 0.0));
-
+  gl_Position = project(vec4(center + vertex, 1.0));
```
ibgreen: project_to_clipspace?

src/viewport/viewport.js
```diff
@@ -64,17 +64,36 @@ export default class WebGLViewport extends Viewport {
...
-      delta);
-
+    const projectedPositionOrigin = [positionOriginPixels[0], positionOriginPixels[1], 0.0, 1.0];
     const projections = this.getProjections();
-    const projectionMatrix = projections._viewProjectionMatrix;
+    const projectionMatrix = projections.viewProjectionMatrix;
+
+    const projectionCenter = [0.0, 0.0, 0.0, 0.0];
+
+    projectionCenter[0] =
```
ibgreen: `vec4` has a `transformMat4` method. Documented towards the end of http://glmatrix.net/docs/vec4.html.

shaderlib/project/project.glsl
```diff
@@ -226,6 +226,9 @@ vec4 project_position(vec4 position) {
 //
 
 vec4 project_to_clipspace(vec4 position) {
+  if (projectionMode == PROJECT_MERCATOR_OFFSETS) {
+    return projectionMatrix * vec4(position.xy, 0.0, 0.0) + projectionCenter;
```
ibgreen: For a separate PR:

We should just be able to build this offset into the `projectionMatrix` in `Viewport.getUniforms`?  

Then we could also apply a model matrix in `getUniforms`. 

Maybe we can build in the scaling as well, and then both `project_scale` and `project_position` become identity functions under OFFSET_METERS mode?

PR #186, opened 2016-11-07
Voronoi Layer fragment depth fix

ibgreen:
@gnavvy @shaojingli 
Partial fix to Voronoi Layer, uses gl_FragDepth extension to populate the depth buffer while rendering a flat image. 
screen shot 2016-11-07 at 6 53 52 am

### DIFF
--- a/src/viewport/old-viewport.js
+++ b/src/viewport/old-viewport.js
@@ -337,8 +337,8 @@ export default class Viewport {
       this._createMat4(),
       2 * Math.atan((this.height / 2) / this.altitude), // fov in radians
       this.width / this.height,                         // aspect ratio
-      0.1,                                              // near plane
-      this.farZ * 10.0                                  // far plane
+      0.2,                                              // near plane
+      this.farZ * 2                                     // far plane
     );
     /* eslint-enable no-inline-comments */
 
--- a/example/layer-examples.js
+++ b/example/layer-examples.js
@@ -15,7 +15,8 @@ import {
 } from '../src/layers/fp64';
 
 import {
-  EnhancedChoroplethLayer
+  EnhancedChoroplethLayer,
+  VoronoiLayer
 } from '../src/layers/samples';
 
 const ArcLayerExample = props =>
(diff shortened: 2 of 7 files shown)


src/viewport/old-viewport.js
```diff
@@ -337,8 +337,8 @@ export default class Viewport {
       this._createMat4(),
       2 * Math.atan((this.height / 2) / this.altitude), // fov in radians
       this.width / this.height,                         // aspect ratio
-      0.1,                                              // near plane
-      this.farZ * 10.0                                  // far plane
+      0.2,                                              // near plane
```
gnavvy: not sure if those constraints are still there, but in some layers we had to add 1.0 to the user provided z positions (usually = 0), otherwise the instances will not get rendered. changing 0.1 to -0.1 was able to fix those.

src/viewport/old-viewport.js
```diff
@@ -337,8 +337,8 @@ export default class Viewport {
...
+      this.farZ * 2                                     // far plane
```
shaojingli: Any specific reason that we change this?

src/viewport/old-viewport.js
ibgreen: > otherwise the instances will not get rendered. changing 0.1 to -0.1 was able to fix those.

I believe Shaojing fixed that by increasing the far plane (the `* 10` multiplier in the code). AFAIK, the near plane must be a positive number. I still kept a `* 2` multiplier which seemed to do the job.

src/viewport/old-viewport.js
ibgreen: > Any specific reason that we change this?

Turned out that the Voronoi layer is extremely sensitive to rounding errors in the depth buffer, and we get some "flickering" while tilting (not panning). Reducing the [range of the depth buffer](https://www.opengl.org/wiki/Depth_Buffer_Precision) from 100 to 10 mitigates this noticeably. On the other hand, the flickering is still not completely solved, and we can hardly reduce the depth buffer range more than that, so it doesn't really matter.

Also please note that this change is in unused code. We are importing the `Viewport` from viewport-mercator-project, not from this file. I was just experimenting here, so others can play with it after we land.

PR #187, opened 2016-11-07
Support drawCount uniform to control polygon offsets (defeat z-fighting)

ibgreen:
@shaojingli @gnavvy 

By adding a `drawCount` uniform we can add small offsets in the vertex shader while rendering to avoid z-fighting.

`drawCount` is simply an integer starting at 0 and incremented for each rendered layer.

### DIFF
--- a/src/lib/layer-manager.js
+++ b/src/lib/layer-manager.js
@@ -115,9 +115,11 @@ export default class LayerManager {
     assert(this.context.viewport, 'LayerManager.drawLayers: viewport not set');
 
     const {uniforms} = this.context;
+    let drawCount = 0;
     for (const layer of this.layers) {
       if (layer.props.visible) {
-        layer.drawLayer({uniforms});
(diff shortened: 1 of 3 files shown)


src/lib/layer-manager.js
```diff
@@ -115,9 +115,11 @@ export default class LayerManager {
     assert(this.context.viewport, 'LayerManager.drawLayers: viewport not set');
 
     const {uniforms} = this.context;
+    let drawCount = 0;
```
gnavvy: minor but shall we use layerIndex or layerZIndex instead? 
drawCount sounds like the accumulated number of rendering calls.

ibgreen: Good point. Ideally it should be the accumulated number of draw calls. Which could be more than 1 for some layers. But that would require layers to return draw call counts, or let model.draw to have this support built in (maybe not a bad idea).

PR #194, opened 2016-11-09
reduce flickering effect on portal page

gnavvy:


demo/src/javascripts/components/demos/trips-layer/trips-layer.js
```diff
@@ -59,11 +59,14 @@ export default class TripsLayer extends Layer {
       isIndexed: true,
       onBeforeRender: () => {
         gl.enable(gl.BLEND);
+        gl.enable(gl.POLYGON_OFFSET_FILL);
```
ibgreen: In deck.gl v3 the recommended approach is to override the Layer.draw method and do these things there, rather than use the model onBeforeRender/onAfterRender. I personally think that looks much cleaner, and will be easier to adapt to the new stateless GL support coming in luma.

PR #205, opened 2016-11-10
transparency fix and mapbox token stuff

person594:
I added the mapbox token to the secondary demo, so that I could use it.

Also, I modified the screen grid layer so that other layers can be seen behind it.

### DIFF
--- a/example/main.js
+++ b/example/main.js
@@ -41,7 +41,7 @@ import DeckGL from '../src/react/deckgl';
 // ---- Default Settings ---- //
 /* eslint-disable no-process-env */
 const MAPBOX_ACCESS_TOKEN = process.env.MAPBOX_ACCESS_TOKEN ||
-  'Set MAPBOX_ACCESS_TOKEN environment variable or put your token here.';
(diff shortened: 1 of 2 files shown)


ibgreen:
Thanks! I think the screen grid layer already supports showing other layers behind it, the minColor should be set to a color with alpha component 0. 

@Pessimistress - your call on the ScreenGridLayer.

Also you need to complete the CLA before we can merge
https://docs.google.com/a/uber.com/forms/d/e/1FAIpQLSfu-DmpJD5wYYUOxY50noHF9BWJwDbV0d2W1w45oEIbIx72RQ/viewform

Pessimistress:
Is this needed on other core layers as well? (e.g. arc, choropleth when you specify color with alpha?) Looks to me DeckGL should set default blending on initialization.

Pessimistress:
@person594 @ibgreen @Apercu Just checked again we do not by default enable blending on the gl context. 
I suggest we keep the ScreenGridLayer as is, and set the default blending mode in src/react.deckgl.js: DeckGL._onRendererInitialized. Maybe before this.props.onWebGLInitialized is called so that the user can override it.

We started accepting alpha values from color accessors in all the layers but never sufficiently tested this feature. We should update the examples to test transparency in all the layers.

PR #206, opened 2016-11-11
syntax highlight README

eklitzke:


Apercu:
In my opinion, the highlighting of bash command is not really needed as it highlights only the `&&`.

I would say to only add it on the basic js example, it looks better.

dawsonbotsford:
lgtm!

eklitzke:
I've removed the bash syntax highlighting.

Apercu:
Cool thanks!

PR #207, opened 2016-11-11
Removes HTML from markdown file

hrabbey:
Removed HTML from markdown file to allow showing mandlebrot animation.
This loses the  centering of the caption, due to lack of support in Github flavored markdown.

This should allow me and others to see the animation via browsing github.

### DIFF
--- a/docs/64-bits.md
+++ b/docs/64-bits.md
@@ -4,12 +4,10 @@ From version `3.0.0`, deck.gl begins to provide basic 64-bit math support
 in GPU shaders. 64-bit maths are used in various 64-bit layers that are provided
 with deck.gl. Please find the sample usage of them in the examples section.
 
-<div align="center">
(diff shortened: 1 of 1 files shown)


Apercu:
Thanks! But in this case, the markup is not to blame, rather the path of the image.

If you want to fix it, you will need to use the same path as in the layer docs ([ex](https://raw.githubusercontent.com/uber/deck.gl/master/docs/layers/arc-layer-64.md)), and switch back the markup to what it was initially. We prefer to keep things consistent.

Also, you'll need to sign the [Uber CLA](https://docs.google.com/a/uber.com/forms/d/e/1FAIpQLSfu-DmpJD5wYYUOxY50noHF9BWJwDbV0d2W1w45oEIbIx72RQ/viewform).

hrabbey:
Thank you for correcting my ignorance and bad assumption.  I will make the suggested edit in another branch.

PR #208, opened 2016-11-11
enabled blending by default

person594:
Enables alpha blending by default, so we can see things behind other things.

### DIFF
--- a/src/react/deckgl.js
+++ b/src/react/deckgl.js
@@ -82,6 +82,9 @@ export default class DeckGL extends React.Component {
   }
 
   @autobind _onRendererInitialized({gl, canvas}) {
+    gl.enable(GL.BLEND)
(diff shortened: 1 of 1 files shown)


src/react/deckgl.js
```diff
@@ -82,6 +82,9 @@ export default class DeckGL extends React.Component {
   }
 
   @autobind _onRendererInitialized({gl, canvas}) {
+    gl.enable(GL.BLEND)
```
Apercu: Missing semi ;)

person594: this linter is tyrranical...

Although, I have to admit this is the cleanest codebase I've ever worked on

Apercu: Everything comes at a certain price I guess 😄

PR #209, opened 2016-11-11
Correct the path of the mandelbrot image

hrabbey:
Corrected the display of the mandelbrot image by updating path, as  [recommended](https://github.com/uber/deck.gl/pull/207#issuecomment-260016592).


Apercu:
Thanks!

PR #212, opened 2016-11-12
Update nvm installation

Kerumen:
`npm install -g nvm` won't install nvm (https://www.npmjs.com/package/nvm)

And if you run `nvm install` it will use automatically the installed version, no need to run `nvm use`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,9 +29,13 @@ use cases, such as scatterplots, choropleths etc in 2 and 3 dimensions.
 npm install --save deck.gl
 ```
 Note: deck.gl has a dependency on node version 0.12 or higher. If you use an older version, you can install a node version manager like nvm and use a separate shell to install and build deck.gl
+
+Please refer to [this](https://github.com/creationix/nvm#installation) 
+to install nvm, then you can run:
 ```
-npm install -g nvm && nvm install 0.12 && nvm use 0.12
+nvm install 0.12
 ```
+It will use this node version automatically for your current shell.
 
 ## Example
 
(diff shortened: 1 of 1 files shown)


Apercu:
Hello there 😄 

Indeed `nvm`is not installed with `npm`, thanks!

I would prefer if you use [this link](https://github.com/creationix/nvm#installation) instead, add a `.nvmrc` to the repo, and drop the last explanation sentence, which would became useless.

You'll also need to sign the [Uber CLA](https://docs.google.com/a/uber.com/forms/d/e/1FAIpQLSfu-DmpJD5wYYUOxY50noHF9BWJwDbV0d2W1w45oEIbIx72RQ/viewform).

Kerumen:
I signed the CLA and updated the link.

But, correct me if I'm wrong, if I add a `.nvmrc` at the root of the repo it would only be used to build `deck.gl` (contributing to it) and not by the guys who `npm install` it right?
I understand it could be useful too but it was not was the README was about I guess.

Apercu:
Just noticed a change in the Readme I wasn't aware of, which makes it a bit confusing. Will get back to you once I know more.

Apercu:
Node `0.12+` is not needed for a basic install, only to build it from source. Having this information in the Installation section was a mistake, and adduced confusion.

There is multiple ways to switch to a compatible version, and detailing one of them was kinda out-of-scope of this project, and should be left as an exercise to the reader. A basic explanation should be enough.

That said, thanks for your report, it was definitely wrong the way it was initially.

Closing in favor of https://github.com/uber/deck.gl/commit/3f61792d2bcc2724230faa92ea4c29694a0de171.

PR #213, opened 2016-11-12
Add test code (attribute-manager)

ovekyc:
* Fix import bug in attribute-manager.spec.js
* Add attrubute manager to test
* Edit path of compare-object.spec.js

### DIFF
--- a/src/test/index.js
+++ b/src/test/index.js
@@ -22,6 +22,5 @@ import 'babel-polyfill';
 
 import './imports-spec';
 import './viewport';
-import '../lib/utils/compare-objects.spec';
 import './lib';
 import './layers';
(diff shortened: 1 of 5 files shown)


ovekyc:
Hi @Apercu 
I read this code and found that there was a import bug in attribute-manager.spec.js
And also there was no entry point for that file. So i added it.
I moved spec file to test directory and restructured some code.
So please give me any comment.

Apercu:
Hey @ovekyc, sorry for not answering.

Thanks for your contribution, I'm looking into it!

In the meantime, would you mind to [sign the CLA](https://docs.google.com/a/uber.com/forms/d/e/1FAIpQLSfu-DmpJD5wYYUOxY50noHF9BWJwDbV0d2W1w45oEIbIx72RQ/viewform)? It is required for external contributors.

Apercu:
Given that only `compare-objects` is tested from `src/lib/utils`, I think we could safely drop the `src/test/lib/utils/index.js` file and import the test directly from `src/test/lib/index.js`.

ovekyc:
Great! I droped it.

Apercu:
Nice thanks! 😺

Issue #221, opened 2016-11-15
attributes are being re-calculated when prop changes instead of only when data changes


abmai:

abmai:
seems like `dataChanged` is set to `true` on `initializeState()` and then never gets set back to `false` after

abmai:
problem seems to be here: https://github.com/uber/deck.gl/blob/master/src/lib/layer.js#L391

`setState` is only called if `willReceiveProps` is defined

shaojingli:
#222

gnavvy:
fixed in v3.0.5

PR #222, opened 2016-11-15
 Fix data update bug

shaojingli:
Fix a bug so that attribute data calculation is fired only when data has changed. #221


abmai:
lgtm. I'm guessing the code path that involves `setState` in `updateLayer` is going to be deprecated altogether eventually?

shaojingli:
Yeah... that's the plan.  updateState in the layer.js and corresponding child classes will replace setState.

PR #230, opened 2016-11-22
ScatterPlot min/max (WIP) - Closes #117

### LINKED ISSUE #117: changing dot radius scaling? (3.x)
Going to take a pass at doing some documentation for all of the layers + examples.

The logic for scaling the dots in scatterplots changed in the latest 3.x beta. Is there any info about how this works/is configured?

I know before the radius was a flat pixel size, I think now it works as meters width. Is this configurable? I can't seem to figure out how to override the scaling to do my own logic (min width/max width in pixels)

contra:
(Excuse the extraneous commits, will be squashed before merge)

Working fine on the 32 bit layer, might need a little help on the 64 bit layer - work in progress on that. Thought I would get this up as a POC to start getting feedback before I finish it off.

Does this look alright?


gnavvy:
LGTM, @shaojingli could you check on the fp64 in case I missed anything?

shaojingli:
32-bit shaders looks good to me. We may need to change the code a bit for it to work on 64-bit scatterplot. Let's merge 32-bit first

contra:
Yeah, 64 bit was only implemented up to the shader - the vars are unused there, figured @shaojingli  would be best for that, seems like the 64 bit stuff is tricky.

PR #233, opened 2016-11-25
Reflection

person594:
Added a subtle reflection effect to the map.  Only really visible on arcs and extruded choropleths, as those are the only things with height at the moment.


contra:
Got any pictures of the result?

src/react/deckgl.js
```diff
@@ -89,6 +93,8 @@ export default class DeckGL extends React.Component {
 
     // Note: avoid React setState due GL animation loop / setState timing issue
     this.layerManager = new LayerManager({gl});
+    this.effectManager = new EffectManager({gl, deckgl: this});
```
ibgreen: This line is a pretty big deal since it breaks the idea that deck.gl is independent of react. (you are giving EffectManager access to the react component). If you can just pass in layerManager that will be solved.

src/lib/effect.js
```diff
@@ -0,0 +1,27 @@
+let counter = 0;
+
+export default class Effect {
```
ibgreen: A test file for this class (`effect.spec.js`), would be trivial to write (just check that the layer symbol imports, that you can construct the class and that you can call `setNeedsRedraw`). 

It will make your PR look much more complete.

src/lib/effect-manager.js
```diff
@@ -0,0 +1,52 @@
+/* eslint-disable no-try-catch */
+
+export default class EffectManager {
```
ibgreen: A small test file (`effect-manager.spec.js`) would be highly desirable here.

src/react/deckgl.js
```diff
@@ -89,6 +93,8 @@ export default class DeckGL extends React.Component {
...
+    this.effectManager.addEffect(new ReflectionEffect());
```
ibgreen: This should be up to the application right? Should we add another `effects` property to the DeckGL component?

<DeckGL layers=[...] effects=[...]/>

ibgreen:
```
<DeckGL layers=[...] effects=[...]/>
```

person594:
Alright, I'll start working on these changes tonight.  I wasn't sure about a lot of the big-picture architecture decisions, so thanks for the input regarding that.

person594:
@contra 
Hereare some screenshots:
image

image
I'm rather happy with the effect on the extruded choropleths, although, due to the relative scale, that is rather subtle unless really zoomed in.  I'm not quite content with how the reflections look on the arc layer at the moment -- I intend to experiment with some gaussian blur, or maybe modifying the blend modes, to make the reflections there a bit less distracting.

Apercu:
Nice! In my opinion it would be even nicer if they were only on water surfaces. Dunno how though :)

person594:
@Apercu That's certainly possible, I just need some way to know where there is water and where there isn't on the underlying map.  Interesting, as this effect is drawn underneath all layers, choropleths will occlude the reflections -- if you look closely at the screenshot of the arc layer, there are no reflections on the United States, as the whole country has its own choropleth.  Providing choropleths for all landmasses would get us that effect for free, but I'm not sure how feasible that would be.

Apercu:
@person594 Ah nice, dunno the perfs implications this would have, but that's a possibility at least.

src/experimental/effects/reflection-effect/reflection-effect-fragment.glsl
```diff
@@ -18,9 +18,6 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-/* fragment shader for the grid-layer */
-#define SHADER_NAME reflection-effect-fs
```
ibgreen: Why are we removing the SHADER_NAMEs? We will still parse these (don't need glslify for that).

src/experimental/index.js
```diff
@@ -0,0 +1,2 @@
+export * from './effects';
+export * from './lib';
```
ibgreen: I don't think we should export lib here. It is not experimental right?

src/react/deckgl.js
```diff
@@ -93,7 +93,7 @@ export default class DeckGL extends React.Component {
 
     // Note: avoid React setState due GL animation loop / setState timing issue
     this.layerManager = new LayerManager({gl});
-    this.effectManager = new EffectManager({gl, deckgl: this});
+    this.effectManager = new EffectManager({gl, layerManager: this.layerManager});
     this.effectManager.addEffect(new ReflectionEffect());
```
ibgreen: Add effects from the prop.
```
for (const effect of this.props.effects) {
  this.effectManager.addEffect(effect);
}
```

src/react/deckgl.js
```diff
@@ -21,10 +21,10 @@ import React, {PropTypes} from 'react';
 import autobind from 'autobind-decorator';
 import WebGLRenderer from './webgl-renderer';
 import {LayerManager} from '../lib';
-import {EffectManager} from '../lib';
+import {EffectManager} from '../experimental/lib';
 import {GL, addEvents} from 'luma.gl';
 
-import {ReflectionEffect} from '../effects';
+import {ReflectionEffect} from '../experimental/effects';
```
ibgreen: I still think you should add a new prop instead, and let the application or example specify the effects.

In this file:
L35: ```effects: PropTypes.array,```
L47:```effects: [],```

src/experimental/index.js
person594: I recreated my original filestructure in the experimental directory.  effect.js and effect-manager.js were originally in src/lib/; now they are in src/experimental/lib/.  Does that seem like a sensible way of doing things?

src/experimental/effects/reflection-effect/reflection-effect-fragment.glsl
person594: oops

src/react/deckgl.js
person594: yeah, I intended to do that before updating the pull request; I forgot that pushing to my branch updates the pull request automatically.  I also intended to add some tests as you requested

person594:
Alright, I've made effects a property like layers are, and I've added some unit tests
Reflection is not currently enabled on any of the demo pages, but it can be by providing the react component with a property effects = [new ReflectionEffect]

Apercu:
Just a note: we removed `glslify` entirely from the project, you should use the same pattern as the other layers, as seen [here](https://github.com/uber/deck.gl/blob/master/src/layers/core/arc-layer/arc-layer.js#L95-L96) for example

ibgreen:
@person594 There was a small remaining conflict so I took over and merged this. I added reflections prop to the dev example but didn't see any effects..

person594:
@ibgreen 
Thanks, I knew there was just a small change in switching away from glslify, but it had gotten rather late over here.  Looking at what you've tried, that seems like it _should_ work; I had done something very similar in demo/ and it worked for me.  I'll try to look into what's going on tomorrow

person594:
@ibgreen I just tried the dev example on master, and the reflection effects work for me:
image

Let me know if they still don't work for you.

ibgreen:
Ah yes, works for extruded choropleth layer. Very subtle on my screen (without the map). 

Is it possible to set the reflection intensity a property on the effect?

Still can't see the reflections on the ArcLayer though.

PR #234, opened 2016-11-27
add cache support for reusing shader programs

gnavvy:
partially solves the extra overheads introduced due to recursive layer initialization when matching composited layers. (https://github.com/uber/deck.gl/blob/master/src/lib/layer-manager.js#L237)

potentially helpful for on-the-fly adding and removing layers (e.g. highlighting selection).

### DIFF
--- a/src/layers/core/arc-layer/arc-layer.js
+++ b/src/layers/core/arc-layer/arc-layer.js
@@ -98,12 +98,17 @@ export default class ArcLayer extends Layer {
       positions = [...positions, i, i, i];
     }
 
+    const program = this.state.program ? {
+      program: this.state.program
+    } : assembleShaders(gl, {
+      vs: glslify('./arc-layer-vertex.glsl'),
+      fs: glslify('./arc-layer-fragment.glsl')
+    });
+
     return new Model({
       gl,
-      ...assembleShaders(gl, {
-        vs: glslify('./arc-layer-vertex.glsl'),
-        fs: glslify('./arc-layer-fragment.glsl')
-      }),
+      id: this.props.id,
+      ...program,
       geometry: new Geometry({
         drawMode: GL.LINE_STRIP,
         positions: new Float32Array(positions)
--- a/src/layers/core/choropleth-layer/choropleth-layer.js
+++ b/src/layers/core/choropleth-layer/choropleth-layer.js
@@ -115,13 +115,17 @@ export default class ChoroplethLayer extends Layer {
   }
 
   getModel(gl) {
+    const program = this.state.program ? {
+      program: this.state.program
+    } : assembleShaders(gl, {
+      vs: glslify('./choropleth-layer-vertex.glsl'),
+      fs: glslify('./choropleth-layer-fragment.glsl')
+    });
+
     return new Model({
       gl,
       id: this.props.id,
-      ...assembleShaders(gl, {
-        vs: glslify('./choropleth-layer-vertex.glsl'),
-        fs: glslify('./choropleth-layer-fragment.glsl')
-      }),
+      ...program,
       geometry: new Geometry({
         drawMode: this.props.drawContour ? GL.LINES : GL.TRIANGLES
       }),
--- a/src/layers/core/line-layer/line-layer.js
+++ b/src/layers/core/line-layer/line-layer.js
@@ -86,13 +86,17 @@ export default class LineLayer extends Layer {
   createModel(gl) {
     const positions = [0, 0, 0, 1, 1, 1];
 
+    const program = this.state.program ? {
+      program: this.state.program
+    } : assembleShaders(gl, {
+      vs: glslify('./line-layer-vertex.glsl'),
+      fs: glslify('./line-layer-fragment.glsl')
+    });
+
     return new Model({
       gl,
       id: this.props.id,
-      ...assembleShaders(gl, {
-        vs: glslify('./line-layer-vertex.glsl'),
-        fs: glslify('./line-layer-fragment.glsl')
-      }),
+      ...program,
       geometry: new Geometry({
         drawMode: GL.LINE_STRIP,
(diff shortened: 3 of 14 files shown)


ibgreen:
@gnavyy - My read is that this is intended to prevent the same GL programs from being compiled and linked multiple times - i.e. the issue is that compiling and linking takes noticeable time. This looks like a nice optimization especially given that our shaders are quite big.

A potentially serious issue I see with reusing programs between layers, is that programs do not just hold shaders, they also hold state, in particular uniform values. These uniform values are being set in `updateState` of each layer, and can thus be different for different instances of the same layer class, depending on props.

If most of the time is spent compiling (rather than linking) the programs, we could cache the `Shader` objects instead of the `Program` objects and give each `Layer` a new Program that reuses the same compiled shaders.

Finally, I'm wondering if long-term it would make sense to build shader caching into the `assembleShaders` module. Seems it could make sense to have it handled in a different architectural component.

ibgreen:
A smaller issue is that the cache is never cleared. Should we make sure that the React `componentWillUnmount` for `DeckGL` calls a `finalize` method on the `LayerManager` in which we call `delete` on all the cached programs?

gnavvy:
Thanks for the comments.

> A potentially serious issue I see with reusing programs between layers ...

The programs are not reused between layers; they are cached and reused at the [per-layer](https://github.com/uber/deck.gl/pull/234/commits/7ca485fca803550614deeb8e4eae8364e9c99262#diff-879c914f558dab92183c5a444f97315bR303) basis. This is not ideal once we have multi-model support for one layer, but it seems to be the MVP based on the current code structure.

> ... programs do not just hold shaders, they also hold state, in particular uniform values.

This is also a reason why the program is cached at the per-layer basis, such that states and uniforms can be carried over for each layer. 

> we could cache the Shader objects instead of the Program objects and give each Layer a new Program that reuses the same compiled shaders.

Will take a second look into this approach. I started with caching the compiled shaders but they are not accessible in the layer class (the [getModel](https://github.com/uber/deck.gl/pull/234/commits/7ca485fca803550614deeb8e4eae8364e9c99262#diff-3e22ea2b65c7024100d4d235707930d2R112) function) unless we provide accessors to them from luma.gl)

> I'm wondering if long-term it would make sense to build shader caching into the assembleShaders module.

The assembleShaders is more of a utility module. We can definitely use it to handle the caching and retrieving actions but it should not own the compiled shaders. The layer-manager (which acts as a scenegraph) seems like the most feasible owner for now.

> Should we make sure that the React componentWillUnmount for DeckGL calls a finalize method on the LayerManager in which we call delete on all the cached programs?

Will do.

gnavvy:
The LayerManager is an ordinary JS class with [no event listener attached](http://stackoverflow.com/a/35120903), set the layerManager instance to null in componentWillUnmount() for explicit GC.

(We do have a `finalize` function in the Layer class, but we don't want to clear the program there as the cached program was meant to be carried over to the new Layer)

ibgreen:
> (We do have a finalize function in the Layer class, but we don't want to clear the program there as the cached program was meant to be carried over to the new Layer)

OK. This helps a little. 

FWIW, I was thinking we would delete the programs explicitly with `program.delete(). The problem with relying on GC is we don't know when it is going to happen - it depends on the state of the JavaScript "heap" and could take a long time, which means that GPU resources can be kept around for longer than we think.

But to be fair, we are not yet very rigorous with such cleanup in the rest of the code base, and programs are probably not the most resource intensive objects.

ibgreen:
> The programs are not reused between layers; they are cached and reused at the per-layer basis. This is not ideal once we have multi-model support for one layer, but it seems to be the MVP based on the current code structure.

We actually have support for multi-model layers in v3, it was part of the big refactoring push just before we published. We just haven't been using them yet. So, it looks like this change would break that. Not too happy about it, let's have another think before we merge.

And I understand now that you key on layer id rather than layer class - Yep, it should work in most normal cases.  But it also seems to me that this optimization might be less critical for single layers. Layers should not be destroyed and created so rapidly that shaders need to be frequently recompiled. Perhaps the issue you are seeing with composite layer matching is simply a bug in the lifecycle code that we could fix?

tehwalris:
Since I use multi-model layers, I implemented my own version of shader caching to improve load times [here](https://github.com/uber/deck.gl/compare/master...teralytics:feature/shader-cache?expand=1). If you like the approach I took, tell me what needs to be adjusted so that I can make a PR.

ibgreen:
@tehwalris This is very cool. Would love to see some of your multi-primitive layers in action!

As to the PR: after some back and forth we  have decided to go with per-shader caching (instead of per-program). I.e. we will cache luma.gl `Shader` objects rather than `Program` objects.

This will give us the finest granularity of compiled shader reuse and avoid issues with `Program` state conflicting across multiple layers (e.g. uniforms being set differently in two instances of the same layer due to the instances being created with different props). 

We just releases a new version of luma.gl that has a simple `ShaderCache` object and allows `Shader` objects to be passed to `Program` instead of string. We'll build this feature on that.

ibgreen:
@tehwalris Out of curiosity, do you have any ballpark numbers for what improvements you saw, before and after implementing your cache?

tehwalris:
@ibgreen 

After reading your comments, I realized my layer is not the a true multi-primitive layer. I started using deck.gl in October before that was a feature, so I handled layer compositing myself - looks like exactly what I needed though.

My composite layer draws a chain of points, connected by lines with directions marked on them. I use a LineLayer, ScatterPlotLayer and a custom MidMarkerLayer for each composite layer. For code and data related reasons I was drawing about 30 of these composite layers at once. This made my browser take about 1 minute to load, because it was compiling shaders. That's down to a couple of seconds with shader caching now.

ibgreen:
Closing this as outdated. Let's open a new PR if we want to bring this back

PR #236, opened 2016-11-29
Replace glslify with brfs

Apercu:
Improve DX by enabling hot reload of vertex and fragments shaders.
Ease usage outside of Browserify.


.gitignore
```diff
@@ -13,5 +13,5 @@ TODO
 _docs
 examples/winds/data/*.html
 node_modules
-npm-debug.log
+*.log*
```
ibgreen: Nit: Maybe this wildcard is a little bit too general?

demo/src/javascripts/components/demos/trips-layer/trips-layer.js
```diff
@@ -1,7 +1,8 @@
-import {Layer, assembleShaders} from '../../../../../../index';
 import {Model, Program, Geometry, glGetDebugInfo} from 'luma.gl';
+import {readFileSync} from 'fs';
+import {join} from 'path';
 
-const glslify = require('glslify');
+import {Layer, assembleShaders} from '../../../../../../index';
```
ibgreen: Nit: Why are we moving this import line down? I think we should follow a logical ordering when including files (e.g. our own code first, then our own modules, then standard modules).

demo/src/static/code-markdown/trips.md
```diff
@@ -66,10 +66,10 @@ export default class HeroDemo extends Component {
 trips-layer.js:
 
 ```
-import {Layer, assembleShaders} from 'deck.gl';
 import {Model, Program, Geometry, glGetDebugInfo} from 'luma.gl';
-
-const glslify = require('glslify');
+import {Layer, assembleShaders} from 'deck.gl';
```
ibgreen: Nit: Maybe have `deck.gl` before `luma.gl`? Higher level to lower level import order.

src/layers/core/arc-layer/arc-layer.js
```diff
@@ -18,11 +18,12 @@
...
 // THE SOFTWARE.
 
-import {Layer} from '../../../lib';
-import {assembleShaders} from '../../../shader-utils';
 import {GL, Model, Geometry} from 'luma.gl';
+import {join} from 'path';
+import {readFileSync} from 'fs';
 
-const glslify = require('glslify');
+import {Layer} from '../../../lib';
```
ibgreen: These are imported differently in other layers (using single import line), let's be consistent.

src/layers/core/choropleth-layer/choropleth-layer.js
```diff
@@ -18,15 +18,15 @@
...
-const glslify = require('glslify');
-
-import earcut from 'earcut';
 import flattenDeep from 'lodash.flattendeep';
 import normalize from 'geojson-normalize';
+import {readFileSync} from 'fs';
+import {join} from 'path';
+import earcut from 'earcut';
+
+import {Layer} from '../../../lib';
```
ibgreen: Ditto, order and combine imports

src/layers/fp64/arc-layer/arc-layer.js
```diff
@@ -18,13 +18,14 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
+import {GL, Model, Geometry} from 'luma.gl';
+import {join} from 'path';
+import {readFileSync} from 'fs';
+
 import {Layer} from '../../../lib';
```
ibgreen: order and combine imports

src/layers/fp64/choropleth-layer/choropleth-layer.js
```diff
@@ -18,16 +18,16 @@
...
-const glslify = require('glslify');
-
-import earcut from 'earcut';
 import flattenDeep from 'lodash.flattendeep';
 import normalize from 'geojson-normalize';
+import {readFileSync} from 'fs';
+import {join} from 'path';
+import earcut from 'earcut';
+
+import {Layer} from '../../../lib';
```
ibgreen: Ditto

demo/src/javascripts/components/demos/trips-layer/trips-layer.js
Apercu: I find it interesting to separate vendors modules from project source files, this way we know for sure what belongs is a dependency or not. Might be obvious right now because of the relative url but I would like to use absolute one in a near future, so I'm just laying to fondations.

demo/src/static/code-markdown/trips.md
Apercu: `luma.gl` being a dependency of `deck.gl`, I find it more logic to put luma first.

src/layers/core/arc-layer/arc-layer.js
Apercu: Any example? Don't see it.

.gitignore
Apercu: Hum yeah `npm-debug.log` sometimes gets suffixed with numbers, maybe `*.log.*` would be better.

src/layers/core/arc-layer/arc-layer.js
ibgreen: The second file in this diff...

contra:
Wouldn't an alternate approach be to prebuild and tell webpack users to rely on the prebuilt version? That seems to be common (mapbox-gl-js does this).

Apercu:
@contra The issue was that even if relying on the builded and minified version of `deck.gl`, `glslify` was still around and causing errors. Besides, having hot-reload of shaders is still not possible with the current `glslify` and is an big benefit.

I can look into integrating the `mapbox-glsl-optimizer` though, it looks promising, thanks!

gnavvy:
The __dirname fails to get the correct path with npm linked modules. (shader files cannot be loaded in such case)

PR #241, opened 2016-11-30
Line layer elevations

ibgreen:


src/layers/fp64/line-layer/line-layer-vertex.glsl
```diff
@@ -32,25 +33,32 @@ uniform float renderPickingBuffer;
...
 
+  // linear interpolation of source & target to pick right coord
   float segmentIndex = positions.x;
   vec2 mixed_temp[2];
 
   vec2_mix_fp64(projectedSourceCoord, projectedTargetCoord, segmentIndex, mixed_temp);
 
+  float mixedElevation =
+    mix(instanceElevations.x, instanceElevations.y, segmentIndex);
+
```
Apercu: Extra newline here

src/layers/fp64/line-layer/line-layer-64.js
```diff
@@ -60,8 +60,18 @@ export default class LineLayer64 extends Layer {
...
+        size: 4,
+        update: this.calculateInstanceSourcePositions
+      },
+      instanceTargetPositionsFP64: {
+        size: 4,
+        update: this.calculateInstanceTargetPositions
+      },
+      instanceElevations: {
+        size: 2,
+        update: this.calculateInstanceElevations
```
Apercu: Can the type here be `GL.UNSIGNED_BYTE` or maybe the elevation can be negative?

Apercu: Ok that makes sense, thanks

PR #242, opened 2016-11-30
Build on Node 6 & 7

Apercu:
Travis now test the project using the Node 6 LTS version and the latest 7, so we can have better vision of our current support and future possible issues coming in a next update.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,9 @@ addons:
     - libgl1-mesa-dri
     - libglapi-mesa
     - libosmesa6
+    - libxi-dev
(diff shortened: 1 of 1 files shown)


ibgreen:
Thanks!
- Should we use the "engines" field in package.json to specify these versions?
- Should we mention our preference in README.md?

Apercu:
Don't think we should use `engines`, as it's only advisory (they deprecated `engineStrict`), and this would also be used upon simple install I believe.

We can add a small statement in the *Node version requirement* section indeed.

PR #245, opened 2016-12-01
Warning no pickable with mouse events

Apercu:
Notify when no pickable layer has been found but one of the two mouse events has been specified.

I personally got caught and figured people could/have too, this is a simple way to address it.

### DIFF
--- a/src/react/deckgl.js
+++ b/src/react/deckgl.js
@@ -23,6 +23,7 @@ import WebGLRenderer from './webgl-renderer';
 import {LayerManager, Layer} from '../lib';
 import {EffectManager, Effect} from '../experimental';
 import {GL, addEvents} from 'luma.gl';
+import {log} from '../lib/utils';
 
 function noop() {}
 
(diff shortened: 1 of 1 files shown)


src/react/deckgl.js
```diff
@@ -91,6 +91,17 @@ export default class DeckGL extends React.Component {
     }
     this._updateLayers(this.props);
 
+    const hasEvent = this.props.onLayerClick !== noop || this.props.onLayerHover !== noop;
+    const hasPickableLayer = this.layerManager.layers.map(l => l.props.pickable).includes(true);
+
+    if (hasEvent && !hasPickableLayer) {
+      console.warn([ //eslint-disable-line
```
ibgreen: Strongly recommend using `log.once` here. If someone creates multiple DeckGL component, or destroys and recreates DeckGL this will quickly become annoying. Grep the code for examples.

ibgreen: Also I can see situations where you e.g. start out with an empty DeckGL layer list and allow the user to add layers from a palette, which would trigger this warning for correct code. Keeping the message tight (a one liner in the console) might help a little.

src/react/deckgl.js
```diff
@@ -91,6 +91,17 @@ export default class DeckGL extends React.Component {
     }
     this._updateLayers(this.props);
 
+    const hasEvent = this.props.onLayerClick !== noop || this.props.onLayerHover !== noop;
```
ibgreen: To make these checks less intrusive when reading this function, I'd add a one line comment on L93 explaining what this does.  And remove the empty line on L96 to make this a tight sub-block of code.

PR #246, opened 2016-12-01
Cleanup dist files and gitignore

Apercu:
Reorganize `.gitignore` and remove legacy references, remove `dist-demo` folder from git.


ibgreen:
About the dist-demo folder. I believe we use `git subtree` to publish it to `gh-pages`. Would that still work if we remove it from git on master? @Pessimistress @manassra

Apercu:
Not familiar with subtrees, but I don't seem to grasp why it should be tracked on `master` when it's needed only in `gh-pages`, but untracked there.

ibgreen:
> I don't seem to grasp why it should be tracked on master when it's needed only in gh-pages, but untracked there.
Good question. I am open to improvements in the way we push to `gh-pages`, our current method is rather fragile.

Apercu:
@ibgreen Any update on this one? Don't see the problem with ignoring it on `master`, since we would only have to copy its content in the root of `gh-pages` once we want to update it.

ibgreen:
> Any update on this one? 

@apercu I completely agree that it would be much preferable to not check in the demo-dist directory on master.

>Don't see the problem with ignoring it on `master`, since we would only have to copy its content in the root of `gh-pages` once we want to update it.

I would just like to understand how we would do the copy of `demo-dist` from `master` to `gh-pages`? We currently use `git-subtree`. Can `git-subtree` access files we have not committed? If it can, then go ahead and remove `demo-dist`.

Finally, note that this work should be done on `dev` instead of `master`. `demo` has moved to a new directory there with its own package.json, so it may be possible to arrange things in a better way now.

ibgreen:
Closing as further work should be done on dev.

PR #256, opened 2016-12-02
onLayerHover and onLayerClick return first match

Pessimistress:
`onLayerHover(info, pickedInfos, event)`
`onLayerHover(info, pickedInfos, event)`
The first argument is more useful if it is the first non-empty layer under the pointer instead of the topmost layer.

### DIFF
--- a/docs/using-with-react.md
+++ b/docs/using-with-react.md
@@ -108,9 +108,10 @@ Callback arguments:
 Callback - called when the mouse moves over the layers.
 
 Arguments:
-- `info` - the `info` object of the topmost
-- `pickedInfos` - an array of info objects for all visible layers that
-matched the picked coordinate, in top to bottom order.
+- `info` - the [info](#remarks) object for the topmost matched layer
+at the coordinate
+- `pickedInfos` - an array of info objects for all visible and pickable 
+layers, in top to bottom order.
 - `event` - the original MouseEvent object
 
 ##### `onLayerClick` (Function, optional)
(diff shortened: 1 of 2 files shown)


docs/using-with-react.md
```diff
@@ -108,19 +108,21 @@ Callback arguments:
 Callback - called when the mouse moves over the layers.
 
 Arguments:
-- `info` - the `info` object of the topmost
-- `pickedInfos` - an array of info objects for all visible layers that
-matched the picked coordinate, in top to bottom order.
+- `info` - the [info](#remarks) object for the topmost matched layer
+at the coordinate
+- `pickedInfos` - an array of info objects for all visible and pickable
```
ibgreen: visible and pickable => `pickable` layers (that are also `visible`)

docs/using-with-react.md
```diff
@@ -108,19 +108,21 @@ Callback arguments:
...
+layers, in top to bottom order.
 - `event` - the original MouseEvent object
 
 ##### `onLayerClick` (Function, optional)
 
 Callback - called when clicking on the layer.
 
 Arguments:
-- `info` - the `info` object of the topmost
-- `pickedInfos` - an array of info objects for all visible layers that
```
ibgreen: Same here - only `pickable` layers that are also `visible` will be considered. Let's use the exact same words.

gnavvy:
any user case we want to surface info with index < 0?

Pessimistress:
@gnavvy No preference here. I can see both model working. However if you use onLayerHover() to replace onHover() for individual layers, then it is easier to track state change (e.g. mouseout) this way.

ibgreen:
> any user case we want to surface info with index < 0?

In current model, index=-1 means nothing was selected.

ibgreen:
Additionally, the reason we include the layers with index=-1 is to make sure that the app can track when a previously selected layer becomes deselected.

PR #261, opened 2016-12-15
Add viewport prop to DeckGL. Use new `Viewport` class from viewport-mercator-project

ibgreen:
Not yet ready for merge but would like feedback:
@shaojingli @gnavvy - This enables the user to supply an arbitrary viewport.
@apercu - there is an issue with the github dependency, please take a look


src/viewport/viewport.js
```diff
@@ -6,26 +6,39 @@
...
   const hiPart = Math.fround(a);
   const loPart = a - Math.fround(a);
   return [hiPart, loPart];
 }
 
-export default class WebGLViewport extends Viewport {
+export default class WebGLViewport {
+
+  constructor({viewport}) {
+    this.viewport = viewport;
```
shaojingli: LayerManager's context.viewport property points to a WebGLViewport, which is now a wrapper around "real" viewport of type WebMercatorViewport. Some `context.viewport` references will be broken. e.g. reflection-effect.js, line 42

src/viewport/viewport.js
```diff
@@ -55,17 +68,20 @@ export default class WebGLViewport extends Viewport {
...
   } = {}) {
     // TODO: move the following line to initialization so that it's done only once
-    const positionOriginPixels = this.projectFlat(positionOrigin);
+    const positionOriginPixels = this.viewport.projectFlat(positionOrigin);
 
     const projectedPositionOrigin =
       [positionOriginPixels[0], positionOriginPixels[1], 0.0, 1.0];
-    const projections = this.getProjections();
+
+    const projections = this.viewport.getMatrices();
```
shaojingli: `const projections = this.viewport.getMatrices({});`

ibgreen:
> One thing that we might need to consider is, after this change, we will have many viewport classes that are situated both along hierarchical chain as well as ownership chain... This probably could be solve just by better naming.

@shaojingli thanks for the good review! Yes I completely agree that the `WebGLViewport` is unnecessarily confusing and should be changed. It could probably be repackaged as a function or two rather than a class, or at least renamed as you suggest. 

I'll see if I get to it before I merge. If not, since this class is completely internal to deck.gl it will be easy to change in a later diff. At the moment I am focusing on getting the externally visible Viewport classes to work correctly again after the split, and sort out intermodule dependencies.

src/viewport/viewport.js
ibgreen: > const projections = this.viewport.getMatrices({});

Fixing in `getMatrices` instead by adding default arg
@autobind getMatrices({modelMatrix = null, ...opts} **= {}**) {

src/lib/layer-manager.js
```diff
@@ -53,6 +54,23 @@ export default class LayerManager {
     Object.seal(this.context);
   }
 
+  setViewport({viewport}) {
+    assert(viewport instanceof Viewport, 'Invalid viewport');
+    const oldViewport = this.context.viewport;
+    const viewportChanged = !oldViewport || !viewport.equal(oldViewport);
```
ibgreen: > equals(oldViewport)

Eagle eyes!

src/viewport/viewport.js
ibgreen: > context.viewport references will be broken. e.g. reflection-effect.js, line 42

Good catch. Have updated but will need some testing
@person594

ibgreen: > context.viewport references will be broken. e.g. reflection-effect.js, line 42

The fix will only work for mercator views, not for general views

PR #262, opened 2016-12-16
Add React Interactions component

ibgreen:
@shaojingli @abmai @vicapow
Copying the interactions component from react-map-gl as a possible basis for supporting interaction with map-less deck.gl views. Removed the dependency on `Point` class.

This is just a proposal for @shaojingli's work on this.


ibgreen:
Closing, we have a more updated component

PR #263, opened 2016-12-16
Reflection improvements

person594:
Two main changes:

1. Reflections will no longer appear washed-out against light colored maps.  This was due to discrepancies
between how CSS and OpenGL handle alpha blending
2. Reflections now support gaussian blurring, specified by a blur parameter.  In the future I'd like to experiment with making this depth dependent -- at the moment the blurring is uniform across the reflection.  This helps makes the distinction between objects in the scene and their reflections more visible.

In addition, I added some dist files to gitignore, as they would change with every build.


src/experimental/effects/reflection-effect/reflection-effect-fragment.glsl
```diff
@@ -25,11 +25,54 @@ precision highp float;
...
+
 uniform float reflectivity;
+uniform float blur;
+
 
 varying vec2 uv;
 
+#define KERNEL_SIZE 7
+
+vec4 sample_gaussian(sampler2D tex, vec2 dim, vec2 uv, float sigma) {
```
ibgreen: Some comments would be nice before each function

src/experimental/effects/reflection-effect/reflection-effect-fragment.glsl
```diff
@@ -25,11 +25,54 @@ precision highp float;
...
+  float sigma = blur / (alpha * (1.0 - blur));
+  //let this be our standard deviation in terms of screen-widths.
+  //rewrite this in terms of pixels.
+  sigma *= float(reflectionTextureWidth);
+  
+  
+  gl_FragColor = sample_gaussian(reflectionTexture, vec2(reflectionTextureWidth, reflectionTextureHeight), vec2(uv.x, 1. - uv.y), sigma);
+  //because our canvas expects alphas to be pre-multiplied, we multiply by whole
+  //color vector by reflectivity, not just the alpha channel
+  gl_FragColor*= reflectivity;
```
ibgreen: Nit: space before operator

src/experimental/effects/reflection-effect/reflection-effect.js
```diff
@@ -67,7 +73,10 @@ export default class ReflectionEffect extends Effect {
   draw({gl, layerManager}) {
```
ibgreen: Some comments before this call would be good.

src/experimental/effects/reflection-effect/reflection-effect.js
```diff
@@ -35,7 +41,7 @@ export default class ReflectionEffect extends Effect {
         vertices: new Float32Array([0, 0, 0, 1, 0, 0, 1, 1, 0, 0, 1, 0])
```
ibgreen: Perhaps `this.model` could have a more specific name here? Like this.unitQuad or something?

Layers have some built-in assumptions about `model` not sure that effects need to.

person594:
@ibgreen Fair enough, I mostly gitignored those files for my own sanity during development.  I'll make this commit deal only with the changes to reflections, and I'll submit the .gitignore addition as a separate pull request.  We really shouldn't be tracking compiled javascript, and nothing seemed to break for my when I added those files to .gitignore

ibgreen:
I agree. Ideally the dist-demo files should be pushed to gh-pages but not to master/dev. But making changes here could break our demo-publish scripts which is why it needs separate review.

Apercu:
We will keep track of this effort in https://github.com/uber/deck.gl/pull/246

PR #265, opened 2016-12-20
Adding the viewMatrix back

shaojingli:
To make deck.gl backward compatible with v2 shaders

### DIFF
--- a/src/viewport/viewport.js
+++ b/src/viewport/viewport.js
@@ -10,7 +10,7 @@
 import Viewport, {COORDINATE_SYSTEM}
   from 'viewport-mercator-project/perspective';
 import autobind from 'autobind-decorator';
-import {mat4, vec4} from 'gl-matrix';
+import {mat4, vec3, vec4} from 'gl-matrix';
 
 export {COORDINATE_SYSTEM}
   from 'viewport-mercator-project/perspective';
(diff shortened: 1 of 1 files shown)


src/viewport/viewport.js
```diff
@@ -99,6 +99,7 @@ export default class WebGLViewport extends Viewport {
   _calculateWebGLMatrices() {
     this.glProjectionMatrix =
       new Float32Array(this.viewProjectionMatrix);
+    this.glViewMatrix = new Float32Array(this.viewMatrix);
```
ibgreen: Are these different matrices? Can we reuse the same Float32Array matrix?

src/viewport/viewport.js
```diff
@@ -79,14 +79,14 @@ export default class WebGLViewport extends Viewport {
...
-
       // Main projection matrices
       projectionMatrix: this.glProjectionMatrix,
       projectionFP64: this.glProjectionMatrixFP64,
       projectionPixelsPerUnit: this.pixelsPerMeter,
       projectionScale: this.scale,
       projectionScaleFP64: fp64ify(this.scale),
-
+      // viewMatrix is needed for backward-compatibility to deck.gl v2 shaders
+      viewMatrix: this.glViewMatrix,
```
ibgreen: Should this go under the "deprecated" comment? As we continue to introduce shader modules, we might want to phase out support for this.

src/viewport/viewport.js
shaojingli: Yep... I just found that... Let's wait a bit to land it

shaojingli:
This PR will be recreated against dev.

PR #274, opened 2017-01-06
Use Buble compiler to simplify example install and build

ibgreen:
@apercu @Pessimistress @abmai @gnavvy 

Switching compiler from babel to [buble](https://buble.surge.sh/guide/) allows us to reduce the number of dev dependencies in the example's package.json further, making the example simpler to understand and set up (it also compiles faster and into a smaller bundle).

Next step would be to remove `redux` and `react-redux` dependencies as these are overkill for a simple example.

Also would be nice if we could make react-map-gl not require immutable as peer dependency. I have some ideas about how we could do that.


ibgreen:
Closing - this PR should go towards dev.

PR #277, opened 2017-01-09
Clean up Example app

Pessimistress:
- remove redux
- remove csv
- separate data loading and rendering
- generic handler for loading/generating data
- generic handler for mouse events
- use accessors to do data formatting
- consolidate element styling to css


example/layer-examples.js
```diff
@@ -9,140 +9,144 @@ import {
...
   new ArcLayer({
-    ...props,
-    id: props.id || 'arcLayer',
-    data: props.arcs,
-    strokeWidth: props.arcStrokeWidth || 1,
+    id: 'arcLayer',
+    data: props.data,
+    getSourcePosition: f => flattenCoords(f.geometry.coordinates).shift(),
+    getTargetPosition: f => flattenCoords(f.geometry.coordinates).pop(),
+    getSourceColor: f => f.properties.DIRECTION === 'INBOUND' ? [0, 128, 200] : [0, 200, 128],
```
ibgreen: Maybe a comment about what INBOUND means in the data?

example/layer-examples.js
```diff
@@ -9,140 +9,144 @@ import {
...
+    data: props.data,
+    getPosition: d => d.COORDINATES,
+    unitWidth: 40,
+    unitHeight: 40,
+    minColor: [0, 0, 80, 0],
+    maxColor: [100, 255, 0, 128],
+    pickable: false
   });
 
 const LineLayerExample = props =>
```
ibgreen: Do we need to send props now? You only seem to be using `data`...

Pessimistress: Also sending `modelMatrix`, `onHover` and `onClick`. The extrudedChoroplethLayer uses `mapViewState` for lighting.

PR #278, opened 2017-01-10
Pure es6 code base

ibgreen:
Updates code to run untranspiled on ES6 compatible environments (tests run on Node 7 and Chrome).


package.json
```diff
@@ -81,14 +82,29 @@
...
+    "tape-catch": "^1.0.4",
+    "webpack": "^2.2.0-rc.3"
   },
   "peerDependencies": {
     "luma.gl": "3.0.0-alpha2",
     "react": "0.14.x - 15.x",
     "react-dom": "0.14.x - 15.x"
   },
   "scripts": {
+    "wp-start": "(cd examples/main && yarn && npm start)",
```
gnavvy: do we need yarn in the dep or devDep? 
I had to npm i -D it, now sure if we should assume users have it installed globally

Apercu: Thus [this line](https://github.com/uber/deck.gl/blob/master/exhibits/package.json#L5) in deck.gl exhibits

ibgreen: > do we need yarn in the dep or devDep?

> Thus this line in deck.gl exhibits

Good question. While very clever, I personally think that the line in the exhibits looks much too intimidating. It makes a simple example and makes it feel complicated - and I think it should be removed. My goal is super simple package.json files that can be understood at a glance.

We could just require that yarn is installed for development of the library, and just make a note of it in the README?

Or we could require the user to go to examples/main and install any way he wants before running start.

Apercu: Since the `package.json` at the root of the exhibits is not something intended for end-users to copy and get into their project, complexity shouldn't be a real issue, although I agree it might seems a bit obnoxious at first. The important parts are the sub-folders.

We could add this requirement, but we have to keep in mind that a lot of people tend to be quite reluctant to install again another package globally. Do we want to scare them off just because it's easier to support only yarn? Although yarn is the best in my opinion, npm still is the most used package manager out there and the official one (for now ;). Just my 2 satoshis.

PR #282, opened 2017-01-12
Updated fp64-shader tests from browserify to webpack

shaojingli:
Make our existing tests on fp64 shaders work.
Uses file loader for shader loading for now... we could determine what to do with them later...


test/fp64-test/fp64-shader.spec.js
```diff
@@ -25,10 +25,7 @@ import 'babel-polyfill';
 
 import {document, window} from 'global';
 import {Buffer, createGLContext, Program} from 'luma.gl';
-import {join} from 'path';
-import {readFileSync} from 'fs';
-
-import {assembleShaders} from 'deck.gl/shader-utils/assemble-shaders';
+import {assembleShaders} from 'deck.gl';///shader-utils/assemble-shaders';
```
ibgreen: Remove the comment

test/fp64-test/fp64-shader.spec.js
```diff
@@ -246,9 +243,10 @@ function test_float_add(gl, testName) {
   const float0_vec2 = fp64ify(float0);
   const float1_vec2 = fp64ify(float1);
   const float_ref_vec2 = fp64ify(float_ref);
+
   const program = new Program(gl, assembleShaders(gl, {
-        vs: readFileSync(join(__dirname, './shader-tests/vs_float_add.glsl'), 'utf8'),
-        fs: readFileSync(join(__dirname, './shader-tests/fs.glsl'), 'utf8'),
+        vs: require('./vs_float_add.glsl'),
```
ibgreen: This won't work in Node.js without bundling which is a pity (and probably won't play well with browserify which is a smaller concern at this point).

Unfortunately, it seems there is no perfect option to inline text files across both node and browser, so this is not a bad way.

test/fp64-test/webpack.config.test-fp64.js
```diff
@@ -0,0 +1,52 @@
...
+  resolve: {
+    alias: {
+      'deck.gl': resolve('../../src')
+    }
+  },
+
+  module: {
+    rules: [
+      {
+        // Mapbox has some unresolved fs calls
```
ibgreen: This comment is actually wrong - it is not needed for mapbox here.

It could have taken care of the readFileSync inlining, if you had added current dir to the `include`s.

A problem with babel-brfs is that we don't seem to get hot reloading of shaders, perhaps your import solution is better there.

test/fp64-test/fp64-shader.spec.js
shaojingli: Removed.

PR #283, opened 2017-01-12
ScatterplotLayer64 as subclass of Scatterplot

ibgreen:
@shaojingli @Pessimistress @gnavvy This refactors `ScatterplotLayer64` as a subclass of `ScatterplotLayer`.

* This enables a significant amount of code sharing between the layers and should make it easier to keep them in sync.
* It should now implement the same behavior as the 32 bit layer including clamping radius to props.
* It removes the custom `radiusFP64` uniform which could potentially affect the rendering issues with this layer reported by @contra - worth testing again once this lands.
* The main drawback at this point is that the 32 bit `instancePositions` attribute is not removed from the `attributeManager` and will be needlessly calculated, causing some overhead in space and time. We'll later want to add a `removeAttribute` method to the `attributeManager` class to avoid this.


contra:
@ibgreen Planning on doing this with all of the 64 bit layers? Makes a lot of sense

src/layers/fp64/scatterplot-layer/scatterplot-layer-64.js
```diff
@@ -18,96 +18,35 @@
...
-      positions.push(
-        Math.cos(Math.PI * 2 * i / NUM_SEGMENTS),
-        Math.sin(Math.PI * 2 * i / NUM_SEGMENTS),
-        0
-      );
-    }
-
-    const shaders = assembleShaders(gl, this.getShaders());
+  initializeState() {
+    // We the model and several attributes of the base layer
```
gnavvy: typo

src/layers/fp64/scatterplot-layer/scatterplot-layer-64.js
```diff
@@ -18,96 +18,35 @@
...
-      geometry: new Geometry({
-        drawMode: GL.TRIANGLE_FAN,
-        positions: new Float32Array(positions)
-      }),
-      isInstanced: true
+    // Add the 64 bit positions
+    const {attributeManager} = this.state;
+    attributeManager.addInstanced({
+      instancePositionsFP64: {size: 4, update: this.calculateInstancePositions},
+      instanceHeightsFP64: {size: 2, update: this.calculateInstanceHeightsFP64}
```
gnavvy: could you remind me why we need instanceHeights for? 
isn't instance height encoded in position.z and we might need a global layerHeight?
(occasional updates on individual z without changing x and y?)

src/layers/fp64/scatterplot-layer/scatterplot-layer-vertex.glsl
```diff
@@ -22,47 +22,49 @@
...
+  vec4_fp64(vec4(positions * radiusPixels, 0.0), vertex_pos_localspace);
 
-  vec2 vertex_pos_modelspace[4];
+  // vec2 pos_mul_radius[4];
+  // vec4_fp64(vec4(positions * radiusPixels, 0.0), pos_mul_radius);
+
+  // vec2 vertex_pos_localspace[4];
+  // vec4_scalar_mul_fp64(pos_mul_radius, radius, vertex_pos_localspace);
 
+  vec2 vertex_pos_modelspace[4];
```
gnavvy: can we add a line of comment on what vertex_pos_modelspace[4] is for?

`vec2(instanceHeightsFP64.x + 1.0, instanceHeightsFP64.y)`
Also, ^ is this + 1.0 corresponding to the comment _need to add one to elevation to show up in untilted mode_? if I recall it was due to the near clipping plane was set to 0.1, so zero z=0 will be clipped and we should have fixed for that.

src/layers/fp64/scatterplot-layer/scatterplot-layer-64.js
shaojingli: This is simply the counterpart of the z-component of instancePositions in 32-bit layers.

src/layers/fp64/scatterplot-layer/scatterplot-layer-64.js
```diff
@@ -18,96 +18,35 @@
...
-      radiusMaxPixels: this.props.radiusMaxPixels,
-      zoomRadiusFP64: this.state.zoomRadiusFP64
-    }));
-  }
-
-  getShaders() {
+  // Override the super class vertex shader
+  getShaders(id) {
     return {
       vs: readFileSync(join(__dirname, './scatterplot-layer-vertex.glsl'), 'utf8'),
```
shaojingli: Do we want to change the name of this file to "scatterplot-layer-vertex-64.glsl"?

src/layers/fp64/scatterplot-layer/scatterplot-layer-64.js
gnavvy: I see, x & y in fp64 will take over a vec4 so we need an extra component, got it.

src/layers/fp64/scatterplot-layer/scatterplot-layer-vertex.glsl
shaojingli: This is just an intermediate step to store calculated vertex coordinates. The name seems to be a bit wrong... I should have called it vertex_pos_worldspace. Will change that later.

ibgreen:
> Planning on doing this with all of the 64 bit layers? 

@contra - Yes. The only real concern here is that we need to improve our test matrix (since changing a 32 bit layer can now silently break the subclass).

Ultimately we would like 64-bit mode to just be controlled with a prop on the standard layers, but that will require us to dynamically generate shader code, so it will have to wait for the next major release.

src/layers/fp64/scatterplot-layer/scatterplot-layer-vertex.glsl
shaojingli: We can also remove commented code here

ibgreen: > Also, ^ is this + 1.0 corresponding to the comment need to add one to elevation to show up in untilted mode? if I recall it was due to the near clipping plane was set to 0.1, so zero z=0 will be clipped and we should have fixed for that.

@gnavvy Shaojing increased the far clipping plane a while ago, so we no longer need to add a fudge factor to the z coordinates.

src/layers/fp64/scatterplot-layer/scatterplot-layer-64.js
ibgreen: > could you remind me why we need instanceHeights for?

Good point. I will rename these attributes toa `instancePositions64xy` and `instancePositions64z` that will hopefully make things clearer.

ibgreen:
> However, since 64-bit layer doesn't have this Meters Mode. We need to figure out an interface for users to specify both "meters" and "pixels" as the size of the scatterplot circles.

@shaojingli - "Meters mode" should only affect the interpretation of "positions" (i.e. `instancePositions...`). Sizes should still be in meters (in the cartographic "projection" modes). At least that is the idea.

Pessimistress:
@ibgreen Why are the attributes called instancePositions when the layer is not drawing instances?

ibgreen:
> Why are the attributes called instancePositions when the layer is not drawing instances?

@Pessimistress The scatterplot layer is instanced. The circle primitive is drawn once for each instance.

Pessimistress:
@ibgreen My bad, I got linked here from a choropleth layer discussion and thought it was the ChoroplethLayer64

PR #286, opened 2017-01-12
FP64 math function improvement

shaojingli:
Fixed a bug that fp64 math occasionally fail to reach expected precision on MacBook with NVIDIA GPUs

Trade some accuracies in exp_fp64 for speed

### DIFF
--- a/src/shaderlib/fp64/math-fp64.glsl
+++ b/src/shaderlib/fp64/math-fp64.glsl
@@ -319,13 +319,13 @@ vec2 sin_taylor_fp64(vec2 a) {
   t = mul_fp64(r, INVERSE_FACTORIAL_5_FP64);
   s = sum_fp64(s, t);
 
-  r = mul_fp64(r, x);
-  t = mul_fp64(r, INVERSE_FACTORIAL_7_FP64);
-  s = sum_fp64(s, t);
+  // r = mul_fp64(r, x);
+  // t = mul_fp64(r, INVERSE_FACTORIAL_7_FP64);
+  // s = sum_fp64(s, t);
 
-  r = mul_fp64(r, x);
-  t = mul_fp64(r, INVERSE_FACTORIAL_9_FP64);
-  s = sum_fp64(s, t);
+  // r = mul_fp64(r, x);
+  // t = mul_fp64(r, INVERSE_FACTORIAL_9_FP64);
+  // s = sum_fp64(s, t);
 
   return s;
 }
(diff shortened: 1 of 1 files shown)


src/shaderlib/fp64/math-fp64.glsl
```diff
@@ -104,9 +104,11 @@ vec2 nint_fp64(vec2 a) {
 
 #if defined(NVIDIA_EQUATION_WORKAROUND) || defined(INTEL_EQUATION_WORKAROUND)
 vec2 twoSum(float a, float b) {
-  float s = (a + b) * ONE;
-  float v = (s - a);
-  float err = (a - (s - v) * ONE) * ONE + (b - v);
+  float s = (a + b);
+  float v = (s * ONE - a) * ONE;
+  float err = (a - (s - v) * ONE) * ONE * ONE * ONE + (b - v);
```
ibgreen: This is not exactly intuitive. How is someone else going to be able to maintain this? I think needs a comment, (maybe a url to an explanation if you have one, or at least an explanation of what the issues are).

If this is just trial and error, maybe document what has been tried and failed?

src/shaderlib/fp64/math-fp64.glsl
```diff
@@ -120,9 +122,9 @@ vec2 twoSum(float a, float b) {
 
 #if defined(NVIDIA_EQUATION_WORKAROUND) || defined(INTEL_EQUATION_WORKAROUND)
 vec2 twoSub(float a, float b) {
-  float s = (a - b) * ONE;
-  float v = (s - a);
-  float err = (a - (s - v) * ONE) * ONE - (b + v);
+  float s = (a - b);
+  float v = (s * ONE - a) * ONE;
+  float err = (a - (s - v) * ONE) * ONE * ONE * ONE - (b + v);
```
ibgreen: Same here. There should perhaps be a small README on the GPU workaround package, with sections on Intel, Nvidia etc. We could reference that here.

src/shaderlib/fp64/math-fp64.glsl
```diff
@@ -247,16 +249,17 @@ vec2 exp_fp64(vec2 a) {
   p = mul_fp64(p, r);
   t = mul_fp64(p, INVERSE_FACTORIAL_5_FP64);
 
-  s = sum_fp64(s, t);
-  p = mul_fp64(p, r);
-  t = mul_fp64(p, INVERSE_FACTORIAL_6_FP64);
+  // s = sum_fp64(s, t);
```
ibgreen: Drop the comments below or add another comment line in front of each comment indicating why we are keeping them.

src/shaderlib/fp64/math-fp64.glsl
```diff
@@ -218,7 +220,7 @@ vec2 sqrt_fp64(vec2 a) {
 }
 
 vec2 exp_fp64(vec2 a) {
-  const float k = 512.0;
+  const float k = 16.0;
```
ibgreen: Maybe a comment that indicates the significance of this constant and possible values. And/or a longer, self explanatory variable name.

src/shaderlib/fp64/math-fp64.glsl
shaojingli: I'll make the comments clearer. 

It's not trial and error. The principle here is to expand the final express to a polynomials of ONE and make each coefficient contains at most one a and one b.

src/shaderlib/fp64/math-fp64.glsl
shaojingli: I'll change this and define a separate macro.

PR #288, opened 2017-01-12
disable blending when render picking buffer

heshan0131:


src/lib/pick-layers.js
```diff
@@ -29,6 +29,11 @@ export function pickLayers(gl, {
   }, () => {
     let zOrder = 0;
```
ibgreen: I'd move this declaration down closer to the loop (i.e. keep the blend stuff first)

src/lib/pick-layers.js
```diff
@@ -29,6 +29,11 @@ export function pickLayers(gl, {
...
 
+    const blend = gl.getParameter(GL.BLEND);
```
ibgreen: I'd keep these three lines tightly together without empty lines, and put the comment as the first of the three lines

Maybe change comment to 
    // disable blending while picking, restore when done

ibgreen:
> We should check the blending mode using gl.isEnabled()

Ah yes, it should be `gl.isEnabled` not `gl.getParameter`. Good catch!

ibgreen:
@heshan0131 Just click to sign the CLA in the "checks" section above and then go ahead and merge.

PR #289, opened 2017-01-14
Lint & beautify demo

### LINKED ISSUE #255: Lint demo dir and eslint disables

Apercu:
Closes #255.


.eslintrc
```diff
@@ -8,7 +8,8 @@
     "eslint-config-uber-es2015"
   ],
   "rules": {
-    "guard-for-in": 0
+    "guard-for-in": 0,
+    "callback-return": 0
```
ibgreen: Do we need to disable `callback-return` here in the root? This is related to node.js style callbacks which we don't use a lot.

Seems related to the demo, could we put it in a comment in the offending file or in a local .eslint for the demo?

Just a nit, the motivation being to keep our top level eslintrc small.

demo/src/javascripts/actions/app-actions.js
```diff
@@ -1,35 +1,39 @@
 import {request, text} from 'd3-request';
+
 import {StreamParser} from '../utils/worker-utils';
 
-export function loadContent(filename) {
+const loadContentSuccess = (name, content) => {
```
ibgreen: I personally find using `function` keyword clearer than arrow syntax for plain function declaration.  But I guess it is the trend in redux apps to use arrow notation for everything, so the change might be appropriate here in the `demo` folder. Up to @Pessimistress to decide.

demo/src/javascripts/components/demos/arc.js
```diff
@@ -108,16 +111,14 @@ export default class ArcDemo extends Component {
     selectedFeature = selectedFeature || features[362];
 
     const {flows, centroid, name} = selectedFeature.properties;
-    const arcs = [];
-
-    for (let toId in flows) {
+    const arcs = Object.keys(flows).reduce((acc, toId) => {
       const f = features[toId];
-      arcs.push({
+      return [...acc, {
```
ibgreen: While elegant, I think there could be a big performance difference here as the new code creates a new array every time. I would consider restoring the old code.

demo/src/javascripts/components/home.js
```diff
@@ -1,16 +1,14 @@
+/* global window */
```
ibgreen: Nit: I would probably have declared `requestAnimationFrame` and `cancelAnimationFrame` as globals instead of `window`. More similar to how we import symbols.

demo/src/javascripts/components/map.js
```diff
@@ -24,21 +23,20 @@ class Map extends Component {
   }
 
   _loadDemo(demo, useTransition) {
-    const {loadData, useParams, updateMap} = this.props;
     const DemoComponent = Demos[demo];
 
     if (DemoComponent) {
-      loadData(demo, DemoComponent.data);
-      useParams(DemoComponent.parameters);
+      this.props.loadData(demo, DemoComponent.data);
```
ibgreen: +1 Agree it is easier to understand this when calling functions.

demo/src/javascripts/components/page.js
```diff
@@ -39,13 +41,13 @@ class Page extends Component {
     if (typeof content !== 'object') {
       content = {content};
     }
-    const {demo, ...docs} = content;
-    const {loadContent} = this.props;
+
+    const {...docs} = content;
```
ibgreen: Isn't this equal to `const docs = content`?

demo/src/javascripts/components/page.js
```diff
@@ -77,28 +79,28 @@ class Page extends Component {
     const {tabs, mapFocus} = this.state;
     const activeTab = location.query.tab || Object.keys(tabs)[0];
 
-    return Object.keys(tabs).map(tabName => {
-      const tab = tabs[tabName];
+    return Object.keys(tabs).map(_tabName => {
+      const _tab = tabs[_tabName];
```
ibgreen: Nit: I see the problem, but underscore always looks awkward. Maybe `selectedTabName` and `selectedTab`, or `activeTabName` at the top would look better.

demo/src/javascripts/components/demos/arc.js
Pessimistress: Using `[...array]` in `reduce` has a big perf hit (`O(N^2)`). 
Babel transforms the following:
```
arr.reduce((acc, d) => [...acc, d], []);
```
To:
```
function _toConsumableArray(arr) { if (Array.isArray(arr)) { for (var i = 0, arr2 = Array(arr.length); i < arr.length; i++) { arr2[i] = arr[i]; } return arr2; } else { return Array.from(arr); } }

arr.reduce(function (acc, d) {
  return [].concat(_toConsumableArray(acc), [d]);
}, []);
```

.eslintrc
Apercu: No we don't I agree, I was just really annoyed by this rule.

demo/src/javascripts/actions/app-actions.js
Apercu: @Pessimistress 🙏

demo/src/javascripts/components/demos/arc.js
Apercu: Although I agree with you both, we have to keep in mind this is only a demo 😃 

I'll make the change back, but this kinda makes me think about the diff I did to remove the `switch` statement and replace it by a hashmap. Although less elegant, the solution had the advantage to be `O(1)`.

My question being what's the line where performance becomes more important than code prettiness?

demo/src/javascripts/components/home.js
Apercu: Hum I would rather declare only one, the less the better. Do you have a strong opinion on this?

demo/src/javascripts/components/page.js
Apercu: Haha yes 😫

demo/src/javascripts/components/page.js
Apercu: Yep, I was really annoyed:

> This is javascript! I know what scopes are! :rage1:

demo/src/javascripts/actions/app-actions.js
Pessimistress: I don't really have a preference, what you're doing here looks fine to me.

demo/src/javascripts/components/demos/arc.js
Pessimistress: It's not usually a problem except in data processing. N in this case can be up to the 3k range so it's reasonable to prioritize performance. If N goes to 10k+ there will be a quite visible freeze.

ibgreen:
@apercu - How hard would it be to rebase this PR to `dev`? It might not be that interesting to land it on `master` (which is essentially in maintenance mode right now), just to have to merge and rebase it into dev...

PR #290, opened 2017-01-16
Set Viewport project/unproject default to screen coordinates

Pessimistress:
- Match default value of option `topLeft` with JSDoc
- Fix bug where y is flipped in `ScreenGridLayer`
- Fix bug where picking info produce wrong `latLng` values
- Add tests

@ibgreen @heshan0131


Pessimistress:
`WebMercatorViewport.unproject` does not seem to pass my tests. I will open an issues for this.

PR #293, opened 2017-01-17
New autocalculating "accessor" option in AttributeManager

ibgreen:
@gnavvy @Pessimistress @shaojingli

* Allows us to remove the `calculateInstance` functions for most instanced layers.

* Also some webpack config cleanup.

Based on an idea from @apercu


ibgreen:
@gnavvy - I removed the change to `ScatterplotLayer`. I'm fine with not touching layers for v3.Next, but would still like to get this change in as basis for additional work on `attributeManager` and `updateTriggers`.

PR #296, opened 2017-01-17
Refactor 64-bit layers to make them subclasses of correspondent 32-bit layers

shaojingli:
@ibgreen @gnavvy @Pessimistress


src/layers/fp64/arc-layer/arc-layer-64.js
```diff
@@ -18,104 +18,51 @@
...
       },
-      instanceTargetColors: {
+      instanceTargetPositions64: {
         size: 4,
-        type: GL.UNSIGNED_BYTE,
-        update: this.calculateInstanceTargetColors
+        update: this.calculateInstanceTargetPositions64
       }
+      // Reuse from base class
+      // instanceSourceColors: {
```
ibgreen: Maybe just mentioning the attributes we reuse is enough - i.e. a one line comment

PR #297, opened 2017-01-18
Cherry pick #286 to fix 64-bit math and reduce its accuracy for better efficiency

shaojingli:


shaojingli:
@Pessimistress @gnavvy when you try to implement 64-bit path layer for your other project, you can try this version of 64-bit math that trades some accuracy with speed.

PR #302, opened 2017-01-23
make compute[Picking]Colors fast

### LINKED ISSUE #301: Low performance from isFlattenable / baseFlatten
Deck.gl version: 3.1.1
Chrome version: 55
System: 2016 12" Macbook

I'm using deck.gl to draw a zoning map of San Francisco. The dataset is not large, only 1600 polygons and a total of 200k vertices.

With picking enabled, I'm getting a blazing 3 FPS.

It looks like that's totally CPU bound. GPU is fine. Most of the scripting time is spent in isFlattenable and baseFlatten:

image

Thoughts?

dcposch:
reduce code duplication, reduce memory allocations, update a Uint8Array directly rather than creating tons of JS arrays and calling flatten() each time. since these functions must run every single frame when picking is enabled and the user is moving their mouse quickly, this makes a big difference.

fixes #301 

see that bug for details on how i measured the following frame rates:

### before: 3 FPS in debug mode

image

CPU bound, with most of the time spent in `isFlattenable` and `baseFlatten`, which are both called from `flatten`.

### after: 20 FPS in debug mode

image

flatten is gone, and most of the time is now spent in `getError`, which goes away once we disable debug mode. (i left it on to get an apples to apples comparison.)

**tldr; mousing over my choropleth map with 1600 regions and 200k total vertices is smooth now**

### DIFF
--- a/src/layers/core/choropleth-layer/choropleth-layer.js
+++ b/src/layers/core/choropleth-layer/choropleth-layer.js
@@ -170,43 +170,83 @@ export default class ChoroplethLayer extends Layer {
 
   calculateColors(attribute) {
     const {data: {features}, getColor} = this.props;
+    const {choropleths} = this.state;
 
-    const colors = this.state.choropleths.map(
-      (choropleth, choroplethIndex) => {
-        const feature = features[choropleth.featureIndex];
-        const color = getColor(feature) || DEFAULT_COLOR;
-
-        if (isNaN(color[3])) {
-          color[3] = DEFAULT_COLOR[3];
-        }
-
-        return choropleth.map(polygon =>
-          polygon.map(vertex => color)
-        );
+    calculateAttributeColors(attribute, choropleths, 4, (choropleth) => {
+      const feature = features[choropleth.featureIndex];
+      const color = getColor(feature) || DEFAULT_COLOR;
+      if (isNaN(color[3])) {
+        color[3] = DEFAULT_COLOR[3];
       }
-    );
-
-    attribute.value = new Uint8Array(flattenDeep(colors));
+      return color;
+    });
   }
 
   // Override the default picking colors calculation
   calculatePickingColors(attribute) {
+    const {choropleths} = this.state;
+    const {drawContour} = this.props;
 
-    const colors = this.state.choropleths.map(
-      (choropleth, choroplethIndex) => {
-        const {featureIndex} = choropleth;
-        const color = this.props.drawContour ? [0, 0, 0] : [
-          (featureIndex + 1) % 256,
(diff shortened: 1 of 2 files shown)


dcposch:
@shaojingli @ibgreen

dcposch:
https://dcpos.ch/yimby/zoning/

src/layers/core/choropleth-layer/choropleth-layer.js
```diff
@@ -170,43 +170,83 @@ export default class ChoroplethLayer extends Layer {
...
+
+  attribute.value = arr;
+}
 
-    attribute.value = new Uint8Array(flattenDeep(colors));
+/**
+ * Takes a Uint8Array. Returns the same array if it's long enough to fit `length` bytes, otherwise
+ * returns a new Uint8Array with the elements copied over from the old array.
+ */
+function resizeIfNeeded(arr, length) {
```
ibgreen: Not a blocker since this already improves perf, but probably worth precalculating array length - at least that's what the new code on `dev` does.

src/layers/core/choropleth-layer/choropleth-layer.js
```diff
@@ -170,43 +170,83 @@ export default class ChoroplethLayer extends Layer {
...
+        (featureIndex + 1) % 256,
+        Math.floor((featureIndex + 1) / 256) % 256,
+        Math.floor((featureIndex + 1) / 256 / 256) % 256];
+      return color;
+    });
+  }
+}
+
+function calculateAttributeColors(attribute, choropleths, colorComponents, colorFn) {
+  if (colorComponents !== 3 && colorComponents !== 4) {
```
ibgreen: Is this being checked twice?

ibgreen:
@dcposch The possible concerns with merging this PR into main would be that:

1. this is already being addressed in a more comprehensive way in the next deck.gl version (the new `PolygonLayer` with the `PolygonTesselator` class that minimizes unnecessary copying and array creation). 
2. there are several `ChoroplethLayers` (32 bit, 64 bit, extruded, enhanced) - doing this for one of them will bring them further from each other.
3. There has been work done on `dev` on making `ChoroplethLayer`s work with `Immutable.js` data - which requires care when iterating over polys and verts. This will likely interfere with that.

However, since the next release is some time out, it might still be reasonable to merge this in as a temporary fix on master. For now please look at the review comments.

ibgreen:
Closing.
* ChoroplethLayers are deprecated in v4 and will likely be removed in v5 so we don't want to change them unless for a critical bugfix. 
* Instead we are trying to incorporate the idea here into the v4 documentation and a sample layer to show how layers can be subclassed to do things like this.

PR #304, opened 2017-01-24
added hexagon layer

heshan0131:
added 3d hexagon layer. add yarn requirement to README.md


README.md
```diff
@@ -53,9 +53,18 @@ or browse directly the [docs folder](./docs).
 
 #### Node Version Requirement
 
-Building deck.gl from source has a dependency on node `0.12` or higher.
+Building deck.gl from source has a dependency on node `4.0` or higher.
 Either upgrade to a newest version, or install something like [nvm](https://github.com/creationix/nvm).
 
+#### Install yarn
+deck.gl uses [yarn] (https://www.npmjs.com/package/yarn) to manage packages.
+To develop deck.gl, install yarn with brew
```
ibgreen: This is mac specific. To avoid confusing external users, maybe best to also provide a link https://yarnpkg.com/en/docs/install

src/layers/core/hexagon-layer/hexagon-layer-vertex.glsl
```diff
@@ -0,0 +1,91 @@
...
+// a magic number to scale elevations so that 1 unit roughly equals 100 meter
+float elevationScaler = 6000.;
+
+void main(void) {
+
+  // rotate primitive position and normal
+  mat2 rotationMatrix = mat2(cos(angle), -sin(angle), sin(angle), cos(angle));
+
+  vec3 rotatedPositions = vec3(
+    vec2(rotationMatrix * positions.xz).x,
```
ibgreen: Use a temporary to avoid rotating twice?

src/layers/core/hexagon-layer/hexagon-layer-vertex.glsl
```diff
@@ -0,0 +1,91 @@
...
+    positions.y,
+    vec2(rotationMatrix * positions.xz).y
+  );
+  vec3 rotatedNormals = vec3(
+    vec2(rotationMatrix * normals.xz).x,
```
ibgreen: Ditto?

src/layers/core/hexagon-layer/hexagon-layer-vertex.glsl
```diff
@@ -0,0 +1,91 @@
...
+uniform mat4 worldInverseTransposeMatrix;
+
+// Custom uniforms
+uniform float opacity;
+uniform vec3 invisibleColor;
+uniform float radius;
+uniform float angle;
+uniform float enable3d;
+
+// Lighting constants, this can also be set via uniform
```
ibgreen: Is this comment correct? I do not see any uniforms...

src/layers/core/hexagon-layer/hexagon-layer-vertex.glsl
```diff
@@ -0,0 +1,91 @@
...
+const vec3 pointColor = vec3(1., 1., 1.);
+const vec3 pointSpecularColor = vec3(0.6, 0.6, 0.6);
+const float shininess = 3.;
+const float pointLightAmbientCoefficient = 0.8;
+
+// Result
+varying vec4 vColor;
+
+// a magic number to scale elevations so that 1 unit roughly equals 100 meter
+float elevationScaler = 6000.;
```
ibgreen: Does `project_scale` not work here?

If not maybe use a #define `ELEVATION_SCALE` to make it clear that this is a constant.

src/layers/core/hexagon-layer/hexagon-layer.js
```diff
@@ -0,0 +1,221 @@
...
+   * DeckGL calls initializeState when GL context is available
+   * Essentially a deferred constructor
+   */
+  initializeState() {
+    const {gl} = this.context;
+    this.setState({model: this.getModel(gl)});
+
+    const {attributeManager} = this.state;
+    attributeManager.addInstanced({
+      instancePositions: {size: 3, update: this.calculateInstancePositions},
```
ibgreen: Note for upcoming API audit:
In this case it is possible that splitting into `positions` and `height` (getPositions and getHeight) would make more sense.
 E.g. being able to animate height as well as color for a set of hexagons.

src/layers/core/hexagon-layer/hexagon-layer.js
```diff
@@ -0,0 +1,221 @@
...
+  id: 'hexagon-layer',
+  dotRadius: 1,
+  elevation: 0,
+  enable3d: true,
+  invisibleColor: {r: 0, g: 0, b: 0},
+  getCentroid: x => x.centroid,
+  getColor: x => x.color,
+  getAlpha: x => 255,
+  getElevation: x => x.elevation
+};
```
ibgreen: Let's list all props, even those with empty defaults, e.g.
    vertices: null

src/layers/core/hexagon-layer/hexagon-layer.js
```diff
@@ -0,0 +1,221 @@
...
+    this.setUniforms({
+      // Calculate angle that the perpendicular hexagon vertex axis is tilted
+      angle: Math.acos(dx / dxy) * -Math.sign(dy) + Math.PI / 2,
+      // Allow user to fine tune radius
+      radius: dxy / 2 * Math.min(1, dotRadius)
+    });
+  }
+
+  getCylinderGeometry(radius) {
+    return new CylinderGeometry({
```
ibgreen: For performance with large data sets in 2D, it could be valuable to support an `extruded` prop, and if false, generate a simpler non-extruded hexagon primitive. For now it would need a simple [0, 1, 0] normal to work with the lighting code in the shader, until we make that optional as well.

examples/main/layer-examples.js
```diff
@@ -160,6 +160,20 @@ const ExtrudedChoroplethLayer64Example = {
...
 
+const HexagonLayerExample = {
+  layer: HexagonLayer,
+  props: {
+    id: 'hexagonLayer',
+    data: dataSamples.hexagons,
+    hexagonVertices: dataSamples.hexagons[0].vertices,
+    getColor: h => [48, 128, h.value * 255],
+    getElevation: h => h.value,
+    enable3d: true,
```
ibgreen: Note for API audit: we should define the name of this prop across layers (`enable3d` vs. `extruded` vs...) and define expected semantics (generate simpler geometry, turn off lighting etc)

src/layers/core/hexagon-layer/hexagon-layer.js
```diff
@@ -0,0 +1,221 @@
...
+  getShaders() {
+    const vertex = readFileSync(join(__dirname, './hexagon-layer-vertex.glsl'), 'utf8');
+    const lighting = readFileSync(join(__dirname, './lighting.glsl'), 'utf8');
+    const picking = readFileSync(join(__dirname, './picking.glsl'), 'utf8');
+    const vs = picking.concat(lighting).concat(vertex);
+
+    return {
+      vs,
+      fs: readFileSync(join(__dirname, './hexagon-layer-fragment.glsl'), 'utf8')
+    };
```
ibgreen: About to merge a PR where you can just add 

    modules: ['lighting']

to the getShaders object.

src/layers/core/hexagon-layer/hexagon-layer.js
```diff
@@ -0,0 +1,221 @@
...
+    super.draw({uniforms: Object.assign({}, {viewMatrix}, uniforms)});
+  }
+
+  calculateInstancePositions(attribute) {
+    const {data, getCentroid, getElevation} = this.props;
+    const {value, size} = attribute;
+    let i = 0;
+    for (const object of data) {
+      const [lon, lat] = getCentroid(object);
+      const elevation = getElevation(object);
```
ibgreen: Note for API audit:
- Another reason to split `positions` into two attributes? I.e. to maintain a one-to-one attribute / accessor mapping.
- Also wonder if `getCentroid` should be `getPosition` (if vertices prop is supplied it may not be the centroid) and if `getElevation` should be `getHeight`

src/layers/core/hexagon-layer/hexagon-layer.js
```diff
@@ -0,0 +1,221 @@
...
+  calculateInstanceColors(attribute) {
+    const {data, getColor, getAlpha} = this.props;
+    const {value, size} = attribute;
+    let i = 0;
+    for (const object of data) {
+      const color = getColor(object);
+      value[i + 0] = color[0];
+      value[i + 1] = color[1];
+      value[i + 2] = color[2];
+      value[i + 3] = getAlpha(object);
```
ibgreen: API audit - most layers just handle this by allowing `getColor` return a 4 element array.

PR #308, opened 2017-01-24
GeojsonLayer cleanup

ibgreen:
@gnavvy Updates per review


src/layers/core/geojson-layer/geojson.js
```diff
@@ -0,0 +1,95 @@
...
+    case 'LineString':
+      lineFeatures.push(feature);
+      break;
+    case 'MultiLineString':
+      // Break multipolygons into multiple polygons with same properties
+      Container.forEach(coordinates, path => {
+        lineFeatures.push({geometry: {coordinates: path}, properties, feature});
+      });
+      break;
+    case 'Polygon':
```
gnavvy: So we don’t add lineFeatures for Polygon and MultiPolygon types now?
I don’t see the polygonFeatures being used with the PathLayer, how do we generate the contour then?

src/layers/core/geojson-layer/geojson.js
```diff
@@ -0,0 +1,95 @@
...
+        pointFeatures.push({geometry: {coordinates: point}, properties, feature});
+      });
+      break;
+    case 'LineString':
+      lineFeatures.push(feature);
+      break;
+    case 'MultiLineString':
+      // Break multipolygons into multiple polygons with same properties
+      Container.forEach(coordinates, path => {
+        lineFeatures.push({geometry: {coordinates: path}, properties, feature});
```
gnavvy: why do we need **feature** together with the **geometry** and **properties** at the same time?

src/layers/core/geojson-layer/geojson-layer.js
```diff
@@ -155,64 +120,116 @@ export default class GeoJsonLayer extends Layer {
...
+        wireframe: true,
+        onHover: this._onHoverPolygon,
+        onClick: this._onClickPolygon
       }));
-
-    const pathLayer = pathLayerData &&
-      new PathLayer(Object.assign({}, this.props, {
-        id: `${id}-path-sublayer`,
-        data: pathLayerData,
-        getPaths: this._extractPaths,
```
gnavvy: ref

src/layers/core/geojson-layer/geojson-layer.js
```diff
@@ -155,64 +120,116 @@ export default class GeoJsonLayer extends Layer {
...
+        onHover: this._onHoverPolygon,
+        onClick: this._onClickPolygon
       }));
+    }
 
-    return [polygonLayer, pathLayer, pointLayer].filter(Boolean);
+    const lineLayer = drawPolygons && drawLines && new PathLayer(Object.assign({}, this.props, {
+      id: `${id}-line-paths`,
+      data: lineFeatures,
+      getPaths: f => Container.get(f, 'geometry.coordinates'),
```
gnavvy: LineString, Polygon, MultiLineString, and MultiPolygon types in geoJson store the coordinates differently, I don't see why the PathLayer can use a unified get function, nor should it be aware of how the coordinates are stored (that's why we had a _extractPaths function previously, marked as ref on the left)

src/layers/core/geojson-layer/geojson.js
ibgreen: Yes good question. This change was related to picking - my assumption here was that we want the `onHover` and `onClick` methods on `GeoJsonLayer` to be able to include the original `feature` object from the geojson blob as part of the `info` item in those callbacks. 

If we e.g. just wanted to return `properties` we could drop this.

Regardless of what we decide, I should update the layer docs and add comments.

src/layers/core/geojson-layer/geojson.js
ibgreen: Yes, this is related to the new props, `drawPoints`, `drawLines`, `drawPolygons` and `fillPolygons`. To be able to control these separately, we now generate two `PathLayers`: once for lines and one for polygons. My assumption was that I could then just reuse the polygon geometry for the second path layer. Based on your next comment, that may not be true, if so we could add a `polygonOutlineFeatures`.

src/layers/core/geojson-layer/geojson-layer.js
ibgreen: Understood. A problem with heavily transforming the data in the GeoJson layer (apart from possible performance issues) is that it gets harder to track the connection to the original features, e.g. during picking as discussed above. I will clearly have to give this more thought.

src/layers/core/geojson-layer/geojson.js
gnavvy: the geometry and properties were from the original feature a few lines above, and they should contain all the info within the feature.

src/layers/core/geojson-layer/geojson-layer.js
gnavvy: I agree we should try to avoid unnecessary data transform when possible, my hunch is that the performance loss due to array repacking won't be as salient (what was the absolute delta time for the 5x speedup in the bench test?)

Regarding picking, I agree at the geoJson layer level, we should keep the original feature, but it seems strange that the general polygon and path layer need to understand geoJson and surface a geoJson feature when hovering. I will think about it too and come back in the later comment.

ibgreen: > what was the absolute delta time for the 5x speedup in the bench test?

Creating geometry for all SF zip codes went from 0.1s to 0.02s
- It's a noticeable difference in how quickly the layer pops up when clicking on the layer, and that is not a super large data set. 
- That said, a lot of the speedup (maybe 75%) was in the geometry generation, not in avoiding this particular copy. 

> the general polygon and path layer need to understand geoJson and surface a geoJson feature when hovering

Agreed, those layers shouldn't understand geojson. The way I wanted to make them understand geojson was through having GeoJson layer instantiate them with accessors that understand geojson. I'll review it again to make sure there is no geojson knowledge in `PolygonLayer` and `PathLayer`

PR #312, opened 2017-01-25
Update all layers to use new defaultProps mechanism. Add docs.

ibgreen:
Add default props to all layers

### DIFF
--- a/src/layers/core/icon-layer/icon-layer.js
+++ b/src/layers/core/icon-layer/icon-layer.js
@@ -22,6 +22,28 @@ import {GL, Model, Geometry, Texture2D, loadTextures} from 'luma.gl';
 import {readFileSync} from 'fs';
 import {join} from 'path';
 
+/*
+ * @param {object} props
+ * @param {Texture2D | string} props.iconAtlas - atlas image url or texture
+ * @param {object} props.iconMapping - icon names mapped to icon definitions
+ * @param {object} props.iconMapping[icon_name].x - x position of icon on the atlas image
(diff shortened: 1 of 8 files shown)


src/layers/core/icon-layer/icon-layer.js
```diff
@@ -22,6 +22,28 @@ import {GL, Model, Geometry, Texture2D, loadTextures} from 'luma.gl';
...
+ * @param {object} props.iconMapping[icon_name].mask - whether icon is treated as a transparency
+ *   mask. If true, user defined color is applied. If false, original color from the image is
+ *   applied. Default to false.
+ * @param {number} props.size - icon size in pixels
+ * @param {func} props.getPosition - returns anchor position of the icon, in [lng, lat, z]
+ * @param {func} props.getIcon - returns icon name as a string
+ * @param {func} props.getScale - returns icon size multiplier as a number
+ * @param {func} props.getColor - returns color of the icon in [r, g, b, a]. Only works on icons
+ *   with mask: true.
+ */
```
gnavvy: http://usejsdoc.org/howto-es2015-classes.html
is there a specific reason we want to move the jsDoc here after the import and before the consts?

ibgreen: > is there a specific reason we want to move the jsDoc here after the import and before the consts?

Yes, the JSDoc of a class is associated with the class constructor. When I converted the class to the defaultProps, there was no longer a need for a constructor. Since most layers document their props in markdown, rather than in the code, it didn't seem like a big issue. I just moved the docs up next to the props for now. They can probably be deleted once we have reviewed the docs.

PR #315, opened 2017-01-25
Fix project_fp64 issue under linux + nvidia

shaojingli:
Fix a bug in the project_fp64 that only shows up in linux + Nvidia;
Added unit test for radians_fp64;
@Pessimistress


shaderlib/project64/project64.glsl
```diff
@@ -5,7 +5,11 @@ uniform vec2 projectionFP64[16];
 
 void mercatorProject_fp64(vec4 lnglat_fp64, out vec2 out_val[2]) { //longitude: lnglat_fp64.xy; latitude: lnglat_fp64.zw
 
+#if defined(NVIDIA_FP64_WORKAROUND)
```
shaojingli: This is the line that actually solves the problem...

PR #321, opened 2017-01-26
use the test data in the geojson example

gnavvy:
Point & multiPoint still break, looking at it now.


gnavvy:
duplicated, abandon.

PR #324, opened 2017-01-26
Fix mutation issue in defaultProps handling.

ibgreen:


ibgreen:
@Pessimistress thanks for spotting this.

PR #325, opened 2017-01-26
Simple parse colors implementation with benchmarking

ibgreen:
@gnavvy - this is an experimental color converter implementation based on our discussion. Benchmarking the color conversions in isolation looks good, although we need to integrate them into attribute updaters and measure the impact.

### DIFF
--- a/src/lib/utils/color.js
+++ b/src/lib/utils/color.js
@@ -1,0 +1,39 @@
+
+// Parse array or string color
+export function parseColor(color) {
+  if (Array.isArray(color)) {
+    if (color.length === 3) {
+      return [color[0], color[1], color[2], 255];
+    }
+    return color;
+  }
+  return parseHexColor(color);
+}
+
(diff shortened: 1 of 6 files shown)


src/lib/utils/color.js
```diff
@@ -0,0 +1,39 @@
...
+    array[0] = value / 65536;
+    array[1] = (value / 256) % 256;
+    array[2] = value / 65536;
+    array[3] = 255;
+  } else if (color.length === 9) {
+    const value = parseInt(color.substring(1), 16);
+    array[0] = value / 16777216;
+    array[1] = (value / 65536) % 256;
+    array[2] = (value / 256) % 256;
+    array[3] = value % 256;
```
gnavvy: +1

gnavvy:
I don't have a strong preference but would be great if we can finish all the 3.1 features before adding hex color support for all other layers.

PR #326, opened 2017-01-26
PathLayer perf improvements

Pessimistress:
- do not construct intermediate arrays
- replace `prevPositions` and `nextPositions` with delta (prep for 64bit version)
- add benchmark

### DIFF
--- a/src/layers/core/path-layer/path-layer.js
+++ b/src/layers/core/path-layer/path-layer.js
@@ -92,7 +92,7 @@ export default class PathLayer extends Layer {
   }
 
   calculateIndices(attribute) {
-    const {paths, IndexType, model, indexCount} = this.state;
+    const {paths, IndexType, model, indexCount, pointCount} = this.state;
 
     if (IndexType === Uint16Array && indexCount > 65535) {
       throw new Error('Vertex count exceeds browser\'s limit');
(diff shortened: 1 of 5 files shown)


src/layers/core/path-layer/path-layer.js
```diff
@@ -97,7 +82,11 @@ export default class PathLayer extends Layer {
...
 
   calculateIndices(attribute) {
-    const {paths, IndexType, model, indexCount} = this.state;
+    const {paths, IndexType, model, pointCount} = this.state;
+
+    // each path with length n has n-1 line segments
+    // * 2 * 3: each segment is rendered as 2 triangles with 3 vertices
+    const indexCount = (pointCount - paths.length) * 2 * 3;
 
     if (IndexType === Uint16Array && indexCount > 65535) {
```
ibgreen: It's not the indexCount but the vertexCount you need to check (i.e. an Uint16Array can have more than 65K elements, but each value can only be max 65K).

src/layers/core/path-layer/path-layer.js
```diff
@@ -153,69 +135,72 @@ export default class PathLayer extends Layer {
         positions[i++] = point[0];
         positions[i++] = point[1];
         positions[i++] = point[2] || 0;
-        positions[i++] = point[0];
-        positions[i++] = point[1];
-        positions[i++] = point[2] || 0;
       })
     );
+    positions.copyWithin(i, 0, i);
```
ibgreen: +1 for using copyWithin. But maybe add a comment about how this works. Why do we need two copies of the entire path array.

src/layers/core/path-layer/path-layer.js
```diff
@@ -153,69 +135,72 @@ export default class PathLayer extends Layer {
...
+      path.reduce((prevPoint, point) => {
+        if (prevPoint) {
+          leftDeltas[i++] = point[0] - prevPoint[0];
+          leftDeltas[i++] = point[1] - prevPoint[1];
+          leftDeltas[i++] = (point[2] - prevPoint[2]) || 0;
+        }
+        return point;
+      }, null);
     });
+    leftDeltas.copyWithin(i, 0, i);
```
ibgreen: ditto

src/layers/core/path-layer/path-layer.js
```diff
@@ -153,69 +135,72 @@ export default class PathLayer extends Layer {
...
       if (isNaN(w)) {
         w = 1;
       }
-
-      path.forEach(() => {
-        directions[i++] = w;
-        directions[i++] = -w;
-      });
+      const count = path.length;
+      fillArray({target: directions, source: [w], start: i, count});
```
ibgreen: If it is just one element, `Array.fill` might be faster (and standard JS). `fillArray` was written to quickly duplicate a subarray.

src/layers/core/path-layer/path-layer.js
```diff
@@ -224,60 +209,42 @@ export default class PathLayer extends Layer {
   calculateColors(attribute) {
     const {data, getColor} = this.props;
     const {paths, pointCount} = this.state;
-    const colors = new Uint8Array(pointCount * attribute.size * 2);
+    const {size} = attribute;
+    const colors = new Uint8Array(pointCount * size * 2);
```
ibgreen: `Uint8ClampedArray`?

src/layers/core/path-layer/path-layer.js
Pessimistress: Oops. My bad.

Pessimistress:
After some poking around, unfortunately `copyWith` has a huge hit on perf. I'm favoring for loops for now.

PR #330, opened 2017-01-27
Build script fixes, example cleanup

ibgreen:


ibgreen:
@gnavvy 

@Pessimistress wants to remove strokeWidth completely for GL.LINE based rendering since Chrome is withdrawing support. I'm fine with that. GL.LINE will basically be useful for wireframes only.

gnavvy:
for GL.LINE based rendering, sure I 100% for it.

PR #331, opened 2017-01-27
Add per-layer settings to example app

Pessimistress:
Prop types are inferred from names and values in `layer.defaultProps` and `example.props`.
Only accessors supported are color accessors. Update triggers are keyed to the prop name.
Updating the `effects` prop on `DeckGL` does not seem to work.

@ibgreen

### DIFF
--- a/examples/main/app.js
+++ b/examples/main/app.js
@@ -14,7 +14,7 @@ import {FPSStats} from 'react-stats';
 import LayerInfo from './layer-info';
 import LayerSelector from './layer-selector';
 import LayerControls from './layer-controls';
-import LAYER_CATEGORIES, {DEFAULT_ACTIVE_LAYERS} from './layer-examples';
+import LAYER_CATEGORIES from './layer-examples';
 
 /* eslint-disable no-process-env */
 const MAPBOX_ACCESS_TOKEN = process.env.MAPBOX_ACCESS_TOKEN || // eslint-disable-line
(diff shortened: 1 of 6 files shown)


examples/main/layer-examples.js
```diff
@@ -17,9 +17,9 @@ import {
 
   ChoroplethLayer,
   ChoroplethLayer64,
-  ExtrudedChoroplethLayer64,
+  ExtrudedChoroplethLayer64
 
-  Container
+  // Container
```
ibgreen: We're exporting the `get` function now separately. This could be nice to showcase as a way to get nested data.

Agree we should not use the Container class, it's being removed from the exports.

examples/main/utils/color-picker.js
```diff
@@ -0,0 +1,57 @@
+/* global document */
+import React, {PureComponent} from 'react';
+
+function pad2(str) {
```
ibgreen: While the functions in this file are somewhat self explanatory, it would be good to have at least one line of comment for each function describing what they do.

examples/main/app.js
```diff
@@ -150,7 +171,7 @@ class App extends PureComponent {
           {...mapViewState}
           onWebGLInitialized={ this._onWebGLInitialized }
           layers={this._renderExamples()}
-          effects={this._effects}
+          effects={effects ? this._effects : undefined}
```
ibgreen: Can we use `null` instead of undefined to make it clear that we are returning a `null` object.

Pessimistress: Can't pass null to effects...

PR #332, opened 2017-01-27
Allow accessor names as update trigger keys

Pessimistress:
- allow accessor names as update trigger keys
- fix bug where undefined update trigger throws an error
- add accessor names to all core layer attributes
- add test cases

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -8,8 +8,7 @@
   ],
   "rules": {
     "guard-for-in": 0,
-    "no-inline-comments": 0,
-    "no-continue": 0
+    "no-inline-comments": 0
   }
 }
 
--- a/.eslintrc
+++ b/.eslintrc
@@ -8,7 +8,8 @@
   ],
   "rules": {
     "guard-for-in": 0,
-    "no-inline-comments": 0
+    "no-inline-comments": 0,
+    "no-continue": 0
   }
 }
 
--- a/examples/main/app.js
+++ b/examples/main/app.js
@@ -70,21 +70,12 @@ class App extends PureComponent {
 
   _onToggleLayer(exampleName, example) {
     const activeExamples = {...this.state.activeExamples};
-
-    if (activeExamples[exampleName]) {
-      activeExamples[exampleName] = null;
-    } else {
-      activeExamples[exampleName] = {
-        ...example.layer.defaultProps,
-        ...example.props
-      };
-    }
+    activeExamples[exampleName] = !activeExamples[exampleName];
     this.setState({activeExamples});
   }
 
   _onUpdateLayerSettings(exampleName, settings) {
     const activeExamples = {...this.state.activeExamples};
-
     activeExamples[exampleName] = {
       ...activeExamples[exampleName],
       ...settings
@@ -130,11 +121,17 @@ class App extends PureComponent {
     for (const categoryName of Object.keys(LAYER_CATEGORIES)) {
       for (const exampleName of Object.keys(LAYER_CATEGORIES[categoryName])) {
 
-        const settings = this.state.activeExamples[exampleName];
+        let settings = this.state.activeExamples[exampleName];
         // An example is a function that returns a DeckGL layer instance
         if (settings) {
           const example = LAYER_CATEGORIES[categoryName][exampleName];
-          layers.push(this._renderExampleLayer(example, settings, index++));
+          const layer = this._renderExampleLayer(example, settings, index++);
+
+          if (typeof settings !== 'object') {
+            this.state.activeExamples[exampleName] = layer.props;
+          }
+
+          layers.push(layer);
         }
       }
     }
--- a/examples/main/app.js
+++ b/examples/main/app.js
@@ -115,6 +115,7 @@ class App extends PureComponent {
     return new Layer(layerProps);
   }
 
+  /* eslint-disable max-depth */
   _renderExamples() {
     let index = 1;
     const layers = [];
@@ -125,21 +126,21 @@ class App extends PureComponent {
 
         const settings = activeExamples[exampleName];
         // An example is a function that returns a DeckGL layer instance
-        if (!settings) {
-          continue;
-        }
-        const example = LAYER_CATEGORIES[categoryName][exampleName];
-        const layer = this._renderExampleLayer(example, settings, index++);
+        if (settings) {
+          const example = LAYER_CATEGORIES[categoryName][exampleName];
+          const layer = this._renderExampleLayer(example, settings, index++);
 
-        if (typeof settings !== 'object') {
-          activeExamples[exampleName] = layer.props;
-        }
+          if (typeof settings !== 'object') {
+            activeExamples[exampleName] = layer.props;
+          }
 
-        layers.push(layer);
+          layers.push(layer);
+        }
       }
     }
     return layers;
   }
+  /* eslint-enable max-depth */
 
   _getModelMatrix(index, offsetMode) {
     const {settings: {separation, rotationZ, rotationX}} = this.state;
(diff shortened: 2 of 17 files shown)


.eslintrc
```diff
@@ -8,7 +8,8 @@
   ],
   "rules": {
     "guard-for-in": 0,
-    "no-inline-comments": 0
+    "no-inline-comments": 0,
+    "no-continue": 0
```
ibgreen: This lint rule might be better to disable on a case by case basis. Avoiding `continue` in loops is a good practice.

examples/main/app.js
```diff
@@ -127,15 +118,24 @@ class App extends PureComponent {
...
     for (const categoryName of Object.keys(LAYER_CATEGORIES)) {
       for (const exampleName of Object.keys(LAYER_CATEGORIES[categoryName])) {
 
-        const settings = this.state.activeExamples[exampleName];
+        const settings = activeExamples[exampleName];
         // An example is a function that returns a DeckGL layer instance
-        if (settings) {
-          const example = LAYER_CATEGORIES[categoryName][exampleName];
-          layers.push(this._renderExampleLayer(example, settings, index++));
+        if (!settings) {
```
ibgreen: Is the `continue` really adding anything here? It could be better to wrap in an if statement? If nested to deep, maybe an additional function is in order?

examples/main/layer-examples.js
```diff
@@ -381,7 +369,6 @@ export default {
   'Core Layers': {
     'GeoJsonLayer': GeoJsonLayerExample,
     'GeoJsonLayer (Extruded)': GeoJsonLayerExtrudedExample,
```
ibgreen: For later - do we even need these two? Could it just be one?

src/layers/core/arc-layer/arc-layer.js
```diff
@@ -41,15 +41,20 @@ export default class ArcLayer extends Layer {
 
     const {attributeManager} = this.state;
     attributeManager.addInstanced({
-      instancePositions: {size: 4, update: this.calculateInstancePositions},
+      instancePositions: {
```
ibgreen: Why no accessor? Shouldn't this be an array?

src/layers/core/geojson-layer/geojson-layer.js
```diff
@@ -114,6 +114,7 @@ export default class GeoJsonLayer extends Layer {
       onHover: this._onHoverSublayer.bind(this),
       onClick: noop
     };
+    const updateTriggers = this.props.updateTriggers || {};
```
ibgreen: The `{}` should be set by `Layer.defaultProps` - we should not need it here?

src/layers/core/geojson-layer/geojson-layer.js
```diff
@@ -124,7 +125,11 @@ export default class GeoJsonLayer extends Layer {
         getHeight,
         getColor: getFillColor,
         extruded,
-        wireframe: false
+        wireframe: false,
+        updateTriggers: {
+          getHeight: updateTriggers.getHeight,
```
ibgreen: Might make the forwarding clearer to type out the full props line: 
```
    updateTriggers: {
          getHeight: this.props.updateTriggers.getHeight,
```

src/layers/core/grid-layer/grid-layer.js
```diff
@@ -89,10 +89,14 @@ export default class GridLayer extends Layer {
 
     const {attributeManager} = this.state;
     attributeManager.addInstanced({
-      instancePositions: {size: 4, update: this.calculateInstancePositions},
+      instancePositions: {
+        size: 4,
+        update: this.calculateInstancePositions
```
ibgreen: Accessor missing?

src/layers/core/hexagon-layer/hexagon-layer.js
```diff
@@ -92,8 +92,16 @@ export default class HexagonLayer extends Layer {
 
     const {attributeManager} = this.state;
     attributeManager.addInstanced({
-      instancePositions: {size: 3, update: this.calculateInstancePositions},
-      instanceColors: {size: 4, update: this.calculateInstanceColors}
+      instancePositions: {
+        size: 3,
+        update: this.calculateInstancePositions
```
ibgreen: ditto

src/layers/core/path-layer/path-layer.js
```diff
@@ -41,15 +41,41 @@ export default class PathLayer extends Layer {
...
-      leftDeltas: {size: 3, update: this.calculateLeftDeltas},
-      rightDeltas: {size: 3, update: this.calculateRightDeltas},
-      directions: {size: 1, update: this.calculateDirections},
-      colors: {size: 4, type: GL.UNSIGNED_BYTE, update: this.calculateColors},
-      pickingColors: {size: 3, type: GL.UNSIGNED_BYTE, update: this.calculatePickingColors,
-        noAlloc: true}
+      indices: {
+        size: 1,
+        isIndexed: true,
+        update: this.calculateIndices, noAlloc
```
ibgreen: 1) lots of missing accessors. 
2) I really think we should try to keep the attribute definitions on single lines. We have increased the max length in the linter to 100, so no need to break on 80 characters.

src/lib/attribute-manager.js
```diff
@@ -81,10 +81,13 @@ export default class AttributeManager {
...
 
-  // Marks an attribute for update
+  /* Marks an attribute for update
+   * @param {string} attributeName: attribute or accessor name
+   */
   invalidate(attributeName) {
     const {attributes} = this;
-    const attribute = attributes[attributeName];
+    const attribute = attributes[attributeName] ||
+      Object.values(attributes).find(attr => attr.accessor === attributeName);
```
ibgreen: This happens quite frequently. I believe we discussed pre-calculating a map to avoid searches here?

At least would like to see a comment mentioning this as a TODO or possible concernt.

examples/main/app.js
Pessimistress: This is because lint complains about nesting too deep. I can either disable `no-continue` or increase `max-depth`.

examples/main/layer-examples.js
Pessimistress: They are using different datasets. If we can find a good data that contains all feature types, I don't see why not.

src/layers/core/arc-layer/arc-layer.js
Pessimistress: Accessor is assumed to be a string across `AttributeManager`. If we allow it to be an array more inspection and testing are needed. Maybe in a separate diff?

src/layers/core/grid-layer/grid-layer.js
Pessimistress: Same as ArcLayer here - GridLayer use multiple accessors to generate instancePositions.

src/layers/core/hexagon-layer/hexagon-layer.js
Pessimistress: HexagonLayer uses multiple accessors to generate instanceColors. `getAlpha` prop should be removed for consistency, though. @heshan0131

src/layers/core/path-layer/path-layer.js
Pessimistress: PathLayer and PolygonLayer extract coordinates in `updateState` when data changes. These coordinates are then consumed by multiple updaters. Triggering the updaters along will not get updated data.

Pessimistress:
@ibgreen Added support for multiple accessors on one attribute.
Regarding your comment on missing accessors for multiple attributes in PathLayer:
UpdateTriggers do not work with the coordinate-dependent attributes in PathLayer and PolygonLayer, because `getPath` and `getPolygon` are called to extract the coordinates once in `updateState()` when `dataChanged` flag is true. If the nested coordinates have changed, the correct response needs to be re-extract the coordinates and invalidate all attributes. Calling updaters alone will not produce updated results. Currently there's no way to trigger this without reconstructing the `data` array.

PR #333, opened 2017-02-02
Rendering test

shaojingli:
Added rendering tests for deck.gl.


src/react/deckgl.js
```diff
@@ -50,7 +51,8 @@ const defaultProps = {
   effects: [],
   onWebGLInitialized: noop,
   onLayerClick: noop,
-  onLayerHover: noop
+  onLayerHover: noop,
+  onLayerDrawn: noop
```
ibgreen: Is the name of this callback (i.e. `onLayerDrawn`) appropriate? It aligns with the existing callbacks, but this is not called for a specific layer right? It is more like `onAfterRender`, `onAllLayersDrawn`, etc.

Also, since this is the public API, it should probably be described in the documentation.

src/react/deckgl.js
```diff
@@ -67,6 +69,10 @@ export default class DeckGL extends React.Component {
     this._updateLayers(nextProps);
   }
 
+  _onFrameDrawn(canvas) {
```
ibgreen: I think you could just pass the callback (i.e. the value of the prop) down to the parent component's prop. No need to define this function to wrap it.

src/react/deckgl.js
```diff
@@ -158,6 +164,7 @@ export default class DeckGL extends React.Component {
       debug,
       viewport: {x: 0, y: 0, width, height},
       onRendererInitialized: this._onRendererInitialized,
+      onFrameDrawn: this._onFrameDrawn,
```
ibgreen: We used to have `onBeforeRender` and `onAfterRender` callbacks in the WebGL renderer but they were recently removed during a cleanup. Looks like we are adding one of them back now, which is fine if we need it.

test/rendering-test/webpack.config.test-rendering.js
```diff
@@ -25,16 +25,20 @@ module.exports = {
...
   module: {
     rules: [
       {
-        include: [resolve('./src')],
+        include: [resolve('../../src')],
         loader: 'transform-loader',
         options: 'brfs-babel'
+      },
+      {
+        test: /\.png$/,
```
ibgreen: Maybe add some comment on what this loader does that can't be handled by brfs. Is it correct to assume that these test cases don't work directly under node.js, since they depend on a loader?

src/react/deckgl.js
shaojingli: I knew we once have onAfterRender and onBeforeRender but later deprecate it. I'm only add this function back for the test purposes. We need to exact the rendered canvas to check correctness. We shouldn't let normal developers to use this for feature development.

As for the name, I'm open to any suggestions. onAfterRender() seems to be better

test/rendering-test/webpack.config.test-rendering.js
shaojingli: I'm not exactly sure how these loaders and brfs works... will need to sync with you when you have time

Issue #300, opened 2017-01-22
deck.gl keeps rendering when all layers are off and the reflection is on


shaojingli:
EffectManager needs some tweaking to prevent this from happening

PR #335, opened 2017-02-03
Fixed the unnecessary layer draw calls when there is no layer #300

shaojingli:
Fixed an issue in the layer manager that even when no layer is selected, deck.gl still issues draw calls. #300

### DIFF
--- a/src/lib/layer-manager.js
+++ b/src/lib/layer-manager.js
@@ -42,7 +42,7 @@ export default class LayerManager {
     this.layers = [];
     // Tracks if any layers were drawn last update
     // Needed to ensure that screen is cleared when no layers are shown
-    this.layerCleared = false;
+    this.screenCleared = false;
     this.oldContext = {};
     this.context = {
       gl,
@@ -142,14 +142,14 @@ export default class LayerManager {
 
     // Make sure that buffer is cleared once when layer list becomes empty
     if (this.layers.length === 0) {
-      if (this.layerCleared === false) {
+      if (this.screenCleared === false) {
         redraw = true;
-        this.layerCleared = true;
+        this.screenCleared = true;
         return true;
       }
     } else {
-      if (this.layerCleared === true) {
-        this.layerCleared = false;
+      if (this.screenCleared === true) {
+        this.screenCleared = false;
       }
     }
 
--- a/src/lib/layer-manager.js
+++ b/src/lib/layer-manager.js
@@ -42,7 +42,7 @@ export default class LayerManager {
     this.layers = [];
     // Tracks if any layers were drawn last update
     // Needed to ensure that screen is cleared when no layers are shown
-    this.drewLayers = false;
+    this.layerCleared = false;
     this.oldContext = {};
     this.context = {
       gl,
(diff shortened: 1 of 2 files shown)


src/lib/layer-manager.js
```diff
@@ -42,7 +42,7 @@ export default class LayerManager {
     this.layers = [];
     // Tracks if any layers were drawn last update
     // Needed to ensure that screen is cleared when no layers are shown
-    this.drewLayers = false;
+    this.layerCleared = false;
```
ibgreen: Naming? This is not tracked per layer, right. Should it be called `screenCleared`?

src/lib/layer-manager.js
```diff
@@ -142,19 +141,20 @@ export default class LayerManager {
...
-      this.redrawNeeded = false;
-      redraw = true;
+    if (this.layers.length === 0) {
+      if (this.layerCleared === false) {
+        redraw = true;
+        this.layerCleared = true;
+        return true;
+      }
+    } else {
+      if (this.layerCleared === true) {
```
ibgreen: Looks like we don't need this if statement, i.e. always assigning layerCleared to false has the same effect?

src/lib/layer-manager.js
```diff
@@ -142,19 +141,20 @@ export default class LayerManager {
...
-      redraw = true;
-      return true;
-    }
-
-    if (this.redrawNeeded) {
-      this.redrawNeeded = false;
-      redraw = true;
+    if (this.layers.length === 0) {
+      if (this.layerCleared === false) {
+        redraw = true;
```
ibgreen: I realize that this is a relic from before the change, but looks like we are setting `redraw` but not using it before we return.

src/lib/layer-manager.js
```diff
@@ -51,7 +51,6 @@ export default class LayerManager {
       viewportChanged: true,
       pickingFBO: null
     };
-    this.redrawNeeded = true;
```
shaojingli: this.redrawNeeded seems to be redundant... right? @ibgreen

ibgreen: Yes in principle it looks like it is redundant, although this was originally modeled after redraw flag handling in other objects, where you could call a setRedrawNeeded and force a redraw.

getRedrawNeeded would return true if either redrawNeeded was set, or other flags indicated that redraw was needed, and this would clear redrawNeeded.

The reason there was no returns in the middle of the function was that it was important to clear all redraw flags of subobjects like layers, otherwise they could trigger a redraw on next call.

src/lib/layer-manager.js
shaojingli: yes. you are right... it's the same no matter this if statement is here or not. 

I just wanted to make this line reads like "switch flip"... basically "if layerCleared is true, then flip it to false".

src/lib/layer-manager.js
shaojingli: Agree.

shaojingli:
@ibgreen totally share your concern about the function being too long and the logic being to convoluted to follow. I'm investigating ways to refactor it and make it much clearer

PR #336, opened 2017-02-04
PathLayer: support color-per-point

dcposch:
Small change. Allow `PathLayer#getColor` to optionally return an array of colors instead of a single color. If it returns an array, the length must match the number of points returned by `getPath`.

This lets you draw paths that change color along their length, or paths where the alpha channel fades out along the length of the path.

### DIFF
--- a/docs/layers/path-layer.md
+++ b/docs/layers/path-layer.md
@@ -9,10 +9,9 @@ Inherits from all [Base Layer](/docs/layers/base-layer.md) properties.
 
 ## Layer-specific Properties
 
-
 #### `getPath` (Function, optional)
 
-- Default: `(object, index) => object.paths`
+- Default: `(object, index) => object.path`
 
 Returns the specified path for the object.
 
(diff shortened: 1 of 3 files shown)


dcposch:
image

ibgreen:
@dcposch - Love this, you are totally getting the spirit of deck.gl!

The code looks great. About merging, this is mainly an API audit question which is something we are working through right now for the new release. Basically, we want our accessors to work consistently across layers, and we want to consider future API growth, and so we at least need to ask ourselves:
- does adding this function here mean that we should add similar facilities for other layers
- if we chose to overload accessors like this, what for does that mean for other ways of overloading the accessors.

We use layer sub-classing to organized this kind of features internally (i.e. small improvements to layers that haven't yet made it into deck.gl). You should be able to just override the calculateColors method like so:
```
class MultiColorPathLayer extends PathLayer {
   calculateColors(....
```
If we decide not to merge, this might be a great example for docs about layer subclassing.

dcposch:
Sure, seems reasonable.

> does adding this function here mean that we should add similar facilities for other layers

I don't think color-per-point makes sense for the polygon layer. The user doesn't have control over how the polygon is triangulated, so unless all verts have the same color, it may shade weirdly.

The path, otoh, is a basically a squiggly line, so varying color along its length shades pretty nicely.

Good call on the subclassing, thx!

ibgreen:
Closing as this is being included as a sample layer in #356

PR #338, opened 2017-02-07
Hexagon layer api

heshan0131:
- use radiusScale instead of dotRadius
- use extruded instead of enable3d
- calculate default hexagonVertices if not provided

### DIFF
--- a/docs/layers/hexagon-layer.md
+++ b/docs/layers/hexagon-layer.md
@@ -20,14 +20,14 @@ Inherits from all [Base Layer](/docs/layers/base-layer.md) properties.
 Primitive hexagon vertices as an array of six [lon, lat] pairs,
 in either clockwise or counter clouckwise direction.
 
-##### `dotRadius` (Number, optional)
+##### `radiusScale` (Number, optional)
 
 - Default: `1`
 
 Hexagon radius multiplier, between 0 - 1. It defines the gap between each hexagon.
 when set to 1, all hexagons will be closely placed to each other.
 
-##### `enable3d` (Boolean, optional)
+##### `extruded` (Boolean, optional)
 
 - Default: `true`
 
--- a/docs/layers/hexagon-layer.md
+++ b/docs/layers/hexagon-layer.md
@@ -24,20 +24,14 @@ in either clockwise or counter clouckwise direction.
 
 - Default: `1`
 
-Hexagon radius multiplier, between 0 - 1. It defines the gap between each hexagon.
-when set to 1, all hexagons will be closely placed to each other.
+Hexagon radius multiplier, between 0 - 1. The radius of hexagon is calculated by 
+radiusScale * getRadius(d)
 
 ##### `extruded` (Boolean, optional)
 
 - Default: `true`
 
 Whether to extrude hexagon. If se to false, all hexagons will be set to flat.
-
-#### `invisibleColor` (Number[3], optional)
-
-- Default: `[0, 0, 0]`
-
-To make one hexagon invisible, set its color to equal to invisibleColor
 
 #### `getCentroid` (Function, optional)
 
@@ -49,14 +43,8 @@ Method called to retrieve the centroid of each hexagon. Centoid should be set to
 
 - Default: `object => object.color`
 
-Method called to retrieve the color of each object. Color should be set to [r, g, b]
+Method called to retrieve the color of each object. Color should be set to [r, g, b, a]
 with each number between 0 -255
-
-#### `getAlpha` (Function, optional)
-
-- Default: `object => 255`
-
-Method called to retrieve the alpha of each object. Color should be set between 0 to 255
 
 #### `getElevation` (Function, optional)
 
--- a/examples/main/layer-controls.js
+++ b/examples/main/layer-controls.js
@@ -38,7 +38,8 @@ export default class LayerControls extends PureComponent {
 
   _renderSlider(settingName, value) {
     let max = 1;
-    if (/radius|width|height|pixel|size/i.test(settingName)) {
+    if (/radius|width|height|pixel|size/i.test(settingName) &&
+      !(/^((?!scale).)*$/).test(settingName)) {
       max = 100;
     }
 
--- a/src/layers/core/hexagon-layer/hexagon-layer-vertex.glsl
+++ b/src/layers/core/hexagon-layer/hexagon-layer-vertex.glsl
@@ -19,7 +19,6 @@ uniform mat4 worldInverseTransposeMatrix;
 
 // Custom uniforms
 uniform float opacity;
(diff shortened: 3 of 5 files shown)


src/layers/core/hexagon-layer/hexagon-layer-vertex.glsl
```diff
@@ -22,7 +22,8 @@ uniform float opacity;
 uniform vec3 invisibleColor;
 uniform float radius;
 uniform float angle;
-uniform float enable3d;
+uniform float extruded;
+uniform float radiusScale;
 
 // Lighting constants
 const vec3 ambientColor = vec3(1., 1., 1.);
```
Pessimistress: We should migrate all 3d layers to use the same lighting code that `PolygonLayer` is using.

docs/layers/hexagon-layer.md
```diff
@@ -20,14 +20,14 @@ Inherits from all [Base Layer](/docs/layers/base-layer.md) properties.
 Primitive hexagon vertices as an array of six [lon, lat] pairs,
 in either clockwise or counter clouckwise direction.
 
-##### `dotRadius` (Number, optional)
+##### `radiusScale` (Number, optional)
 
 - Default: `1`
 
 Hexagon radius multiplier, between 0 - 1. It defines the gap between each hexagon.
```
ibgreen: Could the sentence about "gap" otentially be confusing as this also depends on the centroids, which can be arbitrary (although we would expect them to be generated by a hex grid)?

examples/main/layer-controls.js
```diff
@@ -38,7 +38,8 @@ export default class LayerControls extends PureComponent {
 
   _renderSlider(settingName, value) {
     let max = 1;
-    if (/radius|width|height|pixel|size/i.test(settingName)) {
+    if (/radius|width|height|pixel|size/i.test(settingName) &&
+      !(/^((?!scale).)*$/).test(settingName)) {
```
ibgreen: These stacked regular expressions are not exactly easy to read, especially if more cases are added.

Some relief could come from defining them in well-named constants, or through creating a helper function, or at minimum adding a comment.

src/layers/core/hexagon-layer/hexagon-layer-vertex.glsl
ibgreen: > We should migrate all 3d layers to use the same lighting code that PolygonLayer is using.

While I don't have a strong opinion on which lighting implementation we should use. I agree we should have same lighting code, and same support same settings in all the extruded layers. Should be packaged as a shader module, like the lighting code in polygon. 

Finally, this might be longer term but it would be really neat if we could set the lighting settings globally, e.g. through a LightingEffect, instead of having to pass them in as props to every lighting-enabled layer.

src/layers/core/hexagon-layer/hexagon-layer.js
```diff
@@ -120,25 +121,39 @@ export default class HexagonLayer extends Layer {
...
 
-    // distance between two close centroids
-    const dx = spaceCoord0[0] - spaceCoord3[0];
-    const dy = spaceCoord0[1] - spaceCoord3[1];
-    const dxy = Math.sqrt(dx * dx + dy * dy);
+      const vertex0 = vertices[0];
+      const vertex3 = vertices[3];
+
+      // transform to space coordinates
+      const spaceCoord0 = this.projectFlat(vertex0);
```
ibgreen: Ideally we should have a test case for a METER_OFFSET mode HexagonLayer to make sure these JS project functions work in both cases. I can help with that.

src/layers/core/hexagon-layer/hexagon-layer.js
```diff
@@ -120,25 +121,39 @@ export default class HexagonLayer extends Layer {
...
-    this.setUniforms({
       // Calculate angle that the perpendicular hexagon vertex axis is tilted
-      angle: Math.acos(dx / dxy) * -Math.sign(dy) + Math.PI / 2,
-      // Allow user to fine tune radius
-      radius: dxy / 2 * Math.max(0, Math.min(1, dotRadius))
+      angle = Math.acos(dx / dxy) * -Math.sign(dy) + Math.PI / 2;
+      radius = dxy / 2;
+    }
+
+    this.setUniforms({
```
ibgreen: My assumption was to just generate a list of hexagon vertices from angle and radius (if hexagonVertices was not provided). In this version, we would not need angle and radius as uniforms.

This function would be called `calculateDefaultHexagonVertices` or similar.

src/layers/core/hexagon-layer/hexagon-layer.js
```diff
@@ -156,11 +171,12 @@ export default class HexagonLayer extends Layer {
   }
 
   updateUniforms() {
-    const {opacity, enable3d, invisibleColor} = this.props;
+    const {opacity, extruded, invisibleColor, radiusScale} = this.props;
     this.setUniforms({
-      enable3d: enable3d ? 1 : 0,
+      extruded: extruded ? 1 : 0,
```
ibgreen: Just curious, do we need to do this with boolean uniforms? Doesn't luma handle that?

src/layers/core/hexagon-layer/hexagon-layer.js
```diff
@@ -222,5 +238,34 @@ export default class HexagonLayer extends Layer {
   }
 }
 
+// if hexagon vertices not provided, calculate a 1km radius hexagon
+// at the first object's centroid location
+function _getDefaultVertices(data, getCentroid) {
```
ibgreen: Again I would have used the provided radius (in meters) and angle here to generate verts, rather than just assuming 1 kilometer.

The Viewport should have distance mapping functions now that convert meters to lnglat etc. This would be a good test to make sure they work the same in JS and GLSL.

src/layers/core/hexagon-layer/hexagon-layer-vertex.glsl
```diff
@@ -51,12 +52,13 @@ void main(void) {
 
   // calculate elevation, if 3d not enabled set to 0
   // cylindar gemoetry height are between -0.5 to 0.5, transform it to between 0, 1
-  float elevation =  mix(0., project_scale(instancePositions.z  * (positions.y + 0.5) * ELEVATION_SCALE), enable3d);
+  float elevation =  mix(0., project_scale(instancePositions.z  * (positions.y + 0.5) * ELEVATION_SCALE), extruded);
```
ibgreen: As an alternative, we could also generate a simpler primitive geometry when extruded=false, i.e. not have any side triangles.

This could impact the hexagon layer rendering performance as there would be less invocations of the vertex shader in non-extruded mode.

I don't have a strong opinion here, other than that we might want this to work somewhat similarly between layers.

PR #340, opened 2017-02-07
ScatterplotLayer bugfix: update when data changes

dcposch:
Currently on the dev branch, when you change the `data` prop on a `ScatterplotLayer`, it doesn't invalidate attributes.

@ibgreen @Pessimistress

### DIFF
--- a/src/layers/core/scatterplot-layer/scatterplot-layer.js
+++ b/src/layers/core/scatterplot-layer/scatterplot-layer.js
@@ -58,10 +58,12 @@ export default class ScatterplotLayer extends Layer {
     /* eslint-enable max-len */
   }
 
-  updateState({props, oldProps}) {
+  updateState(info) {
+    const {props, oldProps} = info;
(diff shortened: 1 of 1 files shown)


ibgreen:
@Pessimistress @gnavvy @shaojingli

We need to get this fix landed. While @dcposch 's fix is correct and reasonable, I personally don't feel great about having to remember to call `super` when subclassing a layer that doesn't have an `updateState` method. Sounds like an unnecessary "gotcha" that will get users (and us) in trouble.

As an alternative we could change the code that calls `updateState` to automatically look at `changeFlags.dataChanged` and call `attributeManager.invalidateAll()`.

Then this would be called regardless of whether `updateState` was defined or not. The only issue would be in the rare case when someone wanted to completely defeat this. They would need to modify `dataChanged` before returning. Can't think of a case, but you never know...

Pessimistress:
I agree that the user should not have to remember to call `super`.
I'm merging this fix for now, and we can figure out a generic solution before the next release.

PR #341, opened 2017-02-08
update grid layer props based on api audit

heshan0131:


docs/layers/grid-layer.md
```diff
@@ -15,28 +15,24 @@ Inherits from all [Base Layer](/docs/layers/base-layer.md) properties.
 
 ## Layer-specific Properties
 
-##### `latDelta` (Number, optional)
+##### `latOffset` (Number, optional)
```
ibgreen: I think we need to drop the lat/lng here. But we didn't conclude on it in the review.

docs/layers/grid-layer.md
```diff
@@ -15,28 +15,24 @@ Inherits from all [Base Layer](/docs/layers/base-layer.md) properties.
...
+##### `extruded` (Boolean, optional)
 
 - Default: `true`
 
 Whether to enable grid elevation. If se to false, all grid will be flat.
 
-##### `opacity` (Number, optional)
-
-- Default: `0.6`
-
```
ibgreen: Just realized that there may be another decision point here - is it better to have elevation as a separate accessor rather than as z coord of position (like in line layer)?

src/layers/core/grid-layer/grid-layer.js
```diff
@@ -25,31 +25,14 @@ import {readFileSync} from 'fs';
...
- * @param {array} props.data -
- * @param {boolean} props.enable3d - enable grid height
- * @param {number} props.latDelta - grid cell size in lat delta
- * @param {number} props.lngDelta - grid cell size in lng delta
- * @param {number} props.opacity - opacity
- * @param {function} props.getPosition - position accessor, returned as [minLng, minLat]
- * @param {function} props.getElevation - elevation accessor
- * @param {function} props.getColor - color accessor, returned as [r, g, b]
- */
+const DEFAULT_COLOR = [255, 0, 255, 255];
```
ibgreen: Did we decide on black, i.e. 0, 0, 0, 255 as default color for layers?

src/layers/core/grid-layer/grid-layer.js
```diff
@@ -59,6 +42,20 @@ mat4.lookAt(viewMatrixCompat, [0, 0, 0], [0, 0, -1], [0, 1, 0]);
 const viewMatrix = new Float32Array(viewMatrixCompat);
 
 export default class GridLayer extends Layer {
+  /**
```
ibgreen: No big deal, but it looks a bit weird to have function param JSDoc floating like this, when we no longer have a constructor / params. 

We could potentially decide that we don't need these in the code, because all our props are documented separately in markdown, or make these as inline comments on the defaultProps "struct", or change the notation here.

src/layers/core/grid-layer/grid-layer.js
```diff
@@ -132,7 +130,7 @@ export default class GridLayer extends Layer {
       value[i + 0] = color[0];
       value[i + 1] = color[1];
       value[i + 2] = color[2];
-      value[i + 3] = isNaN(color[3]) ? DEFAULT_COLOR[3] : color[3];
+      value[i + 3] = Number.isFinite(color[3]) ? color[3] : DEFAULT_COLOR[3];
```
ibgreen:
```
value[i + 3] = Number.isFinite(color[3]) ? color[3] : 255;
```

Issue #337, opened 2017-02-06
Mitering bug: PathLayer produces uneven strokes if the path has acute angles


dcposch:
Deck.gl verson: 46add427dd7d2f132f5699e38edcc90802f3fc30 (`dev` branch)

![image](https://cloud.githubusercontent.com/assets/169280/22670048/4d52e2ae-ec7c-11e6-9a93-9a2648565b53.png)

### Steps to reproduce

* Create a `PathLayer` as follows:
```
const path = [[-81.29868341618506,41.1053559416184],[-80.97694,41.10014],[-80.82915,41.10922],[-80.84265,41.11511],[-80.76570077085131,41.05483762281574]]

const layer = new PathLayer({
      data: [{path}],
      opacity: 1,
      miterLimit: 1,
      strokeWidth: 6e3
})
```
* Zoom in on Ohio

### Expected result

Smooth path.

### Actual result

Uneven-width path. See screenshot above.

dcposch:
@ibgreen @Pessimistress

Pessimistress:
@dcposch You can fix this by increasing `miterLimit`. This value indicates the max size of a corner in ratio to the width. It's meant to limit super long corners created by sharp angles. In other words any corner with angle smaller than `asin(1 / miterLimit)` will get capped and have uneven appearance.

In the latest beta `miterLimit` is default to `4`. I will update the document for better clarity.

Pessimistress:
@dcposch @ibgreen A proper `miterLimit` should create a bevel when exceeded:

![miter limit](http://www.mikeswanson.com/blog/images/Miter%20Limit%20Illustration.gif)

The reason I have not implement it in the current `PathLayer` is because it will double the vertex count and make perf even worse (this layer is already the slowest in all 3 GeoJSON sub layers).

dcposch:
@Pessimistress thx!

Increasing `miterLimit` causes other issues on my dataset:

![image](https://cloud.githubusercontent.com/assets/169280/22673623/22c31374-ec90-11e6-8bbd-dfdc1f31a3d9.png)

### Three thoughts
- Proper mitering shouldn't double the number of verts. At most, it will increase the number of verts by 50%, since acute angles that reach the miter limit now get two verts on the outside, and still have just one vert on the inside. And that's if every angle in every path is acute. For most paths, the number of verts will be unchanged.

[...]

PR #342, opened 2017-02-08
PathLayer Miter Fix -- POC

Pessimistress:
Re: [Mitering bug: PathLayer produces uneven strokes if the path has acute angles #337](https://github.com/uber/deck.gl/issues/337)

Current PathLayer:
path-old

Fixed miter (still some z-fighting issues):
path-miter

Basically one extra point is added to each corner to control the joint.
This also allows us to do other line joints:
path-rounded

In this POC I'm switching to an instancing model. Comparing instanced against indexed draws:
- removed the requirement of the `OES_element_index_uint` extension
- reduce attribute length to 1/3
- the number of vertex shader calls per path increased from `numPoints * 3` to `(numPoints - 1) * 6`

Setting the perf concerns aside, it does open some pretty nice opportunities for us such as dashed lines. 
@ibgreen @contra @dcposch

### DIFF
--- a/examples/main/data-samples.js
+++ b/examples/main/data-samples.js
@@ -16,6 +16,15 @@ export const meterPoints = pointGrid(1e3, [-5000, -5000, 5000, 5000]);
 
 export const worldGrid = pointsToWorldGrid(points, 0.5);
 
+export const zigzag = [
+  {path: (new Array(12)).fill(0).map(
+    (d, i) => [
+      positionOrigin[0] + i * i * 0.001,
+      positionOrigin[1] + Math.cos(i * Math.PI) * 0.2 / (i + 4)
+    ])
+  }
+];
+
 // time consuming - only generate on demand
 let _points100K = null;
 export function getPoints100K() {
--- a/examples/main/layer-controls.js
+++ b/examples/main/layer-controls.js
@@ -38,8 +38,8 @@ export default class LayerControls extends PureComponent {
 
   _renderSlider(settingName, value) {
     let max = 1;
-    if (/radius|width|height|pixel|size/i.test(settingName) &&
-      !(/^((?!scale).)*$/).test(settingName)) {
+    if (/radius|width|height|pixel|size|miter/i.test(settingName) &&
+      (/^((?!scale).)*$/).test(settingName)) {
       max = 100;
     }
 
--- a/src/layers/core/path-layer/path-layer-vertex.glsl
+++ b/src/layers/core/path-layer/path-layer-vertex.glsl
@@ -21,6 +21,8 @@ varying vec4 vColor;
 varying vec2 vCornerOffset;
 varying float vMiterLength;
 
+const float EPSILON = 0.001;
+
 float flipIfTrue(bool flag) {
   return -(float(flag) * 2. - 1.);
 }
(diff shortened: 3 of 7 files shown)


examples/main/data-samples.js
```diff
@@ -16,6 +16,15 @@ export const meterPoints = pointGrid(1e3, [-5000, -5000, 5000, 5000]);
 
 export const worldGrid = pointsToWorldGrid(points, 0.5);
 
+export const zigzag = [
+  {path: (new Array(12)).fill(0).map(
```
ibgreen: Nit: Is the `fill` needed here?

examples/main/layer-controls.js
```diff
@@ -38,8 +38,8 @@ export default class LayerControls extends PureComponent {
 
   _renderSlider(settingName, value) {
     let max = 1;
-    if (/radius|width|height|pixel|size/i.test(settingName) &&
-      !(/^((?!scale).)*$/).test(settingName)) {
+    if (/radius|width|height|pixel|size|miter/i.test(settingName) &&
```
ibgreen: This is getting a little hard on the eyes. Feels like we'll need a propTypes concept soon to avoid endlessely adding to the "guess list" here.

src/layers/core/path-layer/path-layer-vertex.glsl
```diff
@@ -1,71 +1,197 @@
 #define SHADER_NAME path-layer-vertex-shader
 
-attribute float directions;
 attribute vec3 positions;
-attribute vec3 leftDeltas;
-attribute vec3 rightDeltas;
 
-attribute vec4 colors;
-attribute vec3 pickingColors;
+attribute vec3 instanceStartPositions;
```
ibgreen: We've discussed this off line, so most of this is clear now, but will still post my initial write-up since I wrote it.

I'm not reviewing this vertex shader in detail now, but as a general question, now that we use GL.TRIANGLES and pre-tesselate the line, is my assumption that this code is mainly there support rounded joins correct? 

On the perf side, I note that this shader contains a couple of if statements which is generally considered detrimental to GPU performance, especially when those if statements take different paths for different vertices which they clearly do here.

As an idea, it might be valuable to have @shaojingli review this shader, and also to instrument it with his GPU timer support and then experiment with various changes to understand how performance is affected.

Those changes could e.g. involve reorganizing the code or moving some calculations to JS in the tesselation stage to avoid branching here in GLSL, and then measuring again.

dcposch:
**Nice!** That was fast

src/layers/core/path-layer/path-layer-vertex.glsl
Pessimistress: We have always been calculating the line joints in the vertex shader. The layer itself is projection agnostic. I share your concern about the if statements. As the title suggests this is a quick POC, and there's a lot of room for optimization.

examples/main/data-samples.js
Pessimistress: Yes :)
```
(new Array(3)).fill(0).map(function(d, i) { return i })
[0, 1, 2]
(new Array(3)).map(function(d, i) { return i })
[undefined × 3]
```

ibgreen: Yes you are right, this one keeps getting me. Even though the length is reported as 4, there are no values and the iteration doesn't happen. 

I do know that we don't need the parenthesis though, i.e `new Array(12).fill(0).map` works fine and looks cleaner to me. It is also possible to drop the `new` for extra terseness but I personally prefer to have it.

src/layers/core/path-layer/path-layer-vertex.glsl
Pessimistress: I pushed a new commit that simplifies the vertex shader. Will run some perf test tomorrow.

PR #345, opened 2017-02-09
IconLayer API Audit

Pessimistress:
- renamed `getScale` to `getSize`
- renamed `size` to `sizeScale`
- moved color fallback to the default accessor

@shaojingli @gnavvy @ibgreen @heshan0131 @apercu @abmai


Pessimistress:
@ibgreen `getSize` returns a number in pixels - how is it not an absolute size?

ibgreen:
> getSize returns a number in pixels - how is it not an absolute size?
Yes now that reread the code it looks like that is indeed the case. I probably read the old comment. We're all good then!

PR #352, opened 2017-02-15
Fix stroke width on LineLayer and ArcLayer

Pessimistress:
Because of the deprecation of `gl.lineWidth()` in Chrome and Firefox, change LineLayer and ArcLayer to draw meshes instead of lines. Lines are extruded in screen space.
Eventually we want to draw cylinders/tubes for proper extrusion in 3d.

@ibgreen @shaojingli These `strokeWidth` are in pixels, while PathLayer's `strokeWidth` is in meters. Do we want to name them differently?


src/layers/core/line-layer/line-layer-vertex.glsl
```diff
@@ -26,21 +26,30 @@ attribute vec3 instanceTargetPositions;
...
   // linear interpolation of source & target to pick right coord
   float segmentIndex = positions.x;
-  vec3 p = mix(source, target, segmentIndex);
+  vec4 p = mix(source, target, segmentIndex);
 
-  gl_Position = project_to_clipspace(vec4(p, 1.));
+  // extrude by strokeWidth
+  vec2 dir = normalize((target.xy - source.xy) * screenSize);
+  vec2 perp = vec2(-dir.y, dir.x);
+  vec2 offset = perp * positions.y * strokeWidth / screenSize;
```
ibgreen: My first thought was that this always extrudes to one size (rather than centered?) If so it would probably only look good for smaller `strokeWidths`. If this is not true please add some comment explaining how the centering works.

src/layers/core/arc-layer/arc-layer.js
```diff
@@ -70,23 +68,41 @@ export default class ArcLayer extends Layer {
...
 
     const shaders = assembleShaders(gl, this.getShaders());
 
-    return new Model({
+    const model = new Model({
       gl,
       vs: shaders.vs,
       fs: shaders.fs,
       geometry: new Geometry({
-        drawMode: GL.LINE_STRIP,
```
ibgreen: We now rely on default `drawMode` here? Maybe we could still spell it out for clarity? 

Also, would `GL.TRIANGLE_STRIP` work? It might be faster than using indices and align more with `LineLayer`?

src/layers/core/line-layer/line-layer.js
```diff
@@ -49,15 +49,13 @@ export default class LineLayer extends Layer {
...
-    // Setting line width back to 1 is here to workaround a Google Chrome bug
-    // gl.clear() and gl.isEnabled() will return GL_INVALID_VALUE even with
-    // correct parameter
-    // This is not happening on Safari and Firefox
-    gl.lineWidth(1.0);
+    const {strokeWidth} = this.props;
+    const {viewport: {width, height}} = this.context;
+
+    this.state.model.render(Object.assign({}, uniforms, {
+      screenSize: [width, height],
```
ibgreen: Does this consider high DPI displays or do we need to multiply with DPI here?

src/layers/core/line-layer/line-layer-vertex.glsl
```diff
@@ -26,21 +26,30 @@ attribute vec3 instanceTargetPositions;
...
+  vec4 target = project_to_clipspace(vec4(targetPos, 1.0));
 
   // linear interpolation of source & target to pick right coord
   float segmentIndex = positions.x;
-  vec3 p = mix(source, target, segmentIndex);
+  vec4 p = mix(source, target, segmentIndex);
 
-  gl_Position = project_to_clipspace(vec4(p, 1.));
+  // extrude by strokeWidth
+  vec2 dir = normalize((target.xy - source.xy) * screenSize);
```
ibgreen: Should there be a division or multiplication with `2` somewhere here? Clip space goes from [-1, 1] so I would expect the full screen would have width `2` in clip space.

src/layers/core/line-layer/line-layer.js
```diff
@@ -68,7 +66,15 @@ export default class LineLayer extends Layer {
...
-    const positions = [0, 0, 0, 1, 1, 1];
+    /*
+     *  (1)-------------_(2)
+     *   |          _,-"  |
+     *   o      _,-"      o
+     *   |  _,-"          |
+     *  (0)"-------------(3)
+     */
+    const positions = [0, -1, 0, 0, 1, 0, 1, -1, 0, 1, 1, 0];
+    const indices = [0, 1, 2, 3];
```
ibgreen: Do these indices add anything when they just list the vertices in order?

src/layers/core/arc-layer/arc-layer-vertex.glsl
```diff
@@ -43,22 +44,41 @@ float paraboloid(vec2 source, vec2 target, float ratio) {
...
+  
+  vec3 currPos = getPos(source, target, segmentRatio);
+  vec3 nextPos = getPos(source, target, nextSegmentRatio);
+  vec4 curr = project_to_clipspace(vec4(currPos, 1.0));
+  vec4 next = project_to_clipspace(vec4(nextPos, 1.0));
+
+  // extrude by strokeWidth
+  vec2 dir = normalize((next.xy - curr.xy) * indexDir * screenSize);
+  vec2 perp = vec2(-dir.y, dir.x);
+  vec2 offset = perp * positions.y * strokeWidth / screenSize;
```
ibgreen: Might be nice to define a small well-named and well-commented function that does the move in pixel space.

ibgreen:
> These strokeWidth are in pixels, while PathLayer's strokeWidth is in meters. Do we want to name them differently?

@Pessimistress Good point! I think we need to be extra clear about what units we are using in different places as we have seen this cause confusion to both new and experienced users. If we called this `strokeWidthPixels` it would seem to align well with similar props in e.g. `ScatterplotLayer` that also talk about pixels.

contra:
@ibgreen I think it would be nice to standardize that across all layers - strokeWidth/radius = meters, but can specify min/max in pixels?

ibgreen:
> I think it would be nice to standardize that across all layers - strokeWidth/radius = meters, but can specify min/max in pixels?

@contra Agreed. The major complication that we are debating internally is whether the pixel width approach is a good fit with perspective mode. E.g. a fixed pixel width does not correctly show perspective e.g. on an arc, it should taper away in the distance. 

To deal with this growing conflict of requirements we've been playing with the idea of defining groups of layers, e.g. a group of 2D focused layers, a group of 2.5D extrusion focused layers, etc. In such as world (deck.gl v5?), this proposal would make a lot of sense as a "standard" for the 2D layers.

@Pessimistress We should at least consider what it would mean to go with @contra's proposal for v4 (i.e. have the strokeProp in meters plus min max pixel props).

src/layers/core/line-layer/line-layer-vertex.glsl
Pessimistress: `dir` is normalized, so the `2` is irrelevant.
We offset the center line by `strokeWidth / 2` to each side, which cancels out with the `2` when translated back to clip space. I'll add more comment for this.

src/layers/core/line-layer/line-layer-vertex.glsl
Pessimistress: `positions.y` is `-1` or `1`, which indicates the offset direction.

src/layers/core/line-layer/line-layer.js
Pessimistress: My understanding is that the same DPI multiplier will be applied to both screen size and stroke width, then cancel each other out. Is that correct?

Pessimistress:
@ibgreen @contra having the strokeProp in meters plus min max pixel props in ArcLayer will require drawing tubes as I mentioned in the original PR. Say we draw the arcs as a tube with decagonal sections. That will increase the number of projections in vertex shader for each arc from `50` (v3) or `200` (this PR) to `1000`.
We have a separate discussion around moving the mercator pre-projection from vertex shader to the CPU. I would be more comfortable with this change if that is the case.

src/layers/core/line-layer/line-layer.js
ibgreen: will be applied to both screen size and stroke width, then cancel each other out
> Ahh, you are probably right. Still, I think this code risks being hard to maintain if we make too many clever assumptions like that - even if we document them.

Unless there is a real perf gain (which I don't think there is here), it is usually better to work with the quantities and coordinate system that one would intuitively expect. In this case if we have a `screenSize` uniform, we'd have it specified in the amount of pixels the shader actually sees. That would make the same uniform easily usable in other contexts.

In fact `screenSize` (in DPI adjusted pixels) might be a common uniform that the `project` shader package could prepare by default (alongside `projectionMatrix` etc). Maybe longer term, your pixel space calculation function could also be added to that package (if general enough).

Not necessarily required for this diff, just sharing my thoughts... @shaojingli

Pessimistress: I would love it if the screen dimensions are provided as default uniforms. `IconLayer` and `ScreenGridLayer` use them as well.

PR #353, opened 2017-02-15
ScatterplotLayer perf improvement + lineWidth fix

Pessimistress:
- Switch from drawing 16-sided polygon to square + fragment shader, about 3x faster
- Fix lineWidth

@ibgreen @shaojingli The `strokeWidth` here is also in pixels, which is consistent with v3 but not the new PathLayer. We probably need another pass of API audit.

### DIFF
--- a/src/layers/core/scatterplot-layer/scatterplot-layer-fragment.glsl
+++ b/src/layers/core/scatterplot-layer/scatterplot-layer-fragment.glsl
@@ -25,7 +25,11 @@ precision highp float;
 #endif
 
 varying vec4 vColor;
+varying vec2 uv;
 
 void main(void) {
+  if (length(uv) > 1.0) {
+    discard;
+  }
   gl_FragColor = vColor;
 }
--- a/src/layers/core/scatterplot-layer/scatterplot-layer-fragment.glsl
+++ b/src/layers/core/scatterplot-layer/scatterplot-layer-fragment.glsl
@@ -24,17 +24,15 @@
 precision highp float;
 #endif
 
-uniform float drawOutline;
-
 varying vec4 vColor;
-varying vec2 uv;
-varying float strokeWidthRatio;
+varying vec2 unitPosition;
+varying float innerUnitRadius;
 
 void main(void) {
 
-  float distToCenter = length(uv);
+  float distToCenter = length(unitPosition);
 
-  if (distToCenter <= 1.0 && (drawOutline == 0.0 || distToCenter >= 1.0 - strokeWidthRatio)) {
+  if (distToCenter <= 1.0 && distToCenter >= innerUnitRadius) {
     gl_FragColor = vColor;
   } else {
     discard;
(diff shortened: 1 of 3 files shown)


src/layers/core/scatterplot-layer/scatterplot-layer.js
```diff
@@ -61,54 +61,37 @@ export default class ScatterplotLayer extends Layer {
...
-
+    const positions = [-1, -1, 0, -1, 1, 0, 1, 1, 0, 1, -1, 0];
     const shaders = assembleShaders(gl, this.getShaders());
 
     return new Model({
       gl,
       id: this.props.id,
       vs: shaders.vs,
       fs: shaders.fs,
+      // timerQueryEnabled: true,
```
shaojingli: probably just remove this?

src/layers/core/scatterplot-layer/scatterplot-layer-fragment.glsl
```diff
@@ -24,8 +24,19 @@
...
 varying vec4 vColor;
+varying vec2 uv;
+varying float strokeWidthRatio;
 
 void main(void) {
-  gl_FragColor = vColor;
+
+  float distToCenter = length(uv);
+
+  if (distToCenter <= 1.0 && (drawOutline == 0.0 || distToCenter >= 1.0 - strokeWidthRatio)) {
```
ibgreen: Instead of a `strokeWidthRatio` varying, maybe just precalculate `innerUnitRadius` defaulting to `0` and  do 
```
if (distToCenter <= 1.0 && distToCenter >= innerUnitRadius) {
```
A little more self documenting and possibly a tad more performant.

src/layers/core/scatterplot-layer/scatterplot-layer-vertex.glsl
```diff
@@ -32,16 +32,22 @@ uniform float radius;
...
+varying vec2 uv;
+varying float strokeWidthRatio;
 
 void main(void) {
   // Multiply out radius and clamp to limits
   float radiusPixels = clamp(
     project_scale(radius * instanceRadius),
     radiusMinPixels, radiusMaxPixels
-  );
+  ) + drawOutline * strokeWidth / 2.0;
```
ibgreen: This addition... are we adding the outline outside of the radius? 
Is that what we want, and if so is the doc updated to clarify this?

Pessimistress: The outline stoke is centered at the radius.

ibgreen: > The outline stoke is centered at the radius

OK that is a reasonable choice. Probably still reasonable to mention it in the docs.

Also, is there a risk that a wide stroke overflows the containing square and gets truncated due to fragment shader not being called? Or are you scaling down the entire thing to make sure it fits?

Pessimistress: I scale the whole thing to include radius + half stroke width. I'm renaming this value to `outerRadiusPixels` for clarity.

PR #355, opened 2017-02-16
Blinn-Phong lighting for our 3D layers

shaojingli:
Implemented a brand new lighting system. Also unified the lighting shaders across layers to make it look similar. 

I'm doing the lighting calculation in the world space since the viewMatrix from Mapbox is non-linear. It scales significantly along the z direction so the entire camera space is squeezed... I tried to fix this, but after the fix, the flickering of 32-bit layers become more prominent (close to unusable) so I choose the current approach

The shader assembling code is not final. We will still make changes to it before the v4 release to have a better shader management system.

@ibgreen @Pessimistress @heshan0131 @gnavvy

feb-16-2017 13-09-40


src/layers/core/grid-layer/grid-layer-vertex.glsl
```diff
@@ -16,25 +15,12 @@ attribute vec3 instancePickingColors;
...
-// Lighting constants
-const vec3 ambientColor = vec3(0.8, 0.8, 0.8);
-const vec3 pointLocation = vec3(1.5, 1.5, 5.);
-const vec3 pointColor = vec3(0.7, 0.7, 0.7);
-const vec3 pointSpecularColor = vec3(0.6, 0.6, 0.6);
-const float shininess = 3.;
-const float pointLightAmbientCoefficient = 1.;
-
 // A magic number to scale elevation so that 1 unit approximate to 100 meter
 #define ELEVATION_SCALE 80.
```
shaojingli: @heshan0131 can we somehow remove this magic number or put it to getElevations() / getPositions()... I believe right now all the height of all our 3d layers are is in meters. But in hexagon and grid layers, we are providing height in this 1 unit = 100 meters way

src/layers/core/grid-layer/grid-layer-vertex.glsl
```diff
@@ -16,25 +15,12 @@ attribute vec3 instancePickingColors;
...
 uniform float lonOffset;
 uniform float latOffset;
 uniform float opacity;
 
-// Lighting constants
-const vec3 ambientColor = vec3(0.8, 0.8, 0.8);
-const vec3 pointLocation = vec3(1.5, 1.5, 5.);
-const vec3 pointColor = vec3(0.7, 0.7, 0.7);
-const vec3 pointSpecularColor = vec3(0.6, 0.6, 0.6);
-const float shininess = 3.;
```
shaojingli: I removed the shininess... I feel that it might be better for us to hardcode it until we have a proper materials system for our users to use

src/layers/core/grid-layer/grid-layer-vertex.glsl
```diff
@@ -43,41 +29,42 @@ varying vec4 vColor;
...
 
-
   // cube gemoetry vertics are between -1 to 1, scale and transform it to between 0, 1
-  vec2 ptPosition = instancePositions.xy + vec2((positions.x + 1.0 ) * lonOffset / 2., (positions.y + 1.) * latOffset / 2.);
+  vec2 ptPosition = instancePositions.xy + vec2((positions.x + 1.0 ) * lonOffset / 2.0, (positions.y + 1.0) * latOffset / 2.0);
 
   vec2 pos = project_position(ptPosition);
 
-  // if 3d not enabled set elevation to 0
-  float elevation =  mix(0., project_scale(instancePositions.w  * (positions.z + 1.) * ELEVATION_SCALE), extruded);
```
shaojingli: I changed all mix() function to if() for better code readability. I don't think there will be any difference in performance... I'll do the perf test later using my latest perf query and see.

src/layers/core/grid-layer/grid-layer.js
```diff
@@ -95,15 +102,20 @@ export default class GridLayer extends Layer {
     });
   }
 
-  draw({uniforms}) {
-    const {extruded, latOffset, lonOffset} = this.props;
+  updateUniforms() {
```
shaojingli: Added updateUniforms() as Hexagon layer

src/layers/core/grid-layer/grid-layer-vertex.glsl
ibgreen: I believe that this is being addressed in @heshan0131 's coming Grid and HexagonLayer update.

src/layers/core/grid-layer/grid-layer-vertex.glsl
ibgreen: Would be good to get a solid conclusion on `if` vs `mix` (solid = perf testing on a couple of platforms) as it could help us establish guidelines for how we write shaders in general. So far we have preferred `mix` over `if` but we don't have any data to back it up.

src/layers/core/grid-layer/grid-layer-vertex.glsl
```diff
@@ -43,41 +29,42 @@ varying vec4 vColor;
...
-  vec3 lightWeightedColor = mix(vec3(1.), lightWeighting, extruded) * (instanceColors.rgb / 255.);
-  vec4 color = vec4(lightWeightedColor, instanceColors.a * opacity / 255.);
+    vec3 lightWeightedColor = lightWeight * (instanceColors.rgb / 255.);
+    vec4 color = vec4(lightWeightedColor, instanceColors.a * opacity / 255.);
+    vColor = color;
 
-  vec4 pickingColor = vec4(instancePickingColors / 255., 1.);
+  } else {
 
-  vColor = mix(color, pickingColor, renderPickingBuffer);
```
shaojingli: I'm benchmarking mix() vs if() right now. I hope to get a solid conclusion on this today.

However, at least we should remove this mix because if we are rendering picking buffer, we should skip the expansive lighting calculations for sure.

shaojingli:
I'm benchmarking mix vs if right now. 

However, for these changes specific to the lighting shaders, "if" should be the way to go as the expansive lighting calculations are completely unnecessary when we are rendering the picking buffer or the entire structure is not extruded.

ibgreen:
> So the actual execution of the shader is slightly faster (or similar). In addition, replacing "mix" with "if" saves us from doing lighting in the picking pass or non-extruded geometry rendering. 

Sounds good. At some point it would be cool to take timings with
1) lighting on
2) lighting off
3) lighting code completely removed from shader. 

> So I'll say the new lighting code is an overall improvement.

Awesome. Better functionality and faster. Sounds like we should just drop the old implementation then.

PR #356, opened 2017-02-17
Fixes sample-layers example

ibgreen:
@Pessimistress @gnavvy @shaojingli 

Idea is to have a separate example that contains a bunch of sample layers (with the same layer property browsing UI as the `main` example has for the core layers). 

At the moment it copies a number of files from the `main` example, in order for this `layer-samples` example to remain as stand alone as possible.

### DIFF
--- a/examples/sample-layers/layers/multi-color-path-layer/multi-color-path-layer.js
+++ b/examples/sample-layers/layers/multi-color-path-layer/multi-color-path-layer.js
@@ -49,3 +49,5 @@ colors returned doesn't match the number of points in the path. Index ${index}`)
     attribute.value = colors;
   }
 }
+
+MultiColorPathLayer.layerName = 'MultiColorPathLayer';
--- a/examples/sample-layers/layers/multi-color-path-layer/multi-color-path-layer.js
+++ b/examples/sample-layers/layers/multi-color-path-layer/multi-color-path-layer.js
@@ -1,5 +1,6 @@
 import {PathLayer} from 'deck.gl';
 
+// PathLayer subclass by @dcposch
 export default class MultiColorPathLayer extends PathLayer {
   calculateColors(attribute) {
     const {data, getColor} = this.props;
--- a/examples/sample-layers/layers/multi-color-path-layer/multi-color-path-layer.js
+++ b/examples/sample-layers/layers/multi-color-path-layer/multi-color-path-layer.js
@@ -1,0 +1,50 @@
+import {PathLayer} from 'deck.gl';
+
+export default class MultiColorPathLayer extends PathLayer {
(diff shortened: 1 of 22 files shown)


examples/sample-layers/layers/multi-color-path-layer/multi-color-path-layer.js
```diff
@@ -0,0 +1,51 @@
+import {PathLayer} from 'deck.gl';
+
+// PathLayer subclass by @dcposch
+export default class MultiColorPathLayer extends PathLayer {
```
ibgreen: @dcposch FYI - Using your idea as an example for how to subclass layers, as discussed in #336

dcposch: 👍

examples/sample-layers/package.json
```diff
@@ -4,10 +4,13 @@
...
     "babel-polyfill": "^6.16.0",
-    "deck.gl": "3.1.0-beta.16",
+    "deck.gl": "^4.0.0-beta.2",
+    "extrude-polyline": "^1.0.6",
     "immutable": "^3.8.1",
-    "luma.gl": "3.0.0-alpha.4",
+    "lodash.flattendeep": "^4.4.0",
+    "luma.gl": "3.0.0-beta.9",
     "react": "^15.4.1",
+    "react-autobind": "^1.0.6",
```
contra: BTW - this module is unmaintained and has [some issues](https://github.com/cassiozen/React-autobind/issues/3)

I think most people are rolling w/ `core-decorators` `autobind` instead

ibgreen: > BTW - this module is unmaintained and has some issues

Ah, good catch.

> I think most people are rolling w/ core-decorators autobind instead

For what it's worth, in v4 we've moved away from decorators and other advanced JS features in favor of a pure ES6 code base that can run without transpilation (modulo JSX) on the latest browsers. 

Ironically, we actually have our own small `autobind` implementation inside deck.gl (in order to minimize package dependencies) but we don't export it from the lib (in order to keep a small API surface).

It would of course be good if our examples follow best practices. Will have to investigate if core-decorators can be used without actual decorator syntax or if there is another suitable replacement. I guess we could just manually bind the methods in the constructor.

ibgreen:
> But even with that, it generates "gl.INVALID_OPERATION" from draws.

Yes it looks like we will need to update @dcposch code to match the latest `PathLayer`. Since this PR is already a big improvement to the `sample-layers` example, I'll get it landed and will do that fix in a separate PR.

Also, I added a color array generator to trigger the multi color functionality in the new layer.

PR #357, opened 2017-02-17
Update doc pages order. Add initial doc on customizing existing layers.

ibgreen:
@gnavvy @Pessimistress @shaojingli @abmai 

This PR attempts to get us close to a final doc structure for v4.

It adds a new doc about customizing existing layers. This is still in draft and will need one more revision.


docs/v4.0/layers/README.md
```diff
@@ -2,31 +2,50 @@
...
+code, but they are not listed here in the official documentation.
+
+To use one of the sample layers an application simply needs to copy it into
+its own source tree, as these layers are not `export`ed by deck.gl.
 
-**Warning**:
-The sample layers are not as actively supported as core layers,
-and could even have breaking changes between minor releases of deck.gl.
+Remarks:
+* Sample layers are not as actively supported as core layers,
```
shaojingli: I believe we should use more positive words here. like "Sample layers are demoed here as examples to show what can be achieved by modifying existing core layers of deck.gl. They are not part of the deck.gl and are intended to be used as the final products instead of building blocks of other layers.

ibgreen: > I think we rephrase this section and make it a bit more positive.

Good point! Dropping the remark.

PR #358, opened 2017-02-17
Fix webpack demo hot-reload

Apercu:


demo/webpack/package.json
```diff
@@ -0,0 +1,6 @@
+{
```
ibgreen: Delete this file?

demo/webpack/prod.js
```diff
@@ -1,12 +1,10 @@
-const devConfig = require('./dev');
+const config = require('./config');
```
ibgreen: This is now so small that it almost feels like we could use `--env.prod` and `--env.dev` webpack options instead of having several config files...

demo/webpack/package.json
Apercu: yes

PR #362, opened 2017-02-21
Added Point Density Grid Layer (happy to accept other better naming)

heshan0131:
density_grid

- DensityGridLayer is a layer wrapper that aggregates points into an array of grid cells
- It takes in an array of colors to render cell color based on number of points in each cell
- It override the onHover event of the subLayer and add aggregated cell info to hover.info


docs/v4.0/layers/point-density-grid-layer.md
```diff
@@ -0,0 +1,63 @@
...
+Inherits from all [Base Layer](/docs/layers/base-layer.md) properties.
+
+
+## Layer-specific Properties
+
+##### `cellSize` (Number, optional)
+
+- Default: `1.0`
+
+Size of each cell in kilometers
```
ibgreen: I think we should standardize on meters. Just set the default to 1000 instead.

examples/main/src/layer-examples.js
```diff
@@ -209,6 +210,26 @@ const GridLayerExample = {
...
+const PointDensityGridLayerExample = {
+  layer: PointDensityGridLayer,
+  props: {
+    id: 'pointDensityGridLayer',
+    data: dataSamples.points,
+    extruded: true,
+    pickable: true,
+    opacity: 1,
+    cellSize: 0.3,
+    lightSettings: {
```
ibgreen: For now, should there be one global lightSettings object in this file, so that all layers in the example get the same lighting?

@pessimistress For longer term, I have suggested that we define a deck.gl `Effect` to contain the global lighting props, then we could have a single control panel for the global lighting props just like we have a control panels for all layer props. Will provide some more detail on this idea separately.

examples/main/src/layer-controls.js
```diff
@@ -43,6 +43,10 @@ export default class LayerControls extends PureComponent {
       max = 100;
     }
 
+    if (settingName === 'cellSize') {
```
ibgreen: Feels like we really will need a propTypes system soon...

src/layers/core/point-density-grid-layer/index.js
```diff
@@ -0,0 +1 @@
+export {default as default} from './point-density-grid-layer';
```
ibgreen: FWIW, if you look at the top level `index.js` files, we've started exporting layers directly from xyz-layer/xyz-layer.js, that way we could drop all these one line index files.

src/layers/core/point-density-grid-layer/point-density-grid-layer.js
```diff
@@ -0,0 +1,138 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
```
ibgreen: Date?

src/layers/core/point-density-grid-layer/point-density-grid-layer.js
```diff
@@ -0,0 +1,138 @@
...
+  elevationRange: defaultElevationRange
+};
+
+function noop() {}
+
+function _needsReProjectPoints(oldProps, props) {
+  return oldProps.cellSize !== props.cellSize;
+}
+
+function _ordinalScale(domain, range, value) {
```
ibgreen: I suspect that `scale` support is something we will want to think through and standardize across layers pretty soon. For now placing these functions in a utility file (even with the same directory) would be a decent first step.

src/layers/core/point-density-grid-layer/point-density-grid-layer.js
```diff
@@ -0,0 +1,138 @@
...
+
+function _linearScale(domain, range, value) {
+
+  return (value - domain[0]) / (domain[1] - domain[0]) * (range[1] - range[0]) + range[0];
+}
+
+export default class PointDensityGridLayer extends Layer {
+  initializeState() {
+    this.state = {
+      gridOffset: {latOffset: 0.0089, lonOffset: 0.0113},
```
ibgreen: Here we are still tied to a specific projection mode. Even if we need more time to generalize the calculations, it would be good if we could avoid tying the names to lat/lon.

src/layers/core/point-density-grid-layer/point-density-grid-layer.js
```diff
@@ -0,0 +1,138 @@
...
+    const {elevationRange} = this.props;
+    const elevationDomain = this.props.elevationDomain || [0, this.state.countRange[1]];
+    return _linearScale(elevationDomain, elevationRange, cell.count);
+  }
+
+  renderLayers() {
+    const {id} = this.props;
+
+    // Override user's onHover and onClick props
+    const handlers = {
```
ibgreen: In my opinion defining this `handlers` variable does not make the code easier to read.

As opposed to just listing these in the `Object.assign` below with a comment in between.

We don't have multiple sublaysers like in `GeoJsonLayer`, so I'd consider merging the two objects.

src/layers/core/point-density-grid-layer/grid-aggregator.js
```diff
@@ -0,0 +1,112 @@
+const R_EARTH = 6378;
```
ibgreen: My vision here is that we would use the `Viewport` classes to help us define a grid in any coordinate mode (lng/lat, meter offsets, identity). I realize that would take a bit of work. In the mean time, trying to avoid tying variable names to lng/lat/kilometers etc would be a good first step.

docs/v4.0/layers/point-density-grid-layer.md
```diff
@@ -0,0 +1,63 @@
+<p align="right">
+  <img src="https://img.shields.io/badge/extruded-yes-blue.svg?style=flat-square" alt="64-bit" />
+</p>
+
+# PointDensityGridLayer
+
+The PointDensityGridLayer renders a grid heatmap based on an array of points.
+It takes the constant size all each cell, projects points into cells. The color 
+and height of the cell is scaled by number of points it contains.
```
ibgreen: At the moment my understanding is that this layer only works with `COORDINATE_SYSTEM.LNG_LAT` so we need to make this restriction quite clear here.

docs/v4.0/layers/point-density-grid-layer.md
```diff
@@ -0,0 +1,63 @@
...
+
+
+<div align="center">
+  <img height="300" src="/demo/src/static/images/point-density-grid.png" />
```
ibgreen: - The gif is cool but pretty big @ 316K demo/src/static/images/point-density-grid.js 
- Also, what's up with the file type - should it have another type (looks like it's called .js above?). 
- General question - Should we consider having smaller fast loading screenshot for the docs, or should we go with gifs for more layers?

heshan0131: This is just a static image actually. The gif is only for this pull request. Yea you are right about the name

src/layers/core/point-density-grid-layer/point-density-grid-layer.js
heshan0131: changed to xOffset and yOffset. I would like to revisit this once I made the grid-layer truely applicable to different projection mode

src/layers/core/point-density-grid-layer/point-density-grid-layer.js
heshan0131: agreed. less assign => easier to read

src/layers/core/point-density-grid-layer/point-density-grid-layer.js
heshan0131: didn't know it was a thing, probably also need to change it in grid and hexagon layer

heshan0131: 2016?

PR #363, opened 2017-02-21
Add PointCloudLayer

Pessimistress:
- add PointCloudLayer, renders LiDAR data with `position` `normal` and `color` properties
- add doc
- replace scatterplot (meter) example with point cloud (meter) example

### DIFF
--- a/src/layers/core/point-cloud-layer/point-cloud-layer.js
+++ b/src/layers/core/point-cloud-layer/point-cloud-layer.js
@@ -1,0 +1,140 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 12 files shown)


contra:
@Pessimistress Do you have a screenshot of what this looks like? Curious to check it out, sounds really cool.

src/layers/core/point-cloud-layer/point-cloud-layer.js
```diff
@@ -0,0 +1,140 @@
...
+    const {radius, lightSettings} = this.props;
+    const {viewport: {width, height}} = this.context;
+    this.state.model.render(Object.assign({}, uniforms, {
+      radius,
+      screenSize: [width, height]
+    }, lightSettings));
+  }
+
+  _getModel(gl) {
+    // a triangle that minimally cover the unit circle
```
ibgreen: Just a though - A circle-fitting triangle generates more fragments (to discard) than a a circle fitting square - so maybe a square is actually faster?

Pessimistress:
@contra @ib screenshot from the example app:
screen shot 2017-02-21 at 2 57 00 pm

src/layers/core/point-cloud-layer/point-cloud-layer.js
Pessimistress: I'll run some perf experiments.

Pessimistress: Rendering triangles is about 1/6 faster than squares.

ibgreen: > Rendering triangles is about 1/6 faster than squares.

@pessimistress Interesting! Out of curiosity, did you try different `radius` values? Intuitively it would make sense if: 
* small on-screen circles (that rasterize into few frags) benefitted from a primitive with less verts to process since the number of wasted fragment shader invocations will be small (this is consistent with your measurements),
* big on-screen circles that rasterize into a lot of frags benefitted from a primitive with less wasted frags. 

For the point cloud layer `radius` would typically be small, then a single triangle should be the right choice here.

But for `ScatterplotLayer` it might make sense to make another measurement before changing to a triangle primitive.

src/layers/core/point-cloud-layer/point-cloud-layer.js
```diff
@@ -0,0 +1,140 @@
...
+import {readFileSync} from 'fs';
+import {join} from 'path';
+
+const DEFAULT_COLOR = [0, 0, 0, 255];
+
+const defaultProps = {
+  getPosition: x => x.position,
+  getNormal: x => x.normal,
+  getColor: x => x.color || DEFAULT_COLOR,
+  radius: 10,  //  point radius in pixels
```
ibgreen: Should we call this `radiusPixels`?

shaojingli:
@ibgreen you are right. Model matrix should be transferred separately to the GPU and multiplied to normals. I'll create a fix for that.

PR #364, opened 2017-02-22
64-bit polygon layer

shaojingli:
Implemented 64-bit polygon layer and also 64-bit geojson layer (partially, as 64-bit path layer will come later)

The 64-bit version of polygon layer is implemented differently as other 64-bit layers as the polygon tessellator uses a fp64 flag to switch between 32-bit and 64-bit geometries. 

I'm investigating if this approach is the way to go for unifying 64 and 32 bit line, arc and scatterplot layers.

@ibgreen @gnavvy @Pessimistress

### DIFF
--- a/examples/main/src/layer-examples.js
+++ b/examples/main/src/layer-examples.js
@@ -16,6 +16,7 @@ import {
   ScatterplotLayer64,
   ArcLayer64,
   LineLayer64,
+  GeoJsonLayer64,
 
   ChoroplethLayer,
   ChoroplethLayer64,
@@ -67,6 +68,31 @@ const IconLayerExample = {
 
 const GeoJsonLayerExample = {
   layer: GeoJsonLayer,
+  props: {
+    id: 'geojsonLayer',
+    data: dataSamples.geojson,
+    // TODO change to use the color util when it is landed
+    getPointColor: f => parseColor(f.properties['marker-color']),
+    getPointSize: f => MARKER_SIZE_MAP[f.properties['marker-size']],
+    getStrokeColor: f => {
+      const color = parseColor(f.properties.stroke);
+      const opacity = f.properties['stroke-opacity'] * 255;
+      return setOpacity(color, opacity);
+    },
+    getStrokeWidth: f => f.properties['stroke-width'],
+    getFillColor: f => {
+      const color = parseColor(f.properties.fill);
+      const opacity = f.properties['fill-opacity'] * 255;
+      return setOpacity(color, opacity);
+    },
+    strokeWidth: 10,
+    strokeWidthMinPixels: 1,
+    pickable: true
+  }
+};
+
+const GeoJsonLayer64Example = {
+  layer: GeoJsonLayer64,
   props: {
     id: 'geojsonLayer',
     data: dataSamples.geojson,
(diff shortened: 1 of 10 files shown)


examples/main/src/layer-examples.js
```diff
@@ -88,6 +89,31 @@ const GeoJsonLayerExample = {
   }
 };
 
+const GeoJsonLayer64Example = {
+  layer: GeoJsonLayer64,
+  props: {
```
ibgreen: Just a thought, we could update this so that each example had a

   `layer`: 
   `layer64`: 

To avoid a never ending list. But would take some UI work. Maybe wait until we fully integrate the two variants.

src/layers/core/polygon-layer/polygon-layer-64-vertex.glsl
```diff
@@ -0,0 +1,72 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+#define SHADER_NAME polygon-layer-vertex-shader
+
+attribute vec3 positions;
+attribute vec2 positions64xyLow;
```
ibgreen: Have you separated into high and low attributes? That's should be a big step towards unification.

In that case I think we could simplify naming and just call the attributes:

 `positions` and `positions64` (or `positions64xy` if we want to emphasize it is of a different length)

My thinking is that the fact that the attribute is named `64` implicitly means that it contains the additional precision for a corresponding 32 bit attribute, and eventually all our layers follow this convention.

src/layers/core/polygon-layer/polygon-layer.js
```diff
@@ -82,6 +87,12 @@ export default class PolygonLayer extends Layer {
       pickingColors: {size: 3, type: GL.UNSIGNED_BYTE, update: this.calculatePickingColors, noAlloc}
     });
     /* eslint-enable max-len */
+
+    if (this.props.fp64) {
```
ibgreen: +1.  Unfortunately we can't remove it at the moment if switching back but that is a rather unusual use case. Maybe leave a comment to that effect.

src/layers/fp64/geojson-layer/geojson-layer-64.js
```diff
@@ -0,0 +1,134 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
```
ibgreen: Assuming this 64 bit version can be deleted once we apply the same configurability to Scatterplot and Path as you have done to PolygonLayer.

src/layers/core/polygon-layer/polygon-layer-64-vertex.glsl
shaojingli: Yes. I forgot to mention that - I separated the 64-bit values into 2 attributes so that no redundant happens for vertex attributes. I expect to do the same for other 64-bit layers.

PR #367, opened 2017-02-24
Added point density hexagon layer (That's it, the name is sealed)

heshan0131:
point-density-hexagon

Changes included
**PointDensityHexagonLayer**
- PointDensityHexagonLayer is a layer wrapper that aggregates points into hexagonal bins
- It takes an array of points and the radius of hexagon bin in meters
- Color and height of each hexagon are scaled by the density of the bins, that is, the number of points it contains.
- By default PointDensityHexagonLayer uses d3-hexbin as hexagon aggregator it also takes hexagonAggregator as a prop

**HexagonLayer**
- HexagonLayer now also takes in `radius` in meter and `angle` in radian as props to calculate primitive hexagon. If `radius` and `angle` are provided, `hexagonVertices` will be ignored.
- Change `radiusScale` to `coverage` for clarity. `coverage` will be clamped between 0 - 1
- Added `elevationScale` to scale elevation. 

**GridLayer**
- Added `elevationScale` to scale elevation. 

**LayerControl**
- Added ColorRangePicker


ibgreen:
> By default PointDensityHexagonLayer uses d3-hexbin as hexagon aggregator it also takes hexagonAggregator as a prop

This is really cool. We should consider whether deck.gl should include d3-hexbin at all, or if that should just be included in the example. In that case the HexagonLayer would not be usable on its own since no default aggregator.

> HexagonLayer now also takes in radius in meter and angle in radian as props to calculate primitive hexagon. 

> If radius and angle are provided, hexagonVertices will be ignored.

I might have done it the other way (i.e. `hexagonVertices` would override `angle` and `radius`) as it seems a little more intuitive for a new user to have the more complicated and powerful API be the optional one.

> Change `radiusScale` to `coverage` for clarity. coverage will be clamped between 0 - 1

Not completely onboard with this. I'll just drop a bunch of thoughts and we can discuss offline:
* First a nit: per the API guidelines, all multiplicative properties should have the suffix `...Scale` (unless they are established concepts like `opacity`). Accordingly this should be called `coverageScale`?
* But since we have `radiusScale` in other layers and this essentially does the same thing, I think there is a real value in calling it the same thing.
* Not sure why we would want to enforce a clamp on this. The general hexagon layer does not enforce centroid spacing, so larger than 1 hexagons could potentially be useful. Also other layers that support radius scale support overlap and overdraw, so not doing it here could be more surprising than helpful in some cases.
* For the demo, I do see that want to restrict the range of the slider. We could do that without renaming the attribute. We could check the `layerName` when we determine the type of a prop. We could add a `propTypes` field to the example objects in `layer-examples.js`.

> Added elevationScale to scale elevation.

Note that elevation scaling could also be achieved using the `modelMatrix` prop. 
* That said, having a separate prop for this is certainly easier for casual users who might not be well versed in linear algebra.
* If we add `elevationScale` to these layers we should at least consider whether we want to update our API guidelines to state that all extruded layers should support `elevationScale`

> Added ColorRangePicker

* This is a really nice addition. At some point we should consider if there are there other layers for which color ranges would make sense?

examples/main/src/layer-examples.js
```diff
@@ -241,14 +242,36 @@ const HexagonLayerExample = {
...
+    id: 'PointDensityHexagonLayer',
+    data: dataSamples.points,
+    extruded: true,
+    pickable: true,
+    radius: 1000,
+    opacity: 1,
+    elevationScale: 0.8,
+    coverage: 1,
+    getPosition: d => d.COORDINATES,
+    lightSettings: {
```
ibgreen: Should we use a global object for `lightSettings` so that all layers in the example get lit the same way?

examples/main/src/layer-examples.js
```diff
@@ -218,10 +219,10 @@ const PointDensityGridLayerExample = {
   props: {
```
ibgreen: We could consider adding a separate `propTypes` object here, instead of adding special cases to the type deducer. When mature this support could be moved into the base `Layer` so that we would support a static `propTypes` field, somewhat similar to React), and this component could just read that field. @Pessimistress 

```
const PointDensityGridLayerExample = {
   layer: PointDensityGridLayer,
   propTypes: {
      coverage: {type: 'range', min: 0, max: 1}
   },
    props: {
```

docs/v4.0/layers/point-density-grid-layer.md
```diff
@@ -37,19 +37,27 @@ Color scale domain, default is set to the range of point counts in each cell.
 
 - Default: <img src="/demo/src/static/images/colorbrewer_YlOrRd_6.png"/></a>
 
-Cell color ranges as an array of colors formatted as `[255, 255, 255]`. Default is colorbrewer `6-class YlOrRd`.
```
ibgreen: Color ranges cannot have opacity? Not a concern, just a question.

shaojingli:
> HexagonLayer now also takes in radius in meter and angle in radian as props to calculate primitive hexagon.
If radius and angle are provided, hexagonVertices will be ignored.

We need to discuss this offline.

> First a nit: per the API guidelines, all multiplicative properties should have the suffix ...Scale (unless they are established concepts like opacity). Accordingly this should be called coverageScale?
But since we have radiusScale in other layers and this essentially does the same thing, I think there is a real value in calling it the same thing.
Not sure why we would want to enforce a clamp on this. The general hexagon layer does not enforce centroid spacing, so larger than 1 hexagons could potentially be useful. Also other layers that support radius scale support overlap and overdraw, so not doing it here could be more surprising than helpful in some cases.

I don't quite get this part. It feels like the "coverage" props shows the ratio between the underlying "binning cell" and actually "rendered cell". Changing it to radiusScale or allow larger than 1 ratio would make this parameter unintuitive...

> Note that elevation scaling could also be achieved using the modelMatrix prop.

That said, having a separate prop for this is certainly easier for casual users who might not be well versed in linear algebra.

Agree. We could decide to use modelMatrix to implement the scaling of heights but probably don't want anyone except power users to directly set it. Exposing elevationScale and/or rotation sounds much better to me. 

> If we add elevationScale to these layers we should at least consider whether we want to update our API guidelines to state that all extruded layers should support elevationScale

We should.

shaojingli:
I feel "binSize" or even "cellSize" as we have in the pointDensityGridLayer is more appropriate than "radius"

ibgreen:
> I don't quite get this part. It feels like the "coverage" props shows the ratio between the underlying "binning cell" and actually "rendered cell". Changing it to radiusScale or allow larger than 1 ratio would make this parameter unintuitive...

OK I see where you are coming from. Since there are two layers involved, different props might make sense for them. The basic `HexagonLayer` does not know anything about binning, or centroids etc, and it uses `radius` as the primary property defining a hexagon. On this layer, things like a `radiusScale` (and even a `getRadius` accessor) might make perfect sense. 

In the `PointDensityHexagonLayer` I agree that the semantics are different and that it could make sense to define a different prop. Where I still struggle is that if `coverage` is less than 1, per the docs this does not affect what is being binned, so the link between the name and semantics of the prop and binning isn't that obvious to me. Somehow I keep feeling that this is a visual prop. 

But in lieu of a better name, I could go with this (we should of course make a call on the `coverage` vs `coverageScale` question)

ibgreen:
> I feel "binSize" or even "cellSize" as we have in the pointDensityGridLayer is more appropriate than "radius"

This direction makes sense if we can find names that are reasonable in both cases. 

And perhaps this could be a good way to attack the prop naming problem discussed above, i.e. clearly separate binning related props and visual props.

docs/v4.0/layers/point-density-grid-layer.md
heshan0131: yes they could, good catch

PR #368, opened 2017-02-24
Add link to old v3 docs in new v4 docs.

ibgreen:
@Pessimistress @apercu @shaojingli

I've created a new branch `3.0-release` from `master`, in preparation for merging dev into master.

Since @apercu has made sure that docs can be browsed directly on github, I just added a link to the [docs on that branch](https://github.com/uber/deck.gl/tree/3.0-release/docs).

screen shot 2017-02-24 at 8 11 21 am

Also renamed `docs/v4.0` directory back to `docs` since in this model we don't need to maintain multiple doc versions in the same tree.

@apercu - the link works when I open in new window but behaves slightly strange when opening in same window, maybe you can take a look?


shaojingli:
> I've created a new branch 3.0-release from master, in preparation for merging dev into master.

Thanks for this Ib. I plan to merge dev back to master after I complete my 64-bit work.

ibgreen:
> Thanks for this Ib. I plan to merge dev back to master after I complete my 64-bit work.

@shaojingli FWIW, I just did two attempts to rebase dev on master. Fairly long way into the rebase, git crashed with a segmentation fault in both cases.

It is weird because the additional commits on master have very little conflicts with dev. It may be that I made a bad choice during my first run, and that my saved "rerere" resolutions kept applying it automatically on my second go, so maybe you'll have better luck.

PR #369, opened 2017-02-24
Api guidelines

ibgreen:
Publishing initial version of API guidelines, to increase transparency of the deck.gl API Audit process. @shaojingli


ibgreen:
@shaojingli a question is if this should be accessible from the docs. I assumed no, since this is more related to contributing, so I kept this in the root folder.

PR #374, opened 2017-02-25
Api guidelines polishing

shaojingli:


API-GUIDELINES.md
```diff
@@ -1,93 +1,83 @@
...
+  makes the effects bring by them clear to the users.
+  Users can choose to learn WebGL and shader programming through deck.gl.
+  This aligns with luma.gl, which exposes one class for every WebGL
   type.
-* **Reactive first** - deck.gl is designed to work in a
+* **Reactive first** - deck.gl is primarily designed to work in a
   [“Reactive architecture”](https://en.wikipedia.org/wiki/Reactive_programming)
-  application. It is great if it works well in other paradigms as well
-  but this is a secondary goal.
+  application. It can be used in other programming paradigms but its support
```
ibgreen: such support comes secondary

API-GUIDELINES.md
```diff
@@ -1,93 +1,83 @@
...
+  come secondary.
 
 ## Design Guidelines
 
-This is a list of specific design rules that have been agreed upon.
-Exceptions to all rules can always be made if there is a good justification
-(which should be documented somewhere if possible, in code comments, docs or
-similar).
+This is a list of design rules that general applies to everything in the framework.
```
ibgreen: generally

API-GUIDELINES.md
```diff
@@ -1,93 +1,83 @@
...
-  similar layers)
-* **Avoid Overly Large/Complicated Layers** - Flexibility, but not at the
-  expense of making our layers large and complicated so that they become
-  hard to subclass and extend - that is a signal that the layer should be
-  divided.
+* **“Small” Set of Core Layers** - A small selection of versatile core layers
+  that are simple to understand and easy to extend through subclassing are provided
+  with deck.gl.
+* **Groups of Related Layers** - The intention is to have groups of layers that
+  each of the groups addresses one type of problem (e.g. general geospatial visualizations,
```
ibgreen: where each group addesses

API-GUIDELINES.md
```diff
@@ -1,93 +1,83 @@
...
+  as examples rather than supported layers with the intention to help users implement similar
+  functionality for themselves.
 
 ## Rules for Naming Props, Attributes, Uniforms and updateTriggers
 
-Summary: Custom layer props usually refer to either attributes or uniforms.
+Summary: Custom layer props usually corresponds to either attributes or uniforms
+of the underlying shaders.
 An accessor (like "getColor") always sets a WebGL attribute (whether a value
 or a function), and a normal prop (like "color") always sets a uniform
```
ibgreen: >  normal prop (like "color") always sets a uniform

This case is not true, e.g. `fp64` flag (there are other examples). So I'd say "usually" rather than "always" in the uniform case.

shaojingli: You are right.. I didn't catch this

PR #375, opened 2017-02-25
Enable lower and upper percentile to filter bins and re calculate color

heshan0131:
hexagon_aggregation_remove_percentile

- Pass in percentile as upperPercentile and lowerPercentile prop
- Sort counts of bins on first pass when data is passed in and save to state. No need to recalculate data when percentile changed.
- User count domain calculate by lower and upper percentile as color domain to calculate color. If colorDomain is passed in as a prop, use it instead.
- Set alpha of bins outside percentile to 0. In vertex shader, if alpha = 0, set hexagon radius to 0. This will prevent hexagon from being rendered and pickable.
- Add upperPercentile and lowerPercentile to getColor updateTrigger, so that only colors are recalculated when percentile changed.


src/layers/core/point-density-hexagon-layer/point-density-hexagon-layer.js
```diff
@@ -24,20 +24,25 @@ import {log} from '../../../lib/utils';
...
 import {defaultColorRange} from '../../../utils/color-utils';
+import {getSortedCounts, getCountRangeFromPercentile} from '../../../utils/aggregation-utils';
 import {pointToHexbin} from './hexagon-aggregator';
 
 const defaultRadius = 1000;
 const defaultCoverage = 1;
 const defaultElevationRange = [0, 1000];
 const defaultElevationScale = 1;
 const defaultAggregator = pointToHexbin;
+const defaultLowerPercentile = 0;
```
ibgreen: Does defining these defaults separately here instead of adding them directly in the `defaultProps` object improve readability? I think it might be cleaner to just drop these.

src/layers/core/hexagon-layer/hexagon-layer-vertex.glsl
```diff
@@ -46,7 +46,7 @@ void main(void) {
     elevation = project_scale(instancePositions.z * (positions.y + 0.5) * ELEVATION_SCALE * elevationScale);
 }
 
-  float dotRadius = radius * clamp(coverage, 0.0, 1.0);
```
ibgreen: Nit: coverage is a uniform, it could be clamped in JS

src/layers/core/point-density-hexagon-layer/point-density-hexagon-layer.js
```diff
@@ -48,13 +53,10 @@ function _needsReProjectPoints(oldProps, props) {
   return oldProps.radius !== props.radius;
 }
 
-function _getCountRange(hexagons) {
-  return [
-    Math.min.apply(null, hexagons.map(bin => bin.points.length)),
-    Math.max.apply(null, hexagons.map(bin => bin.points.length))
-  ];
+function _percentilChanged(oldProps, props) {
```
ibgreen: Typo: percentileChanged.

Also why have this outside the layer instead of a method? It deals with layer specific things, like props so looks like it would fit nicely as a method in the layer.

src/layers/core/point-density-hexagon-layer/point-density-hexagon-layer.js
```diff
@@ -64,14 +66,38 @@ export default class PointDensityHexagonLayer extends Layer {
       props.radius = defaultRadius;
     }
 
+    if (Number.isFinite(props.upperPercentile) &&
```
ibgreen: Not necessary to fix for this PR but still wanted to comment:

That's a lot of error checking code in the constructor. This makes me feel we really need the `propTypes` feature you created in the example to be added to the library.  For now, you could add your own utility function or file here in this layer, use it in the constructor, and we could build on that.

```
props = checkPropTypes(propTypes, props);
```

Also, we are not supposed to modify props (or any object the caller passes in). `checkPropTypes` could take care of that by cloning when changing (and maintain performance by only cloning when issuing a warning and actually replacing a prop).

src/layers/core/point-density-hexagon-layer/point-density-hexagon-layer.js
```diff
@@ -110,15 +150,26 @@ export default class PointDensityHexagonLayer extends Layer {
...
   }
 
   _onGetSublayerElevation(cell) {
-    const {elevationRange} = this.props;
-    const elevationDomain = this.props.elevationDomain || [0, this.state.countRange[1]];
-    return linearScale(elevationDomain, elevationRange, cell.points.length);
+    const {elevationDomain, elevationRange} = this.props;
+    const {sortedCounts} = this.state;
+
+    // elevation is not affected by percentile
```
ibgreen: Should elevation affected by percentile be controlled by a prop?

src/utils/aggregation-utils.js
```diff
@@ -0,0 +1,27 @@
+
+/**
+ * Get an array of object with sorted count and index of bins
+ * @param {Array} bins
+
+ * @return {Array} array of count and index lookup
+ */
+export function getSortedCounts(bins) {
```
ibgreen: I wonder if this would be easier to read and work if it was organized as a class with methods?
```
class SortedBins {
   constructor(bins)
   getRangeFromPercentiles([lo, hi]);
}
```

src/layers/core/point-density-hexagon-layer/point-density-hexagon-layer.js
```diff
@@ -64,14 +66,38 @@ export default class PointDensityHexagonLayer extends Layer {
...
   }
 
   initializeState() {
     this.state = {
       hexagons: [],
-      countRange: null,
-      pickedCell: null
+      pickedCell: null,
+      sortedCounts: null,
+      countDomainByPercentile: null
```
ibgreen: Naming of this variable seems long - can it be shortened? is there any other domain than count?

Would something simple like just `domain` or `valueDomain` work?

src/layers/core/point-density-hexagon-layer/point-density-hexagon-layer.js
```diff
@@ -102,6 +135,13 @@ export default class PointDensityHexagonLayer extends Layer {
     });
   }
 
+  _onPercentileChange() {
+    const [lower, upper] = [this.props.lowerPercentile, this.props.upperPercentile];
```
ibgreen: This seems like a slighly surprising way to decompose. Why not use the standard idiom?
```
const {lowerPercentile, upperPercentile} = this.props;
...getCountRangeFromPercentile(this.state.sortedCounts, [lowerPercentile, upperPercentile])
```

src/layers/core/point-density-hexagon-layer/point-density-hexagon-layer.js
```diff
@@ -110,15 +150,26 @@ export default class PointDensityHexagonLayer extends Layer {
 
   _onGetSublayerColor(cell) {
     const {colorRange} = this.props;
-    const colorDomain = this.props.colorDomain || this.state.countRange;
+    const {countDomainByPercentile} = this.state;
+    const colorDomain = this.props.colorDomain || countDomainByPercentile;
+    const count = cell.points.length;
+
+    const color = ordinalScale(colorDomain, colorRange, count);
```
ibgreen: Maybe some comment about what `ordinalScale` does and why it is used here. Even if we know our D3, always good to leave a hint for the next person to read this.

src/layers/core/point-density-hexagon-layer/point-density-hexagon-layer.js
heshan0131: My thinking is any method that doesn't require the `this` context can be a static method that's outside the layer

ibgreen: > My thinking is any method that doesn't require the this context can be a static method that's outside the layer

Understood. That's a good general programming principle. 

It's your call of course, but to explain my thinking further: in this case I do feel that the props are associated with the layer instance, partly because `props` is really just a short hand for `this.props`, and partly because the logic of diffing props does not seem to fit the description of an independent function, it is really a part of the internal logic of the layer. 

That said, if you extracted the individual props from the props objects and passed those instead of the props object, it would be easier for me to convince myself that this is a utility method (but then it probably wouldn't be small and clean).

src/layers/core/point-density-hexagon-layer/point-density-hexagon-layer.js
heshan0131: I think I did that to avoid lint max-len error. Need to switch mode between framework contributor and app developer.

src/layers/core/point-density-hexagon-layer/point-density-hexagon-layer.js
heshan0131: done

src/layers/core/point-density-hexagon-layer/point-density-hexagon-layer.js
heshan0131: Maybe, I find have both color and elevation affected by percentile is very confusing. I would rather wait until user request to add it

src/layers/core/hexagon-layer/hexagon-layer-vertex.glsl
heshan0131: done

ibgreen:
@heshan0131 Since we have now moved development to master, to minimize rebase problems, I recommend opening up a new PR from latest master. 

Since things are little easier if this a single commit, you can still land this PR on dev (to squash it), or manually do a `rebase -i` to squash your commits here into a single commit. 

Then create a new branch off latest master, and cherry-pick this commit into that branch, and open a new PR.

shaojingli:
Yeah. Shan, please rebase this PR to dev, squash and merge it to dev, and then cherry-pick to apply it on our master and open a new PR against master

heshan0131:
Create https://github.com/uber/deck.gl/pull/403 to fix conflict

PR #376, opened 2017-02-25
Add non-geospatial example (graph explorer)

Pessimistress:
An advanced example of:
- composite layer (`PlotLayer`)
- multiple primitives in one layer (`AxesLayer`)
- non-geospatial viewport

screen shot 2017-02-24 at 5 32 57 pm

@ibgreen need review on the change in src/


demo/src/components/demos/plot-layer/surface-vertex.glsl
```diff
@@ -0,0 +1,50 @@
...
+varying float shouldDiscard;
+
+void main(void) {
+
+  // fit into a unit cube that centers at [0, 0, 0]
+  vec4 position_modelspace = vec4((positions.xyz - modelCenter) * modelScale, 1.0);
+  gl_Position = project_to_clipspace(position_modelspace);
+
+  vec4 center_viewspace = project_to_clipspace(vec4(0.0, 0.0, 0.0, 1.0));
+  float fadeFactor = 1.0 - (gl_Position.z - center_viewspace.z) * lightStrength;
```
shaojingli: Very nice attempt to fade the drawing content according to the clip-space depth! 

First a nit, center_viewspace should be renamed to center_clipspace.

Another point you might want to consider is that the clips-space depth is nonlinear and (gl_Position.z - center_viewspace.z) depends on the near and far values when creating the perspective projection matrix.

But I believe this is good for now. You might put more comments to the shader though.

demo/src/components/canvas3d.js
```diff
@@ -0,0 +1,211 @@
...
+    evt.preventDefault();
+    let value = evt.deltaY;
+    // Firefox doubles the values on retina screens...
+    if (firefox && evt.deltaMode === window.WheelEvent.DOM_DELTA_PIXEL) {
+      value /= window.devicePixelRatio;
+    }
+    if (evt.deltaMode === window.WheelEvent.DOM_DELTA_LINE) {
+      value *= 40;
+    }
+    if (value !== 0 && value % 4.000244140625 === 0) {
```
shaojingli: what's this....

Pessimistress: This originated from react-map-gl. Chrome on desktop produces wheel deltas like this, and it's (apparently) an effective way to differentiate between mouse and trackpad input.

src/index.js
```diff
@@ -53,6 +53,9 @@ export {default as ExtrudedChoroplethLayer64} from './layers/deprecated/extruded
 export {default as DeckGL} from './react/deckgl';
 export {default as default} from './react/deckgl';
 
+// Viewports
+export {default as PerspectiveViewport} from './lib/viewports/perspective-viewport';
```
ibgreen: I'd pair this up with the `COORDINATE_SYSTEM` export, they seem related

demo/src/components/canvas3d.js
```diff
@@ -0,0 +1,211 @@
...
+  ];
+}
+
+const ua = typeof window.navigator !== 'undefined' ?
+  window.navigator.userAgent.toLowerCase() : '';
+const firefox = ua.indexOf('firefox') !== -1;
+
+/* Interaction */
+
+export default class Canvas3D extends Component {
```
shaojingli: This should be named as FreeSpaceViewportController or something similar

demo/src/components/demos/plot-layer/grid-vertex.glsl
```diff
@@ -0,0 +1,75 @@
...
+uniform float offset;
+uniform vec4 strokeColor;
+
+varying vec4 vColor;
+varying float shouldDiscard;
+
+float frontFacing(vec3 v) {
+  vec4 p0_viewspace = project_to_clipspace(vec4(0.0, 0.0, 0.0, 1.0));
+  vec4 p1_viewspace = project_to_clipspace(vec4(v, 1.0));
+  return step(p1_viewspace.z, p0_viewspace.z);
```
shaojingli: I believe this is where Xiaoji tries to determine which grid to show and which to hide. We can see by rotating the demo that grids and labels in front of the 3D surface automatically hide itself.

demo/src/components/demos/graph.js
```diff
@@ -0,0 +1,153 @@
+import React, {Component} from 'react';
```
shaojingli: I'm nitpicking here ... Let's change this file to plot.js?

demo/src/components/demos/graph.js
```diff
@@ -0,0 +1,153 @@
...
+import autobind from 'autobind-decorator';
+import {Parser} from 'expr-eval';
+
+import Canvas3D from '../canvas3d';
+import PlotLayer from './plot-layer/plot-layer';
+import DeckGL from 'deck.gl';
+
+export default class GraphDemo extends Component {
```
shaojingli: And also, PlotDemo here

demo/src/components/canvas3d.js
ibgreen: > This should be named as FreeSpaceViewportController or something similar

Is this an "orbit controller" as in THREE.js etc? I.e. we can move around in an orbit always looking towards the center?

demo/src/components/demos/plot-layer/grid-vertex.glsl
ibgreen: > I believe this is where Xiaoji tries to determine which grid to show and which to hide. We can see by rotating the demo that grids and labels in front of the 3D surface automatically hide themselves.

Agreed! My points were mainly that:
1) normally one would get this effect for "free" by flipping on back face culling for this layer.
2) Also, even if face culling doesn't work (e.g. because we are dealing with GL.LINES), I don't immediately see why we would need to do a dual projection to clip space. Just a single viewMatrix projection of the normal and direction checking should be enough?
Why is the z buffer involved? Is it because we didn't provide the viewMatrix as uniform? Or is there a distance element to the hiding logic as well?

Implementing it like this is not common so comments would be really helpful, both to help with review and to help with future maintenance of this code.

PR #377, opened 2017-02-27
Render all picking buffers in one pass

Pessimistress:
- Improve perf by calling `gl.readPixels()` only once
- Breaking: occluded objects can no longer be picked even if on a separate layer.
- Breaking: `onHover` and `onClick` callbacks are invoked by `layer.pick()`. This allows layers to block an event if necessary. Layers that override the default `pick()` method must now call `super.pick()` *after* augmenting the info object, or call the callback functions themselves.

### DIFF
--- a/examples/main/src/app.js
+++ b/examples/main/src/app.js
@@ -19,8 +19,6 @@ import LAYER_CATEGORIES from './layer-examples';
 const MAPBOX_ACCESS_TOKEN = process.env.MAPBOX_ACCESS_TOKEN || // eslint-disable-line
   'Set MAPBOX_ACCESS_TOKEN environment variable or put your token here.';
 
-const noop = () => {};
-
 // ---- View ---- //
 class App extends PureComponent {
   constructor(props) {
@@ -91,17 +89,17 @@ class App extends PureComponent {
     this.setState({settings});
   }
 
+  _onHover(info) {
+    this.refs.infoPanel.onItemHovered(info);
+  }
+
+  _onClick(info) {
+    this.refs.infoPanel.onItemClicked(info);
+  }
+
   _renderExampleLayer(example, settings, index) {
     const {layer: Layer, props, getData} = example;
-    const {infoPanel} = this.refs;
     const layerProps = Object.assign({}, props, settings);
-
-    if (props.pickable) {
-      Object.assign(layerProps, {
-        onHover: infoPanel ? infoPanel.onItemHovered : noop,
-        onClick: infoPanel ? infoPanel.onItemClicked : noop
-      });
-    }
 
     if (getData) {
       Object.assign(layerProps, {data: getData()});
@@ -170,6 +168,8 @@ class App extends PureComponent {
           width={width} height={height}
           {...mapViewState}
           onWebGLInitialized={ this._onWebGLInitialized }
+          onLayerHover={ this._onHover }
+          onLayerClick={ this._onClick }
           layers={this._renderExamples()}
           effects={effects ? this._effects : []}
         />
--- a/examples/main/src/app.js
+++ b/examples/main/src/app.js
@@ -39,7 +39,9 @@ class App extends PureComponent {
         separation: 0,
         rotationZ: 0,
         rotationX: 0
-      }
+      },
+      hoveredItem: null,
+      clickedItem: null
     };
 
     this._effects = [new ReflectionEffect()];
@@ -90,11 +92,11 @@ class App extends PureComponent {
   }
 
   _onHover(info) {
-    this.refs.infoPanel.onItemHovered(info);
+    this.setState({hoveredItem: info});
   }
 
   _onClick(info) {
-    this.refs.infoPanel.onItemClicked(info);
+    this.setState({clickedItem: info});
   }
 
   _renderExampleLayer(example, settings, index) {
@@ -180,7 +182,7 @@ class App extends PureComponent {
   }
 
   render() {
-    const {settings, activeExamples} = this.state;
+    const {settings, activeExamples, hoveredItem, clickedItem} = this.state;
 
     return (
       <div>
@@ -196,7 +198,7 @@ class App extends PureComponent {
             onToggleLayer={this._onToggleLayer}
             onUpdateLayer={this._onUpdateLayerSettings} />
         </div>
-        <LayerInfo ref="infoPanel"/>
+        <LayerInfo ref="infoPanel" hovered={hoveredItem} clicked={clickedItem} />
       </div>
     );
   }
--- a/examples/main/src/layer-examples.js
+++ b/examples/main/src/layer-examples.js
@@ -93,6 +93,7 @@ const GeoJsonLayerExample = {
       const opacity = f.properties['fill-opacity'] * 255;
       return setOpacity(color, opacity);
     },
+    strokeWidthScale: 10,
     strokeWidthMinPixels: 1,
     pickable: true
   }
--- a/examples/main/src/layer-examples.js
+++ b/examples/main/src/layer-examples.js
@@ -93,7 +93,6 @@ const GeoJsonLayerExample = {
       const opacity = f.properties['fill-opacity'] * 255;
       return setOpacity(color, opacity);
     },
-    strokeWidth: 10,
     strokeWidthMinPixels: 1,
     pickable: true
   }
--- a/src/layers/core/geojson-layer/geojson-layer.js
+++ b/src/layers/core/geojson-layer/geojson-layer.js
@@ -97,7 +97,7 @@ export default class GeoJsonLayer extends CompositeLayer {
     // this is called before the onHover/onClick of sublayers
     // pickInfo is used to collect the pick results of all sublayers
     this.state.pickInfos.length = 0;
-    return null;
+    return true;
   }
 
   renderLayers() {
--- a/src/layers/core/geojson-layer/geojson-layer.js
+++ b/src/layers/core/geojson-layer/geojson-layer.js
@@ -53,14 +53,13 @@ const defaultProps = {
   fp64: false
 };
 
-function noop() {}
-
 const getCoordinates = f => get(f, 'geometry.coordinates');
 
 export default class GeoJsonLayer extends CompositeLayer {
   initializeState() {
     this.state = {
       subLayers: null,
+      subLayerCount: 0,
       pickInfos: []
     };
   }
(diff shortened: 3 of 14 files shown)


examples/main/src/layer-examples.js
```diff
@@ -95,7 +95,6 @@ const GeoJsonLayerExample = {
       const opacity = f.properties['fill-opacity'] * 255;
       return setOpacity(color, opacity);
     },
-    strokeWidth: 10,
```
shaojingli: Why we remove this?

src/layers/core/geojson-layer/geojson-layer.js
```diff
@@ -72,27 +72,32 @@ export default class GeoJsonLayer extends Layer {
     }
   }
 
-  _onHoverSublayer(info) {
```
shaojingli: GeoJsonLayer64 still relies on _onHoverSublayer of its base layer so this change of name breaks it. I'll updated it for you.

src/lib/draw-and-pick.js
```diff
@@ -1,29 +1,31 @@
...
+const EMPTY_PIXEL = new Uint8Array(4);
 let renderCount = 0;
 
 export function drawLayers({layers, pass}) {
   log.log(2, `DRAWING ${layers.length} layers`);
 
-  let layerIndex = 0;
-  for (const layer of layers) {
+  let visibleCount = 0;
+  layers.forEach((layer, layerIndex) => {
```
shaojingli: Consider add a comment here: "render layers in normal colors"

src/lib/draw-and-pick.js
```diff
@@ -53,91 +56,88 @@ export function pickLayers(gl, {
...
-    let zOrder = 0;
-
-    for (let i = layers.length - 1; i >= 0; --i) {
-      const layer = layers[i];
-
-      if (layer.props.visible) {
-        layerIndex++;
-      }
-
+    // Clear the frame buffer
```
shaojingli: Add a comment here: "picking process start"

src/lib/draw-and-pick.js
```diff
@@ -53,91 +56,88 @@ export function pickLayers(gl, {
...
+    gl.clear(GL.COLOR_BUFFER_BIT | GL.DEPTH_BUFFER_BIT);
+    // Save current blend settings
+    const oldBlendMode = getBlendMode(gl);
+    // Set blend mode for picking
+    // always overwrite existing pixel with [r,g,b,layerIndex]
+    gl.enable(gl.BLEND);
+    gl.blendFuncSeparate(gl.ONE, gl.ZERO, gl.CONSTANT_ALPHA, gl.ZERO);
+    gl.blendEquation(gl.FUNC_ADD);
+
+    // Render all pickable layers
```
shaojingli: Consider change the comment to "Render all pickable layers in picking colors"

src/lib/draw-and-pick.js
```diff
@@ -53,91 +56,88 @@ export function pickLayers(gl, {
...
+    const pickedLayer = layers[color[3] - 1];
+    const baseInfo = Object.assign(
+      createInfo([x, y], viewport),
+      {
+        devicePixel: [deviceX, deviceY],
+        pixelRatio
+      }
+    );
+
+    layers.forEach(layer => {
```
shaojingli: Didn't really get this piece. Since we already know which layer is picked, why we still want to loop over all layers and send the picking info every unpicked layer to decode?

src/lib/layer.js
```diff
@@ -116,37 +116,29 @@ export default class Layer {
...
   }
 
-  // If state has a model, draw it with supplied uniforms
-  /* eslint-disable max-statements */
-  pick({info, uniforms, pickEnableUniforms, pickDisableUniforms, mode}) {
-    const {gl} = this.context;
-    const {model} = this.state;
+  // called on mouse event.
+  // @returns {bool} handled
+  pick({info, mode}) {
```
shaojingli: Consider change the name of this function to "processPickingInfo". This function on longer "picks" any more - the pick action itself is happening in "drawLayers" now.

src/lib/utils/blend.js
```diff
@@ -0,0 +1,23 @@
+// get current blending settings
+export function getBlendMode(gl) {
```
shaojingli: Good! We should check and restore the original blend mode at the start and end of picking!

However, when profiling, I found this list of getParameter calls takes an large proportion of execution time. Don't attempt to solve it now. This is just a note.

ibgreen: >  I found this list of getParameter calls takes an large proportion of execution time

@shaojingli Yes `getParameter` requires a roundtrip from the GPU and slows things down, just like `getError` does in debug mode. I have a solution in a luma.gl PR, a WebGL state manager that allows the app to push and pop state without ever hitting `getParameter`. Perhaps we could work together on getting this landed soon?

examples/main/src/app.js
```diff
@@ -91,18 +89,18 @@ class App extends PureComponent {
     this.setState({settings});
   }
 
+  _onHover(info) {
+    this.refs.infoPanel.onItemHovered(info);
```
ibgreen: Do we need to use refs here? Doesn't look very "reactive". Are we just trying to avoid setting app state or is there a reason why it needs to be done like this? Ether way, add a comment?

src/layers/core/geojson-layer/geojson-layer.js
```diff
@@ -52,14 +52,14 @@ const defaultProps = {
...
-function noop() {}
-
 const getCoordinates = f => get(f, 'geometry.coordinates');
 
 export default class GeoJsonLayer extends Layer {
+
   initializeState() {
     this.state = {
       subLayers: null,
+      subLayerCount: 0,
```
ibgreen: This looks confusing, next to `subLayers`? Since we can't use `subLayers.length` I assume `subLayers` is something else (e.g. `subLayerData`). Some cleanup of naming could be helpful.

src/layers/core/geojson-layer/geojson-layer.js
```diff
@@ -110,8 +115,8 @@ export default class GeoJsonLayer extends Layer {
 
     // Override user's onHover and onClick props
     const handlers = {
-      onHover: this._onHoverSublayer.bind(this),
-      onClick: noop
+      onHover: this._onPickSublayer.bind(this, 'hover'),
```
ibgreen: Nit: I'd drop the `handlers` object and copy these lines into each layer below. I think the readability would be better (Composition using `Object.assign` doesn't read as well as `...`), but could just be my taste.

We could still bind them here before creating the objects.

src/layers/core/geojson-layer/geojson-layer.js
```diff
@@ -72,27 +72,32 @@ export default class GeoJsonLayer extends Layer {
...
+  }
 
-    Object.assign(opts.info, info, {
-      layer: this,
-      feature: get(info, 'object.feature') || info.object
-    });
+  pick(opts) {
+    // this is called before the onHover/onClick of sublayers
+    // pickInfo is used to collect the pick results of all sublayers
+    this.state.pickInfos.length = 0;
```
ibgreen: Longer term: If this is the way picking needs to be handled for composite layers, it feels like this should be supported by the base `Layer` or by a `CompositeLayer` subclass. Looks too complicated to have to implement this in each composite layer. Let's discuss in person.

src/layers/core/geojson-layer/geojson-layer.js
```diff
@@ -72,27 +72,32 @@ export default class GeoJsonLayer extends Layer {
...
-      this.state.pickInfos = [];
-    }
+      pickInfo = Object.assign({}, pickInfo, {
+        layer: this,
+        feature: get(pickInfo, 'object.feature') || pickInfo.object
+      });
 
-    if (!info) {
-      return;
+      switch (mode) {
```
ibgreen: Nit: I'd probably drop the switch statement, create two methods that call a common method. An argument for this is that hover and click are separate methods everywhere else, this breaks that pattern. Also, as suggested below, the common code might be a candidate for moving up to the base layer.

src/lib/draw-and-pick.js
```diff
@@ -53,91 +56,88 @@ export function pickLayers(gl, {
...
-        // Clear the frame buffer, render and sample
-        gl.clear(GL.COLOR_BUFFER_BIT | GL.DEPTH_BUFFER_BIT);
-        const info = createInfo({
-          layer,
-          pixel: [x, y],
-          devicePixel: [deviceX, deviceY],
-          pixelRatio
-        });
+        // Encode layerIndex with alpha
+        gl.blendColor(0, 0, 0, (layerIndex + 1) / 255);
```
ibgreen: Just curious: Do we need to add 1 to alpha if blending is turned off?

src/lib/draw-and-pick.js
ibgreen: >  Since we already know which layer is picked, why we still want to loop over all layers and send the picking info every unpicked layer to decode?

One tricky part is letting other layers know that they are no longer picked (so that they can remove highlights etc). We either have to remember which layer was picked last time and send a deselect to that layer, or just keep informing all layers that they are not picked. Some comments to this effect would be good.

src/lib/draw-and-pick.js
```diff
@@ -53,91 +56,88 @@ export function pickLayers(gl, {
...
-      unhandledPickInfos.push(info);
-    }
-  }
+      if (!handled) {
+        unhandledPickInfos.push(info);
+      }
+    });
+
+    // restore blend mode
+    setBlendMode(gl, oldBlendMode);
```
ibgreen: @shaojingli Note that an exception during picking leaves us in a bad state here. The proposed new state management in luma.gl would also handle this case.

src/lib/layer.js
```diff
@@ -116,37 +116,29 @@ export default class Layer {
...
+
+    // TODO - selectedPickingColor should be removed?
+    if (mode === 'hover') {
+      const {color} = info;
+      const selectedPickingColor = new Float32Array(3);
+      selectedPickingColor[0] = color[0];
+      selectedPickingColor[1] = color[1];
+      selectedPickingColor[2] = color[2];
+      this.setUniforms({selectedPickingColor});
```
ibgreen: Per discussions with @shaojingli, this should be handled by the JS part of a shader picking shader package. Could change the TODO comment to indicate that.

Pessimistress:
@ibgreen @shaojingli I incorporated your feedbacks in the latest commit. 
Need review: using `LayerManager.context` to remember the last picked color. This allows us to only invoke mouse events on layers that are affected. As a result:
- `onHover` is only called on enter/leave events.
- `onClick` is only called on the picked layer.
- `GeoJSONLayer` can just pass `onHover` and `onClick` props down to sublayers and no longer has to maintain an internal state.

ibgreen:
> Need review: using `LayerManager.context` to remember the last picked color. This allows us to only invoke mouse events on layers that are affected.

@Pessimistress Sounds like a great idea, however would it be possible to do this as a separate PR? It is a bit hard to review as an add-on to this PR.  I think there might be some complications here (e.g. what if the layer data, or list of layers, change between picks etc...)

shaojingli:
> Need review: using LayerManager.context to remember the last picked color. This allows us to only invoke mouse events on layers that are affected. 

I think that this is the correct place to put it. 

a quick question:

> GeoJSONLayer can just pass onHover and onClick props down to sublayers and no longer has to maintain an internal stat

is GeoJSONLayer able to add more info to the picking info object in this model?

Pessimistress:
@ibgreen @shaojingli I'm going to land the approved changes and open a new PR.

PR #378, opened 2017-02-27
initial S2 layer

urosi2453:


examples/sample-layers/layers/s2-layer/s2-layer.js
```diff
@@ -0,0 +1,138 @@
...
+
+  // Cell fill accessors
+  getFillColor: f => f.fillColor || defaultFillColor
+};
+
+export default class S2Layer extends Layer {
+  initializeState() {
+  }
+
+  updateState({oldProps, props, changeFlags}) {
```
ibgreen: Delete

examples/sample-layers/layers/s2-layer/s2-layer.js
```diff
@@ -0,0 +1,138 @@
...
+      getHeight,
+      getColor: getFillColor,
+      extruded,
+      wireframe,
+      updateTriggers: Object.assign({}, this.props.updateTriggers, {
+        getColor: this.props.updateTriggers.getFillColor
+      })
+    }));
+
+    // // Polygon outline or wireframe
```
ibgreen: Let's decide - should we delete or implement these?

examples/sample-layers/layers/s2-layer/s2-utils.js
```diff
@@ -0,0 +1,50 @@
...
+import converter from 'hex2dec';
+
+/**
+ * Given a S2 hex token this function returns cell level
+ * cells level is a number between 1 and 30
+ *
+ * S2 cell id is a 64 bit number
+ * S2 token removed all trailing zeros from the 16 bit converted number
+ * */
+function getLevelFromToken(token) {
```
ibgreen: +1, now we can handle any level!

examples/sample-layers/package.json
```diff
@@ -6,14 +6,16 @@
     "babel-polyfill": "^6.16.0",
     "deck.gl": "^4.0.0-beta.2",
     "extrude-polyline": "^1.0.6",
+    "hex2dec": "^1.0.1",
```
ibgreen: I generally try to avoid trivial dependencies. They tend to add up.

Is there any reason we can't just use [`parseInt(token, 16)`](http://stackoverflow.com/questions/1981760/parse-hex-ascii-into-numbers). If there is, add a comment to the import or call site.

PR #379, opened 2017-02-27
fp64 flag is offered by Geojson layer to dynamically change between 3…

shaojingli:
…2-bit or 64-bit models.

Now GeoJson layer changes between 32-bit and 64-bit versions dynamically through a prop provided to the user: "props.fp64"

To achieve this, 1) The Geojson layer undergos geometry update, attributes changes and shader recompilation when props.fp64 changed; 2) a new method is added to AttributeManager to delete attributes from attribute map.


contra:
@shaojingli Planning on doing this to all layers? I think it would be a nice API choice. Automatically detecting if a viewer supports 64 bit and flipping a flag would be a lot easier.

shaojingli:
@contra yes. that is the current plan

contra:
@shaojingli Sweet - nice choice!

PR #381, opened 2017-02-28
Merge 64-bit and 32-bit scatterplot layer

shaojingli:
The user can switch between two versions using props.fp64.

This PR is on top of #379. It should be merged after #379 got merged.

### DIFF
--- a/examples/main/src/layer-examples.js
+++ b/examples/main/src/layer-examples.js
@@ -14,7 +14,6 @@ import {
   // PolygonLayer,
   PathLayer,
 
-  ScatterplotLayer64,
   ArcLayer64,
   LineLayer64,
 
@@ -336,20 +335,6 @@ const LineLayer64Example = {
   }
 };
 
-const ScatterplotLayer64Example = {
-  layer: ScatterplotLayer64,
-  props: {
-    id: 'scatterplotLayer64',
-    data: dataSamples.points,
-    getPosition: d => d.COORDINATES,
-    getColor: d => [255, 128, 0],
-    getRadius: d => d.SPACES,
-    pickable: true,
-    radiusMinPixels: 1,
-    radiusMaxPixels: 30
-  }
-};
-
 const ChoroplethLayerContourExample = {
   layer: ChoroplethLayer,
   props: {
@@ -435,7 +420,7 @@ const ScatterplotLayerPerfExample = (id, getData) => ({
 });
 
 const ScatterplotLayer64PerfExample = (id, getData) => ({
-  layer: ScatterplotLayer64,
+  layer: ScatterplotLayer,
   getData,
   props: {
     id: `scatterplotLayer64Perf-${id}`,
@@ -443,7 +428,8 @@ const ScatterplotLayer64PerfExample = (id, getData) => ({
     getColor: d => [0, 128, 0],
     // pickable: true,
     radiusMinPixels: 1,
-    radiusMaxPixels: 5
+    radiusMaxPixels: 5,
+    fp64: true
   }
 });
 
@@ -469,9 +455,7 @@ export default {
 
   '64-bit Layers': {
     ArcLayer64: ArcLayer64Example,
-    ScatterplotLayer64: ScatterplotLayer64Example,
-    LineLayer64: LineLayer64Example,
-    GeoJsonLayer64: GeoJsonLayer64Example
+    LineLayer64: LineLayer64Example
   },
 
   'Deprecated Layers': {
--- a/examples/main/src/layer-examples.js
+++ b/examples/main/src/layer-examples.js
@@ -99,32 +99,6 @@ const GeoJsonLayerExample = {
   }
 };
 
-const GeoJsonLayer64Example = {
-  layer: GeoJsonLayer,
-  props: {
-    id: 'geojsonLayer',
-    data: dataSamples.geojson,
-    // TODO change to use the color util when it is landed
-    getPointColor: f => parseColor(f.properties['marker-color']),
-    getPointSize: f => MARKER_SIZE_MAP[f.properties['marker-size']],
-    getStrokeColor: f => {
-      const color = parseColor(f.properties.stroke);
-      const opacity = f.properties['stroke-opacity'] * 255;
-      return setOpacity(color, opacity);
-    },
-    getStrokeWidth: f => f.properties['stroke-width'],
-    getFillColor: f => {
-      const color = parseColor(f.properties.fill);
-      const opacity = f.properties['fill-opacity'] * 255;
-      return setOpacity(color, opacity);
-    },
-    strokeWidth: 10,
-    strokeWidthMinPixels: 1,
-    pickable: true,
-    fp64: true
-  }
-};
-
 const GeoJsonLayerExtrudedExample = {
   layer: GeoJsonLayer,
   props: {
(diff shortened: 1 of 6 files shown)


examples/main/src/layer-examples.js
```diff
@@ -122,7 +120,8 @@ const GeoJsonLayer64Example = {
     },
     strokeWidth: 10,
     strokeWidthMinPixels: 1,
-    pickable: true
+    pickable: true,
+    fp64: true
```
ibgreen: If we add this to `defaultProps` then it should show up as a checkbox and we'll only need one example per layer?

src/index.js
```diff
@@ -45,7 +45,7 @@ export {default as GeoJsonLayer} from './layers/core/geojson-layer/geojson-layer
 export {default as ScatterplotLayer64} from './layers/fp64/scatterplot-layer/scatterplot-layer-64';
 export {default as ArcLayer64} from './layers/fp64/arc-layer/arc-layer-64';
 export {default as LineLayer64} from './layers/fp64/line-layer/line-layer-64';
-export {default as GeoJsonLayer64} from './layers/fp64/geojson-layer/geojson-layer-64';
+// export {default as GeoJsonLayer64} from './layers/fp64/geojson-layer/geojson-layer-64';
```
ibgreen: delete

src/layers/core/geojson-layer/geojson-layer.js
```diff
@@ -124,6 +125,7 @@ export default class GeoJsonLayer extends Layer {
         getColor: getFillColor,
         extruded,
         wireframe: false,
+        fp64: this.props.fp64,
```
ibgreen: This is already being forwarded on line 120 Object.assign({}, this.props, ...

src/layers/core/polygon-layer/polygon-layer.js
```diff
@@ -89,14 +89,39 @@ export default class PolygonLayer extends Layer {
...
+    if (props.fp64 !== oldProps.fp64) {
+      const {gl} = this.context;
+      this.setState({model: this.getModel(gl)});
+    }
+  }
+
+  updateAttribute({props, oldProps, changeFlags}) {
+    if (props.fp64 !== oldProps.fp64) {
+      const {attributeManager} = this.state;
+      if (props.fp64 === true) {
```
ibgreen: Could be nice to have an enable/disable function instead.

src/layers/core/polygon-layer/polygon-layer.js
```diff
@@ -89,14 +89,39 @@ export default class PolygonLayer extends Layer {
...
+        attributeManager.add({
```
ibgreen: Nit: Keep as single line to avoid this simple logic taking a lot of lines?

src/layers/core/polygon-layer/polygon-layer.js
```diff
@@ -89,14 +89,39 @@ export default class PolygonLayer extends Layer {
...
+          positions64xyLow: {size: 2, update: this.calculatePositionsLow}
+        });
+      } else {
+        attributeManager.remove([
```
ibgreen: Keep as single line?

src/layers/core/scatterplot-layer/scatterplot-layer.js
```diff
@@ -35,15 +36,21 @@ const defaultProps = {
...
 };
 
 export default class ScatterplotLayer extends Layer {
   getShaders(id) {
-    return {
-      vs: readFileSync(join(__dirname, './scatterplot-layer-vertex.glsl'), 'utf8'),
-      fs: readFileSync(join(__dirname, './scatterplot-layer-fragment.glsl'), 'utf8')
-    };
+
+    const shaders = this.props.fp64 ? {
```
ibgreen: no need for a temporary, just return the result

src/layers/core/scatterplot-layer/scatterplot-layer.js
```diff
@@ -35,15 +36,21 @@ const defaultProps = {
...
+      vs: readFileSync(join(__dirname, './scatterplot-layer-64-vertex.glsl'), 'utf8'),
+      fs: readFileSync(join(__dirname, './scatterplot-layer-fragment.glsl'), 'utf8'),
+      modules: ['lighting', 'fp64', 'project64']} : {
```
ibgreen: Nit: I'd put a new line before the `} : {` to more clearly show the two objects.

src/layers/core/scatterplot-layer/scatterplot-layer.js
```diff
@@ -65,9 +72,47 @@ export default class ScatterplotLayer extends Layer {
...
+      const {gl} = this.context;
+      this.setState({model: this._getModel(gl)});
+    }
+  }
+
+  updateAttribute({props, oldProps, changeFlags}) {
+    if (props.fp64 !== oldProps.fp64) {
+      const {attributeManager} = this.state;
+
+      // all attributes needs to be set again after the model changes
```
ibgreen: While I assume that you are correct, this does not sound like the way the attribute manager was intended this to work. Let's discuss.

src/layers/core/scatterplot-layer/scatterplot-layer.js
```diff
@@ -109,6 +154,17 @@ export default class ScatterplotLayer extends Layer {
     }
   }
 
+  calculateInstancePositions64xyLow(attribute) {
+    const {data, getPosition} = this.props;
+    const {value} = attribute;
+    let i = 0;
+    for (const point of data) {
+      const position = getPosition(point);
+      value[i++] = fp64ify(position[0])[1];
```
ibgreen: I think I added an `fp64lowPart`. Not sure if I merged yet. Maybe clearer than indexing this result?

src/lib/attribute-manager.js
```diff
@@ -110,6 +110,25 @@ export default class AttributeManager {
...
+   * Removes attributes
+   * Takes an array of attribute names and delete them from
+   * the attribute map if they exists
+   *
+   * @example
+   * attributeManager.remove(['position']);
+   *
+   * @param {Object} attributeNameArray - attribute name array (see above)
+   */
+  remove(attributeNameArray) {
```
ibgreen: +1. Any way we could add a test case for this new method?

src/layers/core/scatterplot-layer/scatterplot-layer.js
shaojingli: @ibgreen This is how attribute manager works now, and I agree that this sounds like a bug to me.
Right now, luma.gl's render function get buffers to bind to attributes from Model.buffers, whose content is populated in model.setAttributes() function. Model.setAttributes() is in turn called by Layer.updateAttributes() with attributes dictionary generated by attributeManager.getChangedAttributes().

So only attributes with changed data will be set to Model. In scatterplot's case, changing between 32-bit and 64-bit model does not change the underlying attribute data, so the attribute manager failed to populate those attributes to new model. We have to do this manually instead.

Be noted that this is not the case for polygon layer as the polygon tessellator regenerate the underlying data when 32-bit and 64-bit change happens so all attributes got automatically populated.

src/layers/core/scatterplot-layer/scatterplot-layer.js
shaojingli: No... you haven't merged it yet. I'll change to that after the method is in.

examples/main/src/layer-examples.js
shaojingli: Yes. That's what we are doing right now.

PR #382, opened 2017-02-28
Reorg

Pessimistress:
- Remove sample-layers app, use sample-layers directory for layers code only
- Add readme docs for TripsLayer and PlotLayer
- Reference sample-layers from demo app
- Add examples for sample-layers in the main example app


examples/main/src/layer-examples.js
```diff
@@ -25,6 +25,12 @@ import {
 
 } from 'deck.gl';
 
+import {
+  EnhancedChoroplethLayer,
+  // PlotLayer,
```
ibgreen: Delete comment?

examples/main/src/layer-examples.js
```diff
@@ -25,6 +25,12 @@ import {
 
 } from 'deck.gl';
 
+import {
```
ibgreen: Could we split this file into three?
* official layer examples
* demo layer examples
* sample layer examples

The reason is that I'd like to keep the distinction between these layer types clear if possible.
I am also not sure we should include the demo layers here.  It kind of gives the impression that these are prepackaged layers ready to consume, rather than just examples provided for hacking. We should look at this live and decide.

examples/sample-layers/index.js
```diff
@@ -0,0 +1,3 @@
+export {default as EnhancedChoroplethLayer} from './enhanced-choropleth-layer';
+export {default as PlotLayer} from './plot-layer';
```
ibgreen: Mainly a feeling but I think this is a strange mix of layers. They are so different in scope and ambition. I would have expected the demo layers to be their own stand alone examples (complete apps) that could be cloned, installed and run.

PR #384, opened 2017-03-01
Only invoke mouse event on affected layers

Pessimistress:
- `onHover` is only called upon enter/leave
- `onClick` is only called on the picked layer

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ Ref: http://keepachangelog.com/en/0.3.0/
 
 #### [v4.0.0-beta.3]
 - Add PointCloudLayer
+- FIX: `onHover` and `onClick` are no longer called on layers that are not affected
 - BREAKING: `layer.pick()` is renamed to `layer.getPickingInfo()`, must return info object
 
 #### [v4.0.0-beta.2]
--- a/src/layers/core/geojson-layer/geojson-layer.js
+++ b/src/layers/core/geojson-layer/geojson-layer.js
@@ -58,9 +58,7 @@ const getCoordinates = f => get(f, 'geometry.coordinates');
 export default class GeoJsonLayer extends CompositeLayer {
   initializeState() {
     this.state = {
-      subLayers: null,
-      subLayerCount: 0,
-      pickInfos: []
+      subLayers: null
     };
   }
 
@@ -72,32 +70,18 @@ export default class GeoJsonLayer extends CompositeLayer {
     }
   }
 
-  _onPickSublayer(mode, info) {
-    const {pickInfos, subLayerCount} = this.state;
-    pickInfos.push(info);
-
-    if (pickInfos.length === subLayerCount) {
-      // all sublayers have been accounted for
-      let pickInfo = pickInfos.find(i => i.index >= 0) || pickInfos[0];
-
-      pickInfo = Object.assign({}, pickInfo, {
-        layer: this,
-        feature: get(pickInfo, 'object.feature') || pickInfo.object
-      });
-
-      switch (mode) {
-      case 'click': this.props.onClick(pickInfo); break;
-      case 'hover': this.props.onHover(pickInfo); break;
-      default: throw new Error('unknown pick type');
-      }
+  _onHoverSubLayer(info) {
+    if (info.object) {
+      info.object = info.object.feature || info.object;
     }
+    this.props.onHover(info);
   }
 
-  getPickingInfo(opts) {
-    // this is called before the onHover/onClick of sublayers
-    // pickInfo is used to collect the pick results of all sublayers
-    this.state.pickInfos.length = 0;
-    return null;
+  _onClickSubLayer(info) {
+    if (info.object) {
+      info.object = info.object.feature || info.object;
+    }
+    this.props.onClick(info);
   }
 
   renderLayers() {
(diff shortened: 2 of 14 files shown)


Pessimistress:
The way I'm storing previously picked info in `layerManager.context` (`lastPickedColor` + `lastPickedLayerId`) accounts for adding/removing layers, but not changes in `layer.props.data`. There's no quick fix for this unless we store a `dataChanged` state in each layer just for hover. On the other hand the chance of data changes and the picking color under the cursor stays the same sounds pretty low...

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ Ref: http://keepachangelog.com/en/0.3.0/
 
 #### [v4.0.0-beta.3]
 - Add PointCloudLayer
+- FIX: `onHover` and `onClick` are no longer called on layers that are not affected
```
ibgreen: Updating the changelog is good but we really need to update the `What's New` and `Upgrade Guide` docs as well. And review the `Layer` and `Layer Lifecycle` docs that mention these props.

src/lib/draw-and-pick.js
```diff
@@ -32,11 +33,10 @@ export function drawLayers({layers, pass}) {
 /* eslint-disable max-depth, max-statements */
 export function pickLayers(gl, {
   layers,
-  pickingFBO,
+  context,
```
ibgreen: The way I see it, `context` is internal to the `LayerManager`/`Layer` classes. Probably cleaner to make this function take explicit args and not know anything about the context object.

src/lib/draw-and-pick.js
```diff
@@ -87,27 +87,53 @@ export function pickLayers(gl, {
...
+    gl.readPixels(deviceX, deviceY, 1, 1, GL.RGBA, GL.UNSIGNED_BYTE, pickedColor);
 
     // restore blend mode
     setBlendMode(gl, oldBlendMode);
     // Picking process end
 
     // Process picked info start
     // Decode alpha to layer index
-    const pickedLayer = layers[color[3] - 1];
+    const pickedLayer = pickedColor[3] ? layers[pickedColor[3] - 1] : null;
```
ibgreen: Maybe split in two lines to make more self documenting:
```
const pickedLayerIndex  = pickedColor[3] - 1;
const pickedLayer = pickedLayerIndex > -1  ? layers[pickedLayerIndex] : null;
```

Side note: I've also considered making the background picking color white (0xFFFFFF) instead of black, so that we don't have to add and subtract one to the picking colors. Would make them a little more intuitive

src/lib/draw-and-pick.js
```diff
@@ -87,27 +87,53 @@ export function pickLayers(gl, {
...
+    const affectedLayers = [pickedLayer];
```
ibgreen: What if pickedLayer is null? I guess we are filtering below. Since we only have one or two layers now, I think it could be easier to read this code if we explicitly only add the layer here if it is non null.

src/lib/draw-and-pick.js
```diff
@@ -87,27 +87,53 @@ export function pickLayers(gl, {
...
+
+    if (mode === 'hover') {
+      // only invoke onHover events if picked object has changed
+      const {lastPickedColor, lastPickedLayerId} = context;
+
+      const pickedLayerChanged = (pickedLayer && pickedLayer.props.id) !== lastPickedLayerId;
```
ibgreen: Again maybe split in two lines to make more self documenting?
```
const pickedLayerId = pickedLayer && pickedLayer.props.id;
...
```

src/lib/draw-and-pick.js
```diff
@@ -87,27 +87,53 @@ export function pickLayers(gl, {
...
+      const pickedColorChanged = (lastPickedColor[0] !== pickedColor[0] ||
+        lastPickedColor[1] !== pickedColor[1] || lastPickedColor[2] !== pickedColor[2]);
+
```
ibgreen: We could store a `pickedLayerIndex` instead of a `pickedLayerColor` so we didn't have to work with color components.

src/lib/draw-and-pick.js
```diff
@@ -87,27 +87,53 @@ export function pickLayers(gl, {
...
+      if (pickedLayerChanged) {
+        // We cannot store a ref to lastPickedLayer in the context because
+        // the state of an outdated layer is no longer valid
+        // and the props may have changed
+        const lastPickedLayer = layers.find(l => l.props.id === lastPickedLayerId);
```
ibgreen: Pretty neat.

src/lib/draw-and-pick.js
```diff
@@ -87,27 +87,53 @@ export function pickLayers(gl, {
...
+        // Let leave event fire before enter event
+        affectedLayers.splice(0, 0, lastPickedLayer);
```
ibgreen: Is that equivalent to 
`affectedLayers.unshift(lastPickedLayer)`? 
`affectedLayers = [lastPickedLayer, ...affectedLayers];`
To my eyes, `splice` always seems a little hard to read, having to keep track of indices and such.

src/lib/draw-and-pick.js
```diff
@@ -87,27 +87,53 @@ export function pickLayers(gl, {
...
+      } else if (!pickedColorChanged) {
+        // picked object did not change, no need to proceed
```
ibgreen: This important `return` statement feels a bit hidden down here. I'd move this check up top even if I had to duplicate an if statement.
```
if (!pickedLayerIdChanged && !pickedColorChanged) {
  return;
}
``

File: src/lib/layer-manager.js [hunk 10]
```diff
@@ -49,7 +49,9 @@ export default class LayerManager {
       uniforms: {},
       viewport: null,
       viewportChanged: true,
-      pickingFBO: null
+      pickingFBO: null,
+      lastPickedColor: new Uint8Array(4),
```
ibgreen: Again the decoded index would look better here.

src/layers/core/geojson-layer/geojson-layer.js
```diff
@@ -72,32 +70,18 @@ export default class GeoJsonLayer extends CompositeLayer {
...
-      });
-
-      switch (mode) {
-      case 'click': this.props.onClick(pickInfo); break;
-      case 'hover': this.props.onHover(pickInfo); break;
-      default: throw new Error('unknown pick type');
-      }
+  _onHoverSubLayer(info) {
+    if (info.object) {
+      info.object = info.object.feature || info.object;
```
ibgreen: Maybe make this into a one liner?
```
info.object = (info.object && info.object.feature) || info.object;
```

PR #385, opened 2017-03-01
Add additional sample layers

ibgreen:
@Pessimistress This is the label layer we talked about, plus a few others. 

Note: this PR is for discussion only, not ready for final review. Also EnhancedChoroplethLayer commented out due to build issues.


ibgreen:
> There is a bug on commit 321f3b0, which is where this PR is based on. Please rebase the PR to latest dev.

@shaojingli Rebased. The label layer works now (but updates too frequently).

ibgreen:
@shaojingli Fixed the excessive update issue (was related to the sublayer not having a proper `id`).

@Pessimistress This PR also contains bitmap and mesh layers, we'll need to pick some proper bitmaps and meshes to use in the example layers.

The labels don't look that great for some reason, but at least this is a start. I suspect selecting a heavier font-weight will help. Should investigate if we can control anti-aliasing using this technique.

screen shot 2017-03-01 at 2 15 27 pm

src/layers/deprecated/choropleth-layer/choropleth-layer.js
```diff
@@ -77,9 +77,15 @@ export default class ChoroplethLayer extends Layer {
       attributeManager.invalidateAll();
     }
 
-    if (oldProps.opacity !== props.opacity) {
+    if (props.opacity !== oldProps.opacity) {
```
Pessimistress: Don't need this any more?

examples/sample-layers/label-layer/label-layer.js
```diff
@@ -0,0 +1,54 @@
...
+
+    const {texture, mapping} = makeTextureAtlasFromLabels(gl, {data, getLabel, fontSize});
+    // if (this.state.texture) {
+    //   this.state.texture.delete();
+    // }
+    this.setState({texture, mapping});
+  }
+
+  renderLayers() {
+    return new IconLayer(Object.assign({}, this.props, {
```
Pessimistress: IconLayer doesn't currently preserve aspect ratio. We should fix it in a separate PR.

src/layers/deprecated/choropleth-layer/choropleth-layer.js
ibgreen: > Don't need this any more?

Good catch. Removed.

shaojingli:
@ibgreen You might want to add a fp64 prop to the label layer later (now we have 64-bit icon layer after #387) as the wobbling effect of texts rendered by 32-bit shaders is really prominent.

ibgreen:
> Assuming that we are going to provide a proper mesh / bitmap to mesh / bitmap layers, we can land this change.

Agreed. I could try to scrounge something up but I have a feeling that it will look much nicer if @Pessimistress does it. If we don't get examples in place we can always remove these sample layers before we ship.

> @ibgreen You might want to add a fp64 prop to the label layer later (now we have 64-bit icon layer after #387) as the wobbling effect of texts rendered by 32-bit shaders is really prominent.

Good idea. I believe that I am passing through all props to the icon layer, so a 64bit prop should work if the IconLayer supports it. I added `fp64` to `defaultProps` to make sure that the layer-browser example can discover the prop. Unfortunately I still see a wobble so something is missing. Maybe you can take a look after I land?

PR #387, opened 2017-03-02
More 64-bit layers

shaojingli:
More 64-bit layers coming!

64-bit IconLayer, HexagonLayer, GridLayer, PointDensityHexagonLayer, PointDensityGridLayer added

### DIFF
--- a/src/layers/core/grid-layer/grid-layer.js
+++ b/src/layers/core/grid-layer/grid-layer.js
@@ -116,9 +116,9 @@ export default class GridLayer extends Layer {
 
   updateState({props, oldProps, changeFlags}) {
     super.updateState({props, oldProps, changeFlags});
-    this.updateUniforms();
     this.updateModel({props, oldProps, changeFlags});
     this.updateAttribute({props, oldProps, changeFlags});
+    this.updateUniforms();
   }
 
   _getModel(gl) {
--- a/src/layers/core/grid-layer/grid-layer.js
+++ b/src/layers/core/grid-layer/grid-layer.js
@@ -83,12 +83,6 @@ export default class GridLayer extends Layer {
       instancePositions: {size: 4, accessor: ['getPosition', 'getElevation'], update: this.calculateInstancePositions},
       instanceColors: {size: 4, type: GL.UNSIGNED_BYTE, accessor: 'getColor', update: this.calculateInstanceColors}
     });
-
-    if (this.props.fp64) {
-      this.state.attributeManager.addInstanced({
-        instancePositions64xyLow: {size: 2, accessor: 'getPosition', update: this.calculateInstancePositions64xyLow}
-      });
-    }
     /* eslint-enable max-len */
   }
 
--- a/src/layers/core/grid-layer/grid-layer.js
+++ b/src/layers/core/grid-layer/grid-layer.js
@@ -23,6 +23,7 @@ import {assembleShaders} from '../../../shader-utils';
 import {GL, Model, CubeGeometry} from 'luma.gl';
 import {readFileSync} from 'fs';
 import {join} from 'path';
+import {fp64ify} from '../../../lib/utils/fp64';
 
 const DEFAULT_COLOR = [255, 0, 255, 255];
 
@@ -41,7 +42,8 @@ const defaultProps = {
     specularRatio: 0.8,
     lightsStrength: [1.0, 0.0, 0.8, 0.0],
     numberOfLights: 2
-  }
+  },
+  fp64: false
 };
 
 export default class GridLayer extends Layer {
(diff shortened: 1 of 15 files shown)


src/layers/core/grid-layer/grid-layer-vertex.glsl
```diff
@@ -28,6 +28,13 @@ uniform float elevationScale;
 // Result
 varying vec4 vColor;
 
+// whether is point picked
+float isPicked(vec3 pickingColors, vec3 selectedColor) {
```
ibgreen: Nit: I might be wrong but think it is possible to compare vectors directly without explicitly doing element wise comparison.

```
return pickingColors == selectedColor;
```

src/layers/core/grid-layer/grid-layer.js
```diff
@@ -60,36 +62,66 @@ export default class GridLayer extends Layer {
...
+      if (props.fp64 === true) {
+        attributeManager.addInstanced({
+          instancePositions64xyLow: {
+            size: 2,
+            accessor: 'getPosition',
+            update: this.calculateInstancePositions64xyLow
+          }
+        });
+      } else {
+        attributeManager.remove([
```
ibgreen: Keep the remove on single line, to reduce the number of lines of boilerplate?

src/layers/core/hexagon-layer/hexagon-layer-64-vertex.glsl
```diff
@@ -0,0 +1,113 @@
...
+  vertex_pos_modelspace[1] = sum_fp64(vertex_pos_localspace[1], projected_coord_xy[1]);
+  vertex_pos_modelspace[2] = sum_fp64(vertex_pos_localspace[2], vec2(elevation, 0.0));
+  vertex_pos_modelspace[3] = vec2(1.0, 0.0);
+
+  vec4 position_worldspace = vec4(vertex_pos_modelspace[0].x, vertex_pos_modelspace[1].x, vertex_pos_modelspace[2].x, vertex_pos_modelspace[3].x);
+
+  gl_Position = project_to_clipspace_fp64(vertex_pos_modelspace);
+
+  // render display
+  if (renderPickingBuffer < 0.5) {
```
ibgreen: Just my 2 cents: I personally think we should just set a pickingColor varying here like most layers used to do, and let the fragment shader decide which color to return based on the uniform. I realize that checking the uniforms here makes picking a little faster, so it is a tradeoff of picking speed vs a slightly simpler vertex shader. 

No need to change now, we can revisit this discussion when we design the picking shader module.

src/layers/core/hexagon-layer/hexagon-layer.js
```diff
@@ -82,39 +84,80 @@ export default class HexagonLayer extends Layer {
     super(props);
   }
 
+  getShaders() {
+    return this.props.fp64 ? {
+      vs: readFileSync(join(__dirname, './hexagon-layer-64-vertex.glsl'), 'utf8'),
+      fs: readFileSync(join(__dirname, './hexagon-layer-fragment.glsl'), 'utf8'),
```
ibgreen: This construction will inline the fragment shader twice, adding a long string twice to the code. It could be better to do:
```
const vs = readFileSync(...
const vs64 = readFileSync(
const fs = readFileSync(
```
And then just return compact one liner objects built from those variables. It might be a little easier to read too?

src/layers/core/hexagon-layer/hexagon-layer.js
```diff
@@ -82,39 +84,80 @@ export default class HexagonLayer extends Layer {
...
+    this.setState({model: this._getModel(gl)});
 
     const {attributeManager} = this.state;
     /* eslint-disable max-len */
     attributeManager.addInstanced({
       instancePositions: {size: 3, accessor: ['getCentroid', 'getElevation'], update: this.calculateInstancePositions},
       instanceColors: {size: 4, type: gl.UNSIGNED_BYTE, accessor: 'getColor', update: this.calculateInstanceColors}
     });
+
+    if (this.props.fp64) {
```
ibgreen: We shouldn't need to do this check twice. It should be enough to do it in `updateState`, since updateState is called on first initialization as well (after `initializeState`). If that doesn't work, we might have a bug on our hands.

src/layers/core/grid-layer/grid-layer-vertex.glsl
shaojingli: You are actually right. I'm just blindly copying previous shader code ....

src/layers/core/hexagon-layer/hexagon-layer.js
shaojingli: I agree with you. I'll do a sweep across all fp64 layers of this pattern later.

PR #388, opened 2017-03-02
Preserve aspect ratio in IconLayer

Pessimistress:


src/layers/core/icon-layer/icon-layer-vertex.glsl
```diff
@@ -39,8 +39,10 @@ varying vec4 vColor;
 varying vec2 vTextureCoords;
 
 void main(void) {
+  // scale icon height to match instanceSize
+  float instanceScale = instanceSizes / max(instanceIconFrames.w, 1.0);
   vec3 center = project_position(instancePositions);
-  vec2 vertex = (positions + instanceOffsets * 2.0) * sizeScale * instanceSizes;
+  vec2 vertex = (positions + instanceOffsets * 2.0) * instanceIconFrames.zw * sizeScale * instanceScale;
```
ibgreen: For readability it might be nice to have an intermediary variable?
```
vec2 iconSize = instanceIconFrames.zw;
```
This could be at the top of the function next to instance scale (a section "unpacking" the packed attributes so to speak).

PR #390, opened 2017-03-02
Add viewportSize, devicePixelRatio and modelViewMatrix to default uniforms

Pessimistress:
- add `viewportSize`, `viewMatrix` and `modelMatrix` to default uniforms
- update layers
- update shaders doc

### DIFF
--- a/docs/writing-shaders.md
+++ b/docs/writing-shaders.md
@@ -51,7 +51,7 @@ void main(void) {
 
 #### `mat4 modelViewMatrix`
 
-The model view matrix of the current viewport.
+The model view matrix based on the current viewport and layer.
 
 #### `vec2 viewportSize`
 
@@ -108,7 +108,7 @@ vs (good)
 #### `vec3 selectedPickingColor`
 
 This uniform is set if `props.pickable` is enabled on the layer and reflects the color
-of the last picked pixel. 
+of the last picked pixel. If no pixel is selected, the value will be `[0, 0, 0]`.
 
 ## Build Concerns
 
--- a/docs/writing-shaders.md
+++ b/docs/writing-shaders.md
@@ -55,7 +55,11 @@ The model view matrix based on the current viewport and layer.
 
 #### `vec2 viewportSize`
 
-Viewport width and height in device pixels. Useful when rendering pixel sizes.
+Viewport width and height in physical pixels. Useful when rendering pixel sizes.
+
+#### `float devicePixelRatio`
+
+Device pixel ratio of the browser window.
 
 ### Layer prop uniforms
 
--- a/docs/writing-shaders.md
+++ b/docs/writing-shaders.md
@@ -39,23 +39,42 @@ void main(void) {
 
 ## Uniforms
 
-### `layerIndex` uniform
+### Viewport uniforms
+
+#### `mat4 modelMatrix`
+
+The user supplied model matrix if `props.modelMatrix` is set on the layer,
+otherwise an identity matrix.
+
+#### `mat4 viewMatrix`
+
+The view matrix of the current viewport.
+
+#### `vec2 viewportSize`
+
+Viewport width and height. Useful when rendering pixel sizes.
+
+### Layer prop uniforms
+
+#### `float layerIndex`
 
 The layerIndex is a small integer that starts at zero and is incremented
 for each layer that is rendered. It can be used to add small offsets to
 the z coordinate of layers to resolve z-fighting between overlapping
 layers.
 
-### `opacity` uniform
+#### `float opacity`
 
 In the fragment shader, multiply the fragment color with the opacity
 uniform.
 
-### `picking` uniforms
+### Picking uniforms
+
+#### `float renderPickingBuffer`
 
 If you choose to implement picking through picking colors, make sure
 the `pickingColors` or `instancePickingColors` attribute is correctly set up,
-and ensure that you return the picking color when `renderPickingColors`
+and ensure that you return the picking color when `renderPickingBuffer`
 uniform is set. Alternatively call the `layerColor` method on your
 fragment color before assigning to `gl_FragColor`.
 
@@ -82,6 +101,11 @@ vs (good)
    	renderPickingBuffer
    );
 ```
+
+#### `vec3 selectedPickingColor`
+
+This uniform is set if `props.pickable` is enabled on the layer and reflects the color
+of the last picked pixel. 
 
 ## Build Concerns
 
--- a/docs/writing-shaders.md
+++ b/docs/writing-shaders.md
@@ -22,7 +22,15 @@ follow deck.gl's projection methods, enabling your layer to accept coordinates
 in both [longitude,latitude,altitude] or [metersX,metersY,metersZ] format.
 
 The projection package is included by default by the `assembleShaders` function,
-and offers three functions `project_position`, `project_scale` and `project_to_clipspace`.
+and offers the following functions:
+
+#### `project_position`
+
+#### `project_scale`
+
+#### `project_to_viewspace`
+
+#### `project_to_clipspace`
 
 ```glsl
 attribute vec3 positions;
@@ -41,18 +49,13 @@ void main(void) {
 
 ### Viewport uniforms
 
-#### `mat4 modelMatrix`
+#### `mat4 modelViewMatrix`
 
-The user supplied model matrix if `props.modelMatrix` is set on the layer,
-otherwise an identity matrix.
-
-#### `mat4 viewMatrix`
-
-The view matrix of the current viewport.
+The model view matrix of the current viewport.
 
 #### `vec2 viewportSize`
 
-Viewport width and height. Useful when rendering pixel sizes.
+Viewport width and height in device pixels. Useful when rendering pixel sizes.
 
 ### Layer prop uniforms
 
--- a/examples/sample-layers/plot-layer/grid-vertex.glsl
+++ b/examples/sample-layers/plot-layer/grid-vertex.glsl
@@ -25,6 +25,8 @@ attribute vec3 normals;
 attribute vec2 instancePositions;
 attribute vec3 instanceNormals;
 
+uniform mat4 modelMatrix;
+uniform mat4 viewMatrix;
 uniform vec3 modelCenter;
 uniform vec3 modelDim;
 uniform float gridOffset;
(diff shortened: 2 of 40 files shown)


docs/writing-shaders.md
```diff
@@ -39,23 +39,42 @@ void main(void) {
 
 ## Uniforms
 
-### `layerIndex` uniform
+### Viewport uniforms
+
+#### `mat4 modelMatrix`
```
ibgreen: +1 for updating docs!

examples/sample-layers/plot-layer/grid-vertex.glsl
```diff
@@ -25,6 +25,8 @@ attribute vec3 normals;
 attribute vec2 instancePositions;
 attribute vec3 instanceNormals;
 
+uniform mat4 modelMatrix;
```
ibgreen: I wonder if we need to expose `modelMatrix` separately. See comment below on perf impact.

examples/sample-layers/plot-layer/grid-vertex.glsl
```diff
@@ -33,13 +35,10 @@ uniform vec4 strokeColor;
...
 
-// determines if the grid line is behind or in front of the center.
-// normally we get this by applying viewMatrix to the vector
-// TODO: send viewMatrix via uniform
+// determines if the grid line is behind or in front of the center
 float frontFacing(vec3 v) {
-  vec4 p0_viewspace = project_to_clipspace(vec4(0.0, 0.0, 0.0, 1.0));
-  vec4 p1_viewspace = project_to_clipspace(vec4(v, 1.0));
-  return step(p1_viewspace.z, p0_viewspace.z);
+  vec4 v_viewspace = viewMatrix * modelMatrix * vec4(v, 0.0);
```
ibgreen: I think a matrix multiplication is about 64 GPU clock cycles, so having these pre multiplied into a `modelViewMatrix` could make sense.

examples/sample-layers/plot-layer/grid-vertex.glsl
```diff
@@ -25,6 +25,8 @@ attribute vec3 normals;
 attribute vec2 instancePositions;
 attribute vec3 instanceNormals;
 
+uniform mat4 modelMatrix;
+uniform mat4 viewMatrix;
 uniform vec3 modelCenter;
```
ibgreen: @shaojingli Possible decision point:

* If these uniforms are generated by the `project` package (which is strongly linked to `Viewport.getUniforms`), they could be auto injected by the `project` shader stub. 

* But since they are not actually used by the functions in that package we could just document them as available and require the shader to declare them.

* The `project` package could also just expose a `project_to_viewspace` next to `project_to_clipspace` which would probably fit better with 64 bit support

examples/sample-layers/plot-layer/label-vertex.glsl
```diff
@@ -48,13 +50,10 @@ float sum3(vec3 v) {
...
 
-// determines if the grid line is behind or in front of the center.
-// normally we get this by applying viewMatrix to the vector
-// TODO: send viewMatrix via uniform
+// determines if the grid line is behind or in front of the center
 float frontFacing(vec3 v) {
-  vec4 p0_viewspace = project_to_clipspace(vec4(0.0, 0.0, 0.0, 1.0));
-  vec4 p1_viewspace = project_to_clipspace(vec4(v, 1.0));
-  return step(p1_viewspace.z, p0_viewspace.z);
+  vec4 v_viewspace = viewMatrix * modelMatrix * vec4(v, 0.0);
```
ibgreen: Here we could just write `project_to_viewspace` if we added that function.

examples/sample-layers/plot-layer/label-vertex.glsl
```diff
@@ -115,6 +114,6 @@ void main(void) {
   label_vertex *= fontSize / labelHeight * position_clipspace.w;
 
   gl_Position = position_clipspace + 
-    vec4(label_vertex / screenSize, 0.0, 0.0);
+    vec4(label_vertex / viewportSize, 0.0, 0.0);
```
ibgreen: Not sure about this one, but maybe there is a better naming here. 

`label_point`?  Seemed strange to divide a vertex by a `vec2` until I realized `label_vertex` is a point.

src/lib/viewport-uniforms.js
```diff
@@ -123,7 +123,11 @@ export function getUniformsFromViewport(viewport, {
     projectionMode,
     projectionCenter,
 
-    // modelMatrix: modelMatrix || new Matrix4().identity(),
+    modelMatrix: modelMatrix || new Matrix4().identity(),
```
ibgreen: modelViewMatrix?

docs/writing-shaders.md
```diff
@@ -39,23 +39,42 @@ void main(void) {
...
+The user supplied model matrix if `props.modelMatrix` is set on the layer,
+otherwise an identity matrix.
+
+#### `mat4 viewMatrix`
+
+The view matrix of the current viewport.
+
+#### `vec2 viewportSize`
+
+Viewport width and height. Useful when rendering pixel sizes.
```
ibgreen: We should be really crisp about what coordinates this is specified in. Device coordinates or Window coordinates. I assume the former.

docs/writing-shaders.md
```diff
@@ -39,23 +47,37 @@ void main(void) {
 
 ## Uniforms
 
-### `layerIndex` uniform
+### Viewport uniforms
+
+#### `mat4 modelViewMatrix`
+
+The model view matrix of the current viewport.
```
ibgreen: based on the current viewport and layer

docs/writing-shaders.md
```diff
@@ -22,7 +22,15 @@ follow deck.gl's projection methods, enabling your layer to accept coordinates
...
 The projection package is included by default by the `assembleShaders` function,
-and offers three functions `project_position`, `project_scale` and `project_to_clipspace`.
+and offers the following functions:
+
+#### `project_position`
+
+#### `project_scale`
+
+#### `project_to_viewspace`
+
```
ibgreen: Feels a bit strange not to be able to go to clipspace from viewspace but we can build this out later by adding the raw projectionMatrix.

docs/writing-shaders.md
```diff
@@ -83,6 +105,11 @@ vs (good)
    );
 ```
 
+#### `vec3 selectedPickingColor`
+
+This uniform is set if `props.pickable` is enabled on the layer and reflects the color
+of the last picked pixel.
```
ibgreen: Describe what happens when no pixel is selected?

src/lib/viewport-uniforms.js
```diff
@@ -24,10 +24,15 @@ function calculateMatrixAndOffset({
 
   let projectionCenter;
   let modelViewProjectionMatrix;
+  let modelViewMatrixInv;
 
-  const modelViewMatrixInv = new Matrix4(viewMatrix);
+  const modelViewMatrix = new Matrix4(viewMatrix);
   const viewProjectionMatrix = new Matrix4(projectionMatrix).multiplyRight(viewMatrix);
 
+  if (modelMatrix) {
```
ibgreen: I think this needs to be handled in the cases below. We really need a meter offsets example for one of the layers that use the new uniform.

src/lib/viewport-uniforms.js
```diff
@@ -23,28 +23,13 @@ function calculateMatrixAndOffset({
...
     projectionCenter = ZERO_VECTOR;
-    modelViewProjectionMatrix = viewProjectionMatrix;
-    if (modelMatrix) {
-      // Apply model matrix if supplied
-      // modelViewProjectionMatrix = modelViewProjectionMatrix.clone();
-      modelViewProjectionMatrix.multiplyRight(modelMatrix);
-    }
-    modelViewMatrixInv = new Matrix4(modelViewMatrix);
-    modelViewMatrixInv.invert();
+    modelViewMatrix = new Matrix4(viewMatrix);
```
ibgreen: This now looks correct to me, although I think this code would be clearer if we didn't add the prefix `model` to the variableName until we actually multiply it in (which is now happening half a page down in the code). 

At this stage we should just call the variable `viewMatrix`.

src/lib/viewport-uniforms.js
```diff
@@ -53,26 +38,27 @@ function calculateMatrixAndOffset({
...
   default:
     throw new Error('Unknown projection mode');
   }
 
+  if (modelMatrix) {
+    // Apply model matrix if supplied
+    modelViewMatrix.multiplyRight(modelMatrix);
+  }
+
+  const modelViewMatrixInv = modelViewMatrix.clone().invert();
```
ibgreen: Do we actually want a `modelViewMatrixInv` or just `viewMatrixInv`? I am assuming the latter.

src/lib/viewport-uniforms.js
```diff
@@ -53,26 +38,27 @@ function calculateMatrixAndOffset({
...
+  const modelViewProjectionMatrix = new Matrix4(projectionMatrix).multiplyRight(modelViewMatrix);
   const cameraPos = [modelViewMatrixInv[12], modelViewMatrixInv[13], modelViewMatrixInv[14]];
```
ibgreen: Why do we need the camera position in the model space? We should add a comment. And we almost need to call it `modelCameraPos` in that case.

shaojingli:
I plan to review this change a bit later today.

src/lib/viewport-uniforms.js
shaojingli: Very good catch. This is a bug. cameraPos should be from viewMatrixInv not modelViewMatrixInv.

src/shaderlib/project/project.glsl
```diff
@@ -247,6 +248,10 @@ vec4 project_position(vec4 position) {
 
 //
 
+vec4 project_to_viewspace(vec4 position) {
+  return modelViewMatrix * position;
```
shaojingli: Does project_to_viewspace support meter offset mode?

src/lib/viewport-uniforms.js
shaojingli: You are correct

shaojingli:
@ibgreen thanks for the catch. The cameraPos should be in world space, not model space.

src/shaderlib/project/project.glsl
Pessimistress: Yes.

PR #391, opened 2017-03-02
Add 64-bit path layer

shaojingli:


src/shaderlib/fp64/math-fp64.glsl
```diff
@@ -710,6 +710,10 @@ vec2 radians_fp64(vec2 degree) {
   return mul_fp64(degree, PI_180_FP64);
 }
 
+vec2 mix_fp64(vec2 a, vec2 b, float x) {
+  vec2 range = sub_fp64(b, a);
+  return sum_fp64(a, range * x);
+}
```
ibgreen: Could range * x overflow? Does it need a `mul_scalar`?

shaojingli: You are correct.This function is there by mistake... no one is using it. But since it's there, I'll just fix it.

PR #392, opened 2017-03-03
64bit point cloud layer

shaojingli:
Add 64-bit point cloud layer.
Make the fp64 a prop available for all layers
Add a check so that the fp64 prop is only effective when projectionMode is set to 1.

### DIFF
--- a/src/layers/core/arc-layer/arc-layer.js
+++ b/src/layers/core/arc-layer/arc-layer.js
@@ -24,6 +24,7 @@ import {GL, Model, Geometry} from 'luma.gl';
 import {readFileSync} from 'fs';
 import {join} from 'path';
 import {fp64ify} from '../../../lib/utils/fp64';
+import {COORDINATE_SYSTEM} from '../../../lib';
 
 const DEFAULT_COLOR = [0, 0, 0, 255];
 
@@ -32,12 +33,13 @@ const defaultProps = {
   getTargetPosition: x => x.targetPosition,
   getSourceColor: x => x.color || DEFAULT_COLOR,
   getTargetColor: x => x.color || DEFAULT_COLOR,
-  strokeWidth: 1
+  strokeWidth: 1,
+  fp64: false
 };
 
 export default class ArcLayer extends Layer {
   getShaders() {
-    return this.props.fp64 && this.props.projectionMode === 1 ? {
+    return this.props.fp64 && this.props.projectionMode === COORDINATE_SYSTEM.LNG_LAT ? {
       vs: readFileSync(join(__dirname, './arc-layer-64-vertex.glsl'), 'utf8'),
       fs: readFileSync(join(__dirname, './arc-layer-fragment.glsl'), 'utf8'),
       modules: ['fp64', 'project64']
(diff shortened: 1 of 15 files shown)


src/layers/core/arc-layer/arc-layer.js
```diff
@@ -32,13 +32,12 @@ const defaultProps = {
   getTargetPosition: x => x.targetPosition,
   getSourceColor: x => x.color || DEFAULT_COLOR,
   getTargetColor: x => x.color || DEFAULT_COLOR,
-  strokeWidth: 1,
-  fp64: false
```
ibgreen: Why remove `fp64`? It won't show up in the layer browser if it is not here?

src/layers/core/arc-layer/arc-layer.js
```diff
@@ -32,13 +32,12 @@ const defaultProps = {
...
+  strokeWidth: 1
 };
 
 export default class ArcLayer extends Layer {
   getShaders() {
-    return this.props.fp64 ? {
+    return this.props.fp64 && this.props.projectionMode === 1 ? {
```
ibgreen: 64 only supports one projection mode?If so should we throw an error or a warning if other combinations are applied? This would just silently not render, right?

ibgreen:
As an alternative, we could define a `CoreLayer` or `Layer64` intermediary base class for our core layers. It could contain both the `fp64` prop and a function to add/remove `positions64`.

src/layers/core/arc-layer/arc-layer.js
shaojingli: Currently when using 64-bit under meter offset mode, the 64-bit prop will be ignored and the layer will be rendered in 32-bit. I agree that we should output an warning message in the console, but throwing an exception might not be what users want.

shaojingli:
> I see, you are moving fp64 prop to the base Layer.

I can see why, although I am not 100% sure about this. This adds this prop to all layers, including new user layers that might not be 64 bit enabled (all the layers in our samples folder) and also to screen space layers like the ScreenGridLayer. One could make a case for keeping this in the individual layers I think.

@ibgreen ummm.... okay... i agree. let's revert this back.

src/lib/layer.js
```diff
@@ -39,7 +39,9 @@ const defaultProps = {
   onClick: () => {},
   // Update triggers: a key change detection mechanism in deck.gl
   // See layer documentation
-  updateTriggers: {}
+  updateTriggers: {},
+  projectionMode: 1,
```
Pessimistress: For readability, let's use `COORDINATE_SYSTEM` enum for `projectionMode` whenever possible.

src/layers/core/arc-layer/arc-layer.js
shaojingli: Also, I plan to create another PR for this for better tracking.

src/lib/layer.js
shaojingli: Agree

shaojingli:
@ibgreen For some reason... I mistakenly edited your comments instead of wrote my own... so I restored you comment and I'm putting my reply here

ummm.... okay... i agree. I feel that creating another layer of inheritance just to represent layer with 64-bit support might be too much for now. Let's revert this change back.

PR #394, opened 2017-03-03
Warning message for using 64-bit mode with projection mode other than LON_LAT

shaojingli:
Add a warning message when user tries to draw 64-bit layers in projection mode other than LON_LAT;

Reorganized the shader loading code to save some redundant load;

Fix the 64-bit IconLayer issue brought by #388


shaojingli:
@ibgreen enable64bitsupport() are called twice. Once in the getShaders() function so that correct version of shaders are fetched. Once in the updateAttribute() to attach correct attribute to shaders.

PR #395, opened 2017-03-04
Doc and API audit updates. Make polygon layer composite.

ibgreen:
* Updates `GeoJsonLayer` and `PolygonLayer` props
* Move API reference documentation from articles into separate folder (to keep articles short)
* Separate `Composite Layer` and `Subclassing` Articles
* Update `Performance` section


src/layers/core/polygon-layer/polygon-layer.js
```diff
@@ -18,174 +18,82 @@
...
+        getElevation,
+        getColor,
+        extruded: true,
+        wireframe: true,
+        updateTriggers: Object.assign({}, this.props.updateTriggers, {
+          getColor: this.props.updateTriggers.getFillColor
+        })
+      }));
+    } else if (drawPolygons) {
+      polygonOutlineLayer = new PathLayer(Object.assign({}, this.props, {
```
Pessimistress: This won't work with polygon with holes?

src/layers/core/geojson-layer/geojson-layer.js
```diff
@@ -81,23 +78,22 @@ export default class GeoJsonLayer extends CompositeLayer {
...
-    drawPoints = drawPoints && pointFeatures && pointFeatures.length > 0;
-    drawLines = drawLines && lineFeatures && lineFeatures.length > 0;
-    drawPolygons = drawPolygons && polygonOutlineFeatures && polygonOutlineFeatures.length > 0;
-    fillPolygons = fillPolygons && polygonFeatures && polygonFeatures.length > 0;
+    const {features} = this.state;
+    const {pointFeatures, lineFeatures, polygonFeatures, polygonOutlineFeatures} = features;
+    const {getColor, getFillColor, getRadius, getStrokeWidth, getElevation} = this.props;
+    const {id, stroked, filled, extruded, wireframe} = this.props;
+
+    let {} = this.props;
```
shaojingli: I don't particularly like this syntax in javascript... readers have no idea what variables are assigned here.

src/layers/core/geojson-layer/geojson-layer.js
```diff
@@ -117,33 +113,29 @@ export default class GeoJsonLayer extends CompositeLayer {
...
-        getColor: getStrokeColor,
+        getColor,
         extruded: true,
         wireframe: true,
         onHover,
         onClick,
-        updateTriggers: {
-          getColor: this.props.updateTriggers.getStrokeColor
-        }
+        updateTriggers: Object.assign({}, this.props.updateTriggers, {
```
Pessimistress: This will throw an error when an updateTrigger cannot be matched with an accessor.

src/layers/core/geojson-layer/geojson-layer.js
ibgreen: Good catch - that line didn't do anything.

src/layers/core/geojson-layer/geojson-layer.js
```diff
@@ -21,52 +21,49 @@
 import {CompositeLayer, get} from '../../../lib';
 import ScatterplotLayer from '../scatterplot-layer/scatterplot-layer';
 import PathLayer from '../path-layer/path-layer';
-import PolygonLayer from '../polygon-layer/polygon-layer';
+// Use primitive layer to avoid "Composite Composite" layers for now
+import PrimitivePolygonLayer from '../primitive-polygon-layer/polygon-layer';
```
ibgreen: > SimplePolygonLayer sounds better to me... no matter what name we choose, I think we should update the file names with the folder to be consistent with the new name.

I agree. Since, we already talk about simple and complex polygons (without/with holes), I called this `SolidPolygonLayer` for now and updated all names / file names accordingly.

ibgreen:
Updated per review, but still need to deal with issues with `strokePolygons` and `updateTriggers` forwarding.

PR #399, opened 2017-03-07
ES6 Container/Immutable support cleanup

ibgreen:
This is an "almost ready" PR. Putting it up for discussion purposes, to see if we want this in before we v4 goes code complete.


ibgreen:
Merging this PR as it cleans up code related to iteration and access, and provides simple guidelines that should make it easy in the future to make layers support ES6 containers.

Note that this PR does not contain the required modifications to the individual layers (except `ScatterplotLayer`). I believe that should to happen after the v4 release. It feels a bit late to make such a widespread change now.

PR #409, opened 2017-03-09
replace Object.values conform with ES6 (ES2015)

heshan0131:


examples/layer-browser/src/utils/grid-aggregator.js
```diff
@@ -27,7 +27,7 @@ export function pointsToWorldGrid(points, worldUnitSize) {
     return accu;
   }, {});
 
-  const maxHeight = Math.max.apply(null, Object.values(gridHash));
+  const maxHeight = Math.max.apply(null, Object.keys(gridHash).map(k => gridHash[k]));
```
ibgreen: Logic looks fine.

A concern could be if gridHash could get big? We are generating two unnecessary temporary arrays here to be able to apply `Math.max`. If those arrays could be very large we may want to avoid this.

if instead iterated over the keys using `for in` we could do this without temporaries.

heshan0131: Yea, I am using iterate in the src code. This is for examples. there are only around 500 cells.

shaojingli: Since this is just for the layer-browser example, we can land it.

PR #416, opened 2017-03-12
Wind demo refactor/cleanup

ibgreen:
This "light" refactoring pass focuses on
- passing eslint,
- align code and file structure with rest of repo
- more idiomatic usage of deck.gl

I envision two additional PRs once this lands:
* PR 2) refactor the rather verbose raw WebGL in the layers using luma.gl classes.
* PR 3) refactor the layer properties to make the component layers more generally usable (e.g. `WindLayer` => `VectorFieldLayer`? etc)

@philogb 
* It might be worthwhile checking that things still work as before
* would also recommend deleting the `wind` branch at this point (if no further work has been done on it). This PR is based off the `wind-demo` branch which contains the `wind` branch cherry-picked onto `master`.


gnavvy:
> PR 2) refactor the rather verbose raw WebGL in the layers using luma.gl classes.
There're several functions we need to tweak in luma.gl before we can fully switch over.
e.g. use gl.transformFeedbackVaryings to set the varyings before gl.linkProgram.
Both functions are provided in luma.gl, but set varyings is not in the current pipeline in _compileAndLink in program.js.

I do have some more basic transform feedback examples and will try to list all the places we need to change in luma.gl for fully support of transform feedback cases first.

philogb:
Thanks for taking the time to look at the code. I didn't expect for this to be checked in honestly :) The code running [here](http://philogb.github.io/page/wind/) is the one in _broadcast2_ branch (which is a fast-forward from the wind branch). There aren't that many changes but many variable tweaks. If you want to block some time w/ me to go through it please lmk.

PR #419, opened 2017-03-13
fix orthographic viewport and exports

gnavvy:
only PerspectiveViewport was exported previous, plus a minor bug fix in OrthographicViewport.


src/lib/viewports/orthographic-viewport.js
```diff
@@ -13,15 +13,14 @@ export default class OrthographicViewport extends Viewport {
...
-    fovy = 75, // Field of view covered by camera
     left, // Left bound of the frustum
     top, // Top bound of the frustum
     // automatically calculated
     right = null, // Right bound of the frustum
     bottom = null // Bottom bound of the frustum
   }) {
     right = Number.isFinite(right) ? right : left + width;
-    bottom = Number.isFinite(bottom) ? right : top + height;
+    bottom = Number.isFinite(bottom) ? bottom : top + height;
```
shaojingli: ... how did this work before ~~ -_-##

ibgreen: Ortho viewports were never tested. As I said, the viewport code is still experimental...

src/index.js
```diff
@@ -25,7 +25,13 @@ export {assembleShaders} from './shader-utils';
 // Lib
 export {Layer, CompositeLayer, LayerManager, AttributeManager} from './lib';
 export {COORDINATE_SYSTEM} from './lib';
+// Viewports
+export {default as Viewport} from './lib/viewports/viewport';
+export {default as ViewportMercatorProject} from './lib/viewports/flat-viewport';
+export {default as FlatViewport} from './lib/viewports/flat-viewport';
```
ibgreen: We should delete the FlatViewport completely - it is the old 2D viewport from mercator-project. It will only be confusing to have it here. 

Both of the exports and the file itself should be deleted.

src/lib/viewports/orthographic-viewport.js
shaojingli: Yeah... just realized this is from the orthographic viewport...

PR #421, opened 2017-03-14
Fix the bug that Layer's invalidateAttribute get called for CompositeLayer

shaojingli:
, which has no AttributeManager so no attribute to invalidate;
Clean up the accessor and prop names for Polygon layer and GeoJson layer

### DIFF
--- a/src/lib/composite-layer.js
+++ b/src/lib/composite-layer.js
@@ -7,7 +7,15 @@ export default class CompositeLayer extends Layer {
 
   // initializeState is usually not needed for composite layers
   // Provide empty definition to disable check for missing definition
-  initializeState() {}
+  initializeState() {
+
+  }
+
+  // No-op for the invalidateAttribute function as the composite
+  // layer has no AttributeManager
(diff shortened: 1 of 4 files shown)


src/lib/composite-layer.js
```diff
@@ -7,7 +7,15 @@ export default class CompositeLayer extends Layer {
 
   // initializeState is usually not needed for composite layers
   // Provide empty definition to disable check for missing definition
-  initializeState() {}
+  initializeState() {
+
```
ibgreen: Nit: I could go either way on the braces but I don't think we need an empty line

src/lib/composite-layer.js
```diff
@@ -7,7 +7,15 @@ export default class CompositeLayer extends Layer {
...
+  }
+
+  // No-op for the invalidateAttribute function as the composite
+  // layer has no AttributeManager
+  invalidateAttribute() {
+
```
ibgreen: ditto

PR #422, opened 2017-03-14
Add Scatterplot Layer API demo

Pessimistress:
`createLayerDemoClass` is a generic factory for layer API demos.
The idea is to add demos for all core layers into `components/demos/layers.js`. In this PR there is only one for `ScatterplotLayer`.


ibgreen:
@Pessimistress Is the intention to do this instead of adding the `layer-browser` demo to the website, or is it a complement?

Pessimistress:
@ibgreen yes, we will have one demo on each layer's doc page, instead of a master demo that includes everything.

shaojingli:
Let's discuss it offline

Pessimistress:
This PR has grown quite large. TL;DR it contains the following changes:
- directly loading data from json file without passing through a worker (so that the user can see the data format)
- utility that infers parameter types from layer props
- a LayerDemo factory (`layer-base.js`)
- sample layers (`layers.js`)
- ability to embed a demo in the same tab as a markdown file in the `Page` component
- CSS styling

demo/src/components/demos/layer-base.js
```diff
@@ -0,0 +1,101 @@
...
+  latitude: 37.7,
+  zoom: 9,
+  maxZoom: 20,
+  pitch: 0,
+  bearing: 0
+};
+
+export default function createLayerDemoClass(settings) {
+
+  const renderLayer = (data, params, extraProps = {}) => {
```
ibgreen: I'd avoid local classes and functions unless they are very small as I feel they reduce readability.

The absence of the `export` keyword already makes it very clear what is local to a module.

PR #424, opened 2017-03-14
Polygon layer fix

shaojingli:
Fix the GeoJSON layer demo after the change of default value of the scatterplot layer
Update the GeoJSON layer accessor names according to the API audit
Fix the layer props of the GeoJSON layer and the Polygon layer
Also, removed the updateModel in the base Layer and put the code to each subclassed layer because the criteria for updating models are different for each layer.

### DIFF
--- a/src/layers/core/geojson-layer/geojson-layer.js
+++ b/src/layers/core/geojson-layer/geojson-layer.js
@@ -86,7 +86,7 @@ export default class GeoJsonLayer extends CompositeLayer {
     let {} = this.props;
     const drawPoints = pointFeatures && pointFeatures.length > 0;
     const drawLines = lineFeatures && lineFeatures.length > 0;
-    const drawPolygons = stroked && polygonOutlineFeatures && polygonOutlineFeatures.length > 0;
+    const drawPolygons = polygonOutlineFeatures && polygonOutlineFeatures.length > 0;
     const fillPolygons = filled && polygonFeatures && polygonFeatures.length > 0;
 
     const onHover = this._onHoverSubLayer.bind(this);
@@ -111,7 +111,8 @@ export default class GeoJsonLayer extends CompositeLayer {
 
     // Polygon outline or wireframe
     let polygonOutlineLayer = null;
-    if (drawPolygons && extruded && wireframe) {
+    // If extruded, check the "wireframe" prop
+    if (extruded && wireframe && drawPolygons) {
       polygonOutlineLayer = new SolidPolygonLayer(Object.assign({}, this.props, {
         id: `${id}-polygon-wireframe`,
         data: polygonFeatures,
@@ -127,7 +128,8 @@ export default class GeoJsonLayer extends CompositeLayer {
         onHover,
         onClick
       }));
-    } else if (drawPolygons) {
+    // If non-extruded, check the "stroked" prop
+    } else if (!extruded && stroked && drawPolygons) {
       polygonOutlineLayer = new PathLayer(Object.assign({}, this.props, {
         id: `${id}-polygon-outline`,
         data: polygonOutlineFeatures,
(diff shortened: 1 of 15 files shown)


src/layers/core/polygon-layer/polygon-layer.js
```diff
@@ -83,26 +83,26 @@ export default class PolygonLayer extends CompositeLayer {
...
+          getElevation: updateTriggers.getElevation,
           getColor: updateTriggers.getFillColor
-        })
+        }
       }));
 
     // Polygon outline layer
     let polygonOutlineLayer = null;
-    if (strokePolygons) {
+    if (!extruded && stroked && polygon) {
```
ibgreen: Here `polygon` has only an indirect relation to `paths`. Could be better to check `paths` directly.

src/layers/core/polygon-layer/polygon-layer.js
```diff
@@ -111,9 +111,24 @@ export default class PolygonLayer extends CompositeLayer {
...
         onClick,
-        updateTriggers: Object.assign({}, updateTriggers, {
+        updateTriggers: {
+          getWidth: updateTriggers.getWidth,
           getColor: updateTriggers.getColor
-        })
+        }
+      }));
+    } else if (extruded && wireframe && polygon) {
+      polygonOutlineLayer = new SolidPolygonLayer(Object.assign({},
```
ibgreen: To me `outlineLayer` sounds like it is for the `Paths`, for clarity I think it could be worth a third variable - this would be the `polygonWireframeLayer`...

src/layers/core/polygon-layer/polygon-layer.js
ibgreen: This is a change, we now disable "strokes" when the "fill" is extruded.

src/layers/core/geojson-layer/geojson-layer.js
```diff
@@ -102,16 +101,17 @@ export default class GeoJsonLayer extends CompositeLayer {
...
       onHover,
       onClick
     }));
 
     // Polygon outline or wireframe
     let polygonOutlineLayer = null;
-    if (drawPolygons && extruded && wireframe) {
+    // If extruded, check the "wireframe" prop
+    if (extruded && wireframe && drawPolygons) {
       polygonOutlineLayer = new SolidPolygonLayer(Object.assign({}, this.props, {
```
ibgreen: Like in PolygonLayer below, recommend changing this variable to `polygonWireframeLayer` as outlines and wireframes are different concepts.

src/layers/core/geojson-layer/geojson-layer.js
```diff
@@ -80,20 +80,19 @@ export default class GeoJsonLayer extends CompositeLayer {
...
-    const drawPolygons = stroked && polygonOutlineFeatures && polygonOutlineFeatures.length > 0;
-    const fillPolygons = filled && polygonFeatures && polygonFeatures.length > 0;
+    const drawPolygons = polygonOutlineFeatures && polygonOutlineFeatures.length > 0;
 
     const onHover = this._onHoverSubLayer.bind(this);
     const onClick = this._onClickSubLayer.bind(this);
 
     // Filled Polygon Layer
-    const polygonFillLayer = fillPolygons && new SolidPolygonLayer(Object.assign({}, this.props, {
+    const polygonFillLayer = filled && drawPolygons && new SolidPolygonLayer(Object.assign({}, this.props, {
```
ibgreen: This doesn't look right to me - `drawPolygons` is derived from `polygonOutlineFeatures`, not `polygonFeatures`... See discussion below, maybe we can drop the length check on the data.

src/layers/core/polygon-layer/polygon-layer.js
```diff
@@ -83,26 +83,26 @@ export default class PolygonLayer extends CompositeLayer {
     const {data, id, stroked, filled, extruded, wireframe} = this.props;
     const {paths, onHover, onClick} = this.state;
 
-    const strokePolygons = stroked && data && data.length > 0;
-    const fillPolygons = filled && data && data.length > 0;
+    const polygon = data && data.length > 0;
```
shaojingli: I agree with the "hasData" is the proper name

src/layers/core/geojson-layer/geojson-layer.js
shaojingli: You are right. This is a bug.

PR #425, opened 2017-03-14
Fix the picking for polygon layer and extruded polygon layer

shaojingli:


src/layers/core/solid-polygon-layer/polygon-tesselator-extruded.js
```diff
@@ -184,10 +184,10 @@ function calculateColors({groupedVertices, getColor, wireframe = false}) {
 }
 
 function calculatePickingColors({groupedVertices, color = [0, 0, 0], wireframe = false}) {
-  const colors = groupedVertices.map((vertices, buildingIndex) => {
+  const colors = groupedVertices.map((vertices, polygonIndex) => {
     const numVertices = countVertices(vertices);
-    const topColors = new Array(numVertices).fill([0, 0, 0]);
-    const baseColors = new Array(numVertices).fill([0, 0, 0]);
+    const topColors = new Array(numVertices).fill([polygonIndex, 0, 0]);
```
ibgreen: This will only cover 255 colors. I think we need to use `decodePickingColor`

PR #430, opened 2017-03-15
add interoperability demo + identity viewport bug fix

gnavvy:
A demo to showcase deck.gl is interoperable with SVG-based visualizations.
One can use tools like d3-force to generate the layout for the input data and use deck.gl, instead of d3 for the rendering, which will lead to a huge performance gain.

SVG + Deck.gl rendering sharing the same data.
mar-15-2017 16-01-48


examples/interoperability/app.js
```diff
@@ -0,0 +1,125 @@
...
+
+  render() {
+    const {width, height, viewMode} = this.state;
+
+    const left = -Math.min(width, height) / 2;
+    const top = -Math.min(width, height) / 2;
+    const glViewport = new OrthographicViewport({width, height, left, top});
+
+    return width && height && <div>
+      {viewMode > 0 && <svg viewBox={`0 0 ${width} ${height}`}>
```
shaojingli: Yes, I know this is just a demo, but can we have a clearer viewMode? Using integers and also doing "(viewMode + 1) % 3" is a bit ... hacky ... to me

shaojingli:
For some reason, I'm not able to run this example per se...

screen shot 2017-03-15 at 4 33 33 pm

docs/api-reference/orthographic-viewport.md
```diff
@@ -9,7 +9,7 @@ A subclass of `Viewport` that creates an orthogonal matrix
 | `width`      | `Number`    | 1       | Width of "viewport" or window                 |
 | `height`     | `Number`    | 1       | Height of "viewport" or window                |
 view matrix arguments
-* `eye` (Vector3) - Defines eye position
+* `eye` (Vector3 = [0, 0, 1]) - Defines eye position, default [0, 0, 1]
```
ibgreen: No need to state the default twice?

shaojingli:
Also, I think we need a better name "interoperability" doesn't sound correct to me because deck.gl is not interacting with any other packages... it's shown to be a swappable renderer.

examples/interoperability/app.js
```diff
@@ -0,0 +1,125 @@
+import 'babel-polyfill';
```
ibgreen: Do we still need babel-polyfill?

examples/interoperability/app.js
```diff
@@ -0,0 +1,125 @@
...
+/* global document, window,*/
+/* eslint-disable no-console */
+import React, {PureComponent} from 'react';
+import {render} from 'react-dom';
+import DeckGL, {
+  ScatterplotLayer, OrthographicViewport, COORDINATE_SYSTEM
+} from '../../src';
```
ibgreen: In other expamples, we import `from 'deck.gl'` and use the shared webpack config to optionally "alias" that to `src`.

examples/interoperability/app.js
```diff
@@ -0,0 +1,125 @@
...
+// import DeckGL, {ScatterplotLayer} from 'deck.gl';
```
ibgreen: Cut?

examples/interoperability/app.js
```diff
@@ -0,0 +1,125 @@
...
+
+    const left = -Math.min(width, height) / 2;
+    const top = -Math.min(width, height) / 2;
+    const glViewport = new OrthographicViewport({width, height, left, top});
+
+    return width && height && <div>
+      {viewMode > 0 && <svg viewBox={`0 0 ${width} ${height}`}>
+        { this._renderSVGPoints() }
+      </svg>}
+      {viewMode !== 1 && <DeckGL width={width} height={height} viewport={glViewport}
```
ibgreen: This is quite a bit of logic to put inline in the JSX. Might be more readable to move this to if/switch before the render, use symbolic constants etc.

examples/interoperability/webpack.config.js
```diff
@@ -0,0 +1,31 @@
...
+      options: 'brfs-babel'
+    }]
+  },
+  node: {
+    fs: 'empty'
+  },
+  plugins: [
+    new webpack.HotModuleReplacementPlugin()
+  ]
+};
```
ibgreen: Add the webpack config in the root directory to support building from src?

examples/interoperability/webpack.config.js
```diff
@@ -0,0 +1,31 @@
...
+      test: /\.js$/,
+      loader: 'buble-loader',
+      include: [resolve('.')],
+      exclude: [/node_modules/],
+      options: {
+        objectAssign: 'Object.assign'
+      }
+    }, {
+      // Needed to inline deck.gl GLSL shaders
+      include: [resolve(__dirname, '../../src')],
```
ibgreen: This should no longer be needed?

src/shaderlib/project/project.glsl
```diff
@@ -227,9 +228,9 @@ vec2 project_mercator_(vec2 lnglat) {
 }
 
 vec2 project_position(vec2 position) {
-  // if (projectionMode == PROJECT_LINEAR) {
-  //   return (position + vec2(TILE_SIZE / 2.0)) * projectionScale;
-  // }
+  if (projectionMode == PROJECT_IDENTITY) {
+    return position * projectionScale;
```
ibgreen: The original idea was that `projectionScale` would always be 1 in `IDENTITY` - has that changed?

examples/interoperability/app.js
gnavvy: Yes, I'm aware. There're some changes and one commit after rc.1 and I wanted to test the example before I publish the rc.2. Once those changes are landed, I will change back import from '../../src' to 'deck.gl'.

This is also the reason for the error that @shaojingli saw. We are having two copies of react and I had to npm link it. will be resolved after we switch to import from 'deck.gl'

ibgreen: > Once those changes are landed, I will change back import from '../../src' to 'deck.gl'.

If you look at the other examples, they allow you to choose whether to run from the installed version or from the local source `npm start` vs `npm run start-local`. My point is that by following the same setup, you can have both models working, and you can debug from src now without npm link etc (the `webpack.config.local` in `/examples` parent dir takes care of aliasing all the modules so that npm link is not needed).

examples/interoperability/webpack.config.js
gnavvy: still needed until I finish the shaders in the example.
will cut in the next PR.

examples/interoperability/app.js
gnavvy: using npm run start-local helped me resolved multiple copies of luma.gl, but react still complains. maybe I misconfigured something, will give it another try.

src/shaderlib/project/project.glsl
gnavvy: no. makes sense, will remove the * projectionScale.

PR #431, opened 2017-03-16
more brfs cleanup

gnavvy:
glsl cleanup
brfs cleanup (except those in tests, brfs as devDep)


demo/package.json
```diff
@@ -47,7 +47,6 @@
     "babel-preset-es2015": "^6.3.13",
     "babel-preset-react": "^6.3.13",
     "babel-preset-stage-2": "^6.3.13",
-    "brfs-babel": "^1.0.0",
```
ibgreen: Do we still need the transform-loader after removing this (Per my understanding, the webpack transform loader is used to run browserify transforms under webpack, and brfs-babel is a browserify transform).

examples/layer-browser/package.json
```diff
@@ -5,7 +5,7 @@
     "build": "webpack app bundle.js --env.local --display-error-details"
   },
   "dependencies": {
-    "deck.gl": "^4.0.0-beta.3",
+    "deck.gl": "^4.0.0-rc1",
```
ibgreen: @shaojingli Didn't notice this until now. Correct [semver](www.semver.org) versioning is 4.0.0-rc.1 (note the last dot). We should probably republish the RC (make it a `rc.2`).

Hopefully the wildcards will still work and pick up the latest after we change.

examples/svg-interoperability/webpack.config.js
```diff
@@ -15,11 +15,6 @@ module.exports = {
       options: {
         objectAssign: 'Object.assign'
       }
-    }, {
-      // Needed to inline deck.gl GLSL shaders
-      include: [resolve(__dirname, '../../src')],
-      loader: 'transform-loader',
-      options: 'brfs-babel'
```
ibgreen: Nit: I still don't see the last line in this example that imports the webpack.local.config, like other examples do. Fine if intentional. Just want to make sure the mechanism is understood.

examples/layer-browser/package.json
shaojingli: Yes. I noticed that yesterday. @Pessimistress 's ocs-lite got updated to this version automatically. 

I'm going to republish it today after @gnavvy 's change got landed

demo/package.json
gnavvy: theoretically no we don't need it, but removing transform-loader from devDep completely caused tests to fails. (we are still using fs.readFileSync in the tests, I'd assume it is handled by node.js so we should not need transform-loader but it failed in my first trial.

Since transform-loader is in devDep only, I'd prefer to land this PR first to release the rc.2 (a blocker).

gnavvy:
found a regression in layer-browser that mapbox-gl still needs fs, aka cannot remove the transform-loaders I guess.
will patch together with https://github.com/uber/deck.gl/pull/432

PR #432, opened 2017-03-16
update webpack config for start-local scripts

gnavvy:
add start-local (run against source code) for the svg-compatibility example.
minor change on the addLocalDevSettings logic in webpack.config.local.js.
(^ resolve is not a must have field in webpack configs)

### DIFF
--- a/examples/webpack.config.local.js
+++ b/examples/webpack.config.local.js
@@ -46,7 +46,16 @@ const LOCAL_DEVELOPMENT_CONFIG = {
 };
 
 function addLocalDevSettings(config) {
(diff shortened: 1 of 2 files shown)


examples/webpack.config.local.js
```diff
@@ -46,7 +46,16 @@ const LOCAL_DEVELOPMENT_CONFIG = {
 };
 
 function addLocalDevSettings(config) {
-  Object.assign(config.resolve.alias, LOCAL_DEVELOPMENT_CONFIG.resolve.alias);
+  if (config.resolve) {
```
ibgreen: Idea: If we added this line and we could remove the big if statement:
```
config.resolve = config.resolve || {}
```
We could do the same for rules (although needs to be initialized to epmty array, not object.

PR #433, opened 2017-03-16
Update base-layer.md

shaojingli:


docs/layers/base-layer.md
```diff
@@ -1,27 +1,28 @@
...
 minified code). It is used as the default Layer id as well as for debugging
 and profiling.
 
 #### `defaultProps` (static, object, optional)
 
-Layers are strongly encouraged to define a `defaultProps` static member
-listing their props and default values, rather than programmatically
-declaring the props in constructor parameters etc. Using `defaultProps` improves
+All deck.gl layers define a `defaultProps` static member listing their props and
+default values, rather than programmatically declaring the props in their
```
ibgreen: Should we just drop the "rather than programmatically ..." as this is outdated)?

docs/layers/base-layer.md
```diff
@@ -107,33 +96,26 @@ Layers can be interacted with using these properties.
 
 - Default: `false`
 
-Whether layer responds to mouse events.
+Whether layer responds to mouse pointer picking events.
```
ibgreen: Whether the layer

docs/layers/base-layer.md
```diff
@@ -107,33 +96,26 @@ Layers can be interacted with using these properties.
...
 
 This callback will be called when the mouse hovers over the deck.gl
-viewport. The callback will be called for every layer that has
-both `visible` and `picking` properties set. When hovering over a
-feature drawn by the layer will be detected and mapped to an index.
-
-The function will be called with a single parameter `info`, which is an
-object that contains a variety of fields describing the mouse or touch
+viewport. The callback will be called for the layer that has
+been picked and the feature that user's mouse is hovered on will be
```
ibgreen: Maybe use `object` or `data item`/`data object` here instead of feature - feature is very GeoJson specific.

docs/layers/base-layer.md
```diff
@@ -107,33 +96,26 @@ Layers can be interacted with using these properties.
...
+reported to the caller of onHover through a single parameter `info`.
+This `info` object contains a variety of fields describing the mouse or touch
 event and what was hovered.
 
-If index is -1, no feature was matched for this layer
```
ibgreen: Why delete this? Do we no longer get this callback? @Pessimistress

docs/layers/base-layer.md
```diff
@@ -107,33 +96,26 @@ Layers can be interacted with using these properties.
...
-viewport. The callback will be called for every layer that has
-both `visible` and `picking` properties set. When clicking on a
-feature drawn by the layer this will be detected and mapped to an
-index in the data prop.
-
-The function will be called with a single parameter `info`, which is an
-object that contains a variety of fields describing the mouse or touch
+This callback will be called when the mouse clicks over the deck.gl
+viewport. The callback will be called for the layer that has
+been picked and the feature that user's mouse is clicked on will be
```
ibgreen: feature again -> object (see above)?

docs/layers/base-layer.md
```diff
@@ -147,25 +129,26 @@ that are not Web Mercator projected longitudes/latitudes.
 
 ##### `projectionMode` (Number, optional)
 
-Specifies how layer positions and offsets should be interpreted.
+Specifies how layer positions and offsets should be geographically interpreted.
```
ibgreen: We support a non-geographical mode, so maybe clearer to say 
Specifies how layer positions and offsets should be interpreted (as geospatial or mathematical coordinates).

docs/layers/base-layer.md
```diff
@@ -147,25 +129,26 @@ that are not Web Mercator projected longitudes/latitudes.
...
 
 The default is to interpret positions as latitude and longitude, however it
-is also possible to interpret positions as meter offsets (measured from the
-`projectionCenter` prop).
+is also possible to interpret positions as meter offsets on top of a projection
```
ibgreen: "on top of" => "added to" ?

docs/layers/base-layer.md
```diff
@@ -147,25 +129,26 @@ that are not Web Mercator projected longitudes/latitudes.
...
 
 See the article on Coordinate Systems for details
 
+TODO: This section needs to be rewritten
 ##### `modelMatrix` (Number[16], optional)
 
-An optional 4x4 matrix that is premultiplied into the affine projection
+An optional 4x4 matrix that is multiplied into the affine projection
 matrices used by shader `project` GLSL function and the
 Viewport's `project` and `unproject` JavaScript function.
```
ibgreen: Note that modelMatrix is applied after any non-linear cartographic projections. This means that modelMatrices are easiest to work with in projection modes that do not involve cartographic projections, such as METER_OFFSETS AND IDENTITY.

docs/layers/base-layer.md
```diff
@@ -201,45 +183,18 @@ can be refactored to avoid the need.
...
-The value of `updateTriggers` is a map with fields corresponding to
-attribute names (or `all`). Each field has a value which is an object,
-it can contain any amount of data. The data for each field is compared
-shallowly, and if a change is detected, the attribute is invalidated
-(all attributes are invalidated if the `all` key is used.)
-
-Note: updateTriggers are ignored by the normal shallow comparison, so it is
-ok for the app to mint a new object on every render.
-
-##### "attributes" properties
```
ibgreen: So `attributes` is the prop that we need for the PointCloudLayer (enabling binary PLY loader to just pass in buffers, defeating attribute generation).

It is implemented in attribute manager but not tested. On one hand it makes sense to not include the docs. But in that case we should probably create a new PR that adds it back for next release so we don't have to rewrite this.

docs/layers/base-layer.md
```diff
@@ -201,45 +183,18 @@ can be refactored to avoid the need.
...
-that share the same attributes (typical when using multiple layers to
-create effects like shadows or highlights around the primary primitives),
-making it desirable to avoid having multiple layers repeat the same
-automatic attribute recalculations.
-
-This allows an application to supply its own buffers properties.
-If attributes are shared between layers, a useful technique is to create
-a separate AttributeManager and calculate the attributes once, and then
-pass them in as "overrides" to the layers.
+This prop expects an object with key/value pairs that the keys matching attribute names.
```
ibgreen: typo (that) and functionality change (attribute names => accessors)
...pairs, with keys matching accessors

docs/layers/base-layer.md
```diff
@@ -201,45 +183,18 @@ can be refactored to avoid the need.
...
+If any values in this object are changed between props udpate, the corresponding attribute
```
ibgreen: Typo: update.
Change: the attribute corresponding to the accessor named by the key will be invalidated.

docs/layers/base-layer.md
```diff
@@ -201,45 +183,18 @@ can be refactored to avoid the need.
...
+will be invalidated.
+
+Using this method, the attribute update can happen when the content of `data` prop changed, or
```
ibgreen: Maybe drop the "on the GPU side"? Simpler is better?

Change: can happen not only when ..., but also ...

Final statement: updateTriggers cannot block attribute updates, just trigger them. Overriding updateState can block attribute updates.

docs/layers/base-layer.md
shaojingli: This line is not deleted....

docs/layers/base-layer.md
shaojingli: Thanks for the input. Yes, I'd like to write something similar but decided to do it later. I added a TODO for now.

docs/layers/base-layer.md
shaojingli: Yes. I'd like to keep the "attributes" prop away from the docs as we are not officially supporting it and also this attributes props is not part of the Layer but the LayerManager

ibgreen: >  also this attributes props is not part of the Layer but the LayerManager

While it is implemented by the AttributeManager, it works as a prop on the layer:
```
new Layer({
  attributes: {
    instancePositions: new Float32Array([0, 0, 0]),
    instanceColors: ...
  }
}
```

PR #436, opened 2017-03-17
Layer docs update

shaojingli:
Update layer docs
Removed two props from the GeoJSON layer example of Layer Browser

### DIFF
--- a/docs/layers/geojson-layer.md
+++ b/docs/layers/geojson-layer.md
@@ -7,13 +7,122 @@
 The GeoJson Layer takes in [GeoJson](http://geojson.org/) formatted data and
 renders it as interactive polygons, lines and points.
 
-  <div align="center">
-    <img height="300" src="/demo/src/static/images/demo-thumb-geojson.jpg" />
-  </div>
+<div align="center">
+  <img height="300" src="/demo/src/static/images/demo-thumb-geojson.jpg" />
+</div>
+
+    import {GeoJsonLayer} from 'deck.gl';
+
+## Properties
+
+Inherits from all [Base Layer properties](/docs/layers/base-layer.md),
+however, the `data` prop is interpreted slightly more flexibly to accommodate
+pure GeoJson "payloads".
+
+Also note that the `GeoJsonLayer` is a **composite** layer, that renders points, lines and polygons with its outlines with different "underlying" layers, so developers can sometimes achieve additional control of rendering by controlling
+the props of the underlying layers directly.
+
+#### `data` (Object, Array or iterable)
+
+A GeoJson `FeatureCollection` object, or a list of GeoJson `Feature` objects.
+For more information on how the GeoJson is interpreted see the introduction.
+
+#### `filled` (Boolean, optional)
+
+- Default: `true`
+
+Whether to draw filled polygons (solid fill). Note that for each polygon,
+only the area between the outer polygon and any holes will be filled. This
+prop is effective only when the polygon is NOT extruded.
+
+#### `stroked` (Boolean, optional)
+
+- Default: `false`
+
+Whether to draw an outline around polygons (solid fill). Note that
+for complex polygons, both the outer polygon as well the outlines of
+any holes will be drawn.
+
+#### `extruded` (Boolean, optional)
+
+Extrude Polygon and MultiPolygon features along the z-axis if set to
+true. The height of the drawn features is obtained using the `getElevation` accessor.
+
+- Default: `false`
+
+#### `wireframe` (Boolean, optional)
+
+- Default: `false`
+
+Whether to generate a line wireframe of the hexagon. The outline will have
+"horizontal" lines closing the top and bottom polygons and a vertical line
+(a "strut") for each vertex on the polygon.
+
+Remarks:
+* These lines are rendered with `GL.LINE` and will thus always be 1 pixel wide.
+* Wireframe and solid extrusions are exclusive, you'll need to create two layers
+  with the same data if you want a combined rendering effect.
+* This is only effective if the `extruded` prop is set to true.
+
+#### `fp64` (Boolean, optional)
+
+- Default: `false`
+
+Whether the layer should be rendered in high-precision 64-bit mode
+
+#### `lightSettings` (Object, optional) **EXPERIMENTAL**
+
+This is an object that contains light settings for extruded polygons.
+Be aware that this prop will likely be changed in the future versions of
+deck.gl.
+
+## Accessors
+
+#### `getColor`
+
+- Default: `object => object.strokeColor || [0, 0, 0, 255]`
+
+Called to retrieve the point or line color for a GeoJson feature, depending
+on its type.
+
+#### `getFillColor` (Function, optional)
+
+- Default: `object => object.fillColor || [0, 0, 0, 25]`
+
+Called to retrieve the polygon fill color for a GeoJson feature.
+
+Note: This accessor is only called for `Polygon` and `MultiPolygon` features.
+
+#### `getRadius`
+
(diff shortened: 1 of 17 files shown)


docs/layers/geojson-layer.md
```diff
@@ -80,69 +43,111 @@ Whether to draw an outline around polygons (solid fill). Note that
...
+- Default: `object => object.elevation || 1000`
 
 Called to retrieve the elevation of a polygon feature (when `extruded` is true).
 
 If a cartographic projection mode is used, height will be interpreted as meters,
 otherwise will be in unit coordinates.
 
 Note: This accessor is only called for `Polygon` and `MultiPolygon` features.
+
+## Remarks
```
ibgreen: Maybe call this section Example?

The remarks are usually just bullets with "advanced info"

Maybe move the text above the code?

docs/layers/geojson-layer.md
```diff
@@ -80,69 +43,111 @@ Whether to draw an outline around polygons (solid fill). Note that
...
+#### `wireframe` (Boolean, optional)
 
+- Default: `false`
 
-## Callbacks
+Whether to generate a line wireframe of the hexagon. The outline will have
+"horizontal" lines closing the top and bottom polygons and a vertical line
+(a "strut") for each vertex on the polygon.
 
-#### `onHover` (Function, optional)
```
ibgreen: About onHover, onClick  I think the key things we were trying to say is that the callbacks return a 'feature'. Since the GeoJson is not a standard array format. It could just be a bullet in the remarks.

docs/layers/geojson-layer.md
```diff
@@ -80,69 +43,111 @@ Whether to draw an outline around polygons (solid fill). Note that
...
-
-Called to retrieve the polygon fill color for a GeoJson feature.
-
-Note: This accessor is only called for `Polygon` and `MultiPolygon` features.
-
-
 #### `getElevation` (Function, optional)
 
-- Default: `object => object.elevation`
+- Default: `object => object.elevation || 1000`
```
ibgreen: What happens if object.elevation === 0? We probably need to look at this in many places.

An option could be to use object.elevation | 1000 instead of || 1000.

docs/layers/hexagon-cell-layer.md
```diff
@@ -59,16 +59,27 @@ to scale all hexagon elevations without updating the data.
 
 Whether to extrude hexagon. If se to false, all hexagons will be set to flat.
 
+##### `fp64` (Boolean, optional)
```
ibgreen: Is fp64 being added twice here?

docs/layers/hexagon-cell-layer.md
```diff
@@ -59,16 +59,27 @@ to scale all hexagon elevations without updating the data.
...
+
+- Default: `false`
+
+Whether the layer should be rendered in high-precision 64-bit mode
+
 ##### `selectedPickingColor` (Array, optional) **EXPERIMENTAL**
 
 Shader based highlighting of a selected object
 
+##### `fp64` (Boolean, optional)
```
ibgreen: again?

docs/layers/geojson-layer.md
```diff
@@ -7,70 +7,33 @@
...
-
 #### `filled` (Boolean, optional)
 
 - Default: `true`
 
 Whether to draw filled polygons (solid fill). Note that for each polygon,
-only the area between the outer polygon and any holes will be filled.
-
+only the area between the outer polygon and any holes will be filled. This
+prop is effective only when the polygon is NOT extruded.
```
ibgreen: Aren't extruded polygons always filled unless wireframe is true? While technically correct, the sentence could be read to say they are not.

docs/layers/geojson-layer.md
```diff
@@ -7,70 +7,33 @@
...
-puree GeoJson "payloads".
+pure GeoJson "payloads".
 
-Also note that the `GeoJsonLayer` is a **composite** layer, that renders points
-using `ScatterplotLayer`, lines and polygon outlines (`stroked` polygons)
-using `PathLayer`, and `filled` polygons using a primitive polygon layer.
-The `GeoJsonLayer` passes through most of its props to these underlying layers
-so you can sometimes achieve additional control of rendering by studying
-the props of the generated layers.
+Also note that the `GeoJsonLayer` is a **composite** layer, that renders points, lines and polygons with its outlines with different "underlying" layers, so developers can sometimes achieve additional control of rendering by controlling
```
ibgreen: Just a thought: If we don't document what the generated layers are, I think we need to make an effort to declare all the props from those layers in this layer.

E.g. how is a user to know that he can use `radiusScale` to control the radius of the points, etc, if it is not mentioned here, and the fact that we generate a `ScatterplotLayer` isn't mentioned either.

At the very least, I think we need to list the fp64 prop here, as we pass it down to all the sublayers.

examples/layer-browser/src/examples/core-layers.js
```diff
@@ -85,8 +85,6 @@ const GeoJsonLayerExample = {
     },
     getWidth: f => f.properties['stroke-width'],
     getElevation: f => 500,
-    widthScale: 10,
```
ibgreen: why remove these? no longer supported?

src/layers/core/geojson-layer/geojson-layer.js
```diff
@@ -43,11 +43,11 @@ const defaultProps = {
   // Point and polygon fill color
   getFillColor: f => get(f, 'properties.fillColor') || defaultFillColor,
   // Point radius
-  getRadius: f => get(f, 'properties.radius') || get(f, 'properties.size') || 5,
+  getRadius: f => get(f, 'properties.radius') || get(f, 'properties.size') || 1,
```
ibgreen: Again our radius === 0 issue.

src/layers/core/grid-layer/grid-layer.js
```diff
@@ -32,7 +32,9 @@ const defaultElevationScale = 1;
 const defaultProps = {
   cellSize: defaultCellSize,
   colorRange: defaultColorRange,
+  colorDomain: null,
   elevationRange: defaultElevationRange,
+  elecationDomain: null,
```
ibgreen: Typo elevationDomain

src/layers/core/solid-polygon-layer/solid-polygon-layer.js
```diff
@@ -34,7 +34,7 @@ import solidPolygonVertex64 from './solid-polygon-layer-vertex-64.glsl';
 import solidPolygonFragment from './solid-polygon-layer-fragment.glsl';
 
 const defaultProps = {
-  // Whether to extrude in 2.5D
+  // Whether to extrude in 3D
```
ibgreen: "along z-axis" to match your change in the docs above?

examples/layer-browser/src/examples/core-layers.js
shaojingli: @ibgreen This is where I got confused. This two props are props of the underlying path layer, and they are not defined in the defaultProps of GeoJsonLayer, and we didn't mentioned them in the API audit, so I assume that we don't want to expose it. 

We need to discuss again what props we would like to have for the GeoJsonLayer.

docs/layers/geojson-layer.md
shaojingli: The current implementation is that filled and wireframe props are independent with each other. If the user set 'wireframe' and 'filled' both to 'false', nothing will be drawn to the screen

shaojingli:
Docs updated to reflect the latest API guidelines implemented in #438

PR #437, opened 2017-03-17
fix brfs regressions

gnavvy:
resume brfs and transform-loader for examples / demo depending on mapbox-gl

### DIFF
--- a/examples/webpack.config.local.js
+++ b/examples/webpack.config.local.js
@@ -47,7 +47,8 @@ const LOCAL_DEV_CONFIG = {
 
 function addLocalDevSettings(config) {
   config.resolve = config.resolve || {};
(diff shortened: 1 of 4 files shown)


examples/webpack.config.local.js
```diff
@@ -47,7 +47,8 @@ const LOCAL_DEV_CONFIG = {
 
 function addLocalDevSettings(config) {
   config.resolve = config.resolve || {};
-  Object.assign(config.resolve, {alias: LOCAL_DEV_CONFIG.resolve.alias});
```
gnavvy: can someone enlighten me why js thinks the assign target (config.resolve) is null even if I have line 49?

for reference:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign

PR #438, opened 2017-03-17
Composite layer interface and props transfer

shaojingli:
Change the interface of GeoJson layer and Polygon layer to clear the confusion
Change the way composite layers pass their props to their underlying layers. Now it requires explicitly passing. Layers affected are the GeoJsonLayer, PolygonLayer, GridLayer, and HexagonLayer.

Docs change will be in an updated #436 

@Pessimistress @ibgreen @gnavvy

### DIFF
--- a/src/layers/core/geojson-layer/geojson-layer.js
+++ b/src/layers/core/geojson-layer/geojson-layer.js
@@ -92,7 +92,7 @@ export default class GeoJsonLayer extends CompositeLayer {
 
     const drawPoints = pointFeatures && pointFeatures.length > 0;
     const drawLines = lineFeatures && lineFeatures.length > 0;
-    const hasPolygonLine = polygonOutlineFeatures && polygonOutlineFeatures.length > 0;
+    const hasPolygonLines = polygonOutlineFeatures && polygonOutlineFeatures.length > 0;
     const hasPolygon = polygonFeatures && polygonFeatures.length > 0;
 
     const onHover = this._onHoverSubLayer.bind(this);
@@ -140,7 +140,7 @@ export default class GeoJsonLayer extends CompositeLayer {
 
     const polygonLineLayer = !extruded &&
       stroked &&
-      hasPolygonLine &&
+      hasPolygonLines &&
       new PathLayer({
         id: `${id}-polygon-outline`,
         data: polygonOutlineFeatures,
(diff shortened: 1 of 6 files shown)


src/layers/core/geojson-layer/geojson-layer.js
```diff
@@ -80,13 +84,15 @@ export default class GeoJsonLayer extends CompositeLayer {
...
+      getLineWidth, getElevation, updateTriggers} = this.props;
     const {id, stroked, filled, extruded, wireframe} = this.props;
+    const {lineWidthScale, lineWidthMinPixels, lineWidthMaxPixels,
+      lineJointRounded, lineMiterLimit, fp64} = this.props;
 
-    let {} = this.props;
     const drawPoints = pointFeatures && pointFeatures.length > 0;
     const drawLines = lineFeatures && lineFeatures.length > 0;
-    const hasPolygonOutline = polygonOutlineFeatures && polygonOutlineFeatures.length > 0;
+    const hasPolygonLine = polygonOutlineFeatures && polygonOutlineFeatures.length > 0;
```
ibgreen: hasPolygonLines?

src/layers/core/geojson-layer/geojson-layer.js
```diff
@@ -95,91 +101,106 @@ export default class GeoJsonLayer extends CompositeLayer {
...
-        onHover,
-        onClick
-      }));
+        }
+      });
 
     const polygonWireframeLayer = wireframe &&
       extruded &&
       hasPolygon &&
-      new SolidPolygonLayer(Object.assign({}, this.props, {
```
ibgreen: Looks good to get rid of the `Object.assign` here.

src/layers/core/geojson-layer/geojson-layer.js
```diff
@@ -95,91 +101,106 @@ export default class GeoJsonLayer extends CompositeLayer {
...
 
     return [
       polygonFillLayer,
       polygonWireframeLayer,
-      polygonOutlineLayer,
-      lineLayer,
+      polygonLineLayer,
+      pathLayer,
       pointLayer
     ].filter(Boolean);
```
ibgreen: We should to add the filtering by default inside deck.gl - Because I see so many layers doing `.filter(Boolean)` thing now. React supports `null`/empty elements, we should too. I think it is a one line change, could even go into this PR.

src/layers/core/grid-layer/grid-layer.js
```diff
@@ -35,6 +35,7 @@ const defaultProps = {
   elevationRange: defaultElevationRange,
   elevationScale: defaultElevationScale,
   getPosition: x => x.position,
+  extruded: false,
   fp64: false
   // AUDIT - getWeight ?
```
ibgreen: Remove?

src/layers/core/geojson-layer/geojson-layer.js
shaojingli: Agree. but let's do it in a different PR

Issue #444, opened 2017-03-21
picking issue for non-fullscreen widget


gnavvy:
https://github.com/uber/react-vis/pull/304
picking for scatterplot layer does not work in the above PR.

https://github.com/uber/deck.gl/blob/master/src/react/deckgl.js#L133
here **x** and **y** should be relative positions to the canvas, not screen coordinates.

will fix for v4.0.

ibgreen:
Related to #248?

PR #445, opened 2017-03-22
use offsetX & offsetY instead of x & y to get relative positions

### LINKED ISSUE #248: Mouse event positions incorrect when page scrolled
When the map is partially scrolled out of view, click events do not work as expected (see image). This issue only happens when the main page scroll is scrolling the map out of view, not the scroll of a containing element. From what I can tell, this is a bug in luma.gl. I will also file an issue there.

map bug

gnavvy:
fixes #248 and #444 
tested (ad-hoc) in the svg-interoperability example with a scolled offset.
will add a dedicated example consist of multiple non-full screen widgets later.

### DIFF
--- a/src/react/deckgl.js
+++ b/src/react/deckgl.js
@@ -119,7 +119,11 @@ export default class DeckGL extends React.Component {
 
   // Route events to layers
   _onClick(event) {
-    const {event: {offsetX: x = 0, offsetY: y = 0} = {}} = event;
+    // use offsetX|Y for relative position to the container, drop event if falsy
+    if (!event || !event.event || !Number.isFinite(event.event.offsetX)) {
+      return;
+    }
+    const {event: {offsetX: x, offsetY: y}} = event;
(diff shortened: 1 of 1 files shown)


src/react/deckgl.js
```diff
@@ -119,7 +119,7 @@ export default class DeckGL extends React.Component {
 
   // Route events to layers
   _onClick(event) {
-    const {x, y} = event;
+    const {event: {offsetX: x = 0, offsetY: y = 0} = {}} = event;
```
gnavvy: Just a sanity check. 

I agree it makes more sense to drop the event if either `event.event` is falsy (luma.gl addEvent fails, shouldn't be) or `offsetX` `offsetY` are falsy (browser fails, again shouldn't be according to the [spec](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/offsetX)); it's also heavy and requires changes on existing apps if we surface the error message via the picked object.

will just drop the event if sanity check fails.

PR #447, opened 2017-03-22
Fix the position generation for extruded polygons with holes

shaojingli:
@ibgreen @Pessimistress @gnavvy

### DIFF
--- a/src/layers/core/solid-polygon-layer/polygon-tesselator-extruded.js
+++ b/src/layers/core/solid-polygon-layer/polygon-tesselator-extruded.js
@@ -87,6 +87,7 @@ export class PolygonTesselatorExtruded {
 }
 
 function countVertices(vertices) {
+
   return vertices.reduce((vertexCount, polygon) => vertexCount + count(polygon), 0);
 }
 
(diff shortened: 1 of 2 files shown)


src/layers/core/solid-polygon-layer/polygon-tesselator-extruded.js
```diff
@@ -87,6 +87,7 @@ export class PolygonTesselatorExtruded {
 }
 
 function countVertices(vertices) {
+
```
ibgreen: Delete empty line?

src/layers/core/solid-polygon-layer/polygon-tesselator-extruded.js
```diff
@@ -117,15 +118,15 @@ function calculateIndices({groupedVertices, wireframe = false}) {
 // * each top vertex is on 3 surfaces
 // * each bottom vertex is on 2 surfaces
 function calculatePositionsJS({groupedVertices, wireframe = false}) {
-  const positions = groupedVertices.map(complexPolygon =>
-    complexPolygon.map(vertices => {
-      const topVertices = [].concat(vertices);
+  const positions = groupedVertices.map(
+    vertices => {
+      const topVertices = Array.prototype.concat.apply([], vertices);
```
ibgreen: What's the difference here?

shaojingli: Yeah, the bug is that applying double map() to groupedVertices changes the sequence of the vertices. So I reverted back to a single map(). 

However, using single map(), I need to use concat.apply() to have topVertices in flattened format for the baseVertices to be correctly calculated.

It took me some time to figure out that
[].concat([[1, 2, 3]]) is [[1, 2, 3]],
Array.prototype.concat.apply([], [[1, 2, 3]]) is [1, 2, 3]

Learned a Javascript lesson....

PR #448, opened 2017-03-22
Consistently handle picking in composite layers

Pessimistress:
- Because composite layers do not render anything themselves, they must manually trigger `onHover` and `onClick` callbacks by listening to sublayer events. This was a bug on `GridLayer` and `HexagonLayer`.
- Hover and click events on composite layers should have `info.layer` point to the composite layer and `info.object` point to user data, instead of sublayer and intermediate data. This was a bug on `GeoJsonLayer` and `PolygonLayer`.

### DIFF
--- a/src/layers/core/hexagon-layer/hexagon-layer.js
+++ b/src/layers/core/hexagon-layer/hexagon-layer.js
@@ -94,7 +94,7 @@ export default class HexagonLayer extends Layer {
       object: pickedCell
     });
 
-    this.props[handler](info);
+    return this.props[handler](info);
   }
 
   _onGetSublayerColor(cell) {
--- a/src/layers/core/hexagon-layer/hexagon-layer.js
+++ b/src/layers/core/hexagon-layer/hexagon-layer.js
@@ -41,8 +41,6 @@ const defaultProps = {
   //
 };
 
-function noop() {}
-
 function _needsReProjectPoints(oldProps, props) {
   return oldProps.radius !== props.radius;
 }
@@ -69,8 +67,7 @@ export default class HexagonLayer extends Layer {
   initializeState() {
     this.state = {
       hexagons: [],
-      countRange: null,
-      pickedCell: null
+      countRange: null
     };
   }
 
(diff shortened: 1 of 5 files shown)


Pessimistress:
We were unable to catch some of these bugs before because layer-browser assigns `onHover` and `onClick` handlers to the DeckGL canvas instead of using layer props. The "catch all" callbacks collect all unhandled events from sublayers, so we could not spot behavior problem on the composite layers.

@ibgreen

src/layers/core/hexagon-layer/hexagon-layer.js
```diff
@@ -86,24 +83,18 @@ export default class HexagonLayer extends Layer {
     }
   }
 
-  getPickingInfo(opts) {
-    const info = super.getPickingInfo(opts);
-    const pickedCell = this.state.pickedCell;
+  _onPickSubLayer(handler, info) {
+    const pickedCell = info.picked && info.index > -1 ?
+      this.state.hexagons[info.index] : null;
```
ibgreen: While I can see why this is done, it does look a bit strange: the app doesn't know anything about the aggregated cells... Yet we are suddenly exposing them to the user in the picking callback.

But I guess it doesn't make sense to send something from `props.data` back to user...

src/layers/core/polygon-layer/polygon-layer.js
```diff
@@ -73,14 +73,14 @@ export default class PolygonLayer extends CompositeLayer {
...
 
-  _onHoverSubLayer(info) {
-    info.object = (info.object && info.object.feature) || info.object;
-    this.props.onHover(info);
-  }
+  _onPickSubLayer(handler, info) {
+    Object.assign(info, {
+      layer: this,
+      // override object with picked feature
+      object: (info.object && info.object.feature) || info.object
```
ibgreen: 1) My off-the-cuff idea is that `PolygonLayer` shouldn't know anything about `object.feature` (that being a `GeoJsonLayer` thing). Can we remove the reference?
2) If so, it would be cool if there was a way to provide some generic forwarding of picking callbacks, perhaps on the `CompositeLayer` class. i.e any reason why these couldn't be moved to that layer?

Pessimistress: 1. Good catch, this seems to be another bug - `info.object.feature` should be `info.object.object` (reference back to user data from exploded paths)

Pessimistress: 2. Where I see the most potential in moving this to the base layer is not having to manually bind the callbacks and override sublayer props in `renderLayers`. Is this something `CompositeLayer` can do automatically without knowing about the sublayers?

src/layers/core/hexagon-layer/hexagon-layer.js
Pessimistress: It's reasonable for the user to get some aggregated information (e.g. total counts, location) from picking. This implementation does create a catch: `_onPickSubLayer` must be bound every frame for `this.state` to work.

src/layers/core/polygon-layer/polygon-layer.js
Pessimistress: A way to make this work is if the sublayers keep a reference to the parent layer. Then in picking we can directly call the `getPickingInfo()` and `onHover` of the parent layer.

PR #449, opened 2017-03-22
Fix PointCloudLayer bug where radiusPixels is not working

Pessimistress:
- Use `radiusPixels` instead of `radius` in rendering
- Better data sample in layer browser

### DIFF
--- a/src/layers/core/point-cloud-layer/point-cloud-layer.js
+++ b/src/layers/core/point-cloud-layer/point-cloud-layer.js
@@ -104,7 +104,7 @@ export default class PointCloudLayer extends Layer {
   draw({uniforms}) {
     const {radiusPixels, lightSettings} = this.props;
     this.state.model.render(Object.assign({}, uniforms, {
-      radius: radiusPixels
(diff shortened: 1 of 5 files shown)


src/layers/core/point-cloud-layer/point-cloud-layer.js
```diff
@@ -102,9 +102,9 @@ export default class PointCloudLayer extends Layer {
   }
 
   draw({uniforms}) {
-    const {radius, lightSettings} = this.props;
+    const {radiusPixels, lightSettings} = this.props;
     this.state.model.render(Object.assign({}, uniforms, {
-      radius
+      radius: radiusPixels
```
shaojingli: Let's just update the uniform names in the shader to "radiusPixels"

ibgreen: > Let's just update the uniform names in the shader to "radiusPixels"

Agreed. Let's keep uniforms and prop names aligned when we can.

PR #450, opened 2017-03-22
Simplify picking inplementation in CompositeLayer

Pessimistress:
- add `parentLayer` reference to sublayers
- `pickLayers()` only trigger callbacks on top level layers
- remove the requirement for `onHover` and `onClick` overrides
- use consistent interface (`getPickingInfo`) with the `Layer` class

### DIFF
--- a/src/lib/composite-layer.js
+++ b/src/lib/composite-layer.js
@@ -15,6 +15,10 @@ export default class CompositeLayer extends Layer {
   invalidateAttribute() {
   }
 
+  // called to augment the info object that is bubbled up from a sublayer
+  // override Layer.getPickingInfo() because decoding / setting uniform do
+  // not apply to a composite layer.
+  // @return null to cancel event
   getPickingInfo({info}) {
     return info;
   }
--- a/src/lib/composite-layer.js
+++ b/src/lib/composite-layer.js
@@ -15,8 +15,7 @@ export default class CompositeLayer extends Layer {
   invalidateAttribute() {
   }
 
-  getPickingInfo(opts) {
-    // do not call onHover/onClick on the container
-    return null;
+  getPickingInfo({info}) {
+    return info;
   }
 }
--- a/src/lib/draw-and-pick.js
+++ b/src/lib/draw-and-pick.js
@@ -136,7 +136,6 @@ export function pickLayers(gl, {
 
     affectedLayers.forEach(layer => {
       let info = Object.assign({}, baseInfo);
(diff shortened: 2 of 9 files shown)


src/lib/draw-and-pick.js
```diff
@@ -136,16 +136,21 @@ export function pickLayers(gl, {
...
       }
 
-      // Let layers populate its own info object
-      info = layer.pickLayer({info, mode});
+      // walk up the composite chain and find the owner of the event
+      // sublayers are never directly exposed to the user
+      while (layer && info) {
+        info.layer = layer;
+        // Let layers populate its own info object
+        info = layer.pickLayer({info, mode});
```
ibgreen: So all layers in the chain overwrite the same info object? I can see the pros (in terms of simplicity) but also cons with this (inadvertent clobbering of important fields, confusing mix of residual information from topmost and deeper layers).
Just a thought: Do we need more documentation on the pick info object to make the semantics here clearer to users and layer writers?

src/lib/composite-layer.js
```diff
@@ -15,8 +15,7 @@ export default class CompositeLayer extends Layer {
   invalidateAttribute() {
   }
 
-  getPickingInfo(opts) {
-    // do not call onHover/onClick on the container
-    return null;
+  getPickingInfo({info}) {
```
ibgreen: If this is different from base layer, maybe a short comment on why?

src/lib/draw-and-pick.js
Pessimistress: Chaining is needed here because the sublayer is the one that can interpret the encoded picking color, and the composite layer is the one that knows the original user data.
We definitely should have a dedicated doc about picking. I can take a stab at it once the demo site is fixed.

ibgreen: We absolutely need chaining, the question is more whether these should all modify in the same picking object or we should create a chain of picking objects, one for the parent and one for each sublayer. Imagine:

`info.parent`


I can't see many applications using that chain though, so it is probably overkill. But it would avoid the potential mess of multiple layers overwriting (or not overwriting) each other's fields in the same info object.

Pessimistress: Downside of a nested info object is that if the implementation of a layer is changed (for example, `PolygonLayer` went from a standalone layer to a composite layer), the structure of the picking info may also change. I like the simplicity of our current API where callbacks only need to look at `info.object` for the details of the picked object.

PR #451, opened 2017-03-22
When the iconMapping changes we must invalid any dependent attributes

twojtasz:
This is a thin PR.  I don't see any existing tests for the icon layer.  So I will add one and test the call to accessors when the state changes.

### DIFF
--- a/src/layers/core/icon-layer/icon-layer.js
+++ b/src/layers/core/icon-layer/icon-layer.js
@@ -107,7 +107,14 @@ export default class IconLayer extends Layer {
   updateState({oldProps, props, changeFlags}) {
     super.updateState({props, oldProps, changeFlags});
 
-    const {iconAtlas} = props;
+    const {iconAtlas, iconMapping} = props;
+
+    if (oldProps.iconMapping !== iconMapping) {
+      const {attributeManager} = this.state;
+      attributeManager.invalidate('instanceOffsets');
+      attributeManager.invalidate('instanceIconFrames');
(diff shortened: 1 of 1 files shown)


src/layers/core/icon-layer/icon-layer.js
```diff
@@ -107,7 +107,14 @@ export default class IconLayer extends Layer {
   updateState({oldProps, props, changeFlags}) {
     super.updateState({props, oldProps, changeFlags});
 
-    const {iconAtlas} = props;
+    const {iconAtlas, iconMapping} = props;
+
+    if (oldProps.iconMapping !== iconMapping) {
```
shaojingli: Though in practice, iconMapping and iconAtlas almost always change at the same time. but logically, we might want to do:  if (oldProps.iconMapping !== iconMapping || oldProps.iconAtlas !== iconAtlas) { ...

twojtasz: So invalidate properties if either change, OK.

ibgreen: As far as I can tell, the listed attributes depend only on the `iconMapping` (in the sense that the functions calculating those attributes refer to `iconMapping`) but they do not depend on the `iconAtlas`. So recalculating them if only `iconAtlas` changes should be a noop - i.e generate identical results.

Agree it is better to regenerate the attributes too often (small perf hit) rather than too seldom (incorrect rendering)...

PR #454, opened 2017-03-23
fix projectionMode checks and radiusPixels for point-cloud-layer-64

gnavvy:
bug fixes for typos (COORDINATE_SYSTEM.**LNG_LAT**, instead of COORDINATE_SYSTEM.**LNGLAT**) and raidusPixels for vertex-64 shader of the point-cloud-layer.

### DIFF
--- a/src/layers/core/point-cloud-layer/point-cloud-layer-vertex-64.glsl.js
+++ b/src/layers/core/point-cloud-layer/point-cloud-layer-vertex-64.glsl.js
@@ -31,7 +31,7 @@ attribute vec3 instancePickingColors;
 
 uniform float renderPickingBuffer;
 uniform float opacity;
-uniform float radius;
+uniform float radiusPixels;
 uniform vec2 screenSize;
 
 varying vec4 vColor;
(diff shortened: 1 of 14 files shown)


src/lib/utils/fp64.js
```diff
@@ -10,11 +10,11 @@ export function fp64ify(a) {
 
 export function enable64bitSupport(props) {
   if (props.fp64) {
-    if (props.projectionMode === COORDINATE_SYSTEM.LNG_LAT) {
```
gnavvy: basically, this check always fails and we only used 32bit versions.

src/layers/core/point-cloud-layer/point-cloud-layer-vertex-64.glsl.js
```diff
@@ -31,7 +31,7 @@ attribute vec3 instancePickingColors;
 
 uniform float renderPickingBuffer;
 uniform float opacity;
-uniform float radius;
+uniform float radiusPixels;
```
gnavvy: point-cloud-layer-vertex-64.glsl was never hit as we failed the projectionMode check.

PR #455, opened 2017-03-23
Composite layer to transfer positionOffset and modelMatrix

shaojingli:
Transfer positionOffset and modelMatrix props properly to their underlying layers

@Pessimistress @ibgreen

### DIFF
--- a/src/layers/core/geojson-layer/geojson-layer.js
+++ b/src/layers/core/geojson-layer/geojson-layer.js
@@ -97,7 +97,7 @@ export default class GeoJsonLayer extends CompositeLayer {
     const {opacity, pickable, visible, projectionMode} = this.props;
 
     // viewport props
-    const {positionOffset, modelMatrix} = this.props;
+    const {positionOrigin, modelMatrix} = this.props;
 
     const drawPoints = pointFeatures && pointFeatures.length > 0;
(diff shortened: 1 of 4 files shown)


Pessimistress:
Looks like `positionOffset` has been renamed to `positionOrigin`. The base layer doc needs to be updated.

shaojingli:
> Looks like positionOffset has been renamed to positionOrigin. The base layer doc needs to be updated.

Obviously, I'm not 100% awake on the train to work....

Updated

src/layers/core/geojson-layer/geojson-layer.js
```diff
@@ -82,17 +82,23 @@ export default class GeoJsonLayer extends CompositeLayer {
...
 
+    // Accessor props for underlying layers
+    const {getLineColor, getFillColor, getRadius,
+      getLineWidth, getElevation, updateTriggers} = this.props;
+
     // base layer props
     const {opacity, pickable, visible, projectionMode} = this.props;
 
+    // viewport props
+    const {positionOrigin, modelMatrix} = this.props;
```
Pessimistress: projectionMode is a viewport prop also?

shaojingli:
I agree with you @ibgreen. I'll get this PR landed and then create an issue to mitigate this problem

Issue #443, opened 2017-03-21
Picking bug


dcposch:
**Version:** 4.0.0-rc1

## Steps to reproduce

* Create a ScatterplotLayer
* Zoom and pan

## Observed result

Once per minute or so, it crashes with the following error:

```
Uncaught TypeError: Cannot read property 'model' of null
    at ScatterplotLayer.setUniforms (webpack:///./~/deck.gl/dist/lib/layer.js?:767:21)
    at ScatterplotLayer.getPickingInfo (webpack:///./~/deck.gl/dist/lib/layer.js?:201:14)
    at ScatterplotLayer.pickLayer (webpack:///./~/deck.gl/dist/lib/layer.js?:590:19)
    at eval (webpack:///./~/deck.gl/dist/lib/draw-and-pick.js?:152:20)
    at Array.forEach (native)
    at eval (webpack:///./~/deck.gl/dist/lib/draw-and-pick.js?:141:20)
    at glContextWithState (webpack:///./~/luma.gl/dist-es6/webgl/context.js?:188:13)
    at pickLayers (webpack:///./~/deck.gl/dist/lib/draw-and-pick.js?:63:32)
    at LayerManager.pickLayer (webpack:///./~/deck.gl/dist/lib/layer-manager.js?:188:42)
    at DeckGL._onMouseMove (webpack:///./~/deck.gl/dist/react/deckgl.js?:215:45)
    at EventsProxy.mousemove (webpack:///./~/luma.gl/dist-es6/core/event.js?:352:24)
    at HTMLCanvasElement.eval (webpack:///./~/luma.gl/dist-es6/core/event.js?:160:26)
```

I couldn't find a reliable way to reproduce this, but it happens frequently.

ibgreen:
@dc  Thanks for reporting, we haven't seen this in our testing.
- Can this be reproduced in any of the bundled examples or is this in your app?
- stacktrace shows that this is happening during picking (which was refactored in R4) so that is a starting point...
- Looks like `layer.state` is `null`, this typically happens only when a discarded layer from a previous render cycle is accessed.
- Do you have multiple Scatterplot layers? (perhaps there are some issues with layer `id` prop?).

shaojingli:
From my experience, this is usually due to two Scatterplot layer having the same "id".

dcposch:
@shaojingli @ibgreen thanks!

[...]

PR #458, opened 2017-03-23
Fix the bug that layers got updated in the middle of the picking cycle

shaojingli:
@Pessimistress 
please take a look and see if I understand your request correctly

@ibgreen

### DIFF
--- a/src/lib/draw-and-pick.js
+++ b/src/lib/draw-and-pick.js
@@ -169,28 +169,25 @@ export function pickLayers(gl, {
     });
 
     infos.forEach(info => {
-      // If layer.getPickingInfo() returns null, do not proceed
-      if (info) {
-        let handled = false;
-        // The onClick and onHover functions are provided by the user
-        // and out of control by deck.gl. It's very much possible that
-        // the user calls React lifecycle methods in these function, such as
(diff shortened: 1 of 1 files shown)


src/lib/draw-and-pick.js
```diff
@@ -152,6 +155,12 @@ export function pickLayers(gl, {
         layer = layer.parentLayer;
       }
 
+      // This guarantees that there will be only one copy of info with
+      // the same composite layer
+      infos.set(info.layer.id, info);
```
Pessimistress: Check if `info` is empty

shaojingli:
@Pessimistress

shaojingli:
#443

src/lib/draw-and-pick.js
```diff
@@ -143,21 +150,38 @@ export function pickLayers(gl, {
...
+      // This guarantees that there will be only one copy of info for
+      // one composite layer
+      if (info) {
+        infos.set(info.layer.id, info);
+      }
+    });
+
+    infos.forEach(info => {
       // If layer.getPickingInfo() returns null, do not proceed
       if (info) {
```
Pessimistress: Not necessary since only non-empty info objects were added to the map

PR #460, opened 2017-03-23
replace lat lon offset with cellSize

heshan0131:


examples/layer-browser/src/utils/grid-aggregator.js
```diff
@@ -1,12 +1,12 @@
...
 
 /**
  * Aggregate points for the grid layer
  * @param {array} points
- * @param {number} worldUnitSize - unit size in kilometer
+ * @param {number} cellSize - cell size in meters
  * @returns {object} - grid data and cell dimension
  */
-export function pointsToWorldGrid(points, worldUnitSize) {
+export function pointsToWorldGrid(points, cellSize) {
```
shaojingli: So you changed all worldUnitSize to cellSize here. Should we do the same in src/layers/core/grid-layer/grid-aggregator.js

PR #467, opened 2017-03-26
Add IconLayer Demo

Pessimistress:
- 109k icons dataset
- Clustering
- Hover and click interaction

screen shot 2017-03-26 at 1 03 11 pm copy

@contra @ibgreen


demo/src/actions/app-actions.js
```diff
@@ -41,13 +56,17 @@ export const loadData = (owner, dataArr) => {
       return;
     }
 
-    let resultData = [];
-    let resultMeta = {};
     const isArray = Array.isArray(dataArr);
```
ibgreen: Nit: the name`dataArr` bothers me. If we really wanted to give a hint that this must be an array, then we should type it out and call the parameter `dataArray`. However the first thing we do in the function is to handle non-array values so I think it should just be called `data`.

demo/src/actions/app-actions.js
```diff
@@ -41,13 +56,17 @@ export const loadData = (owner, dataArr) => {
...
 
     if (!isArray) {
       dataArr = [dataArr];
     }
+    const context = {
+      owner,
+      resultData: [],
+      resultMeta: [],
+      isArray
```
ibgreen: Do we need to pass this when we have wrapped the data in an array above?

demo/src/actions/app-actions.js
```diff
@@ -28,10 +28,25 @@ export const loadContent = filename => {
 
 const loadDataStart = owner => ({type: 'LOAD_DATA_START', owner});
 
-const loadDataSuccess = (owner, data, meta) => ({
-  type: 'LOAD_DATA_SUCCESS',
-  payload: {owner, data, meta}
-});
+const loadDataSuccess = (context, index, data, meta) => {
+
+  if (context.isArray) {
```
ibgreen: Apparently this function can handle a couple of different types of data, and expects them to have a certain structure ? Worth documenting?

Also it seems designed to gradually add things to the context parameter (instead of adding them to some state variable). This seems to be outside of the normal redux model and worth commenting).

demo/src/components/demos/icon.js
```diff
@@ -0,0 +1,255 @@
...
+  evt.stopPropagation();
+}
+
+export default class IconDemo extends Component {
+
+  static get data() {
+    return [
+      {
+        url: 'data/film-locations.txt',
+        worker: 'workers/film-locations-decoder.js'
```
ibgreen: Why do we specify a JS file here, rather than reference an import?

demo/src/components/demos/icon.js
```diff
@@ -0,0 +1,255 @@
...
+      viewport.height !== oldViewport.height)) {
+      this._updateCluster(points, nextProps.viewport);
+    }
+  }
+
+  // pre-compute icon clusters
+  _updateCluster(points, viewport) {
+    const tree = this._tree;
+
+    const transform = new WebMercatorViewport({
```
ibgreen: Call it `viewport` and reserve `transform` for matrices?

demo/src/components/demos/icon.js
```diff
@@ -0,0 +1,255 @@
...
+      maxZoom: 20,
+      pitch: 0,
+      bearing: 0
+    };
+  }
+
+  static renderInfo(meta) {
+    return (
+      <div>
+        <h3>Where Films Were Made (2007 - 2017)</h3>
```
ibgreen: Maybe a line break before `(2007` to make the final layout cleaner?

demo/src/components/demos/icon.js
```diff
@@ -0,0 +1,255 @@
...
+        </div>
+      </div>
+    );
+  }
+
+  constructor(props) {
+    super(props);
+
+    // build spatial index
+    this._tree = rbush(9, ['.x', '.y', '.x', '.y']);
```
ibgreen: Seems possible in rbush to specify array format points to avoid destructuring below. Assume that is not used because of the `zoomLevel` field below?

> var tree = rbush(9, ['[0]', '[1]', '[0]', '[1]']); // accept [x, y] points

https://github.com/mourner/rbush#data-format

demo/src/static/code-markdown/icon.md
```diff
@@ -0,0 +1,131 @@
+```
```
ibgreen: In luma.gl we load the actual code using the ACE editor. Here we maintain separate copies of the code in markdown, which is a maintenance problem?

demo/src/actions/app-actions.js
```diff
@@ -41,13 +56,17 @@ export const loadData = (owner, dataArr) => {
...
     }
 
-    let resultData = [];
-    let resultMeta = {};
     const isArray = Array.isArray(dataArr);
 
     if (!isArray) {
       dataArr = [dataArr];
     }
+    const context = {
```
ibgreen: The word `context` has a lot of baggage. Can we be more specific?

The use is not really clear to me but I assume this is something the worker will gradually populate? (`workerContext`, `workerData`, `workerState` etc).

demo/src/static/code-markdown/icon.md
Pessimistress: This needs to be a longer discussion - unlike luma.gl, these demos are not standalone applications, and a lot of the code wrapped into the components do not make sense unless you know how the container uses them. I don't like where it is either - it's hard to maintain and very confusing for people who try to dig into the demo source. I'm willing to make a push to refactor this before v4.

demo/src/components/demos/icon.js
Pessimistress: The source data come in the format of `[{name, scene, coordinates}]`. The projected positions of the coordinates are calculated in `_updateCluster()` and tagged onto each datum as `x` and `y`.

I use the projected positions instead of longitude and latitude to build this spatial index (because this particular dataset is distributed all over the world, we can't use some fixed deltaLon and deltaLat).

Will add more comments for this.

demo/src/components/demos/icon.js
Pessimistress: This is a static JS file used to construct a web worker at runtime.

Issue #466, opened 2017-03-25
Make sublayer instance accessible in `CompositeLayer.getPickingInfo()`


Pessimistress:
### Problem
Because the source sublayer is no longer exposed, user cannot define different hover behaviors for each sublayer by checking `info.layer`. They will instead have to infer this information using `info.object`, which requires them to know the implementation details of the composite layer.

### Proposed feature
When calling `getPickingInfo()` on a composite layer, pass an additional `sourceLayer` argument so that the composite layer can populate `info` with appropriate properties. This change will not affect the existing core layers, but give custom composite layers more context to decide what to expose to the user.

@ibgreen @shaojingli

shaojingli:
#468

PR #468, opened 2017-03-26
Expose sourceLayer to getPickingInfo()

Pessimistress:
Addressing [Issue #466](https://github.com/uber/deck.gl/issues/466)


shaojingli:
The potential issue here is that if a composite layer consists of another composite layer, we might still not able to get the behavior we want. We can discuss this offline.

ibgreen:
> We probably want to make sure that this solution works as expected when a composite layer consists of another composite layer. We can discuss this offline tomorrow

Good point. I still suspect we might have dismissed the idea of `info.parentInfo` too quickly. The core implementation of composite layer picking could set up a new `info` object with the `parent` field pointing to the "primitive" layer and the composite layer's job would be to look at that and make the right choices to populate the new `info` object.

PR #470, opened 2017-03-27
add lowerPercentile & upperPercentile back to hexagonLayer

heshan0131:


src/layers/core/hexagon-layer/hexagon-layer.js
```diff
@@ -94,17 +127,35 @@ export default class HexagonLayer extends Layer {
...
+  }
+
   _onGetSublayerColor(cell) {
     const {colorRange} = this.props;
-    const colorDomain = this.props.colorDomain || this.state.countRange;
+    const {valueDomain} = this.state;
+
+    const colorDomain = this.props.colorDomain || valueDomain;
+    const count = cell.points.length;
+    const color = quantizeScale(colorDomain, colorRange, count);
```
ibgreen: Does `quantizeScale` depend on an additional d3-module being imported by deck.gl? This could be the reason this was pushed out previously?

src/layers/core/hexagon-layer/hexagon-layer.js
```diff
@@ -77,9 +103,16 @@ export default class HexagonLayer extends Layer {
       const {viewport} = this.context;
 
       const hexagons = hexagonAggregator(this.props, viewport);
-      const countRange = _getCountRange(hexagons);
+      const sortedCounts = new SortedBins(hexagons);
```
ibgreen: The name of this class doesn't feel quite right, but not sure I can't come up with anything better. `BinSorter`?

src/utils/sorted-bins.js
```diff
@@ -18,6 +18,7 @@ export default class SortedBins {
   /**
    * Get an array of object with sorted count and index of bins
    * @param {Number} lower
```
ibgreen: Nit: This doesn't correctly document the params (looks like `f(lower, upper)`) . But probably no need to change this now. We'll probably move to something like flow type annotation comments in the future. A more correct version of JSDoc would probably look like:
```
@param {Number[]} range
@param {Number} range[0] - lower bound
...
```

src/layers/core/hexagon-layer/hexagon-layer.js
heshan0131: `quantizeScale` is implemented from scratch in `utils/scale-util`. We are not depend on d3 for this

src/layers/core/hexagon-layer/hexagon-layer.js
```diff
@@ -26,48 +26,74 @@ import {quantizeScale, linearScale} from '../../../utils/scale-utils';
...
   constructor(props) {
     if (!props.radius) {
-      log.once(0, 'PointDensityHexagonLayer: radius in meter is needed to aggregate points into ' +
+      log.once(0, 'HexagonLayer: radius in meter is needed to aggregate points into ' +
         'hexagonal bins, Now using 1000 meter as default');
 
       props.radius = defaultProps.radius;
     }
 
+    if (Number.isFinite(props.upperPercentile) &&
```
heshan0131: I agree we should have a more formal way of checking props. At the moment, this can't be called by checkRequiredProps. because they are not required props. I only check if those props are valid when user provide them.

Maybe we should create a ticket for better prop checking?

src/layers/core/hexagon-layer/hexagon-layer.js
```diff
@@ -26,48 +26,74 @@ import {quantizeScale, linearScale} from '../../../utils/scale-utils';
...
+function _percentileChanged(oldProps, props) {
+  return oldProps.lowerPercentile !== props.lowerPercentile ||
+    oldProps.upperPercentile !== props.upperPercentile;
 }
 
 export default class HexagonLayer extends Layer {
   constructor(props) {
     if (!props.radius) {
-      log.once(0, 'PointDensityHexagonLayer: radius in meter is needed to aggregate points into ' +
+      log.once(0, 'HexagonLayer: radius in meter is needed to aggregate points into ' +
```
heshan0131: This is not a deprecation warning. I simply changed the name of the layer after we changed PointDensityHexagonLayer -> HexagonLayer

src/utils/sorted-bins.js
heshan0131: I'm no quitter

PR #471, opened 2017-03-27
Demo refactor - make examples accessible as standalone apps

Pessimistress:
- Rewrite the trips example to be self-contained and easier to read
- Add `example/trips` with sample data
- Import rendering component from the standalone app into demo site

@ibgreen @ericsoco


demo/webpack/config.js
```diff
@@ -26,20 +41,16 @@ module.exports = {
...
 
   resolve: {
     modules: [resolve(rootDir, 'node_modules'), resolve(demoDir, 'node_modules')],
     alias: {
       'deck.gl': libSources,
-      webworkify: 'webworkify-webpack-dropin'
-      // react: resolve('./node_modules/react'),
-      // 'react-dom': resolve('./node_modules/react-dom'),
-      // 'gl-matrix': resolve('./node_modules/gl-matrix/dist/gl-matrix.js')
+      webworkify: 'webworkify-webpack-dropin',
```
ibgreen: I recommend adding a comment above every line in a webpack config. They are really hard to read otherwise. E.g. explain why we need webworkify, what does it do...

examples/trips/README.md
```diff
@@ -0,0 +1,2 @@
+* [hello-world-webpack2](./hello-world-webpack.md): Bundles a minimal app with
```
ibgreen: Copy paste... Needs content.

Also the README in the parent examples folder should be updated.

examples/trips/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "scripts": {
+    "start-local": "webpack-dev-server --env.local --progress --hot --port 3030 --open",
```
ericsoco: why `:3030` instead of default `:3000`?

Pessimistress: For running this app and the demo at the same time :)

ericsoco:
Makes sense to me, and most things don't yet. So 👍

PR #478, opened 2017-03-28
Add 3d heatmap demo

heshan0131:
- Add 3d heatmap demo using hexagonLayer
- Use UK traffic accident data
- Create a standalone app and import it from demo

3d-heatmap-examples


examples/3d-heatmap/webpack.config.js
```diff
@@ -0,0 +1,40 @@
...
+      options: {
+        objectAssign: 'Object.assign'
+      }
+    }]
+  },
+
+  resolve: {
+    alias: {
+      // From mapbox-gl-js README. Required for non-browserify bundlers (e.g. webpack):
+      'deck.gl': resolve('./node_modules/deck.gl'),
```
ibgreen: This should not be under the comment that relates to mapbox, it should be separate and separately commented (my current position is that every line in a webpack config needs its own comment).

demo/src/components/demos/heatmap.js
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  static renderInfo(meta) {
+
+    const colorRamp = HeatmapOverlay.defaultColorRange.slice()
+      .map(color => `rgb(${color.join(',')})`);
+
+    return (
+      <div>
+        <h3>United Kingdom Road Safty</h3>
```
shaojingli: Safety

examples/3d-heatmap/deckgl-overlay.js
```diff
@@ -0,0 +1,152 @@
...
+  constructor(props) {
+    super(props);
+    this.startAnimationTimer = null;
+    this.intervalTimer = null;
+    this.state = {
+      elevationScale: elevationScale.min,
+      hoveredObject: null
+    };
+
+    this._startAnimate = this._startAnimate.bind(this);
```
shaojingli: I like the animations!

ibgreen:
> I guess that in the future we'll always have to store our large data files twice... once in the example folder and once in the demo folder... 

The idea was that the files in the example folder would be small subsets, quick to load uncompressed.

For the big data files in the demo repository, we really need to get the out of this repository. We can't let deck.gl become a 1GB repo that takes an hour to clone.

I'll open an issue on this.

examples/3d-heatmap/webpack.config.js
heshan0131: This needs to be consistent among all examples. Will make another request to address

PR #482, opened 2017-03-28
JSX wrappers

ibgreen:
Add new example that enables code to be written as follows (allows use of JSX syntax for deck.gl layers):
```
        <Deck {...viewport} width={width} height={height} debug>
          <LineLayer
            data={[
              {sourcePosition: [-122.41669, 37.7883], targetPosition: [-122.41669, 37.781]}
            ]}
            strokeWidth={5}/>
          <ScatterplotLayer
            data={[{position: [-122.41669, 37.79]}]}
            radiusScale={100}
            getColor={x => [0, 0, 255]}/>
        </Deck>
```
Just wanted to make sure this was pushed and put up the PR. I'm not particular on whether this makes it into v4 or not.


ibgreen:
> Should we consider having this in our core library in the future?

Yes, I think this falls squarely under the "make deckgl easier to use", this would cover the level 1 users (React programmers) who can now use a more familiar syntax.

Of course these are not true React components (just Layer masquerading as React components) so there could be surprises for them down the road, but that is a small downside for making the initial learning curve easier I think

ibgreen:
@mcnuttandrew / @apercu - as discussed in react-vis - you could copy the 5 line `Deck` wrapper from this example if you wanted to start using JSX syntax for deck.gl layers right now.

PR #484, opened 2017-03-28
Make composite layers transfer the lightSettings prop

shaojingli:
to its underlying layer

It slipped through because it's not one of the props defined in base layer's default props...


shaojingli:
Thanks to @Pessimistress ... who discovered this through a new dataset she is exploring...

Yes. Forgot the Polygon layer. Will change

PR #485, opened 2017-03-29
Demo site links update

shaojingli:
Clear remaining issues around the change from choropleth to geojson for our demo site


shaojingli:
@Pessimistress @ibgreen

PR #489, opened 2017-03-29
Add LineLayer Example

Pessimistress:
screen shot 2017-03-28 at 10 23 46 pm copy


examples/line/README.md
```diff
@@ -0,0 +1,13 @@
+This is a minimal standalone version of the GeoJsonLayer example
```
shaojingli: copy / paste

demo/src/static/workers/flight-path-data-decoder.js
```diff
@@ -0,0 +1,55 @@
+importScripts('./util.js');
+const FLUSH_LIMIT = 20000;
+let result = [];
+let index = 0;
+let count = 0;
+
+onmessage = function(e) {
```
ibgreen: `e` is not a good parameter name. Simply spelling this out will make this code easier to read.

demo/src/static/workers/flight-path-data-decoder.js
```diff
@@ -0,0 +1,55 @@
...
+  const lines = e.data.text.split('\n');
+
+  lines.forEach(function(line) {
```
ibgreen: Arrow syntax?

demo/src/static/workers/flight-path-data-decoder.js
```diff
@@ -0,0 +1,55 @@
...
+
+    if (!line) {
+      return;
+    }
+
+    var parts = line.split('\t');
+    var coords0 = parts[2].split('\x01').map(function(str) { return decodePolyline(str, 5) });
```
ibgreen: Arrow syntax? Spell out `string`?

demo/src/static/workers/flight-path-data-decoder.js
```diff
@@ -0,0 +1,55 @@
...
+    var coords1 = parts[3].split('\x01').map(function(str) { return decodePolyline(str, 1) });
+
+    coords0.forEach(function(lineStr, i) {
```
ibgreen: Arrow syntax. Spell out parameter `lineString`.

demo/src/static/workers/flight-path-data-decoder.js
```diff
@@ -0,0 +1,55 @@
...
+      for (var j = 1; j < lineStr.length; j++) {
+        var prevPt0 = coords0[i][j - 1],
```
ibgreen: `const`? We clearly have a linter setup issue with the demo. The root lint script should include the demo folder.

shaojingli: > We clearly have a linter setup issue with the demo. The root lint script should include the demo folder.

My linter seems to work properly. Let me fix some of the coding style issues for @Pessimistress

Pessimistress: Workers are not compiled, so we can't use es6 syntax. In fact I was thinking of replacing all `const` and `let` for browser compatibility. That is also why we did not set up linter for them. We could try bring them to es6 and compile each one. It feels that webpack is not designed for this.

Workers has always been the ugly side of the demos - the data is generated outside of this project, and their formats are simply hard coded in these workers.

ibgreen: > Workers are not compiled, so we can't use es6 syntax. 

Ok thanks for clarifying, I didn't realize that. I will need to study the setup then to understand the constraints. 
1) We could certainly just run them all through a babel step - at the cost of losing hot reloading.
```
"build-workers": "babel src/workers --out-dir dist-workers ... --source-maps"
```
2) It is probably a bit early to require ES6 compatible browsers, but essentially all the browsers we want to support from a WebGL perspective anyway (the evergreen browsers) already run untranspiled ES6.

ericsoco:
Possible future suggestion: loading/parsing is not critical to demoing
deck.gl. Could we pre-process the data to save on parsing time, and maybe
even eliminate workers? Not sure about loading still blocking UI.

Pessimistress:
Workers are not necessary for all demos, but in some cases (e.g. the scatterplot demo, the trips demo) we have to aggressively compress the data to avoid downloading a 30M file.

@ibgreen I like your suggestion. I opened an issue [here](https://github.com/uber/deck.gl/issues/494). Will address in a future PR.

PR #490, opened 2017-03-29
Point cloud layer example fix for the layer browser

shaojingli:
Comment out modelMatrix.rotateZ and modelMatrix.rotateX for the point cloud layer in the layer browser as we removed those controls in a previous commit


shaojingli:
@ibgreen

PR #493, opened 2017-03-29
Add MapControls to demo. Partial fix for zoom centering issue. 

ibgreen:
Temporarily copy in `MapControls` component (builds on `EventManager`) from react-map-gl 3.0.0-alpha.

Adds a fix for zoom center (seems good in flat mode, possibly still has issues in perspective mode).


Pessimistress:
We definitely do need to fix interaction in react-map-gl@3, but that's separate from deck.gl.
My concern is that we may not be able to fix this fast enough. On a parallel track I'll dig into fixing webpack with react-map-gl@2.

ibgreen:
Deferring to 4.1 per discussions, will hold landing until we branch 4.0-release.

PR #495, opened 2017-03-29
Rollback to react-map-gl 2 and fix demo build

Pessimistress:


demo/webpack/prod.js
```diff
@@ -1,5 +1,12 @@
 const config = require('./config');
 
+// transpile mapbox-gl utils imported by react-map-gl@2
```
ibgreen: Seeing the solution makes it look so obvious...

Pessimistress: I still don't understand what broke it though. We always excluded node_modules in webpack config and it had never been a problem before.

PR #500, opened 2017-03-29
Default light settings in the defaultProps

shaojingli:
Need to add this default value otherwise the default light settings of composite layers won't work properly

@ibgreen


src/layers/core/geojson-layer/geojson-layer.js
```diff
@@ -51,7 +51,17 @@ const defaultProps = {
   // Line and polygon outline accessors
   getLineWidth: f => get(f, 'properties.lineWidth') || 1,
   // Polygon extrusion accessor
-  getElevation: f => get(f, 'properties.elevation') || 1000
+  getElevation: f => get(f, 'properties.elevation') || 1000,
+
+  // Optional settings for 'lighting' shader module
+  lightSettings: {
```
ibgreen: OK for now. Should really be in lib/constants.js or similar so that all layer take the same light.

PR #508, opened 2017-03-30
Remove destructuring from examples webpack configs so they run in Nod…

ericsoco:
…e < 6.

Per discussion [here](https://github.com/uber/deck.gl/issues/505#issuecomment-290460139).

### DIFF
--- a/examples/3d-heatmap/webpack.config.js
+++ b/examples/3d-heatmap/webpack.config.js
@@ -1,3 +1,7 @@
+// NOTE: To use this example standalone (e.g. outside of deck.gl repo)
+// delete the local development overrides at the bottom of this file
+
+// avoid destructuring for older Node version support
 const resolve = require('path').resolve;
 const webpack = require('webpack');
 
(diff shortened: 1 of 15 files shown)


examples/3d-heatmap/webpack.config.js
```diff
@@ -1,4 +1,4 @@
-const {resolve} = require('path');
```
ibgreen: Might be worth adding a comment stating that we do this to support Node 4 so that some helpful maintainer doesn't revert in a future commit. Probably worth doing in every file.

PS - These days my thinking is that almost every line in webpack configs should have a comment it is not an easy file to read.

ibgreen:
Should we also include an update of the main README that we support Node >=4 for running examples, and Node >= 6 for development of core library?

examples/3d-heatmap/webpack.config.js
shaojingli: > PS - These days my thinking is that almost every line in webpack configs should have a comment it is not an easy file to read.

This is my feeling too.

shaojingli:
> Should we also include an update of the main README that we support Node >=4 for running examples, and Node >= 6 for development of core library?

We should. @ericsoco can you add this to the main README with this change?

examples/3d-heatmap/webpack.config.js
ericsoco: Done.

PR #509, opened 2017-03-30
No token warning

ericsoco:
Per discussion [here](https://github.com/uber/deck.gl/issues/505#issuecomment-290460139).


ericsoco:
screen shot 2017-03-30 at 10 09 46 am

ibgreen:
Very nice. Although this "no token warning" only covers our main example. 

Adding this much code to every minimal example would be too much.

As an alternative we might actually want to consider this as a PR to react-map-gl itself. @abmai.

ericsoco:
Closing in favor of #513.

PR #513, opened 2017-03-30
Add on-screen warning when no Mapbox token exists.

ericsoco:
Also turn on ScatterplotLayer by default; without anything visible on the map, the example feels unfinished IMO. (Especially when there's no basemap visible.) But open to disagreement on that.

Supercedes #509, which had an errant commit from another branch.

### DIFF
--- a/examples/layer-browser/src/app.js
+++ b/examples/layer-browser/src/app.js
@@ -35,7 +35,9 @@ class App extends PureComponent {
         pitch: 30,
         bearing: 0
       },
-      activeExamples: {},
+      activeExamples: {
+        ScatterplotLayer: true
+      },
       settings: {
         // immutable: false,
         // Effects are experimental for now. Will be enabled in the future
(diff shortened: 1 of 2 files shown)


ericsoco:
screen shot 2017-03-30 at 10 09 46 am

examples/layer-browser/src/app.js
```diff
@@ -35,7 +35,9 @@ class App extends PureComponent {
         pitch: 30,
         bearing: 0
       },
-      activeExamples: {},
+      activeExamples: {
+        ScatterplotLayer: true
```
ibgreen: Maybe add a comment here stating the logic (same text you put in this PR)

examples/layer-browser/src/app.js
```diff
@@ -199,12 +201,24 @@ class App extends PureComponent {
...
+    /* eslint-disable max-len */
+  }
+
   render() {
     const {settings, activeExamples, hoveredItem, clickedItem} = this.state;
 
     return (
       <div>
         { this._renderMap() }
+        { !MAPBOX_ACCESS_TOKEN && this._renderNoTokenWarning() }
```
ibgreen: Could move this check inside the function.

examples/layer-browser/src/app.js
```diff
@@ -176,7 +178,7 @@ class App extends PureComponent {
     const {width, height, mapViewState, settings: {effects}} = this.state;
     return (
       <MapboxGLMap
-        mapboxApiAccessToken={MAPBOX_ACCESS_TOKEN}
+        mapboxApiAccessToken={MAPBOX_ACCESS_TOKEN || 'no_token'}
```
ibgreen: What does sending in 'no_token' achieve?

PR #516, opened 2017-03-30
Point cloud example

gnavvy:
es6-filed the loader, the workers are kept es5 and lint ignored for now.


gnavvy:
oops, this is really hard to review. will open another PR.

PR #517, opened 2017-03-31
Point cloud examples

gnavvy:
point cloud examples for both ply and laz format.

orbit-control.js is the same as in the 3D surface Plot Example.
files inside the worker folder can be ignored. (will es6-ify some of them if time permitting)


shaojingli:
Is it intended that point-cloud-ply doesn't have an app.js?

examples/point-cloud-laz/app.js
```diff
@@ -0,0 +1,177 @@
...
+      viewport: {
+        ...viewport,
+        rotationY: viewport.rotationY + 1
+      }
+    });
+    window.requestAnimationFrame(this._onUpdate);
+  }
+
+  _renderLazPointCloudLayer() {
+    return this.state.points.length && new PointCloudLayer({
```
ibgreen: Can this check be avoided?

examples/point-cloud-laz/app.js
```diff
@@ -0,0 +1,177 @@
...
+      radiusPixels: 1
+    });
+  }
+
+  _renderDeckGLCanvas() {
+    const {width, height, viewport} = this.state;
+    const canvasProps = {width, height, ...viewport};
+    const glViewport = OrbitController.getViewport(canvasProps);
+
+    return width && height && <OrbitController {...canvasProps} ref={canvas => {
```
ibgreen: Nit: the line break formatting here isn't very pretty. I'd use a parenthesis around the component, put it on its own line, and then one prop per line.

examples/point-cloud-laz/app.js
```diff
@@ -0,0 +1,177 @@
...
+      this._canvas = canvas;
+    }} onChangeViewport={this._onChangeViewport}>
+      <DeckGL width={width} height={height} viewport={glViewport}
```
ibgreen: One prop per line could be easier to read?

examples/point-cloud-laz/worker/laslaz.js
```diff
@@ -0,0 +1,368 @@
+/* eslint no-use-before-define: 0 */
```
ibgreen: No way to install this as a module? Does it all have to be copied?

examples/point-cloud-ply/utils/orbit-control.js
```diff
@@ -0,0 +1,177 @@
+/* global window */
```
ibgreen: Didn't I see another orbit controller above?

gnavvy:
good catch! not intended, it slipped.

examples/point-cloud-laz/worker/laslaz.js
gnavvy: plas.io, where the decoder originally comes from, did not publish it as a module.
we can polish the code, address the source and put it on npm for sure.
but it is going to take awhile. shall we hold this PR for that?

examples/point-cloud-ply/utils/orbit-control.js
gnavvy: yes, two separate examples. we talked about to consolidate all the viewport + interaction logics, after that I can just import the orbit controller.

gnavvy:
> A lot of duplicated code (that is already duplicated from other places, like OrbitControls). 
> I wonder if we should organize this as one folder that exports two demos?
I agree we should reuse the code as much as possible. But in this case, I deliberately separate them out into two examples. PLY point-cloud (mesh generated) and LAS/LAZ point-cloud (3D scan) have different use cases plus loading and parsing logics.

I see the following two parts can be shared:
1. the position normalization logics, mostly the same.
2. the orbit-control, again, will import instead of duplicate the code once it is part of core deck.gl.

examples/point-cloud-laz/app.js
gnavvy: you mean to create a PointCloudLayer even if the data is empty?

gnavvy:
just noticed resolving conflicts using the GitHub UI will result in a merge instead of rebase.

examples/point-cloud-laz/worker/laslaz.js
ibgreen: Probably not a blocker but let's discuss.

PR #527, opened 2017-03-31
Documentation improvements & updates

Apercu:


shaojingli:
Thanks @Apercu !!! I know that master kept changing today so you are rebasing your changes repeatedly * _ * I'll review your change and help on the rebase

docs/advanced/using-standalone.md
```diff
@@ -10,7 +10,7 @@ applications.
 
 ## LayerManager class
 
-The deck.gl `LayerManager` class handles updates, drawing and picking
+The deck.gl [LayerManager](/src/lib/layer-manager.js) class handles updates, drawing and picking
```
Pessimistress: link to /doc/api-reference/layer-manager.md ?

docs/advanced/subclassed-layers.md
```diff
@@ -7,14 +7,13 @@ the vertex and/or fragment shaders.
 If a small feature is missing from a layer, subclassing can often be a
 good technique to add it.
 
-## Overriding Attribute Calculation
+### Overriding Attribute Calculation
```
ibgreen: About the hashes... normally we would "indent one level at a time. Assuming you decided that this looks better? 

@Pessimistress is there a way in our markdown component to control the size of various headings rather than using non-standard "indentation"?

I have concerns with this - we have more and more maintainers of this library and it is not easy for other engineers to figure out that we are using non-standard indentation.

docs/advanced/tips-and-tricks.md
```diff
@@ -11,9 +11,11 @@ does not give ideal results and the way the deck.gl development team to improve
 the blending effect is by specifying the CSS property `mix-blend-mode`
 in modern browsers to be `multiply`:
 
-    .overlays canvas {
-      mix-blend-mode: multiply;
-    }
+```css
```
ibgreen: Neat

docs/advanced/using-standalone.md
```diff
@@ -32,8 +32,6 @@ handling.
 
 ## Using deck.gl without React
 
-In lieu of better documentation, the React integration in
-`src/react/deckgl.js` uses the `LayerManager` class to render
-the layers, so looking at that code can be a good reference
-for how to use the `LayerManager` class.
-
+In lieu of better documentation, the [React integration](/src/react/deckgl.js)
```
Pessimistress: link to /doc/api-reference/deckgl.md ?

docs/advanced/viewports.md
```diff
@@ -2,9 +2,9 @@
 
 ## Overview
 
-deck.gl calculates projections using a a generic `Viewport` class
-(which is essentially a 3D matrix "camera" class of the type you would find
-in any 3D/WebGL/OpenGL library, holding `view` and `projection` matrices).
+deck.gl calculates projections using a a generic [Viewport class](src/lib/viewports/viewport.js)
```
Pessimistress: link to /docs/api-reference/viewport.js?

docs/advanced/using-standalone.md
ibgreen: We now have a place holder for the `LayerManager` class in the API reference. Better to link to that, and then have that doc link to the src.

docs/advanced/viewports.md
ibgreen: We have docs for viewports in the API reference, we should link to that.

docs/advanced/using-standalone.md
ibgreen: Link to `DeckGL` doc and let that link to src.

docs/api-reference/base-layer.md
```diff
@@ -82,10 +86,10 @@ The opacity of the layer. deck.gl automatically applies gamma to the opacity
 in an attempt to make opacity changes appear linear (i.e. the opacity is
 visually proportional to the value of the prop)
 
-Remarks
-* While it is a recommended convention that all deck.gl layers should
-  support the `opacity` prop, it is up to each layer's fragment shader
-  to implement support for opacity.
+**Remark:**
```
ibgreen: Nit: Fine, we have a pretty consistent style of having bulleted remarks, but this works too.

docs/get-started/using-with-mapbox-gl.md
```diff
@@ -3,7 +3,7 @@
 While a deck.gl application could choose to render on top of any JavaScript
 Map component (or even render without an underlying map),
 deck.gl has been developed in parallel with
-the [react-map-gl](https://uber.github.io/react-map-gl/#/) component,
+the [react-map-gl](https://github.com/uber/react-map-gl) component,
```
ibgreen: Don't we want to link to the docs rather than the github readme? Maybe provide both?

docs/layers/arc-layer.md
```diff
@@ -8,14 +8,16 @@
 
 The Arc Layer renders raised arcs joining pairs of source and target points,
 specified as latitude/longitude coordinates.
+[Source](https://github.com/uber/deck.gl/tree/master/src/layers/core/arc-layer)
```
ibgreen: We previously decided to put source links at the bottom, partially to be consistent with THREE.js, gl-matrix etc, and because it is an additional level of detail, together with remarks etc.

Is this important? Do we need to discuss?

docs/layers/arc-layer.md
```diff
@@ -8,14 +8,16 @@
...
 
-```
+```js
 import DeckGL, {ArcLayer} from 'deck.gl';
 
-render() {
+const App = () => {
   const {data, viewport} = this.props;
```
ibgreen: Should these be arguments to `App` above?

docs/layers/grid-layer.md
```diff
@@ -5,19 +5,21 @@
   <img src="https://img.shields.io/badge/extruded-yes-blue.svg?style=flat-square" alt="extruded" />
 </p>
 
-# GridLayer
+# Grid Layer
```
ibgreen: I think we need to go with one style or another. Split all layer names or none. 

If this is reference documentation, it makes sense to keep the names without spaces to be identical to the actual layer names.

Under any circumstances, the text in the index on the left, and the title of each doc should be the same.

docs/layers/grid-layer.md
```diff
@@ -29,13 +31,13 @@ render() {
...
   });
 
-  return <DeckGL {...viewport} layers={[layer]} />
-}
+  return (<DeckGL {...viewport} layers={[layer]} />);
+};
 ```
 
-## Properties
+**Remark:** GridLayer at the moment only works with `COORDINATE_SYSTEM.LNGLAT`.
```
ibgreen: We usually have remarks at the bottom followed by source code.

docs/layers/hexagon-cell-layer.md
```diff
@@ -120,9 +120,4 @@ Method called to retrieve the color of each object. Color should be set to
 
 - Default: `object => object.elevation`
 
-Method called to retrieve the elevation of each object.
-1 unit approximate to 100 meters.
-
-
-## Source Code
-[HexagonCellLayer](https://github.com/uber/deck.gl/tree/master/src/layers/core/hexagon-cell-layer)
+Method called to retrieve the elevation of each object. 1 unit approximate to 100 meters.
```
ibgreen: @heshan0131 can you check if the prop defaults (100 meter, lng.lat etc are still correct in this doc?)

docs/layers/polygon-layer.md
```diff
@@ -95,7 +93,8 @@ Whether to generate a line wireframe of the hexagon. The outline will have
 "horizontal" lines closing the top and bottom polygons and a vertical line
 (a "strut") for each vertex on the polygon.
 
-Remarks:
+**Remarks:**
```
ibgreen: Remarks should go at the bottom

docs/api-reference/base-layer.md
Apercu: the bold addition was to catch user eye, I don't have a strong opinion on it

docs/get-started/using-with-mapbox-gl.md
Apercu: Given the website is quite outdated, I preferred linking to the repo, where the readme docs are pretty complete

docs/layers/arc-layer.md
Apercu: Not important, but having the source easily available right after the description of the layer will encourage people to look into it more than if it was at the very bottom of the docs after the props.

It takes very little place and is a good entry point if you want to start building your own layer or extending existing ones.

I guess if three.js did it, it could be fine, but I don't think it would be in our case.

docs/layers/arc-layer.md
Apercu: oh yes, definitely.

docs/layers/grid-layer.md
Apercu: I took the arc layer as reference, but it might have been this one that was out of sync

docs/layers/arc-layer.md
ibgreen: One of the problems we have had was that there was a lot of vertical empty space in the beginning of the layer docs. Every line with a tag, link etc pushes down the text about the props that 95% of the people are looking for. 

I'm open to discussion, but if we want it at the top, the layout (in the demo docs, not on github) is critical. The examples have a nice little link icon to the source if we could do that on the same line as the tags then it would be OK.

docs/layers/arc-layer.md
ibgreen: I think this was repeated in several places, perhaps do a search.

docs/layers/grid-layer.md
ibgreen: Ok let's go with the actual layer names then (no spaces). It will match better with the API reference docs, the examples etc.

docs/advanced/subclassed-layers.md
Pessimistress: I suggest we go with css

ibgreen: Apparently @apercu's concern is that CSS only works on demo site. His task is to make this look good when read directly on github.

Still what stops github from changing their formatting at any time. We'd need to change our documentation structure if they do. Better to go with 1,2,3 than 1,3,5 and accept that github docs won't look as good as they could.

docs/get-started/using-with-mapbox-gl.md
ibgreen: > Given the react-map-gl website is quite outdated, I preferred linking to the repo, where the readme docs are pretty complete

OK makes sense we need to update the react-map-gl docs @abmai

PR #532, opened 2017-04-02
Fix getting started sample code

Pessimistress:
- Remove inline comments inside jsx
- Remove unnecessary code

### DIFF
--- a/docs/get-started/using-with-react.md
+++ b/docs/get-started/using-with-react.md
@@ -7,63 +7,46 @@ render it as a child of another component, passing in your list of deck.gl
 layers as a property.
 
 ```jsx
-// Import React
+/// app.js
 import React, {Component} from 'react';
-import {render} from 'react-dom';
-
-// Import Mapbox via react-map-gl
 import MapGL from 'react-map-gl';
-
-// Import deck.gl and the layer you'd like to use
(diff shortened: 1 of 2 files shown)


docs/get-started/using-with-react.md
```diff
@@ -7,63 +7,46 @@ render it as a child of another component, passing in your list of deck.gl
...
+const data = [
+  {sourcePosition: [-122.41669, 37.7853], targetPosition: [-122.41669, 37.781]}
+];
+
+export default class App extends Component {
 
   render() {
-    const {viewport, width, height} = this.state;
+
+    const layers = [
```
ibgreen: I think we are doing users a disservice by constructing the layers outside of the JSX. It makes it less clear that layers are supposed to be created every render cycle, just like React elements.
Therefore I think the example should look like this:
```
<MapGL {...viewport} mapboxApiAccessToken={MAPBOX_ACCESS_TOKEN}>
  <DeckGL {...viewport} layers={[
    new LineLayer({id: 'line-layer', data})
  ]} />
</MapGL>   new LineLayer({id: 'line-layer', data})
```
In the next release we can support JSX syntax for layers, then this will be a non-issue.

PR #535, opened 2017-04-02
Add Arc brushing demo

heshan0131:
- show arcs when target or source of arc is in brush range
- show point when point is in brush range, or other point connected by arcs are in range
brushing


demo/package.json
```diff
@@ -24,7 +24,7 @@
     "expr-eval": "^1.0.0",
     "highlight.js": "^9.7.0",
     "immutable": "^3.7.5",
-    "luma.gl": "^3.0.0",
+    "luma.gl": "^3.0.0-beta.11",
```
shaojingli: Any reason we revert back to luma beta?

demo/src/components/demos/arc-brushing-demo.js
```diff
@@ -0,0 +1,120 @@
...
+        <p className="layout">
+          <span className="col-1-2">Net gain</span>
+          <span className="col-1-2 text-right">Net loss</span>
+        </p>
+
+        <p>Data source: <a href="http://www.census.gov">US Census Bureau</a></p>
+
+        <div className="layout">
+          <div className="stat col-1-2">
+            Counties<b>{ meta.count || 0 }</b>
```
shaojingli: Can we change this to "No. of Counties" and "No. of Arcs" to align with other examples?

demo/src/static/markdown/examples.md
```diff
@@ -1,5 +1,12 @@
 <div>
   <div class="thumb">
+    <div data-title="US County-to-County Migration" data-name="ArcBrushingLayer">
+      <a href="#/examples/custom-layers/arc-brushing-layer">
+        <img src="images/demo-thumb-arc-brushing.jpg" />
```
shaojingli: This file is missing.

examples/arc-brushing/scatterplot-brushing-layer/scatterplot-brushing-layer-vertex.glsl.js
```diff
@@ -0,0 +1,110 @@
...
+void main(void) {
+
+  // if enableBrushing is truthy calculate whether instancePosition is in range
+  float isPtInBrush = isPointInRange(instancePositions.xy, mousePos, brushRadius, enableBrushing);
+
+  // for use with arc layer, if brushTarget is truthy 
+  // calculate whether instanceTargetPositions is in range
+  float isTargetInBrush = isPointInRange(instanceTargetPositions.xy, mousePos, brushRadius, 1.);
+
+  float isInBrush = float(float(isPtInBrush > 0. && brushTarget <= 0.) > 0. ||
```
shaojingli: ... this boolean statement looks rather complicated to me... break them down or put some comments?

shaojingli:
Excellent work! This makes the ArcLayer alive!

Just a few comments for discussion

1) Since this demo contains two custom layers, ArcBrushingLayer and ScatterplotBrushingLayer, we probably should call it "Migration" or similar. And also, let's remove the original arc layer demo.

2) Will it be too much work to add a checkbox to switch between "brushing" mode and "unbrushing" mode? Just like the "clustering" checkout in the icon layer demo. If no, we might consider doing it. If yes, then I'm totally okay with current version.

Currently the example shows all arcs and dots initially when the mouse is not on the map, but when the user starts to move the mouse into the map, all arcs and dots disappeared. This is a bit unexpected for me. Feel that adding a checkbox to enable/disable brushing might be better. 

3) The hover tooltip seems to contain some missing info.

screen shot 2017-04-02 at 2 27 16 pm

What's the unit of the "brushing radius" slide? It doesn't seem to be meters...

BTW, add copy right notice to your work @heshan0131 !




BTW, remember to put the copy right notice to every file.

examples/arc-brushing/scatterplot-brushing-layer/scatterplot-brushing-layer.js
```diff
@@ -0,0 +1,87 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
```
shaojingli: 2015-2017

examples/arc-brushing/arc-brushing-layer/arc-brushing-layer-vertex.glsl.js
```diff
@@ -0,0 +1,145 @@
...
+    mix(source, target, segmentRatio),
+    sqrt(max(0.0, vertex_height))
+  );
+}
+
+void main(void) {
+  vec2 source = project_position(instancePositions.xy);
+  vec2 target = project_position(instancePositions.zw);
+
+  // if not enabled return true
```
shaojingli: Didn't get this comment...

PR #519, opened 2017-03-31
Fix viewport

ibgreen:
Now ready for review!

PR #537, opened 2017-04-03
Fix meter offset mode

### LINKED ISSUE #291: WebMercatorViewport.unproject() produces incorrect results
Test case:
https://github.com/uber/deck.gl/blob/viewport-fix/test/lib/viewports/web-mercator-viewport.spec.js
(uncomment line 44 - 49)

```
not ok 41 unproject (center)
  ---
operator: deepEqual
expected: |-
  [ -122.43, 37.75 ]
actual: |-
  [ -122.43, 37.675038961 ]
```

I expect the result of `unproject` reverses that of `project`.

ibgreen:
Meter offset mode was broken in #519.

This fix should let us finally close #291 

Point Cloud Layer (in Layer Browser) before and after the fix

screen shot 2017-04-03 at 9 07 28 am

screen shot 2017-04-03 at 9 06 47 am

### DIFF
--- a/src/lib/viewport-uniforms.js
+++ b/src/lib/viewport-uniforms.js
@@ -64,7 +64,7 @@ function calculateMatrixAndOffset({
     projectionCenter = viewProjectionMatrix
       .transformVector([positionPixels[0], positionPixels[1], 0.0, 1.0]);
 
-    // Always apply uncentered projection matrix (shader adds center)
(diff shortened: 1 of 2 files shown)


shaojingli:
The Point Cloud Layer sample before the fix actually looks pretty cool 😎 .

Just kidding. Thanks for the fix, Ib !

PR #542, opened 2017-04-03
Fine tweak of docs

shaojingli:
Following https://github.com/uber/deck.gl/issues/496

### DIFF
--- a/docs/api-reference/attribute-manager.md
+++ b/docs/api-reference/attribute-manager.md
@@ -1,6 +1,6 @@
 # AttributeManager Class
 
-The `AttributeManager` class automated attribute allocation and updates.
+The [`AttributeManager`](https://github.com/uber/deck.gl/blob/master/src/lib/attribute-manager.js) class provides automated attribute allocations and updates.
 
 Summary:
 - keeps track of valid state for each attribute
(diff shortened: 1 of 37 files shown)


docs/api-reference/attribute-manager.md
```diff
@@ -13,29 +13,12 @@ Limitations:
   There are currently no provisions for only invalidating a range of
   indices in an attribute.
 
-## Static Methods
```
ibgreen: `setDefaultLogFunctions ` is actually public - Just added it a couple of days ago. My mistake...

docs/api-reference/attribute-manager.md
```diff
@@ -1,6 +1,6 @@
 # AttributeManager Class
 
-The `AttributeManager` class automated attribute allocation and updates.
+The [`AttributeManager`](https://github.com/uber/deck.gl/blob/master/src/lib/attribute-manager.js) class provides automated attribute allocations and updates.
```
ibgreen: Just realized: For source links, should we link to `master` or to `4.0-release`?

docs/api-reference/base-layer.md
```diff
@@ -26,10 +26,11 @@ of deck.gl, a check will be in place to enforce this
 
 ## Constructor
 
-- `props` (Object) - Layer properties.
-
 ### Basic Properties
 
+##### `props` (Object)
```
ibgreen: This looks strange, I think this was intended to be the parameter list of the constructor

I think it is better to move this back and add a heading `Parameters`:

## Constructor

Initializes a new `Layer` instance.

Parameters:
- `props` (Object) - Layer properties.

docs/api-reference/attribute-manager.md
ibgreen: Nit: When done this way, it is not clear what is a doc link and what is a source link.

I thought having a link that actually says `Source` (perhaps at the bottom of the page) would be cleaner. See for instance how THREE.js does it https://threejs.org/docs/api/math/Box2.html

docs/layers/arc-layer.md
```diff
@@ -6,10 +6,6 @@
 
 # ArcLayer
 
-The Arc Layer renders raised arcs joining pairs of source and target points,
```
ibgreen: Just to clarify, when I commented about "source link" should be separated from the paragraph, I meant the link to the original source code in the repo.
I.e this piece:
Source](https://github.com/uber/deck.gl/tree/master/src/layers/core/arc-layer)

docs/api-reference/attribute-manager.md
shaojingli: You are right. We should do 4.0-release. I'll branch 4.0-release after all outstanding PRs got landed today

shaojingli:
All links to source code now points to the 4.0-release branch instead of the master branch.

@Pessimistress When I start the demo site locally, all the links to the source code doesn't open in the same tab. The links are good as they can be opened in another tab without problems. Do you have any ideas why this happens?

screen shot 2017-04-03 at 2 35 41 pm

PR #549, opened 2017-04-05
Change the relative path to absolute path

shaojingli:
in the webpack build script in the demo folder. Webpack 2.3.0 and above no longer handles relative path for build

@Pessimistress @ibgreen

### DIFF
--- a/demo/webpack/prod.js
+++ b/demo/webpack/prod.js
@@ -1,5 +1,5 @@
 const config = require('./config');
-
+const path = require('path');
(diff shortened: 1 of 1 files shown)


demo/webpack/prod.js
```diff
@@ -10,7 +10,7 @@ config.module.rules.push({
 module.exports = Object.assign(config, {
 
   output: {
-    path: './dist',
+    path: __dirname + '/dist',
```
Pessimistress: `resolve(__dirname, 'dist')` ?

PR #552, opened 2017-04-05
Fix finalization bug

Pessimistress:
Check `layer.lifecycle` instead of `layer.state`

### DIFF
--- a/src/lib/constants.js
+++ b/src/lib/constants.js
@@ -40,3 +40,11 @@ export const COORDINATE_SYSTEM = {
   // Positions and distances are not transformed: [x, y, z] in unit coordinates
   IDENTITY: 0.0
 };
+
+export const LIFECYCLE = {
+  NO_STATE: 'Awaiting state',
+  MATCHED: 'Matched. State transferred from previous layer',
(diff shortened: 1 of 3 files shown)


src/lib/constants.js
```diff
@@ -40,3 +40,11 @@ export const COORDINATE_SYSTEM = {
   // Positions and distances are not transformed: [x, y, z] in unit coordinates
   IDENTITY: 0.0
 };
+
+export const LIFECYCLE = {
+  NO_STATE: 'Awaiting state',
+  MATCHED: 'Matched. State transferred from previous layer',
+  INITIALIZED: 'Intialized',
+  OUTDATED: 'Outdated. Awaiting garbage collection',
```
ibgreen: Should this be `Outdated. Awaiting finalization`?

Pessimistress: Outdated layers are not finalized, they are replaced by new ones.

ibgreen: Yes of course. Maybe change to
`Outdated by newer layer with same id. Awaiting garbage collection.`

PR #555, opened 2017-04-06
brush layer - start brushing when mouse is inside map

heshan0131:


examples/brushing/deckgl-overlay.js
```diff
@@ -136,7 +136,9 @@ export default class DeckGLOverlay extends Component {
       opacity, mouseEntered, mousePosition} = this.props;
     const {arcs, targets, sources} = this.state;
 
-    const startBrushing = Boolean(mouseEntered && enableBrushing);
+    // mouseEntered is undefined when mouse is in the component while it first loads
```
ibgreen: Not sure about using a "three-state" boolean, seems like an easy thing for a maintainer to miss... 

Consider tracking using another Boolean  variable instead?

sorianog:
Have these changes been merged in?

sorianog:
@heshan0131 Thank you! 😄

PR #556, opened 2017-04-06
Replace spread operator on array for performance reasons

shaojingli:
@Pessimistress @gnavvy


src/layers/core/solid-polygon-layer/polygon-tesselator-extruded.js
```diff
@@ -277,7 +277,7 @@ function calculateSurfaceIndices(vertices, offset) {
 
     // polygon sides
     for (let i = 0; i < numVertices - 1; i++) {
-      indices.push(...drawRectangle(i));
+      indices.push(drawRectangle(i));
```
gnavvy: drawRectangle returns an array, use concat here?

PR #558, opened 2017-04-06
Add unfurls

ericsoco:
I believe this is all we need for nice unfurls. Once it's live we can test at e.g. https://cards-dev.twitter.com/validator


shaojingli:
I'm not exactly familiar with this... can someone else review this?

PR #569, opened 2017-04-12
Add Hexagon layer methods

heshan0131:
- Add `getHexagons` and `getSortedCounts` methods

### DIFF
--- a/src/layers/core/hexagon-layer/hexagon-layer.js
+++ b/src/layers/core/hexagon-layer/hexagon-layer.js
@@ -108,13 +108,9 @@ export default class HexagonLayer extends Layer {
 
   updateState({oldProps, props, changeFlags}) {
     if (changeFlags.dataChanged || _needsReProjectPoints(oldProps, props)) {
-      const {hexagonAggregator} = this.props;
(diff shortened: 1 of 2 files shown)


heshan0131:
@shaojingli Why are you approving pull requests, aren't you suppose to be on vacation?

Pessimistress:
Update layer docs?

heshan0131:
This change doesn't affect the API. `getHexagons` and `getSortedCounts` are methods of the layer class. separate them out will make it easier to create layer subclasses and override them.

src/layers/core/hexagon-layer/hexagon-layer.js
```diff
@@ -125,6 +121,18 @@ export default class HexagonLayer extends Layer {
     }
   }
 
+  getHexagons() {
+    const {hexagonAggregator} = this.props;
+    const {viewport} = this.context;
+    const hexagons = hexagonAggregator(this.props, viewport);
+    Object.assign(this.state, {hexagons});
```
ibgreen: Nit: we have a `Layer.setState` method that also sets the dirty flag to ensure re-render.

PR #579, opened 2017-04-17
separate out more methods in hexagon-layer, use valueAccessor in BinSorter to enable color / elevation by value

heshan0131:
- add internal `getUpdateTrigger` method
- add `colorValueAccessor` as hexagon layer prop to get the value for bin colors. When undefined, layer will use points.length to calculate color
- change default `hexagonAggregator` output to be `{hexagons: [], hexagonVertices: []}`
- add valueAccessor to bin-sorter to support color / elevation by value

### DIFF
--- a/docs/layers/hexagon-layer.md
+++ b/docs/layers/hexagon-layer.md
@@ -77,14 +77,14 @@ to number of counts by passing in an arbitrary color domain. This property is ex
 Hexagon color ranges as an array of colors formatted as `[[255, 255, 255, 255]]`. Default is
 [colorbrewer](http://colorbrewer2.org/#type=sequential&scheme=YlOrRd&n=6) `6-class YlOrRd`.
 
-##### `colorValueAccessor` (Function, optional)
+##### `getColorValue` (Function, optional)
 
-- Default: `undefined`
+- Default: `points => points.length`
 
-`colorValueAccessor` is the accessor function to get the value that bin color is based on. 
+`getColorValue` is the accessor function to get the value that bin color is based on. 
 It takes an array of points inside each bin as arguments, returns a value. For example, 
-You can use `colorValueAccessor`to color the bins by avg/mean/max of a specific attributes of each point.
-When not specified, color of bins will based on count of points.
+You can pass in `getColorValue`to color the bins by avg/mean/max of a specific attributes of each point.
+By default `getColorValue` is undefined, color of bins will based on count of points.
 
 ##### `coverage` (Number, optional)
 
--- a/docs/layers/hexagon-layer.md
+++ b/docs/layers/hexagon-layer.md
@@ -83,8 +83,8 @@ Hexagon color ranges as an array of colors formatted as `[[255, 255, 255, 255]]`
 
 `getColorValue` is the accessor function to get the value that bin color is based on. 
 It takes an array of points inside each bin as arguments, returns a value. For example, 
-You can pass in `getColorValue`to color the bins by avg/mean/max of a specific attributes of each point.
-By default `getColorValue` is undefined, color of bins will based on count of points.
+You can pass in `getColorValue` to color the bins by avg/mean/max of a specific attributes of each point.
+By default `getColorValue` returns the length of the points array.
 
 ##### `coverage` (Number, optional)
 
--- a/docs/layers/hexagon-layer.md
+++ b/docs/layers/hexagon-layer.md
@@ -77,6 +77,15 @@ to number of counts by passing in an arbitrary color domain. This property is ex
 Hexagon color ranges as an array of colors formatted as `[[255, 255, 255, 255]]`. Default is
 [colorbrewer](http://colorbrewer2.org/#type=sequential&scheme=YlOrRd&n=6) `6-class YlOrRd`.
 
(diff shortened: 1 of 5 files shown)


docs/layers/hexagon-layer.md
```diff
@@ -75,6 +77,15 @@ to number of counts by passing in an arbitrary color domain. This property is ex
 Hexagon color ranges as an array of colors formatted as `[[255, 255, 255, 255]]`. Default is
 [colorbrewer](http://colorbrewer2.org/#type=sequential&scheme=YlOrRd&n=6) `6-class YlOrRd`.
 
+##### `colorValueAccessor` (Function, optional)
```
ibgreen: 1) This function is optional and default is `undefined`? Can this be clarified a bit more?

2) About the naming: we never call our accessors `...Accessor`. If this is a normal accessor, can we just call this `getColorValue`?

On the other hand, if it is not a normal accessor and we want to make that clear, then let's at least not call it `...Accessor`.

Assuming we are trying to match the 'hexagonAggregator` naming pattern, maybe this could be `colorValueExtractor` or `colorValueGetter` or similar.

src/layers/core/hexagon-layer/hexagon-aggregator.js
```diff
@@ -45,10 +45,10 @@ export function pointToHexbin({data, radius, getPosition}, viewport) {
 
   const hexagonBins = newHexbin(screenPoints);
 
-  return hexagonBins.map(hex => ({
+  return {hexagons: hexagonBins.map(hex => ({
```
ibgreen: Nit: add a newline to make the object structure more visible:

    return {
        hexagons: ....
    };

src/layers/core/hexagon-layer/hexagon-layer.js
```diff
@@ -54,19 +55,23 @@ const defaultProps = {
...
+function _needsReSortBins(oldProps, props) {
+  return oldProps.colorValueAccessor !== props.colorValueAccessor;
+}
+
 export default class HexagonLayer extends Layer {
   constructor(props) {
-    if (!props.radius) {
-      log.once(0, 'HexagonLayer: radius in meter is needed to aggregate points into ' +
-        'hexagonal bins, Now using 1000 meter as default');
+    if (!props.radius || !typeof props.hexagonAggregator === 'function') {
```
ibgreen: Is this check for hexagonAggregator needed when we have a default? Is it intended to cover the user putting in some unexpected value in this prop?

I know I have said this before, but we should implement a `propTypes` system to avoid adding ad-hoc checks like this.

src/layers/core/hexagon-layer/hexagon-layer.js
```diff
@@ -144,11 +155,26 @@ export default class HexagonLayer extends Layer {
     });
   }
 
+  getUpdateTrigger() {
```
ibgreen: `getUpdateTriggers`?

src/utils/bin-sorter.js
```diff
@@ -17,43 +17,60 @@
...
    * Get ths max count of all bins
    * @return {Number | Boolean} max count
    */
   getMaxCount() {
-    return this.sortedBins.length && this.sortedBins[this.sortedBins.length - 1].counts;
+    return Math.max.apply(null, this.sortedBins.map(b => b.counts));
+  }
+
+  getCountRange(args) {
+    log.once(0, 'HexagonLayer: BinSorter.getCountRange is deprecated, use getValueRange instead');
```
ibgreen: Do we need this? Was this part of a public API?

docs/layers/hexagon-layer.md
heshan0131: when getColorValue is not defined by layer, the BinSorter used by layer will have getValue = points =>  points.length by default.  I don't know if it's necessary to define it twice

src/utils/bin-sorter.js
heshan0131: I guess we are not exporting BinSorter, so we don't need this.

src/layers/core/hexagon-layer/hexagon-layer.js
heshan0131: if user pass in a hexagonAggregator then radius is not required. If not, our default hexagonAggregator requires radius to be set.

docs/layers/hexagon-layer.md
ibgreen: > I don't know if it's necessary to define it twice

While it is preferable to avoid defining things twice, it can be helpful to take the end user's perspective. Per my understanding, the `BinSorter` is not a public class. So how should the user discover the default behavior here?

src/layers/core/hexagon-layer/hexagon-layer.js
ibgreen: > if user pass in a hexagonAggregator then radius is not required. If not, our default hexagonAggregator requires radius to be set.

OK that makes a lot of sense! Perhaps the condition and the error message could be rewritten to make this a little clearer? Use whatever makes most sense to you.
```
if (!hexagonAggregator && !radius) {
   log.once('Default hexagonAggregator requires radius prop to be set);
}
```

PR #581, opened 2017-04-18
fix(solid-polygon): export layer and fix docs

Apercu:


ibgreen:
Note that we have a `PolygonLayer` that renders using the internal `SolidPolygonLayer` so this layer can be used without having it exported directly.

Apercu:
Sure, I was wondering since it had his own readme. Thanks

PR #583, opened 2017-04-19
Migrate from React.PropTypes to `prop-types` module

samhogg:
`deck.gl` is the last remaining dependency of `react-vis` that isn't using the `prop-types` module. This PR fixes that, and gets rid of console warnings when using React >= 15.5

Related 
* uber/react-vis#368
* uber/react-vis#391


samhogg:
@ibgreen I think it'll be consistent to include the explicit dependency, I'll add that to the `package.json`s in the subfolders

PR #584, opened 2017-04-19
Add support for drag events.

ericsoco:
Note that many mouse events are already supported by luma.gl; this PR extends the support for a subset of these upwards into deck.gl.

I know that we have grander plans for encapsulating interaction models for different scenarios. I don't believe this runs afoul of those, but it is possible the direction taken in this PR is setting us up for a bigger refactor when we tackle new interaction models.

/cc @gnavvy @Pessimistress as potential consumers of these additions.

### DIFF
--- a/src/lib/draw-and-pick.js
+++ b/src/lib/draw-and-pick.js
@@ -23,6 +23,11 @@ import {GL, glContextWithState} from 'luma.gl';
 import {getUniformsFromViewport} from './viewport-uniforms';
 import {log, getBlendMode, setBlendMode} from './utils';
 
+// Note: corresponding touch events, once supported, should be included here as well.
+const MOTION_EVENTS = [
+  'dragmove',
+  'dragend'
+];
 const EMPTY_PIXEL = new Uint8Array(4);
 let renderCount = 0;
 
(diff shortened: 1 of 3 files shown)


ericsoco:
Currently absent from this PR: tests and docs. Will tackle after we discuss appropriateness of the feature + implementation.

ericsoco:
@heshan0131 is [my last commit](https://github.com/uber/deck.gl/pull/584/commits/06dc531cd0a181dd6c10a436f51731cb57234de0?w=1) going to cause problems with e.g. brushing?

The synopsis: "motion" events are those that are initiated on a given object and are intended to manipulate just that object. Therefore, after the initial event (e.g. `dragstart`), subsequent events (`dragmove` and `dragend`) don't run the picking process at all, they just return the originally-picked object.

I think this is ok for brushing, as the logic for what's within the brushed area is something run in the context of the parent/container (which would be the originally-picked object, and returned by all `drag` events). But might be missing something.

heshan0131:
No it will not, currently the brushing example listen to mousemove **outside** of MapGL component

src/lib/draw-and-pick.js
```diff
@@ -69,17 +75,39 @@ export function pickLayers(gl, {
...
+      pickedObjectIndex = lastPickedInfo.index;
+      affectedLayers.push(pickedLayer);
+    }
+  } else {
+    // Make sure we clear scissor test and fbo bindings in case of exceptions
+    // We are only interested in one pixel, no need to render anything else
+    // Note that glContextCallback is called synchronously;
+    // code in this function is executed in the order it's written.
+    // TODO - just return glContextWithState once luma updates
+    glContextWithState(gl, {
```
ibgreen: While I agree that this function need to be split into smaller functions, I think this is split in the wrong place. Move the `glContextWithState` into the sub function and rename it something meaningful.

It will make the diff a little smaller also since indentation will be unaffected. It would help my review if this was the case.

Just in case: what `glContextWithState` does is that it locally sets some values on the gl context and guarantees that they are reset even if we exit with an exception.

src/lib/draw-and-pick.js
```diff
@@ -69,17 +75,39 @@ export function pickLayers(gl, {
...
-    scissorTest: {x: deviceX, y: deviceY, w: 1, h: 1}
-  }, () => {
+  if (MOTION_EVENTS.indexOf(mode) !== -1) {
+    // "Motion events" are those that track the motion of an interaction
+    // after its initiation. In this case, these subsequent events
+    // should be bound to the object that was first picked,
+    // e.g. to enable dragging behaviors.
+    // Therefore, the picking process does not run for these events.
+    const {layerId} = lastPickedInfo;
+    pickedLayer = layers.find(l => l.props.id === layerId);
```
ibgreen: Probably you inherited this code but please type out variables: `l` => `layer`.

src/lib/draw-and-pick.js
```diff
@@ -23,6 +23,12 @@ import {GL, glContextWithState} from 'luma.gl';
 import {getUniformsFromViewport} from './viewport-uniforms';
 import {log, getBlendMode, setBlendMode} from './utils';
 
+const MOTION_EVENTS = [
+  'dragmove',
```
ibgreen: Why do we need to separate between drag and touch? These are abstract modes right?

src/react/deckgl.js
```diff
@@ -100,12 +108,18 @@ export default class DeckGL extends React.Component {
     this._updateLayers(this.props);
 
     // Check if a mouse event has been specified and that at least one of the layers is pickable
-    const hasEvent = this.props.onLayerClick !== noop || this.props.onLayerHover !== noop;
+    const hasEvent =
```
ibgreen: As discussed, we need to move this code out to a non-react dependent helper. There should just be a few calls in here that delegate to that helper.

src/react/deckgl.js
```diff
@@ -147,6 +166,52 @@ export default class DeckGL extends React.Component {
...
+    // use offsetX|Y for relative position to the container, drop event if falsy
+    if (!event || !event.event || !Number.isFinite(event.event.offsetX)) {
+      return;
+    }
+    const {event: {offsetX: x, offsetY: y}} = event;
+    const type = typeof explicitType === 'string' ? explicitType : event.event.type;
+    let mode;
+    let layerEventHandler;
+    switch (type) {
+    case 'mousedown':
```
ibgreen: You defined touch modes but I don't see touch events handled here.

src/lib/draw-and-pick.js
ericsoco: Makes sense, code will be cleaner to do so.

src/lib/draw-and-pick.js
ericsoco: Sorry, I don't understand -- I'm not making a distinction between them, they're both included here...?

src/lib/draw-and-pick.js
ericsoco: For future reference, [`?w=1` is your friend for muting whitespace changes](https://github.com/blog/967-github-secrets).

src/react/deckgl.js
ericsoco: That will be part of the future event handling / controller refactor.

src/react/deckgl.js
ericsoco: This PR doesn't as far as supporting touch events. Since we're planning to refactor event handling out anyway, I'm deferring touch event handling for that refactor and focusing only on drag events here (for the graph-layer needs). I'll remove them from `MOTION_EVENTS` for consistency.

src/react/deckgl.js
```diff
@@ -52,7 +56,11 @@ const defaultProps = {
   onWebGLInitialized: noop,
   onLayerClick: noop,
   onLayerHover: noop,
-  onAfterRender: noop
+  onAfterRender: noop,
```
ibgreen: Keep the event handlers together, separated from the render callback.

src/react/deckgl.js
```diff
@@ -100,12 +108,18 @@ export default class DeckGL extends React.Component {
...
+      this.props.onLayerClick !== noop ||
+      this.props.onLayerHover !== noop ||
+      this.props.onLayerDragStart !== noop ||
+      this.props.onLayerDragMove !== noop ||
+      this.props.onLayerDragEnd !== noop ||
+      this.props.onLayerDragCancel !== noop;
     const hasPickableLayer = this.layerManager.layers.map(l => l.props.pickable).includes(true);
     if (this.layerManager.layers.length && hasEvent && !hasPickableLayer) {
       log.once(
         0,
```
ibgreen: I think we should change the log level to 1.

src/react/deckgl.js
```diff
@@ -114,7 +128,11 @@ export default class DeckGL extends React.Component {
       cachePosition: false,
       centerOrigin: false,
       onClick: this._onClick,
-      onMouseMove: this._onMouseMove
+      onMouseMove: this._onMouseMove,
+      onDragStart: this._onDragEvent,
+      onDragMove: this._onDragEvent,
+      onDragEnd: this._onDragEvent,
+      onDragCancel: this._onDragCancel
```
Pessimistress: We're invoking picking on all of these events as long as one of them is specified. Can we add handlers only when they are not `noop`?

src/react/deckgl.js
```diff
@@ -100,12 +108,18 @@ export default class DeckGL extends React.Component {
...
 
     // Check if a mouse event has been specified and that at least one of the layers is pickable
-    const hasEvent = this.props.onLayerClick !== noop || this.props.onLayerHover !== noop;
+    const hasEvent =
+      this.props.onLayerClick !== noop ||
+      this.props.onLayerHover !== noop ||
+      this.props.onLayerDragStart !== noop ||
+      this.props.onLayerDragMove !== noop ||
+      this.props.onLayerDragEnd !== noop ||
+      this.props.onLayerDragCancel !== noop;
```
Pessimistress: This is an anti-React-pattern: these props are only checked once on itialization.

ericsoco: You're right @Pessimistress, this is an anti-pattern. However, it's only checking to offer a warning, in the case that an `onLayer*` handler is passed when no layers have set `pickable`; I imagine the overwhelmingly most common use case will be to set `onLayer*` handlers on init, not dynamically at some point later in the program. Checking these props on every update just to conditionally issue a warning doesn't seem worth the (probably minor) cost.

My vote is we leave as-is unless it proves to be a problem.

src/react/deckgl.js
ericsoco: Actually, digging deeper into this, we're _always_ running picking, even if no layers have `pickable` set. There's no check for existence of handlers or `pickable` truthiness before calling `addEvents`, we're just running picking on every mouse move and calling noops every frame.

Additionally, we currently have no facility for adding/removing event handlers after initialization; deck.gl calls luma's `addEvents` only once, on init; and `addEvents` itself does not support adding/removing handlers. We need to address this in the event handling refactor, @ibgreen.

All this indicates to me that:
- we should definitely fix this
- we should fix it in the event handling refactor, not in this PR.

If @ibgreen and @Pessimistress agree, I'd like to merge as-is.

Pessimistress: We definitely should not register events based on `pickable` states at initialization. In most of our use cases the layers are not passed in till data is loaded. Since we're only logging a warning it's not broken at the moment.

I agree this should be fixed in a separate PR.

PR #585, opened 2017-04-20
Separate composite and non-composite layer APIs

Pessimistress:
Changes:
- Only composite layers have `renderLayer()` methods
- Only non-composite layers' `draw()` methods are called during render

Goal:
- A cleaner interface
- Improve perf by skipping the uniform calculations and parameter updates for composite layers

### DIFF
--- a/docs/advanced/layer-lifecycle.md
+++ b/docs/advanced/layer-lifecycle.md
@@ -59,7 +59,7 @@ and updating unforms by calling
 `model.setUniforms`.
 By default, when `props.data` changes, all attributes are recalculated.
 
-A layer may use
+A composite layer may use
 [`layer.renderLayers()`](/docs/api-reference/composite-layer.md#-renderlayers-)
 to insert one or more deck.gl layers after itself.
 The generated layers will then be matched and updated,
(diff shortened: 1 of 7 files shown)


src/lib/composite-layer.js
```diff
@@ -25,6 +25,10 @@ export default class CompositeLayer extends Layer {
     super(props);
   }
 
+  get isComposite() {
```
ibgreen: It needs to be a getter because layer is sealed? Maybe just keep it as a function?

Pessimistress:
I checked the docs, we already documented the method `renderLayers()` as composite layer only.

docs/advanced/layer-lifecycle.md
```diff
@@ -59,7 +59,7 @@ and updating unforms by calling
 `model.setUniforms`.
 By default, when `props.data` changes, all attributes are recalculated.
 
-A layer may use
+A composite layer may use
 [`layer.renderLayers()`](/docs/api-reference/composite-layer.md#-renderlayers-)
```
ibgreen: `CompositeLayer.renderLayers`

PR #586, opened 2017-04-20
getViewportUniforms Optimization

Pessimistress:
Changes:
- Switch from using custom math class to `gl-matrix`
- Add benchmark

Before:
```
getUniformsFromViewport#LNGLAT x 54,286 ops/sec ±2.77% (84 runs sampled)
getUniformsFromViewport#METER_OFFSETS x 37,155 ops/sec ±2.44% (84 runs sampled)
```
After:
```
getUniformsFromViewport#LNGLAT x 155,845 ops/sec ±1.70% (88 runs sampled)
getUniformsFromViewport#METER_OFFSETS x 123,868 ops/sec ±0.86% (90 runs sampled)
```

### DIFF
--- a/src/lib/viewport-uniforms.js
+++ b/src/lib/viewport-uniforms.js
@@ -19,7 +19,7 @@
 // THE SOFTWARE.
 
 /* global window */
-import {Matrix4} from 'luma.gl';
+import {mat4, vec4} from 'gl-matrix';
 
 import assert from 'assert';
 import {COORDINATE_SYSTEM} from './constants';
@@ -51,7 +51,7 @@ function calculateMatrixAndOffset({
   case COORDINATE_SYSTEM.IDENTITY:
   case COORDINATE_SYSTEM.LNGLAT:
     projectionCenter = ZERO_VECTOR;
-    modelViewMatrix = new Matrix4(viewMatrix);
+    modelViewMatrix = mat4.copy([], viewMatrix);
     break;
 
   // TODO: make lighitng work for meter offset mode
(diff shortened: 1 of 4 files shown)


Pessimistress:
With `glMatrix.debug` off:
```
getUniformsFromViewport#LNGLAT x 54,471 ops/sec ±5.73% (81 runs sampled)
getUniformsFromViewport#METER_OFFSETS x 40,144 ops/sec ±2.07% (86 runs sampled)
```

Pessimistress:
> Still not totally happy since I really value the readability and debug support of the math classes.

My take is that the math classes are great when we are constructing reusable props, but for the code that runs every frame we should prioritize performance. The Viewport classes use gl-matrix directly, for example.

ibgreen:
> With `glMatrix.debug` off:
OK thanks. That was a surprisingly small effect. Will need to port the perf framework to luma.gl and add this as a test case.

ibgreen:
> My take is that the math classes are great when we are constructing reusable props, but for the code that runs every frame we should prioritize performance. 

Yes, agreed, all is not lost. The Math classes still have a role to play.

But regardless, I would expect the classes to be slower by maybe 20%, but not by a factor of 3x. Perhaps there is something that pulls down performance that can be tweaked.

shaojingli:
> But regardless, I would expect the classes to be slower by maybe 20%, but not by a factor of 3x. Perhaps there is something that pulls down performance that can be tweaked.

This is seems to be due to SIMD acceleration of those linear algebra operations

ibgreen:
> This is seems to be due to SIMD acceleration of those linear algebra operations

Yes `gl-matrix` does have SIMD acceleration. That said, the math classes use `gl-matrix` underneath the hood, so if `gl-matrix` SIMD acceleration is active for the basic library, it should be active for the class wrappers as well. 

On this topic, current impression is that Firefox supports the SIMD extensions but Chrome doesn't just yet - or maybe that has changed?

Finally @Pessimistress did some profiling and it looks like it is a transpilation issue - we are using the ES6 feature of inheriting from JavaScript `Array`s (instead of `Object`s) which is not efficiently poly filled in ES5 environments.

ibgreen:
As stated before, I would just like to see the Math class code preserved in the comments rather than deleted completely so that I can activate it when debugging the viewport, rather than having to reimplement it again.

As usual when keeping commented code, there needs to be a comment explaining why we keep it.

PR #588, opened 2017-04-21
Wind demo

ibgreen:


shaojingli:
@ibgreen Thanks for the PR. Nico's wind demo is definitely fascinating but it obviously follows a VERY different approach.  The way WindLayer is created diverges significantly from what we are describing in our docs and expecting our users to do. I'm wondering if it's a better idea to just keep it in a separate branch and merge it to the master when it can be implemented with a standard pattern so that it indeed serves as an example.

ibgreen:
> I'm wondering if it's a better idea to just keep it in a separate branch and merge it to the master when it can be implemented with a standard pattern so that it indeed serves as an example.

@shaojingli Or maybe just move it from the examples folder to e.g. a webgl2 folder while we work on it, so that it doesn't look like a finished example?

ibgreen:
Merged separately by @1chandu

PR #597, opened 2017-04-26
feat(dev): seer integration

Apercu:
Add integration with the soon-to-be born seer.
No performance drawback for production on normal users given the checks over its presence. 

Pass the layer props and allow their edition with an override object that will persist across renders.
Also invalidate the data to be sure we got the latest version of it.

### DIFF
--- a/src/debug/seer-integration.js
+++ b/src/debug/seer-integration.js
@@ -1,0 +1,91 @@
+import seer from 'seer';
+import {window} from '../lib/utils/globals';
+
+/**
+ * Recursively set a nested property of an object given a properties array and a value
+ */
+const recursiveSet = (obj, path, value) => {
+  if (path.length > 1) {
+    recursiveSet(obj[path[0]], path.slice(1), value);
+  } else {
+    obj[path[0]] = value;
+  }
+};
+
+const overrides = new Map();
+
+/**
+ * Create an override on the specify layer, indexed by a valuePath array.
+ * Do nothing in case Seer as not been initialized to prevent any preformance drawback.
+ */
+export const setOverride = (id, valuePath, value) => {
+  if (!window.__SEER_INITIALIZED__) {
+    return;
+  }
+
+  if (!overrides.has(id)) {
+    overrides.set(id, new Map());
+  }
+
+  const props = overrides.get(id);
+  props.set(valuePath, value);
+
+};
+
+/**
+ * Get the props overrides of a specific layer if Seer as been initialized
+ * Invalidates the data to be sure new ones are always picked up.
+ */
+export const getOverrides = props => {
+  if (!window.__SEER_INITIALIZED__ || !props.id) {
+    return;
+  }
+
+  const overs = overrides.get(props.id);
+  if (!overs) {
+    return;
+  }
+
+  overs.forEach((value, valuePath) => {
+    recursiveSet(props, valuePath, value);
+  });
+
+  props.data = [...props.data];
+
+};
+
+/**
+ * Listen for deck.gl edit events
+ */
+export const layerEditListener = cb => {
+  if (!window.__SEER_INITIALIZED__) {
+    return;
+  }
+
+  seer.listenFor('deck.gl', payload => {
+    if (payload.type !== 'edit') {
+      return;
+    }
+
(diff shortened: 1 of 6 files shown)


src/lib/layer-manager.js
```diff
@@ -77,7 +80,18 @@ export default class LayerManager {
         layerId: null
       }
     };
+
     Object.seal(this.context);
+
+    seer.listenFor('deck.gl', payload => {
```
ibgreen: I think it would be nicer if this was a function imported from `seer-integration.js`

src/lib/layer-manager.js
```diff
@@ -216,6 +230,22 @@ export default class LayerManager {
   }
 
   /* eslint-disable max-statements */
+  _logLayer(newLayer) {
+    if (!window.__SEER_INITIALIZED__) {
+      return;
+    }
+
+    const simpleProps = Object.keys(newLayer.props).reduce((acc, key) => {
```
ibgreen: Same here.

src/lib/layer-overrides.js
```diff
@@ -0,0 +1,45 @@
+import {window} from './utils/globals';
```
ibgreen: Rename to `seer-integration.js`...

src/lib/layer-manager.js
Apercu: That's already what it is? Don't get it.

src/lib/layer-manager.js
Apercu: Huh? This is specific to deck.gl, why put it into seer?

src/lib/layer-overrides.js
Apercu: Rename what? window?

src/lib/layer-manager.js
ibgreen: Just like in luma.gl, I am not suggesting we rename `seer` to `seer-integration`. 

Just like in luma.gl, I am suggesting we collect the new seer integration code here inside deck.gl into a new file `seer-integration.js`. 

That file would export a couple of functions that we call in the layer-manager and layer etc.

The purpose is to avoid adding long `seer`-related blocks of code to other classes. If possible, the code in those classes should focus on the main purpose of those classes, not on instrumentation.

src/lib/layer-manager.js
ibgreen: Again, not into `seer` but into a new deck.gl helper file `seer-integration.js`

src/lib/layer-overrides.js
ibgreen: `src/lib/layer-overrides.js` -> `src/lib/debug/seer-integration.js`

src/debug/seer-integration.js
```diff
@@ -0,0 +1,91 @@
...
+  } else {
+    obj[path[0]] = value;
+  }
+};
+
+/**
+ * Get the props overrides of a specific layer if Seer as been initialized
+ * Invalidates the data to be sure new ones are always picked up.
+ */
+export const getOverrides = props => {
```
ibgreen: should setOverrides and getOverrides be a pair (i.e. next to each other in the code)?

src/debug/seer-integration.js
```diff
@@ -0,0 +1,91 @@
...
+    recursiveSet(props, valuePath, value);
+  });
+
+  props.data = [...props.data];
+
+};
+
+/**
+ * Listen for deck.gl edit events
+ */
```
ibgreen: Doesn't the linter rules check the JSDoc and ask for params here? Maybe that only works for functions.

test/node.js
```diff
@@ -21,6 +21,10 @@
 // Enables ES2015 import/export in Node.js
 require('reify');
 
+// Mock addEventListener on window
```
ibgreen: We were going to add a comment that this was for seer, right?

Apercu: Ah, missed this one yes

src/debug/seer-integration.js
Apercu: They were, but the `recursiveSet` util that is unexported needed to be declared first, I can move it at the very top though

PR #603, opened 2017-04-27
Fix brushing layer demo

heshan0131:
For [issue/600](https://github.com/uber/deck.gl/issues/600)


examples/brushing/app.js
```diff
@@ -106,6 +115,7 @@ class Root extends Component {
             brushRadius={100000}
             opacity={0.7}
             strokeWidth={2}
+            enableBrushing={true}
```
Apercu: You don't need to specify the value when you want to pass true, but that's a neatpick

PR #605, opened 2017-04-27
Graph layers

ericsoco:
This PR adds a new sample layer, a non-geospatial layer for layout and rendering of a force-directed network graph. `graph-layer` is a composite layer that delegates rendering to a `ScatterplotLayer`, `LineLayer`, and (optionally) `IconLayer`; `graph-layout-layer` is an adaptor layer that links a layout (defaults to `graph-simulation`, which uses `d3-force`) to the renderer `graph-layer`.

In this design, the graph layout and rendering are decoupled, to allow flexibility with each.

There is an additional layer of processing in `adaptor/` -- these files process data from different formats for use by the specified layout. It might make more sense to call that folder `parser/`, not sure. These parsers/adaptors allow the layout and rendering of three different sample datasets, which can be toggled via the `DATASET` const in `app.js`.

This PR has not gone very far into documentation, and does not add the sample layer to the layer browser; I'll do that work in a subsequent PR once we land on an implementation here that we're comfortable with.

Some notes:
1. I'm driving the force layout manually, `forceUpdate()`ing the root component and subsequently `tick()`ing the simulation via `requestAnimationFrame()`. Alternately, we could let the layout run independently of rendering, but that feels more magical and hard for devs to understand/manipulate. Also, I'm concerned that running the layout independent of the rendering / React updating might cause stutter in the animation.

2. Because of 1. above, I have to force deck.gl to update on every simulation tick, even when no props passed into `graph-layer` have changed. The simulation `alpha` remains constant while dragging as it's continually being re-warmed, so `alpha` can't be used as an `updateTrigger`. Therefore, I'm creating a new `object` every update while the simulation is still running and using that as an `updateTrigger`. Is there a less hacky, more explicit way to have deck.gl update on-demand?

3. `graph-layout-layer` is passing up the type of picked element via its own `getPickingInfo()` implementation. Within, I'm determining if a link or node is picked by digging into `info.layer.context.lastPickedInfo.layerId`. This feels like the wrong way to determine which layer was picked from among `graph-layer`'s sublayers. Is there a better way to get this information from the picking info object? (I can also tell by the structure of the picked datum/object if necessary, but thought I should be able to find the picked layer lowest in the layer stack from the picking info, and so I went digging...)


shaojingli:
I need some time to review this change.

Also, we recently created a repo for data files (https://github.com/uber-common/deck.gl-data) so those 100k line data files don't need to be in deck.gl. Let's discuss the correct way to utilize it

docs/layers/graph-layer.md
```diff
@@ -0,0 +1,35 @@
+<!-- INJECT:"GraphLayerDemo" -->
```
ibgreen: Since everything else is in the `/examples` folder (a good initial approach I think), this doc would be better as a README.md in the layer folder in the example at this point rather than part of the official layer documentation.

examples/graph/app.js
```diff
@@ -0,0 +1,393 @@
+/* global window,document */
+import React, {Component} from 'react';
+import {render} from 'react-dom';
+import {json as requestJSON, csv as requestCSV} from 'd3-request';
+
+import DeckGLOverlay from './deckgl-overlay';
+import {default as GraphBasic} from './graph-layer/adaptor/graph-basic';
```
gnavvy: shall we add an index.js to the adaptor directory that handles default exports, so it's less verbose in the app.js? 
also, would be helpful to have an inline comment, like in your PR description, briefly mentioning what an adaptor does, in cases users (like me) jump directly into reading the code.

examples/graph/app.js
```diff
@@ -0,0 +1,393 @@
...
+      }
+    });
+
+    // only set up icon accessors for sample datasets that have types to be represented as icons.
+    if (dataConfig[DATASET].hasNodeTypes) {
+      // load icon atlas
+      requestJSON('./data/node-icon-atlas.json', (error, response) => {
+        if (!error) {
+          this.setState({
+            iconMapping: response
```
gnavvy: no default value for iconMapping in state.

examples/graph/app.js
```diff
@@ -0,0 +1,393 @@
...
+    return !this._animationFrame;
+  }
+
+  componentDidUpdate() {
+    /* eslint-disable react/no-did-update-set-state */
+    // lastDragged state lasts only one frame
+    const {lastDragged} = this.state;
+    if (lastDragged) {
+      this.setState({
+        lastDragged: null
```
gnavvy: no default lastDragged in state.

examples/graph/app.js
```diff
@@ -0,0 +1,393 @@
...
+      }
+    ];
+    const loader = dataConfig[DATASET].loader;
+    loader(dataConfig[DATASET].data, (error, response) => {
+      if (!error) {
+        // apply timestamp and push loaded sample data into array
+        const GraphAdaptor = dataConfig[DATASET].adaptor;
+        const graph = new GraphAdaptor(response);
+        this.setState({
+          graph,
```
gnavvy: seems graph is set but never used?

examples/graph/app.js
```diff
@@ -0,0 +1,393 @@
...
+  _renderInteractionElement(el, className, viewport) {
+    // Note: node.x/y, calculated by d3 layout,
+    // is measured from the center of the layout (of the viewport).
+    // Therefore, we offset the node coordinates from the viewport center.
+
+    let element;
+    if (el.source) {
+      // link
+      element = (<line
+        x1={el.source.x + viewport.width / 2}
```
gnavvy: instead of adding the offset to x and y for each element, we can keep move it out by adding a 
<g transform={\`translate(${viewport.width / 2} ${viewport.height / 2})\`}>{...}\</g>
that wraps the content in the <svg> tag in line 295.

examples/graph/app.js
```diff
@@ -0,0 +1,393 @@
...
+      return 'pie';
+    default:
+      return null;
+    }
+  }
+
+  //
+  // rendering
+  //
+  _renderInteractionLayer(viewport, hovered, clicked) {
```
gnavvy: it seems the interactionLayer is an SVG layer used for highlighting the selected & related elements.
we might need some descriptions above for explanation. 

also, (not in this PR), while I agree that using SVG for highlighting should be performant enough given we are only highlighting a limited number of connected/related elements for the current graph use case, we loose the flexibility to interleave them (Interaction layer will also on top) for 3D graphs. let's discuss this later.

examples/graph/app.js
```diff
@@ -0,0 +1,393 @@
...
+    // set flags used below to determine if SVG highlight elements should be rendered.
+    // if truthy, each flag is replaced with the corresponding element to render.
+    const elements = {
+      hovered: hovered && hovered.object,
+      clicked: clicked && clicked.object
+    };
+    const relatedElements = {
```
gnavvy: Is this demonstration correct?

    // (node) ---------(edge)------------------ (node)
    //   ^ related       ^ hovered/clicked       ^ related
    //
    // (node) ----(edge)---- (node) ------------(edge)---- (node)
    //   ^ related  ^ related  ^ hovered/clicked  ^ related  ^ related

examples/graph/app.js
ericsoco: good idea.

examples/graph/app.js
ericsoco: yeah, this whole SVG interaction layer thing is less than ideal. let's discuss later if there's a way to handle this more seamlessly within deck.gl.

examples/graph/app.js
ericsoco: while i find that pattern helpful for e.g. library code, in which the contents of the folder fronted by the `index.js` might grow over time, i think in example code it's less helpful -- there's no reason to assume we'll be adding more adaptors as time goes on, and it's one more step the dev following the example has to process to understand where the imports are coming from.  (also, it's a bit magical for those who don't understand that `import` resolution looks for an `index.js` first.)

agreed on the comment, will add.

examples/graph/app.js
ericsoco: I use it twice, immediately after the declaration.

ericsoco: Ah, you mean `graph` is never pulled off of `this.state`. Good catch.

examples/graph/app.js
ericsoco: That is how highlights are applied, yes. However, the `relatedObjects` field only contains the nodes at either end of a selected link, or the links attached to a selected node.

I added the `relatedObjects` field as a convenience for consumers of `graph-layout-layer`. My implementation of choosing elements to highlight, however, is split across `graph-layout-layer::getPickingInfo()` (which assigns `relatedObjects` as I just described), and `app.js::_renderInteractionLayer()`, which goes one step further to pick nodes attached to links that appear as `relatedObjects`.

As I explain this, I'm thinking perhaps I should move all that logic into `getPickingInfo()`. Do you agree that it's a common use case to desire access to nodes connected to a selected node?

examples/graph/graph-layer/graph-layout-layer.js
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    const {layoutProps} = this.props;
+    const {nodes, isUpdating} = layout.update(layoutData, layoutProps);
+    if (isUpdating) {
+      // update state only if layout is still running
+      this.state.nodes = nodes;
+      this.state.links = data ? data.links : undefined;
+      // create new object for updateTriggers diff
+      this.state.layoutUpdating = {};
```
ericsoco: @gnavvy and I discussed; instead of using an Object, a timestamp is probably more legible to indicate that this is a value changing every update.

shaojingli:
> I'm driving the force layout manually, forceUpdate()ing the root component and subsequently tick()ing the simulation via requestAnimationFrame(). Alternately, we could let the layout run independently of rendering, but that feels more magical and hard for devs to understand/manipulate. Also, I'm concerned that running the layout independent of the rendering / React updating might cause stutter in the animation.

Using forceUpdate() for this simple example is okay, but if we were to make this GraphLayer a workhorse for our applications, we might not want the force calculation to run at 60fps. 

That being said, one more important question to be answered here is whether we actually want to have GraphLayer (which contains graph data) as a **sublayer** of GraphLayoutLayer (which contains layout algorithm). It might be worth considering making layout algorithm pluggable to a graph layer as user may want to frequently swap layout algorithms for a same set of data.

More over, in the future, we might want to implement layout calculations outside of browser (e. g. in middleware), if we'd like to visualize millions of nodes or combine layout with other algorithms like edge bundling.

shaojingli:
> graph-layout-layer is passing up the type of picked element via its own getPickingInfo() implementation. Within, I'm determining if a link or node is picked by digging into info.layer.context.lastPickedInfo.layerId. This feels like the wrong way to determine which layer was picked from among graph-layer's sublayers. Is there a better way to get this information from the picking info object? (I can also tell by the structure of the picked datum/object if necessary, but thought I should be able to find the picked layer lowest in the layer stack from the picking info, and so I went digging...)

You are right. Determining whether a link or node has been picked shouldn't involve digging into layer.context. This depends too much on internal implementations of deck.gl's picking and will sure be broken if the implementation is changed. It's better that the type of picked entity be encoded somehow to the pickInfo.object, just like what GeoJson layer is doing.

examples/graph/deckgl-overlay.js
```diff
@@ -0,0 +1,53 @@
...
+  render() {
+    const {viewport, data} = this.props;
+
+    if (!data) {
+      return null;
+    }
+
+    const {width, height} = viewport;
+    const {layoutProps} = this.props;
+    const {layoutAccessors, linkAccessors, nodeAccessors, nodeIconAccessors} = this.props;
```
ericsoco: I think `get*` is good for an individual accessor, but these are maps of accessors, not functions. "Get" is a verb and so implies a function, so it doesn't quite map right for these maps. If you insisted I'd change them but since you're fine with either, I'll leave as-is :)

PR #606, opened 2017-04-27
Remove /examples from coverage. Add report to npm run cover. 

ibgreen:
Include `npm build` in `npm test` to make sure commits don't break the build (which is otherwise only run on publish).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,5 +20,4 @@ script:
   - npm run lint
   - npm run cover
 after_success:
-  - npm install -g coveralls
-  - cat coverage/lcov.info | coveralls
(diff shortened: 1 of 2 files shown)


ibgreen:
@heshan0131 - unless I did something wrong, removing the examples doesn't seem to improve the coverage. 

@apercu suggested we disable the coveralls integration until we can get coverage up a bit.

heshan0131:
I also realized that removing examples doesn't increase coverage. Will look into it later

heshan0131:
At the mean time I am working on a PR with boilerplate tests that can be used to increase test coverage of all our layers.

PR #607, opened 2017-04-28
Fix radiusMinPixels and radiusMaxPixels for METER_OFFSETS in scatterplot

huaruiwu:


shaojingli:
@huaruiwu Thanks for the PR! 

Yes, this is a bug and your change solves the problem. However, it might be better to have this check on projection mode be implemented in shaderlib/project.glsl.js. This will solve this problem also for other layers and make the code clean.

Could you update your change?

huaruiwu:
Glad to be of help!

great suggestion. I will take a look and try to make the appropriate changes.

huaruiwu:
@shaojingli You were right, looks like there was some weird code causing bugs in `project.glsl.js`. I think I've fixed it. Could you check it over? Thanks!

shaojingli:
This doesn't look right to me... Ib can you review it? @ibgreen

ibgreen:
We need to be very careful with changing project functions. 
- The idea was to convert things to meters, so that the shader can do math in meters regardless of which projection mode is used.
- Here we have a different need, to convert things to pixels, so that the shader can do math in pixels. My assumption is that we need a new set of functions in the project package.

Something like
- project_scale_to_meters
- project_scale_to_pixels

But these can not both be passed to the projectionMatrix, so we need an additional project function in the middle.

As much as I'd love to have a PR on this, it will need very careful thought and review by the deck.gl team. If we really want to do this now, we'll need to take the time to get a proposal up for how we extend `project` package.

huaruiwu:
@ibgreen Would it be okay to just put in the fix I had in initially, where I simply convert `radiusMaxPixels` and `radiusMinPixels` into meters in the `ScatterLayer` shader? This would fix the bug I am concerned about.

shaojingli:
@huaruiwu we plan to land your PR after we add some test cases for the meter offset mode. hopefully very soon

huaruiwu:
okay thank you!

Pessimistress:
With this PR applied, in the scatterplot (meters mode) rendering test, pixel match increased from `12.306%` to `94.068%`. The remaining difference is presumedly from treating pixels per meter as a uniform.

huaruiwu:
awesome ! thank you @Pessimistress

PR #608, opened 2017-04-28
Upgrades to luma.gl v4.0.0-alpha.1 with rendering fixes

ibgreen:


ibgreen:
Will address any comments in separate PR.

PR #610, opened 2017-04-28
Add coverage percentiles getColorValue to grid layer for consistence with hexagonLayer, add layer tests

heshan0131:
- add coverage percentiles getColorValue to grid layer
- add to layer docs
- add gridLayer and hexagonLayer tests
- add more layer tests

coverage is now at 69%
screen shot 2017-04-27 at 7 53 05 pm


ibgreen:
Shouldn't this be two separate PRs?

shaojingli:
This is something tremendous! Thanks @heshan0131 

But, as mentioned by @ibgreen , it seems to be better to break it to two PRs. One for adding `coverage` to the GridLayer and another for adding layer tests so that we can track changes more easily in the future.

BTW, when I saw 'coverage' in the title at first, I thought the PR is for increasing "test coverage"... only later I found that you are adding functionality too ~~~~

heshan0131:
The testing get out of hand.
Will make into 2 PRs

PR #611, opened 2017-04-28
Layer tests

heshan0131:
-  Add layer test utils for base layer test, using jsDom and sinon
- Disable some grid / hexagon layer test for now
- Coverage increase to 68%
screen shot 2017-04-27 at 11 46 41 pm

### DIFF
--- a/test/layers/grid-layer.spec.js
+++ b/test/layers/grid-layer.spec.js
@@ -250,9 +250,9 @@ test('GridLayer#renderSubLayer', t => {
   // should call attribute updater twice
   // because test util calls both initialize and update layer
   t.ok(GridLayer.prototype._onGetSublayerColor.called,
-    'should call _onGetSublayerColor number of hexagons times 2');
+    'should call _onGetSublayerColor');
   t.ok(GridLayer.prototype._onGetSublayerElevation.called,
-    'should call _onGetSublayerElevation number of hexagons times 2');
+    'should call _onGetSublayerElevation');
   GridLayer.prototype._onGetSublayerColor.restore();
   GridLayer.prototype._onGetSublayerElevation.restore();
 
--- a/test/layers/grid-layer.spec.js
+++ b/test/layers/grid-layer.spec.js
@@ -17,12 +17,158 @@
 // LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
-
+/* eslint-disable no-unused-vars */
 import test from 'tape-catch';
-import * as data from '../data';
-import {testInitializeLayer} from '../test-utils';
+import sinon from 'sinon';
 
-import {GridLayer} from 'deck.gl';
+import * as data from '../data';
+import {
+  testInitializeLayer,
+  testLayerUpdates,
+  testSubLayerUpdateTriggers
+} from '../test-utils';
+
+import {GridLayer, GridCellLayer} from 'deck.gl';
+
+const getColorValue = points => points.length;
+const getPosition = d => d.COORDINATES;
+
+const TEST_CASES = {
+  // props to initialize layer with
+  initialProps: {
+    data: data.points,
+    cellSize: 400,
+    getPosition,
+    pickable: true
+  },
+  // list of update props to call and asserts on the resulting layer
+  updates: [{
+    updateProps: {
+      cellSize: 800
+    },
+    assert: (layer, oldState, t) => {
+      t.ok(oldState.layerData !== layer.state.layerData,
+        'should update layer data');
+
+      t.ok(oldState.sortedBins !== layer.state.sortedBins,
+        'should update sortedBins');
+
+      t.ok(oldState.valueDomain !== layer.state.valueDomain,
+        'should update valueDomain');
+    }
+  }, {
+    updateProps: {
+      getColorValue
+    },
+    assert: (layer, oldState, t) => {
+      t.ok(oldState.layerData === layer.state.layerData,
+        'should not update layer data');
+
+      t.ok(oldState.sortedBins !== layer.state.sortedBins,
+        'should update sortedBins');
+
+      t.ok(oldState.valueDomain !== layer.state.valueDomain,
+        'should re calculate valueDomain');
+    }
+  }, {
+    updateProps: {
+      upperPercentile: 90
+    },
+    assert: (layer, oldState, t) => {
+      t.ok(oldState.layerData === layer.state.layerData,
+        'should not update layer data');
+
+      t.ok(oldState.sortedBins === layer.state.sortedBins,
+        'should not update sortedBins');
+
+      t.ok(oldState.valueDomain !== layer.state.valueDomain,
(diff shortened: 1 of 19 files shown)


test/test-utils.js
```diff
@@ -86,3 +90,153 @@ export function testInitializeLayer({gl, layer, viewport}) {
...
+  try {
+    layer.updateLayer({
+      oldProps,
+      props: layer.props,
+      oldContext,
+      context,
+      changeFlags: layer.diffProps(oldProps, layer.props, context)
+    });
+  } catch (error) {
+    failures++;
```
ibgreen: I guess this follows the existing code, but how could failures be more than 0 or 1? Should we use a `Boolean`?

test/layers/hexagon-layer.spec.js
```diff
@@ -59,3 +207,59 @@ test('HexagonLayer#constructor', t => {
...
+
+  t.end();
+});
+
+test('HexagonLayer#updateLayer', t => {
+  testLayerUpdates({LayerComponent: HexagonLayer, testCases: TEST_CASES, t});
+  t.end();
+});
+
+/*
```
ibgreen: Cut or add comment explaining why we keep.

test/layers/hexagon-layer.spec.js
```diff
@@ -19,15 +19,162 @@
...
+      getPosition
+    },
+    assert: (subLayer, spies, t) => {
+      t.ok(spies._onGetSublayerColor.notCalled,
+        'update opacity should not call _onGetSublayerColor');
+      t.ok(spies._onGetSublayerElevation.notCalled,
+        'update opacity  should not call _onGetSublayerElevation');
+    }
+  }, {
+  //   newProps: {
```
ibgreen: Cut commented code, or add comment explaining why we keep.

test/layers/hexagon-layer.spec.js
```diff
@@ -19,15 +19,162 @@
...
+        'should update layer data');
 
-import {HexagonLayer} from 'deck.gl';
+      // t.ok(oldState.sortedBins !== layer.state.sortedBins,
+      //   'should update sortedBins');
+
+      t.ok(oldState.valueDomain !== layer.state.valueDomain,
+        'should update valueDomain');
+    }
+  // }, {
```
ibgreen: Cut?

test/layers/grid-layer.spec.js
```diff
@@ -51,3 +227,58 @@ test('GridLayer#constructor', t => {
...
+  // render sublayer
+  const subLayer = layer.renderLayers();
+  testInitializeLayer({layer: subLayer});
+
+  t.ok(subLayer instanceof GridCellLayer, 'GridCellLayer rendered');
+
+  // should call attribute updater twice
+  // because test util calls both initialize and update layer
+  t.ok(GridLayer.prototype._onGetSublayerColor.called,
+    'should call _onGetSublayerColor number of hexagons times 2');
```
ibgreen: You are not asserting the call count, so strange to call it out in the message?

test/layers/grid-layer.spec.js
```diff
@@ -51,3 +227,58 @@ test('GridLayer#constructor', t => {
...
+  t.ok(GridLayer.prototype._onGetSublayerElevation.called,
+    'should call _onGetSublayerElevation number of hexagons times 2');
+  GridLayer.prototype._onGetSublayerColor.restore();
+  GridLayer.prototype._onGetSublayerElevation.restore();
+
+  t.end();
+});
+
+/*
```
ibgreen: Cut

test/layers/grid-layer.spec.js
```diff
@@ -31,15 +177,45 @@ test('GridLayer#constructor', t => {
...
     pickable: true
   });
   t.ok(layer instanceof GridLayer, 'GridLayer created');
 
   testInitializeLayer({layer});
 
+  const {layerData} = layer.state;
+
+  t.ok(layerData.length > 0, 'GridLayer.state.layerDate calculated');
+  // t.ok(sortedBins, 'GridLayer.state.sortedBins calculated');
```
ibgreen: Cut

test/layers/grid-layer.spec.js
```diff
@@ -17,12 +17,158 @@
...
+
+const SUBLAYER_TEST_CASES = {
+  // props to initialize layer with
+  initialProps: {
+    data: data.points,
+    cellSize: 400,
+    getPosition
+  },
+  // list of update props to call and asserts on the resulting layer
+  updates: [{
```
ibgreen: If these are sublists, use uppercase `UPDATES`

test/layers/grid-layer.spec.js
```diff
@@ -17,12 +17,158 @@
...
+      t.ok(oldState.sortedBins === layer.state.sortedBins,
+        'should not update sortedBins');
+
+      t.ok(oldState.valueDomain !== layer.state.valueDomain,
+        'should re calculate valueDomain');
+    }
+  }]
+};
+
+const SUBLAYER_TEST_CASES = {
```
ibgreen: Looks like the right approach - but doesn't seem used?

test/layers/polygon-layer.spec.js
```diff
@@ -19,50 +19,48 @@
...
-
-  layer = new PolygonLayer({
-    data: data.choropleths
-  });
-  t.ok(layer instanceof PolygonLayer, 'PolygonLayer created');
-
-  testInitializeLayer({layer});
-  // t.ok(layer.state.model, 'PolygonLayer has state');
+  const TEST_CASES = {
+    initialProps: {
```
ibgreen: I'd use upper case here to show that these are not cases but sublists

test/setup.js
```diff
@@ -0,0 +1,22 @@
...
+import global from 'global';
+
+// using jsDom to emulate a subset of a web browser to be useful for testing and
+// scraping real-world web applications
+// this is needed to enable using sinon in the node env
+// to test spies, stubs and mocks on function calls
+
+global.document = jsdom('');
+global.window = global.document.defaultView;
+Object.keys(global.document.defaultView).forEach((property) => {
```
ibgreen: This is a lot of copying. When do we actually need this? What is deck.gl's layer manager's dependency on the DOM here? Does sinon not even work unless JSDom is available? That is a lot of dependencies just to be able to spy on a couple of function.

So, good if we could explain deck.gl's need for this more specifically in comments rather than just a generic statement about sinon.

test/test-utils.js
```diff
@@ -86,3 +90,153 @@ export function testInitializeLayer({gl, layer, viewport}) {
...
+  let failures = 0;
+  try {
+    layer.drawLayer({uniforms});
+  } catch (error) {
+    failures++;
+  }
+
+  return failures;
+}
+// testLayerUpdates test takes a layer component, initialize it, then call testUpdateLayer
```
ibgreen: These functions are pretty central, give them full JSDoc?

heshan0131:
The commented code are tests for the changes I made to hex/grid, which I separated out to another PR. For this time being I simply comment them out. Will put them back later

test/layers/grid-layer.spec.js
heshan0131: was going to include in the hex/grid diff, added it back now

PR #613, opened 2017-04-28
Shader cache

shaojingli:
Adding ShaderAssembler/ShaderCache to deck.gl so that one shader get compiled only once during the entire DeckGL lifetime.

Made some aesthetic changes to the code so that getModel() always spell as _getModel() in all layers. Also cleaned up some shader names along the way. That's why we see 34 changed files in this PR.

@Pessimistress @ibgreen


Pessimistress:
This is a pretty big breaking change. Any way we can avoid modifying the `assembleShaders` interface?
Maybe add an additional option to the second argument, something like:
```
getShaders() {
  return {
    vs: VERTEX_SHADER,
    fs: FRAGMENT_SHADER,
    modules: ['lighting'],
    cache: this.context.shaderCache
  }
}
```
That way existing layers will continue to work, just won't take advantage of the cache.

src/shader-utils/assemble-shaders.js
```diff
@@ -101,9 +100,21 @@ function assembleShader(gl, opts = {}) {
 }
 
 export function assembleShaders(gl, opts) {
+  const vsSource = assembleShader(gl, opts);
```
shaojingli: shaderCache is passed in as a property of `opts`. If this property does not exist, the function takes the original path and simply return shader source strings.

shaojingli:
> Any way we can avoid modifying the assembleShaders interface?

I kept the interface of assembleShaders() so that layers passing in shader in the old way shouldn't be affected.

Now, a ShaderAssembler is created when LayerManager creates its context. ShaderAssembler holds a luma.gl/ShaderCache object and both objects has the same lifetime of the LayerManager. 

Please review again. @Pessimistress

src/layers/core/arc-layer/arc-layer.js
```diff
@@ -104,7 +108,8 @@ export default class ArcLayer extends Layer {
     }));
   }
 
-  _getModel(gl) {
+  _getModel() {
```
ibgreen: I was thinking we could start supplying `gl` as an argument to `initializeState`, `updateState` etc so that the app doesn't have to keep looking at the context. 

```
intializeState({gl}) {
 ...
```

The API would look simpler and less intimidating if we didn't mention context in most layers I think.

The reason I mention it is that this particular change goes the other way, making context more prominent.

src/layers/core/grid-cell-layer/grid-cell-layer.js
```diff
@@ -66,15 +66,20 @@ export default class GridCellLayer extends Layer {
 
   getShaders() {
     return enable64bitSupport(this.props) ? {
-      vs: gridCellVertex64, fs: gridCellFragment, modules: ['fp64', 'project64', 'lighting']
+      vs: gridCellVertex64,
```
ibgreen: As a personal preference, I wish we didn't have to expand into multiple lines - I really try to make the layer code have fewer lines. How about:

```
getShaders() {
      const {shaderCache} = this.context;
      return enable64bitSupport(this.props) ? {
 -      vs: vertex64, fs: fragment, modules: ['fp64', 'project64', 'lighting'], shaderCache
     } : {

File: docs/advanced/primitive-layers.md [hunk 4]
```diff
@@ -58,7 +58,8 @@ be instanced, or use dynamic geometry:
   import {assembleShaders} from 'deck.gl';
   import {GL, Model, Geometry} from 'luma.gl';
 
-  getModel(gl) {
+  _getModel() {
+    const {gl} = this.context;
```
Pessimistress: Is this change still needed? The layers that subclass the core layers and override `_getModel()` will be broken.

ibgreen: > Is this change still needed? The layers that subclass the core layers and override _getModel() will be broken.

Good point. Not sure if there are many layers like that, but the less we change, the less we can break...

src/layers/core/grid-cell-layer/grid-cell-layer.js
shaojingli: I found some lines with multiple modules extending over 100 characters limit... that's why I break them into multiple lines...

PR #614, opened 2017-04-29
- Add coverage percentiles getColorValue prop to grid layer 

heshan0131:
- To make grid layer consistent with hexagon layer, add `coverage`, `lowerPercentile`, `upperPercentile` and `getColorValue` to layer prop
- Fix bug in hex `_onGetSublayerColor`, use bin.value instead of bin.points.length to calculate color

### DIFF
--- a/docs/layers/grid-layer.md
+++ b/docs/layers/grid-layer.md
@@ -62,6 +62,23 @@ Color scale domain, default is set to the range of point counts in each cell.
 Color ranges as an array of colors formatted as `[255, 255, 255]`. Default is
 [colorbrewer](http://colorbrewer2.org/#type=sequential&scheme=YlOrRd&n=6) `6-class YlOrRd`.
 
+##### `getColorValue` (Function, optional)
+
+- Default: `points => points.length`
+
+`getColorValue` is the accessor function to get the value that cell color is based on. 
+It takes an array of points inside each cell as arguments, returns a value. For example, 
+You can pass in `getColorValue` to color the cells by avg/mean/max of a specific attributes of each point.
+By default `getColorValue` returns the length of the points array.
+
+##### `coverage` (Number, optional)
+
+- Default: `1`
+
+Cell size multiplier, clamped between 0 - 1. The final size of cell
+is calculated by `coverage * cellSize`. Note: coverage does not affect how points
+are binned.
+
 ##### `elevationDomain` (Array, optional)
 
 - Default: `[0, max(count)]`
@@ -87,6 +104,20 @@ Cell elevation multiplier. The elevation of cell is calculated by
 - Default: `true`
 
 Whether to enable cell elevation. Cell elevation scale by count of points in each cell. If set to false, all cell will be flat.
+
+##### `upperPercentile` (Number, optional)
+
+- Default: `100`
+
+Filter cells and re-calculate color by `upperPercentile`. Cells with value
+larger than the upperPercentile will be hidden.
+
+##### `lowerPercentile` (Number, optional)
+
+- Default: `0`
+
+Filter bins and re-calculate color by `lowerPercentile`. Cells with value
+smaller than the lowerPercentile will be hidden.
 
 ##### `fp64` (Boolean, optional)
 
--- a/docs/layers/grid-layer.md
+++ b/docs/layers/grid-layer.md
@@ -67,9 +67,31 @@ Color ranges as an array of colors formatted as `[255, 255, 255]`. Default is
 - Default: `points => points.length`
 
 `getColorValue` is the accessor function to get the value that cell color is based on. 
-It takes an array of points inside each cell as arguments, returns a value. For example, 
+It takes an array of points inside each cell as arguments, returns a number. For example, 
 You can pass in `getColorValue` to color the cells by avg/mean/max of a specific attributes of each point.
 By default `getColorValue` returns the length of the points array.
+
+Note: grid layer compares whether `getColorValue` has changed to
+recalculate the value for each bin that its color based on. You should
+pass in the function defined outside the render function so it doesn't create a 
+new function on every rendering pass. 
+
+```
+ class MyGridLayer {
+    getColorValue (points) {
+        return points.length;
+    }
+    
+    renderLayers() {
+      return new GridLayer({
+        id: 'grid-layer',
+        getColorValue: this.getColorValue // instead of getColorValue: () => { return points.length; }
+        data,
+        cellSize: 500
+      });
+    }
+ }
+```
 
 ##### `coverage` (Number, optional)
 
@@ -77,7 +99,7 @@ By default `getColorValue` returns the length of the points array.
 
 Cell size multiplier, clamped between 0 - 1. The final size of cell
 is calculated by `coverage * cellSize`. Note: coverage does not affect how points
-are binned.
+are binned. Coverage are linear based.
 
 ##### `elevationDomain` (Array, optional)
 
--- a/docs/layers/grid-layer.md
+++ b/docs/layers/grid-layer.md
@@ -85,7 +85,7 @@ new function on every rendering pass.
     renderLayers() {
       return new GridLayer({
         id: 'grid-layer',
-        getColorValue: this.getColorValue // instead of getColorValue: () => { return points.length; }
+        getColorValue: this.getColorValue // instead of getColorValue: (points) => { return points.length; }
         data,
         cellSize: 500
       });
--- a/docs/layers/hexagon-layer.md
+++ b/docs/layers/hexagon-layer.md
@@ -84,9 +84,31 @@ Hexagon color ranges as an array of colors formatted as `[[255, 255, 255, 255]]`
 - Default: `points => points.length`
 
 `getColorValue` is the accessor function to get the value that bin color is based on. 
-It takes an array of points inside each bin as arguments, returns a value. For example, 
+It takes an array of points inside each bin as arguments, returns a number. For example, 
 You can pass in `getColorValue` to color the bins by avg/mean/max of a specific attributes of each point.
 By default `getColorValue` returns the length of the points array.
+
+Note: hexagon layer compares whether `getColorValue` has changed to
+recalculate the value for each bin that its color based on. You should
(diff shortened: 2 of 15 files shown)


docs/layers/grid-layer.md
```diff
@@ -62,6 +62,23 @@ Color scale domain, default is set to the range of point counts in each cell.
 Color ranges as an array of colors formatted as `[255, 255, 255]`. Default is
 [colorbrewer](http://colorbrewer2.org/#type=sequential&scheme=YlOrRd&n=6) `6-class YlOrRd`.
 
+##### `getColorValue` (Function, optional)
```
ibgreen: Is there a reason not to just call this `getValue`?

The description can explain that this is used to calculate value. 

```
##### `getColorValue` (Function, optional)
```
Note to self: It would be nice to be able to show more clearly the expected return type of the accessors (in this case number)

docs/layers/grid-layer.md
```diff
@@ -62,6 +62,23 @@ Color scale domain, default is set to the range of point counts in each cell.
...
+
+- Default: `points => points.length`
+
+`getColorValue` is the accessor function to get the value that cell color is based on. 
+It takes an array of points inside each cell as arguments, returns a value. For example,
```
ibgreen: "returns a value" => "returns a number"?

docs/layers/grid-layer.md
```diff
@@ -62,6 +62,23 @@ Color scale domain, default is set to the range of point counts in each cell.
...
+`getColorValue` is the accessor function to get the value that cell color is based on. 
+It takes an array of points inside each cell as arguments, returns a value. For example, 
+You can pass in `getColorValue` to color the cells by avg/mean/max of a specific attributes of each point.
+By default `getColorValue` returns the length of the points array.
+
+##### `coverage` (Number, optional)
+
+- Default: `1`
+
+Cell size multiplier, clamped between 0 - 1. The final size of cell
```
ibgreen: Specify whether coverage is linear or area based? I assume this is just a clamped `radiusScale` factor, i.e. linear.

docs/layers/hexagon-layer.md
```diff
@@ -131,15 +131,15 @@ Whether to enable cell elevation. Cell elevation scale by count of points in eac
 
 - Default: `100`
 
-Filter bins and re-calculate color by `upperPercentile`. Hexagons with counts
-bigger than the upperPercentile counts will be hidden.
+Filter bins and re-calculate color by `upperPercentile`. Hexagons with value
+larger than the upperPercentile will be hidden.
```
ibgreen: Wouldn't it be cool if they were just made really transparent / ghosted?

examples/layer-browser/src/examples/core-layers.js
```diff
@@ -209,14 +209,29 @@ const GridCellLayerExample = {
   }
 };
 
+function getMean(pts, key) {
+  const filtered = pts.filter(pt => Number.isFinite(pt[key]));
+
+  return filtered.length ?
```
ibgreen: Why the check?
```
[].reduce((acc, elt) => acc + elt, 0) => 0
```

examples/layer-browser/src/examples/core-layers.js
```diff
@@ -209,14 +209,29 @@ const GridCellLayerExample = {
...
+
+function getColorValue(points) {
+  return getMean(points, 'SPACES');
+}
+
 const GridLayerExample = {
   layer: GridLayer,
   propTypes: {
-    cellSize: {type: 'number', min: 0, max: 1000}
+    cellSize: {type: 'number', min: 0, max: 1000},
```
ibgreen: Time to put up a `propTypes` PR for deck.gl itself?

src/layers/core/grid-layer/grid-layer.js
```diff
@@ -25,18 +25,20 @@ import {pointToDensityGridData} from './grid-aggregator';
 import {linearScale, quantizeScale} from '../../../utils/scale-utils';
 import {defaultColorRange} from '../../../utils/color-utils';
 
-const defaultCellSize = 1000;
```
ibgreen: +1, looks cleaner after the change

src/layers/core/grid-layer/grid-layer.js
```diff
@@ -54,25 +56,62 @@ function _needsReProjectPoints(oldProps, props) {
   return oldProps.cellSize !== props.cellSize;
 }
 
+function _percentileChanged(oldProps, props) {
+  return oldProps.lowerPercentile !== props.lowerPercentile ||
+    oldProps.upperPercentile !== props.upperPercentile;
+}
+
+function _needsReSortBins(oldProps, props) {
+  return oldProps.getColorValue !== props.getColorValue;
```
ibgreen: I would recommend giving this update criterion another round of thought, as it is inconsistent with updateTriggers which we use to make sure we don't update when functions change.

This is to handle the common cases of ad-hoc functions and bound functions:
```
new Layer({
   getColorValue: () => {}, // new function every render,
   getColorValue: this._getColorValue.bind(this), // new function every render,
});
```

src/layers/core/grid-layer/grid-layer.js
```diff
@@ -54,25 +56,62 @@ function _needsReProjectPoints(oldProps, props) {
...
+
+      this.getSortedBins();
+      this.onPercentileChange();
 
-      const {layerData, countRange} =
-        pointToDensityGridData(data, cellSize, getPosition);
+    } else if (_percentileChanged(oldProps, props)) {
 
-      Object.assign(this.state, {layerData, countRange});
+      this.onPercentileChange();
```
ibgreen: `onPercentileChanged` is quite suggestive of a callback (i.e. calling a function from another component). I'd change this to `percenticlesChanged();` or similar

src/layers/core/grid-layer/grid-layer.js
```diff
@@ -84,21 +123,48 @@ export default class GridLayer extends CompositeLayer {
...
+
+    const colorDomain = this.props.colorDomain || valueDomain;
+    const color = quantizeScale(colorDomain, colorRange, value);
+
+    // if cell value is outside domain, set alpha to 0
+    const alpha = value >= valueDomain[0] && value <= valueDomain[1] ?
+      (Number.isFinite(color[3]) ? color[3] : 255) : 0;
 
-    return quantizeScale(colorDomain, colorRange, cell.count);
+    return color.concat([alpha]);
```
ibgreen: Not sure how frequently this is called, but cloning arrays (Array.concat) seems like a recipe for bad performance. Can `quantizeScale` just take an `alpha` argument?

src/layers/core/hexagon-layer/hexagon-layer.js
```diff
@@ -171,34 +170,34 @@ export default class HexagonLayer extends CompositeLayer {
     };
   }
 
-  _onPercentileChange() {
+  onPercentileChange() {
```
ibgreen: Are these callbacks? If not, I'd avoid the `on...` naming convention.

docs/layers/grid-layer.md
heshan0131: Because the value is only used to calculate color, not elevation, so I thought it would be nice to make it clear in the name

heshan0131: And in the future when I add another feature to calculate elevation based on points value I can use `getElevationValue`

docs/layers/hexagon-layer.md
heshan0131: This reason of doing this is to filter out really big numbers and re calculate color. When colors for are recalculated, the larger bins that fall outside the percentile range will not have accurate color any more, so we have to hide it. Also if we want to display transparence, we have to always enable depthTest

examples/layer-browser/src/examples/core-layers.js
heshan0131: if value is null or undefined, we do not include it when calculate Mean, so we need to get the total number of elements that are not null to divide the sum by...

examples/layer-browser/src/examples/core-layers.js
heshan0131: +1

docs/layers/grid-layer.md
ibgreen: Sounds like we are establishing a new proposed pattern here. Maybe add a section with proposed rules to [API GUIDELINES](https://github.com/uber/deck.gl/blob/master/API-GUIDELINES.md)

docs/layers/hexagon-layer.md
ibgreen: > the larger bins that fall outside the percentile range will not have accurate color any more, so we have to hide i

I thought we could make the filtered out hexes a white transparent color - i.e. they would desaturate and just become ghosts, but could still retain their elevation and give the user a feeling for what he has filtered out.

> Also if we want to display transparency, we have to always enable depthTest. 

I guess this could be an optional feature. A prop...

examples/layer-browser/src/examples/core-layers.js
ibgreen: > if value is null or undefined, we do not include it when calculate Mean, so we need to get the total number of elements that are not null to divide the sum by

Then the comparison should not use `filtered.length`, but just `filtered` right?  Perhaps a good opportunity to add a test case that triggers this bug, since you are working on coverage anyway?

src/layers/core/grid-layer/grid-layer.js
heshan0131: I added comment to layer doc.
If people pass in a new function every time, the layer is just going to sort the bins on every render, which won't result in error showing colors. But if we don't check the function change, and not re sort the bins when it actually changed. we have no way to detect whether some one pass in a new function to calculate color value.

ibgreen: > we have no way to detect whether some one pass in a new function to calculate color value.

Unless we expand `updateTriggers` to cover this case (but that would probably require a discussion and a separate PR).

OK for now, but I think it is a potential API design issue that we have different behavior for accessors and other functions. We might want to call it out in one of our documents (possibly API guidelines if you decided to update there).

PR #622, opened 2017-05-02
Remove all yarn.lock except top one (enables yarn on travis)

ibgreen:
These yarn files are thousands of lines, wasting space in the repo. Keep only top one as it triggers yarn in Travis CI for faster unit test installs.


Apercu:
What do you mean by "space"? Are you referring to the bytes the files are taking, the files in themselves or maybe the diff it creates when we update dependencies?

I understand the concern if it's the diffing issue, although we could treat them as a binary too, but size-wise it doesn't make much sense to me, I think we should instead remove the examples and demo from the files that are published on npm since they shouldn't be in the first place (or is there something I don't know about?) and that would resolve this problem.

What do you think?

ibgreen:
> What do you mean by "space"? Are you referring to the bytes the files are taking, the files in themselves or maybe the diff it creates when we update dependencies?

Yes size, they are 3000 lines long (177K) and seem to change every diff. With time that will take space - for what purpose?

> I understand the concern if it's the diffing issue, although we could treat them as a binary too
, but size-wise it doesn't make much sense to me, I think we should instead remove the examples and demo from the files that are published on npm since they shouldn't be in the first place (or is there something I don't know about?) and that would resolve this problem.

They examples are not published on npm. we only publish library source and dist folders. Its controlled by the "files" field in package.json.

> What do you think?

I just haven't seen the value of committing the yarn.lock files for examples and demo. They might just as well be temporary files, ignored by git.

PR #623, opened 2017-05-02
Add babel-plugin-remove-comments to reduce bundle size

ibgreen:
Also update the tree-shaking example to show graphical view of dependencies.
screen shot 2017-05-01 at 7 05 15 pm

### DIFF
--- a/examples/tree-shaking/webpack.config.js
+++ b/examples/tree-shaking/webpack.config.js
@@ -1,6 +1,9 @@
 // avoid destructuring for older Node version support
 const resolve = require('path').resolve;
 const webpack = require('webpack');
+const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin;
+
+// ...
 
 const CONFIG = {
   entry: {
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -55,6 +55,7 @@
     "babel-core": "^6.22.1",
     "babel-loader": "^6.2.10",
     "babel-plugin-istanbul": "^4.1.1",
+    "babel-plugin-remove-comments": "^2.0.0",
```
Apercu: We can get rid of this plugin and get the same behavior by setting `"comments": "false"` in our babel config.

examples/tree-shaking/webpack.config.js
```diff
@@ -72,5 +75,10 @@ module.exports = env => {
     );
   }
 
-  return CONFIG;
+  CONFIG.plugins.push(
```
Apercu: Maybe extend instead of using the mutating `push`?

package.json
ibgreen: Ahh yes, I see there is a [compact](https://blog.mariusschulz.com/2016/07/12/speeding-up-babel-transpilation-with-compact-mode) option too, worth a try.

PR #625, opened 2017-05-02
Icon layer with rotation

shaojingli:


Pessimistress:
Update layer doc?

src/layers/core/icon-layer/icon-layer.js
```diff
@@ -60,7 +60,8 @@ const defaultProps = {
   getPosition: x => x.position,
   getIcon: x => x.icon,
   getColor: x => x.color || DEFAULT_COLOR,
-  getSize: x => x.size || 1
+  getSize: x => x.size || 1,
+  getAngle: x => x.angle || 0
```
Pessimistress: Looks like it expects angles to be in radian. I personally think degrees is more user friendly (and usually the default unit when not specified). @ericsoco @gnavvy @ibgreen Thoughts?

ibgreen: > Looks like it expects angles to be in radian. I personally think degrees is more user friendly 

- Unfortunately, regardless of what we choose, I think this will always trip people up, unless we call this `angleRadians` or similar. While ugly, it could be the right choice.
- Whatever we choose we need to be consistent across our layers - this needs to go into API design guidelines doc.

src/layers/core/icon-layer/icon-layer-vertex.glsl.js
```diff
@@ -40,15 +41,22 @@ varying float vColorMode;
...
+
 void main(void) {
   vec2 iconSize = instanceIconFrames.zw;
   vec2 iconSize_clipspace = iconSize / viewportSize * 2.0;
   // scale icon height to match instanceSize
   float instanceScale = iconSize.y == 0.0 ? 0.0 : instanceSizes / iconSize.y;
   vec3 center = project_position(instancePositions);
-  vec2 vertex = (positions / 2.0 + instanceOffsets) * iconSize_clipspace *
+  vec2 vertex = rotate_by_angle(positions, instanceAngles);
+  vertex = (vertex / 2.0 + instanceOffsets) * iconSize_clipspace *
```
Pessimistress: Icon should be rotated around the anchor point (specified by `instanceOffsets`). You can test this with the marker icon in layer-browser -- when zooming in and out, the base of the pin should remain at a fixed location.

src/layers/core/icon-layer/icon-layer.js
shaojingli: I have no problem with either. Let's make it degrees for now as suggested by @Pessimistress

ibgreen: > Let's make it degrees **for now** 

As long as we don't publish this in a release (including 4.0 patch), that's OK.

If this is going into official release it needs to be audited like all other props.

ericsoco:
Chiming in late, but I've actually found radians to be a lot more conducive
to the trigonometric operations necessary to do animation/interaction.
Degrees seem more common in design applications.

I would call it `angle` regardless of the unit. Much more common than
"degree".

Whichever you choose, document it and devs will figure it out via a
combination of doc reading and trial and error.

shaojingli:
@tangential

shaojingli:
I plan to communicate the PR better through a better formulated feature proposal document before I land it to the master.

PR #626, opened 2017-05-03
Added create-react-app example

espiegel:
Edit App.js in line 7 to add your mapbox token:
```
// Set your mapbox access token here
const MAPBOX_ACCESS_TOKEN = 'MAPBOX_ACCESS_TOKEN';
```

Then:
```
cd examples/create-react-app
npm install
npm start
```

Then when you navigate to `localhost:3000` you will get the following error in the console:
```
Uncaught SyntaxError: Identifier 'Buffer' has already been declared
    at __webpack_require__ (bootstrap 54f7148…:555)
    at fn (bootstrap 54f7148…:86)
    at Object.<anonymous> (buffer_group.js:4)
    at __webpack_require__ (bootstrap 54f7148…:555)
    at fn (bootstrap 54f7148…:86)
    at Object.<anonymous> (bucket.js:4)
    at __webpack_require__ (bootstrap 54f7148…:555)
    at fn (bootstrap 54f7148…:86)
    at Object.<anonymous> (circle_bucket.js:3)
    at __webpack_require__ (bootstrap 54f7148…:555)
```


Apercu:
Since it's not working yet, there's not really a point to add it to our examples. We have to figure out a way to resolve the build error

ibgreen:
> Since it's not working yet, there's not really a point to add it to our examples. We have to figure out a way to resolve the build error

Agreed. Until we do, it might be worth mentioning in the readme that this is not possible today since create-react-app seems to be growing in popularity.

PR #629, opened 2017-05-04
fix(seer): improve invalidate and log

Apercu:
Only invalidate the `data` if it has been set as an override.
Also log a limit of 20 data items instead of the full list.

### DIFF
--- a/src/debug/seer-integration.js
+++ b/src/debug/seer-integration.js
@@ -96,7 +96,7 @@ const transformData = data => {
 };
 
 /**
- * Log a layer properties to Seer
+ * Log layer's properties to Seer
  */
 export const logLayer = layer => {
   if (!window.__SEER_INITIALIZED__) {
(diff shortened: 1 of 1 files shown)


src/debug/seer-integration.js
```diff
@@ -71,6 +72,23 @@ export const layerEditListener = cb => {
   });
 };
 
+const dataBlackList = ['zoomLevels'];
+
+const transformData = data => {
```
ibgreen: Function header please, explain what kind of transformation we are doing

src/debug/seer-integration.js
```diff
@@ -71,6 +72,23 @@ export const layerEditListener = cb => {
...
+    if (dataBlackList.includes(key)) {
+      return acc;
+    }
+    acc[key] = item[key];
+    return acc;
+  }, {}));
+};
+
 /**
  * Log a layer properties to Seer
```
ibgreen: layer => layer's

PR #631, opened 2017-05-05
Optimize encodePickingColor 

shaojingli:
by replacing the naive math with bit-wise operations

### DIFF
--- a/test/bench/index.js
+++ b/test/bench/index.js
@@ -48,6 +48,7 @@ const VIEWPORT_PARAMS = {
 };
 
 let testIdx = 0;
+const testLayer = new ScatterplotLayer({data: data.points});
 
 // add tests
 suite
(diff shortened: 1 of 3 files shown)


shaojingli:
Old impl with naive math: 
encoding picking color x 20,022,207 ops/sec ±1.34% (63 runs sampled)

New impl with bitwise ops:
encoding picking color x 52,929,168 ops/sec ±3.09% (61 runs sampled)

ibgreen:
> 20,022,207 ops/sec -> 52,929,168 ops/sec 

Thanks. While it is a very nice improvement indeed, it sounds like we want to understand why this is called so often - to show up on the perf radar this must have been called millions of times which doesn't sound right.

test/bench/index.js
```diff
@@ -130,6 +132,19 @@ suite
 //   const layer = new ScatterplotLayer({data: data.points});
 //   testInitializeLayer({layer});
 // })
+// .add('encoding picking color', () => {
```
ericsoco: Why not leave this in the test suite?

ericsoco:
@ib I'm not sure how the benchmark test is set up, so perhaps irrelevant. But -- remember the discovery @Pessimistress and I noted last week that picking currently runs every frame, regardless of if mouse event callbacks are declared. Don't think that's related here but JIC.

PR #636, opened 2017-05-08
Modularize support for raw and gestural input events

ericsoco:
First pass at new EventManager.
Provides single API for both raw and gestural events; leverages hammer.js for both.
Replaces use of luma.gl's addEvents.
Does not yet reimplement support for dragging (i.e. dragging is broken as of this commit).

The plan is to implement this in `react-map-gl` first, but I built out the proof-of-concept in deck.gl just to get started.


src/utils/event-manager.js
```diff
@@ -0,0 +1,137 @@
...
+    };
+  }
+}
+
+export function createEventManager(element, options = {}) {
+  // TODO: should this be a Singleton?
+  // conversely, is actually a good reason to use the Factory pattern here,
+  // or should we allow direct instantiation?
+  options = Object.assign({}, DEFAULT_OPTIONS, options);
+  return new EventManager(element, options);
```
shaojingli: I don't really understand the reason having a separate function for creating EventManager here either.

yarn.lock
```diff
@@ -24,8 +24,8 @@ acorn-dynamic-import@^2.0.0:
     acorn "^4.0.3"
```
shaojingli: Please remove yarn.lock file from your diff

ericsoco:
Some bits for discussion:

1. This PR offers a single interface for subscribing to both "raw" input events (e.g. `mousemove`, `click`) and "gestural" events (those with semantic meaning, like `panstart` and `tap`). For that reason and for cross-browser compatibility, I'm wrapping returned events to ensure a consistent event object shape. [The wrapper is currently pretty minimal](https://github.com/uber/deck.gl/pull/636/files#diff-c368acec23e457d00361634bb6ab399dR111); we should consider using an out-of-the-box option.

2. Hammer.js' setup for adding support for events that it doesn't already have is not very clean. I'm currently extending `PointerInputEvent` following [@Pessimistress' example](https://github.com/uber/react-map-gl/compare/hammer?expand=1), but this will fail on environments without support for [pointer events](https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events). Also, I'm adding support for mouse wheel and for `mousemove`/`pointermove` while the pointer is not down, and broke those out into two separate classes. Both of these issues show that a composition setup would work a lot better than the inheritance setup hammer.js uses. We should think through the best way to handle this: reimplementing hammer.js logic in our own code, trying to get PRs into hammer.js, something else.

src/utils/event-manager.js
ericsoco: @Pessimistress had a conversation about how a standalone `event-manager` would work when it's required by multiple packages, e.g. `deck.gl` and `react-map-gl` both. I implemented it (and `gesture-manager`) as a Factory to allow us to turn it into a Singleton if necessary, or enable any other communication between instances to avoid handler collisions. Haven't fully thought through this yet.

ibgreen:
Thanks for putting this up. Not a review, just some idle thoughts:

* Looks like we will end up with a lot of event handling code. In spite of my initial opinion to the contrary, perhaps moving this out of deck.gl would be appropriate after all. It would be preferable in some ways if we could keep deck.gl fairly neutral on the event handling topic.

* Another potential question is: Should hammerjs be a dependency of the core deck.gl library itself, or is hammerjs + additional code something that is brought into the examples via a package dependency. Always nice to have loose coupling but that might be impractical.

shaojingli:
> Looks like we will end up with a lot of event handling code. In spite of my initial opinion to the contrary, perhaps moving this out of deck.gl would be appropriate after all. It would be preferable in some ways if we could keep deck.gl fairly neutral on the event handling topic.

I also think we should do that. But before react-map-gl exports its EventManager (which I assume will be implemented as the same here), I can leave with this piece of code for now.

ericsoco:
Hm, yeah....both raw and gestural events here use hammer.js, so it probably
makes sense for hammer to be a dep of this as a separate package. Don't see
a simple/clean way to break hammer.js out to specify as a dep of deck.gl /
react-map-gl and to compose it into the future event-manager package given
that.

src/react/deckgl.js
```diff
@@ -126,6 +128,18 @@ export default class DeckGL extends React.Component {
       );
     }
 
+    // TODO: add handlers on demand at runtime, not all at once on init
+    this.eventManager = createEventManager(canvas, {})
+    .on({
+      click: this._onClick,
```
shaojingli: I assume in the future, we will be generating user intentions on higher level of abstraction instead of calling callbacks at "mousemove" level, right? I mean, deck.gl will distinguishing events that can handle by itself (events lead to viewport change) from those that need to be handled by the app (e. g. events lead to data change). Is my understanding right?

ericsoco: Per the conversation last week about supporting something like Mapbox's `queryRenderedFeatures`, I think we'll need to support both. E.g. an app dev might want an `onHover` callback that lets the app know when the hover target changes, but might also want to implement marquee selection manually with an `onMouseMove` handler that queries features within the marquee.

shaojingli: Yeah... app will want to when the hover target change and / or when data changed, etc... but is probably don't want to know when camera / viewport should be changed. 

This is just a discussion, not a request to change the code. I was just wondering if it's reasonable to provide the callback API at higher level... say instead of giving them onClick(), we should give them onItemsPicked() and onItemsPicked() will get called when the user tries to pick items in various ways, clicking, dragging, keyboard selection, etc..

ericsoco: @shaojingli and I discussed. He'll add detail to the RFC, but to recap: here's a suggestion for rethinking how application developers may handle user input:

1. Selection-driven: App dev adds selection handlers that return elements selected, deselected, or soft-selected (e.g. hovered). This is a more declarative style of handling user input.

2. Traditional event handling: App dev adds input event handlers that return screen or model-space coordinates, and then uses those coords to query features. This is a more imperative style of handling user input.

If we add support for selection events as outlined in 1. above, we should probably still maintain the traditional event handling paradigm for backwards compatibility and because it's a more familiar paradigm. But we might consider shifting focus toward e.g. `onItemsPicked` and away from `onClick`.

src/utils/gesture-manager.js
```diff
@@ -0,0 +1,120 @@
...
+
+  on(event, handler) {
+    const recognizerEvent = this._getRecognizerEvent(event);
+    if (!recognizerEvent) {
+      return;
+    }
+
+    // add recognizer for this event if not already added.
+    if (!this.manager.get(recognizerEvent)) {
+      this.manager.add(RECOGNIZERS[recognizerEvent]);
```
Pessimistress: I believe the order of recognizers matter unless you use
[recognizer.recognizeWith](http://hammerjs.github.io/recognize-with/) and
[recognizer.requireFailure](http://hammerjs.github.io/require-failure/). Even then it would be nice to fire events with a consistent order.

Alternatively we can create the Manager instance with all recognizers and `enable: false`. Then just [toggle it on](http://hammerjs.github.io/toggle-recognizer/) here.

src/utils/event-manager.js
```diff
@@ -0,0 +1,137 @@
...
+
+    // If specified, use GestureManager for raw event handling
+    // as well as for gesture recognition.
+    if (options.handleRawEventsWithGestureManager) {
+      this._onRawInput = this._onRawInput.bind(this);
+      this.gestureManager.startHandlingRawEvents(this._onRawInput);
+    }
+  }
+
+  on(event, handler) {
```
Pessimistress: Is there a reason we need to implement our own event handling? Since we're creating a hammer manager already, we could register and remove handlers with `manager.on` and `manager.off`, then call `manager.emit` for our own events.

src/utils/gesture-manager.js
ericsoco: Toggling sounds like the way to go, didn't realize that was an option.

src/utils/event-manager.js
ericsoco: It's the `hammer.input` thing -- we need to manage our own handlers because hammer doesn't maintain a one-to-one mapping of `eventName: eventHandler` for the events we pull out of `hammer.input` events. See: `_onRawInput`.

ericsoco: That said, it looks like I'm storing handlers here for the events that _are_ mapped one-to-one by hammer (those present in `EVENT_RECOGNIZER_MAP`, and that's probably unnecessary...

shaojingli:
@ibgreen we'd better name it wrench.js ... nail.js will be hammered by hammer.js.... 😝

ericsoco:
Wrong Hammer metaphor, should be more like boyzIImen.js. See [here](https://en.wikipedia.org/wiki/MC_Hammer) for more inspiration.

ericsoco:
A serious answer, tho...

I imagine `event-manager` and `gesture-manager` to be the standalone package used by both `deck.gl` and `react-map`gl`. So, together they add the functionality missing from hammer.js and provide a unified API for listening to events. Due to hammer's lack of an API for non-gestural events (requirement of listening for `'hammer.input'` and switching on `event.type`), I don't see a clean way to unify the API (to allow consumers to just `.on/off(eventName, eventHandler)` without wrapping...but let's discuss.

src/react/deckgl.js
```diff
@@ -126,56 +128,55 @@ export default class DeckGL extends React.Component {
...
-      cachePosition: false,
-      centerOrigin: false,
-      onClick: this._onClick,
-      onMouseMove: this._onMouseMove,
-      onDragStart: this._onDragEvent,
-      onDragMove: this._onDragEvent,
-      onDragEnd: this._onDragEvent,
-      onDragCancel: this._onDragCancel
+    // TODO: add handlers on demand at runtime, not all at once on init
+    this.eventManager = createEventManager(canvas, {})
```
ibgreen: Some API design comments:
* Not sure I see a need for the `createEventManager` wrapper. Shouldn't we just export the `EventManager` class?
* Especially since it is a class with methods that the user is expected to call, we'll want to document it as a class next to our other classes.
* Also seems taking the first set of handlers directly in the constructor would be nice - perhaps then the `on`/`off` methods might not be needed for casual users...
```
new `EventManager(canvas, {
   click: this._onClick,
   ...
})
```

src/utils/pointer-move-event-input.js
```diff
@@ -0,0 +1,40 @@
+import {PointerEventInput} from 'hammerjs';
+
+const POINTER_MOVE = 'pointermove';
+
+// Copied from Hammer.js' pointerevent.js
+const IE10_POINTER_TYPE_ENUM = {
```
ibgreen: Question for @shaojingli - Should we draw the line at IE11? I am not convinced that IE10 reliably supports WebGL anyway. I think we never published our supported platforms doc.

src/utils/event-manager.js
```diff
@@ -0,0 +1,226 @@
...
+};
+
+/**
+ * Single API for subscribing to events about both
+ * basic input events (e.g. 'mousemove', 'touchstart', 'wheel')
+ * and gestural input (e.g. 'click', 'tap', 'panstart').
+ * Delegates event registration and handling to Hammer.js.
+ */
+class EventManager {
+  constructor(element, options) {
```
ibgreen: This file would be much easier to review if there was a README.md containing an initial version of the documentation for this class, documenting options etc for this class.

src/utils/pointer-move-event-input.js
shaojingli: @ibgreen @ericsoco Yes, we should only support IE11. IE10 doesn't support WebGL.

ericsoco: Great, that will clean things up.

src/react/deckgl.js
ericsoco: Passing a map into the constructor is a good idea and simple to implement.

src/utils/event-manager.js
```diff
@@ -0,0 +1,226 @@
...
+        // Alias to a recognized gesture as necessary.
+        const gestureEventAlias = GESTURE_EVENT_ALIASES[recognizerEvent];
+        if (gestureEventAlias) {
+          this.gestureAliases[event] = this._wrapAliasedGestureHandler(gestureEventAlias, handler);
+          this.manager.on(gestureEventAlias, this.gestureAliases[event]);
+        }
+      }
+
+      // Register event handler.
+      this.manager.on(recognizerEvent || event, handler);
```
Pessimistress: The event should be registered using user-supplied `event` name.

src/utils/event-manager.js
```diff
@@ -0,0 +1,226 @@
...
+      if (recognizerEvent) {
+        // Add recognizer for this event if not already added.
+        if (!this.manager.get(recognizerEvent)) {
+          this.manager.add(RECOGNIZERS[recognizerEvent]);
+        }
+
+        // Alias to a recognized gesture as necessary.
+        const gestureEventAlias = GESTURE_EVENT_ALIASES[recognizerEvent];
+        if (gestureEventAlias) {
+          this.gestureAliases[event] = this._wrapAliasedGestureHandler(gestureEventAlias, handler);
```
Pessimistress: Should check if a wrapper is already created. Otherwise the same event will be emitted multiple times.

Also arguments supplied to `_wrapAliasedGestureHandler` does not match the method (should be `event`.

ericsoco:
I brought in @Pessimistress' updates from `react-map-gl` and addressed other comments. I also noticed a click delay due to `doubletap` handling and eliminated it (except when an application handles both single and double tap, in which case we do the best we can but there's still a slight delay).

I think this is just about ready to land now; I see the following items as followups or, possibly, blockers to landing:
- [ ] Currently, only one handler for an aliased event (only an issue for `'click'` for the foreseeable future) can exist at a time. Local state should be expanded to support an array of handlers against each aliased event.
- [ ] Allow overriding recognizers in constructor.
- [ ] Unit tests.

Once landed, the following events-related work remains:
- [ ] Fix [picking-on-every-event bug](https://github.com/uber/deck.gl/issues/634).
- [ ] Move event handling from `deckgl.js` React wrapper into `layer-manager` so that non-React deck.gl consumers get event handling out of the box.
- [ ] This PR breaks drag event handling, as drag events are not supported by Hammer.js out-of-the-box. That's ok, because drag events are [currently implemented incorrectly anyway](https://github.com/uber/deck.gl/issues/648). So, drag event handling needs to be refactored.

src/utils/events/event-manager.js
```diff
@@ -0,0 +1,170 @@
...
+   */
+  off(event, handler) {
+    if (typeof event === 'string') {
+      // Clean up aliased gesture handler as necessary.
+      const recognizerEvent = EVENT_RECOGNIZER_MAP[event];
+      if (recognizerEvent) {
+        const eventAlias = GESTURE_EVENT_ALIASES[event];
+        if (eventAlias && this.aliasedEventHandlers[event]) {
+          this.manager.off(eventAlias, this.aliasedEventHandlers[event]);
+          delete this.aliasedEventHandlers[event];
```
Pessimistress: This may cause incorrect removal of aliased handlers if the user registers multiple handlers for `click`.

ericsoco: Multiple `click` handlers will currently not work correctly as only one handler per alias can exist right now. See my TODO above ;)

src/utils/events/event-manager.js
```diff
@@ -0,0 +1,170 @@
...
+   * The single tap recognizer must wait for the double tap recognizer
+   * to fail before resolving. Note that enabling both incurs a slight delay
+   * on tap/click handler resolution.
+   */
+  _reconcileSingleAndDoubleTap(event) {
+    if (event === 'tap' || event === 'doubletap') {
+      const singletapRecognizer = this.manager.get('tap');
+      const doubletapRecognizer = this.manager.get('doubletap');
+      if (singletapRecognizer.options.enable && doubletapRecognizer.options.enable) {
+        singletapRecognizer.requireFailure('doubletap');
```
Pessimistress: You can add this to the `RECOGNIZERS` definition. http://hammerjs.github.io/jsdoc/manager.js.html#line26

src/utils/events/wheel-input.js
```diff
@@ -0,0 +1,133 @@
...
+    let value = event.deltaY;
+    // Firefox doubles the values on retina screens...
+    if (firefox && event.deltaMode === window.WheelEvent.DOM_DELTA_PIXEL) {
+      value /= window.devicePixelRatio;
+    }
+    if (event.deltaMode === window.WheelEvent.DOM_DELTA_LINE) {
+      value *= WHEEL_DELTA_PER_LINE;
+    }
+
+    let type = this._state.mouseWheelType;
```
ibgreen: Nit: Why call these different things? I think I'd prefer seeing
```
const {type, timeout, lastValue, time} = this.state;
```
Less "aliasing" and less lines of code.

src/utils/events/wheel-input.js
```diff
@@ -0,0 +1,133 @@
...
+      value = value * SHIFT_MULTIPLIER;
+    }
+
+    // Only fire the callback if we actually know what type of scrolling device
+    // the user uses.
+    if (type) {
+      this._wheel(event, -value, pos);
+    }
+
+    this._setState({
```
ibgreen: Nit: I actually think this `setState` function might not be needed and could even be a little confusing here. It doesn't add much to the a small class like this and for a moment makes me feel that I am reading a React component or a deck.gl layer. 

It is probably a matter of taste, but I personally try to carefully match the amount of "machinery" in a certain piece of code to be in proportion to the complexity of the code in question.

src/utils/events/move-input.js
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Hammer.js swallows 'move' events (for pointer/touch/mouse)
+ * when the pointer is not down. This class sets up a handler
+ * specifically for these events to work around this limitation.
+ * Note that this could be extended to more intelligently handle
+ * move events across input types, e.g. storing multiple simultaneous
+ * pointer/touch events, calculating speed/direction, etc.
+ */
+export default class MoveInput {
```
ibgreen: Nit: This class really doesn't do much. It seems to be there mostly to match `WheelInput`. I wonder if they should just be combined? Or is there an intention to add more of these?

src/utils/events/wheel-input.js
```diff
@@ -0,0 +1,133 @@
...
+
+  constructor(element, callback) {
+    this.element = element;
+    this.callback = callback;
+
+    this.handler = this.handler.bind(this);
+
+    WHEEL_EVENTS.forEach(eventName => element.addEventListener(eventName, this.handler));
+
+    this._state = {
```
ericsoco: TBH I just copied this code from `react-map-gl` and haven't really given it much thought. @Pessimistress any thoughts on @ibgreen's suggestions?

src/utils/events/move-input.js
ericsoco: They could be combined, yes.

ericsoco: Actually, let's hold on combining until we've had a chance for this to land and simmer for a bit. Might turn out we've missed some other things that Hammer doesn't support, or have some other events that need special cases, and there may be a pattern to handle them all together.

Also, I just cleaned up `wheel-input` quite aggressively and would like to keep that separate for now since it has a significant amount of state and logic on its own; will be easier to identify any problems in that refactor if that code stays by itself in its own module.

src/utils/events/event-manager.js
ericsoco: My concern was that requiring failure of a disabled recognizer would cause the single tap to never fire. Not clear on a skim of hammer's source if this is the case, or if hammer is smart enough to skip the disabled recognizer in the `requireFail` check...

Though I _think_ [this](https://github.com/hammerjs/hammer.js/blob/master/src/recognizerjs/recognizer-constructor.js#L233) means that if the fail-required recognizer hasn't started running yet (which is the case when `enable: false`) it will be skipped.

Which means you're right, I can just do this via config. Let me test and circle back.

PR #641, opened 2017-05-12
Add `pickingRadius` prop

Pessimistress:
- Add `pickingRadius` prop to `DeckGL` component. The old picking behavior is equivalent to `pickingRadius: 0`. The closest object to the pointer within `pickingRadius` pixels is picked.

### DIFF
--- a/examples/layer-browser/src/app.js
+++ b/examples/layer-browser/src/app.js
@@ -102,6 +102,12 @@ class App extends PureComponent {
     setImmutableDataSamples(settings.immutable);
   }
 
+  _onQueryFeatures() {
+    const {deckgl} = this.refs;
+    const {width, height} = this.state;
+    console.log(deckgl.elementsInScope([0, 0], [width, height])); // eslint-disable-line
+  }
+
   _onHover(info) {
     this.setState({hoveredItem: info});
   }
@@ -186,6 +192,7 @@ class App extends PureComponent {
 
         <DeckGL
           debug
+          ref="deckgl"
           id="default-deckgl-overlay"
           width={width} height={height}
           {...mapViewState}
@@ -220,6 +227,10 @@ class App extends PureComponent {
         { this._renderMap() }
         { !MAPBOX_ACCESS_TOKEN && this._renderNoTokenWarning() }
         <div id="control-panel">
+          <div>
+            <h4 />
+            <button onClick={this._onQueryFeatures} >Query Rendered Features</button>
+          </div>
           <LayerControls
             title="Composite Settings"
             settings={settings}
--- a/examples/layer-browser/src/app.js
+++ b/examples/layer-browser/src/app.js
@@ -42,7 +42,8 @@ class App extends PureComponent {
         // immutable: false,
         // Effects are experimental for now. Will be enabled in the future
         // effects: false,
-        separation: 0
+        separation: 0,
+        pickingRadius: 0
         // the rotation controls works only for layers in
         // meter offset projection mode. They are commented out
         // here since layer browser currently only have one layer
@@ -100,12 +101,6 @@ class App extends PureComponent {
   _onUpdateContainerSettings(settings) {
     this.setState({settings});
     setImmutableDataSamples(settings.immutable);
-  }
-
-  _onQueryFeatures() {
-    const {deckgl} = this.refs;
-    const {width, height} = this.state;
-    console.log(deckgl.elementsInScope([0, 0], [width, height])); // eslint-disable-line
   }
 
   _onHover(info) {
@@ -181,7 +176,7 @@ class App extends PureComponent {
   }
 
   _renderMap() {
-    const {width, height, mapViewState, settings: {effects}} = this.state;
+    const {width, height, mapViewState, settings: {effects, pickingRadius}} = this.state;
     return (
       <MapboxGLMap
         mapboxApiAccessToken={MAPBOX_ACCESS_TOKEN || 'no_token'}
@@ -192,10 +187,10 @@ class App extends PureComponent {
 
         <DeckGL
           debug
-          ref="deckgl"
           id="default-deckgl-overlay"
           width={width} height={height}
           {...mapViewState}
+          pickingRadius={pickingRadius}
           onWebGLInitialized={ this._onWebGLInitialized }
           onLayerHover={ this._onHover }
           onLayerClick={ this._onClick }
@@ -227,10 +222,6 @@ class App extends PureComponent {
         { this._renderMap() }
         { !MAPBOX_ACCESS_TOKEN && this._renderNoTokenWarning() }
         <div id="control-panel">
-          <div>
-            <h4 />
-            <button onClick={this._onQueryFeatures} >Query Rendered Features</button>
-          </div>
           <LayerControls
             title="Composite Settings"
             settings={settings}
--- a/src/lib/draw-and-pick.js
+++ b/src/lib/draw-and-pick.js
@@ -96,7 +96,7 @@ export function pickLayers(gl, {
 
   } else {
     // For all other events, run picking process normally.
-    const pickResult = pickFromBuffer(gl, {
+    const pickInfo = pickFromBuffer(gl, {
       layers,
       pickingFBO,
       deviceX,
@@ -104,9 +104,9 @@ export function pickLayers(gl, {
       deviceRadius
     });
 
-    pickedColor = pickResult.color;
-    pickedLayer = pickResult.layer;
-    pickedObjectIndex = pickResult.objectIndex;
+    pickedColor = pickInfo.pickedColor;
+    pickedLayer = pickInfo.pickedLayer;
+    pickedObjectIndex = pickInfo.pickedObjectIndex;
     affectedLayers = pickedLayer ? [pickedLayer] : [];
 
     if (mode === 'hover') {
@@ -193,65 +193,10 @@ export function pickLayers(gl, {
 
   return unhandledPickInfos;
 }
-/* eslint-enable max-depth, max-statements */
-
-export function pickLayersByBoundingBox(gl, {
-  layers,
-  pickingFBO,
-  boundingBox: {x, y, width, height},
-  viewport
-}) {
-
-  // Convert from canvas top-left to WebGL bottom-left coordinates
-  // And compensate for pixelRatio
-  const pixelRatio = typeof window !== 'undefined' ?
-    window.devicePixelRatio : 1;
-  const deviceBox = {
-    x: Math.round(x * pixelRatio),
-    y: Math.round(gl.canvas.height - (y + height) * pixelRatio),
-    width: Math.round(width * pixelRatio),
-    height: Math.round(height * pixelRatio)
-  };
-
-  // For all other events, run picking process normally.
-  const pickResults = pickFromBufferByBoundingBox(gl, {
-    layers,
-    pickingFBO,
-    deviceBox
-  });
-
-  const baseInfo = createInfo([x, y], viewport);
-  baseInfo.pixelRatio = pixelRatio;
-
-  // Convert each result to pickingInfo object
-  const infos = {};
-
-  pickResults.filter(Boolean)
-  .forEach(({color, layer, objectIndex}) => {
-    // Get unique objects
-    const objectId = `${layer.id}:${objectIndex}`;
-
-    if (!infos[objectId]) {
-      const info = Object.assign({}, baseInfo, {
-        color,
-        index: objectIndex,
-        picked: true
-      });
-
-      infos[objectId] = getLayerPickingInfo({
-        layer,
-        info,
-        mode: 'query'
-      });
-    }
-  });
-
-  return Object.values(infos);
-}
 
 /**
  * Pick at a specified pixel with a tolerance radius
- * Returns the closest object to the pixel in shape `{color, layer, objectIndex}`
+ * Returns the closest object to the pixel in shape `{pickedColor, pickedLayer, pickedObjectIndex}`
  */
 function pickFromBuffer(gl, {
   layers,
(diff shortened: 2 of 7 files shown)


examples/layer-browser/src/app.js
```diff
@@ -102,6 +102,12 @@ class App extends PureComponent {
     setImmutableDataSamples(settings.immutable);
   }
 
+  _onQueryFeatures() {
```
ibgreen: Should we call it onQueryElements or onQueryObjects for consistency? We should review or documentation of the `data` prop and use the same nomenclature for the individual objects...

examples/layer-browser/src/app.js
```diff
@@ -220,6 +227,10 @@ class App extends PureComponent {
         { this._renderMap() }
         { !MAPBOX_ACCESS_TOKEN && this._renderNoTokenWarning() }
         <div id="control-panel">
+          <div>
+            <h4 />
+            <button onClick={this._onQueryFeatures} >Query Rendered Features</button>
```
ibgreen: Features => Elements/Objects etc

src/lib/draw-and-pick.js
```diff
@@ -206,11 +195,122 @@ export function pickLayers(gl, {
 }
 /* eslint-enable max-depth, max-statements */
 
+export function pickLayersByBoundingBox(gl, {
```
ibgreen: See my concern in the RFC. This will not catch hidden objects which I think it expected in an area selection operation. But maybe I am missing the use case. So depending on what we decide, we should ask ourselves if it is worth including a possibly flawed picking function.

shaojingli:
> Do we need to rerender the picking buffer on every mouse event and API call? Shall we add a needRedraw flag for the picking buffer?

We should not rerender picking buffer on every mouse event.

src/lib/draw-and-pick.js
```diff
@@ -206,11 +195,122 @@ export function pickLayers(gl, {
...
+    y,
+    width: Math.min(pickingFBO.width, deviceX + deviceRadius) - x + 1,
+    height: Math.min(pickingFBO.height, deviceY + deviceRadius) - y + 1
+  };
+
+  const pickResults = pickFromBufferByBoundingBox(gl, {layers, pickingFBO, deviceBox});
+
+  // Traverse all pixels in picking results and find the one closest to the supplied
+  // [deviceX, deviceY]
+  let minDistanceToCenter = deviceRadius;
```
ibgreen: Would be nice to have a bench that tested the perf on this on a relatively big area.

src/lib/draw-and-pick.js
```diff
@@ -206,11 +195,122 @@ export function pickLayers(gl, {
...
+  let closestResultToCenter = null;
+  /* eslint-disable max-depth */
+  for (let col = 0; col < deviceBox.width; col++) {
+    for (let row = 0; row < deviceBox.height; row++) {
+      const pickResult = pickResults[row * deviceBox.width + col];
+      if (pickResult) {
+        const dx = deviceBox.x + col - deviceX;
+        const dy = deviceBox.y + row - deviceY;
+        const d = Math.sqrt(dx * dx + dy * dy);
```
ibgreen: Often possible and cheaper to not draw roots and just compare squares. E.g. calling the variable `minSquaredDistance`.

src/react/deckgl.js
```diff
@@ -146,7 +159,12 @@ export default class DeckGL extends React.Component {
       return;
     }
     const {event: {offsetX: x, offsetY: y}} = event;
-    const selectedInfos = this.layerManager.pickLayer({x, y, mode: 'click'});
+    const selectedInfos = this.layerManager.pickLayer({
```
ibgreen: Assign radius before the call and keep as one line?

src/react/deckgl.js
```diff
@@ -161,7 +179,12 @@ export default class DeckGL extends React.Component {
       return;
     }
     const {event: {offsetX: x, offsetY: y}} = event;
-    const selectedInfos = this.layerManager.pickLayer({x, y, mode: 'hover'});
+    const selectedInfos = this.layerManager.pickLayer({
```
ibgreen: ditto

src/react/deckgl.js
```diff
@@ -201,7 +224,12 @@ export default class DeckGL extends React.Component {
     }
 
     if (mode) {
-      const selectedInfos = this.layerManager.pickLayer({x, y, mode});
+      const selectedInfos = this.layerManager.pickLayer({
```
ibgreen: ditto

Pessimistress:
> We should not rerender picking buffer on every mouse event
We currently do so for hover and click. Hover does update a uniform `selectedPickingColor`, which will cause a redraw (though not making any difference for picking in any of the core layers).

src/react/deckgl.js
```diff
@@ -78,6 +80,17 @@ export default class DeckGL extends React.Component {
     this._updateLayers(nextProps);
   }
 
+  /* Public API */
+  elementsInScope(topLeft, bottomRight) {
```
ibgreen: To support the review, write up the doc pages for the public APIs, either as temp comment here or in the doc structure.

src/lib/draw-and-pick.js
```diff
@@ -95,16 +96,18 @@ export function pickLayers(gl, {
...
     });
-    pickedColor = pickInfo.pickedColor;
-    pickedLayer = pickInfo.pickedLayer;
-    pickedObjectIndex = pickInfo.pickedObjectIndex;
-    affectedLayers = pickInfo.affectedLayers;
+
+    pickedColor = pickResult.color;
+    pickedLayer = pickResult.layer;
+    pickedObjectIndex = pickResult.objectIndex;
+    affectedLayers = pickedLayer ? [pickedLayer] : [];
```
shaojingli: any reason we move this logical out of pickFromBuffer() function now?

src/lib/draw-and-pick.js
```diff
@@ -251,21 +351,33 @@ function pickFromBuffer(gl, {
...
-      pickedColor,
-      pickedLayer,
-      pickedObjectIndex: pickedLayer ? pickedLayer.decodePickingColor(pickedColor) : -1,
-      affectedLayers: pickedLayer ? [pickedLayer] : []
-    };
+    const pickResults = [];
+
+    for (let i = 0; i < pickedColors.length; i += 4) {
+      // Decode picked color
+      const pickedLayerIndex = pickedColors[i + 3] - 1;
```
ibgreen: Can we update the utilities to handle this, e.g:
```
const {pickedLayerIndex, pickedColor} = decodePickingColor(pickedColors[...])
```

src/lib/draw-and-pick.js
Pessimistress: The only logic that has been moved out is `affectedLayers`. `pickFromBuffer()` and `pickFromBufferByBoundingBox()` now return similarly shaped objects.

src/lib/layer-manager.js
```diff
@@ -173,6 +170,26 @@ export default class LayerManager {
     });
   }
 
+  pickLayerByBoundingBox({x, y, width, height}) {
```
shaojingli: I assume this function serves as an extension of the pickLayer() function right above it and it will replace pickLayer(), right?

Pessimistress: No. `pickLayer()` is used by mouse event picking (returns a single info object) and `pickLayerByBoundingBox()` is used by the query API (returns an array of info objects).

src/lib/draw-and-pick.js
```diff
@@ -206,11 +195,122 @@ export function pickLayers(gl, {
...
+  let closestResultToCenter = null;
+  /* eslint-disable max-depth */
+  for (let col = 0; col < deviceBox.width; col++) {
+    for (let row = 0; row < deviceBox.height; row++) {
+      const pickResult = pickResults[row * deviceBox.width + col];
+      if (pickResult) {
+        const dx = deviceBox.x + col - deviceX;
+        const dy = deviceBox.y + row - deviceY;
+        const d = Math.sqrt(dx * dx + dy * dy);
+        if (d <= minDistanceToCenter) {
```
shaojingli: This comparison should be implemented using distance squared, instead of distance to save that super expansive sqrt operations

shaojingli: distance comparison is better to be implemented using squared values to save on those super expansive Math.sqrt() operation

src/lib/draw-and-pick.js
```diff
@@ -206,11 +195,122 @@ export function pickLayers(gl, {
...
+          minDistanceToCenter = d;
+          closestResultToCenter = pickResult;
+          pickResult.x = deviceBox.x + col;
```
shaojingli: x, y values are the location of the closest pixel of the picked element not the location of mouse pointer. We probably want to have better names of these two variables to help the user and also documented.

src/lib/draw-and-pick.js
```diff
@@ -219,7 +319,7 @@ function pickFromBuffer(gl, {
   return glContextWithState(gl, {
     frameBuffer: pickingFBO,
     framebuffer: pickingFBO,
-    scissorTest: {x: deviceX, y: deviceY, w: 1, h: 1}
+    scissorTest: {x, y, w: width, h: height}
```
shaojingli: So are we going to provide default values for x and y or not?

src/lib/layer-manager.js
shaojingli: Yeah, I knew that. pickLayer() and pickLayerByBoundingBox() share a lot of code, so do pickLayers() and pickLayersByBoundingBox() in draw-and-pick.js.

I'm wondering if mouse event picking can also be implemented via pickLayerByBoundingBox() with width and height both set to 1. We'll need to handle picking "mode" better.

examples/layer-browser/src/app.js
ericsoco: This isn't a handler, is it? I thought the intent here was to respond to an input event by calling `queryFeatures` (/elements/objects) manually from within the input event handler. We agreed this path would serve as a complement to the `onClick` / `onHover` callback-style interaction that we already have, to allow app devs more flexibility with how they handle input events.

src/lib/draw-and-pick.js
```diff
@@ -71,8 +71,9 @@ export function pickLayers(gl, {
   // And compensate for pixelRatio
   const pixelRatio = typeof window !== 'undefined' ?
     window.devicePixelRatio : 1;
-  const deviceX = x * pixelRatio;
-  const deviceY = gl.canvas.height - y * pixelRatio;
+  const deviceX = Math.round(x * pixelRatio);
+  const deviceY = Math.round(gl.canvas.height - y * pixelRatio);
+  const deviceRadius = Math.round(radius * pixelRatio);
```
ericsoco: In my local pass at this feature (you made it further than I did), I added `deviceW` / `deviceH`, to enable marquee selection. Is there any benefit to a circular selection instead of a rectangular one? What's the specific use case for circular other than "fat-finger"?

src/lib/draw-and-pick.js
ericsoco: Yep, I already moved this out in the drag events PR because we don't always pick -- drag events that indicate the continuation of a drag event that started on a specific element don't need to re-pick. In that case, the user is assumed to be dragging an element around and so the element picked on `dragstart` is returned without picking.

src/lib/draw-and-pick.js
ericsoco: @ibgreen sorry, I'm having trouble finding this in the RFC. What does it mean for an element to be "hidden"? We should try to follow the DOM model for this -- if it's "turned off" / "hidden", it's not selectable. If its opacity is 0, it is selectable. I'm not clear on exactly how this maps to deck.gl (or even if it does), nor am I clear on how we'd handle the `opacity === 0` case.

ibgreen: > @ibgreen sorry, I'm having trouble finding this in the RFC. 

As mentioned, this is discussed in the comments in the RFC. But I will explain again:

Let's say you have 2 objects. But you are so zoomed out that they all render on a single pixel. You select a generous bounding box around them, expected the list of selected objects to include both objects. But only one of them will detected by this picking approach. 

Now repeat the thought experiment with more pixels and more objects. As long as there is a risk of one object overlapping with another, there is a risk that this picking method does not correctly identify all objects within the region, as it depends on a pixel representing that object actually being rendered into the framebuffer.

src/lib/draw-and-pick.js
ericsoco: Ok, reading further I see you're using a box despite the name `radius` here. Maybe note here that `radius` really means "half-size"? Or maybe it's fine.

src/lib/draw-and-pick.js
ericsoco: This was a concern of mine too...

src/lib/draw-and-pick.js
shaojingli: @ericsoco I believe @ibgreen means "occluded" by "hidden". So, if an element is completely occluded. Users won't expect to select it by pointing at it, but they will expect to select it by box selecting.

src/lib/draw-and-pick.js
ericsoco: Alternatively, is there an algo we can use that starts at the center of the bounding box and moves outward until it hits a pick, then bail? If we're only trying to find one result, closest to center..

src/lib/draw-and-pick.js
```diff
@@ -206,11 +195,122 @@ export function pickLayers(gl, {
...
+          closestResultToCenter = pickResult;
+          pickResult.x = deviceBox.x + col;
+          pickResult.y = deviceBox.y + row;
+        }
+      }
+    }
+  }
+  /* eslint-enable max-depth */
+
+  return closestResultToCenter || {
```
ericsoco: So if I'm reading this correctly, we only return one picked element? What if the intent is to marquee-select all elements within bounds?

examples/layer-browser/src/app.js
ericsoco: Agreed with @ibgreen this should be `elements`.

src/lib/layer-manager.js
ericsoco: This is the plan per our last conversation:
> add queryRenderedFeatures-style functionality to layer-manager to allow applications to handle events themselves and run picking on an arbitrary point/bbox. deck.gl will (later) use this feature for event-based picking under the hood

src/react/deckgl.js
```diff
@@ -38,6 +38,7 @@ const propTypes = {
   effects: PropTypes.arrayOf(PropTypes.instanceOf(Effect)),
   gl: PropTypes.object,
   debug: PropTypes.bool,
+  pickingRadius: PropTypes.number,
```
ericsoco: Out of curiosity, what use cases do you imagine for exposing this? Couldn't hurt, I suppose, just thinking out loud

src/lib/draw-and-pick.js
ericsoco: Right, I understand now. Maybe already covered in the RFC (reading now) but this would require us to maintain coordinates/bounds of all elements in model space...

gnavvy:
> Add elementsInScope(topLeft, bottomRight) API to DeckGL component. The method returns an array of unique objects inside the supplied box.

Would you consider to support lasso selection with this elementsInScope() API later? 
or to add another lower level API that takes in an array of path points?

> Add pickingRadius prop to DeckGL component. The old picking behavior is equivalent to pickingRadius: 0. The closest object to the pointer within pickingRadius pixels is picked.

How do we handle picking for mouse position within the pickingRadius for more than one point now?
(will check the code)
The pickingRadius can be extended later for more general voronoi-based pickings.
pickingRadius: 0 => old picking behavior
pickingRadius: a number => the closest point within the radius
pickingRadius: Infinity => the closest point within the viewport

src/lib/draw-and-pick.js
gnavvy: @ericsoco maybe out of the scope of this PR but a voronoi layer would do that, and it is almost free (a few lines tweak based on the scatterplot layer) in deck.gl.

PR #649, opened 2017-05-20
Reduce z-fighting between layers

Pessimistress:
When multiple layers are rendered on the same plane, [z-fighting](https://en.wikipedia.org/wiki/Z-fighting) may create undesirable artifects (e.g. [Issue#411](https://github.com/uber/deck.gl/issues/411)). This PR offsets each layer by a small amount to improve the visual quality of composition. Advanced users can modify the default behavior by specifying their own `getPolygonOffset` accesor.

polygon-offset

Remaining issues:
- Does not solve in-layer z-fighting when objects overlap each other
- At a higher zoom level, this solution is undone by precision error. Can only be fixed by turning on `fp64` flag.

### DIFF
--- a/docs/api-reference/base-layer.md
+++ b/docs/api-reference/base-layer.md
@@ -216,6 +216,25 @@ if the app to mint a new object on every render, all attributes will be automati
 updateTriggers cannot block attribute updates, just trigger them. To block the attribute updates,
 developers need to override the updateState.
 
+---
+
+### Render Properties
+
+##### `getPolygonOffset` (Function, optional)
+
+- Default: `({layerIndex}) => [-layerIndex, -1]`
+
+When multiple layers are rendered on the same plane, [z-fighting](https://en.wikipedia.org/wiki/Z-fighting)
+may create undesirable artifects. By default, deck.gl uses `gl.polygonOffset` to offset each layer by
+a small amount to improve the visual quality of composition.
+
+This accessor takes a single parameter `uniform` - an object that contains the current render uniforms,
+and returns an array of two numbers `factor` and `units`. For more information, refer to the
+[documentation](https://www.opengl.org/archives/resources/faq/technical/polygonoffset.htm).
+
+If the accessor is assigned a falsy value, `GL.POLYGON_OFFSET_FILL` will be disabled.
+
+
 ## Members
 
 *Remarks: Layer members are designed to support the creation of new layers or
--- a/docs/api-reference/base-layer.md
+++ b/docs/api-reference/base-layer.md
@@ -222,18 +222,26 @@ developers need to override the updateState.
 
 ##### `getPolygonOffset` (Function, optional)
 
-- Default: `({layerIndex}) => [-layerIndex, -1]`
+- Default: `({layerIndex}) => [0, -layerIndex * 100]`
 
 When multiple layers are rendered on the same plane, [z-fighting](https://en.wikipedia.org/wiki/Z-fighting)
-may create undesirable artifects. By default, deck.gl uses `gl.polygonOffset` to offset each layer by
-a small amount to improve the visual quality of composition.
+may create undesirable artifacts. To improve the visual quality of composition,
+deck.gl allows layers to use `gl.polygonOffset` to apply an offset to its depth.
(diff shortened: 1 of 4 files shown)


src/lib/layer.js
```diff
@@ -409,6 +413,17 @@ export default class Layer {
 
   // Calculates uniforms
   drawLayer({uniforms = {}}) {
+    const {gl} = this.context;
+
+    // Apply polygon offset to avoid z-fighting
+    if (this.props.getPolygonOffset) {
+      gl.enable(GL.POLYGON_OFFSET_FILL);
```
ibgreen: A minor concern is that layer that uses this will now "hard" disable fill mode for the entire application.

Idle thought:  What about setting `polygonOffset(0, 0)` for layers that don't need it (or have falsy values).  But that assumes that `polygonOffset(0,0)` is equivalent to disabling it, and that the perf penalty of using `polygonOffset` is small. So perhaps safer this way.

Alternatively, we can land this PR and update when state management becomes available in luma.gl, as that is the complete solution to the state overwriting issue.

docs/api-reference/base-layer.md
```diff
@@ -216,6 +216,25 @@ if the app to mint a new object on every render, all attributes will be automati
...
+---
+
+### Render Properties
+
+##### `getPolygonOffset` (Function, optional)
+
+- Default: `({layerIndex}) => [-layerIndex, -1]`
+
+When multiple layers are rendered on the same plane, [z-fighting](https://en.wikipedia.org/wiki/Z-fighting)
+may create undesirable artifects. By default, deck.gl uses `gl.polygonOffset` to offset each layer by
```
ibgreen: artifects => artifacts

src/lib/layer.js
```diff
@@ -47,6 +47,10 @@ const defaultProps = {
   onDragMove: noop,
   onDragEnd: noop,
   onDragCancel: noop,
+  // Offset depth based on layer index to avoid z-fighting.
+  // Negative values pull layer towards the camera
```
ibgreen: I think the comment about direction could go into docs instead of/in addition to here.

docs/api-reference/base-layer.md
```diff
@@ -216,6 +216,25 @@ if the app to mint a new object on every render, all attributes will be automati
...
+a small amount to improve the visual quality of composition.
+
+This accessor takes a single parameter `uniform` - an object that contains the current render uniforms,
+and returns an array of two numbers `factor` and `units`. For more information, refer to the
+[documentation](https://www.opengl.org/archives/resources/faq/technical/polygonoffset.htm).
```
ibgreen: While the GL ref is good, more information about the semantics of the two values would be great, e.g:
* Add the note on direction mentioned below.
* explain that the default layer indices make layers draw from bottom to top.

src/lib/layer.js
```diff
@@ -47,6 +47,10 @@ const defaultProps = {
...
+  // https://www.opengl.org/archives/resources/faq/technical/polygonoffset.htm
+  getPolygonOffset: ({layerIndex}) => [-layerIndex, -1],
```
shaojingli: According to [https://www.khronos.org/registry/OpenGL-Refpages/es2.0/xhtml/glPolygonOffset.xml], we have

`
void glPolygonOffset(GLfloat factor, GLfloat units);
`
The value of the offset is factor × DZ + r × units , where DZ is a measurement of the change in depth relative to the screen area of the polygon, and r is the smallest value that is guaranteed to produce a resolvable offset for a given implementation. 

Shouldn't the params be [-1, -layerIndex]?

shaojingli:
Also, we should noted in the docs or via comments around the call to polygonOffset that, in order for this change to really reduce the z-fighting, fp64 mode needs to be enabled. The inaccuracy in 32-bit projection matrix itself would cause more flickering than this polygon offset technique can handle.

src/lib/layer.js
Pessimistress: Whatever is being set here only applies to the following `layer.draw()` call. The offset mode can be overwritten in the layer's `draw` method if needed.

PR #654, opened 2017-05-22
Fix single/double tap workaround.

ericsoco:
This update to the single/double-tap workaround from `events-refactor` allows hammer.js to handle `requireFailure()` on its own, and instead drops the single tap / click recognizer interval to 0 by default and bumps it up when doubletap functionality is required.

### DIFF
--- a/src/utils/events/constants.js
+++ b/src/utils/events/constants.js
@@ -80,7 +80,7 @@ export const RECOGNIZERS = [
   [Swipe, {enable: false}],
   [Press, {enable: false}],
   [Tap, {event: 'doubletap', taps: 2, enable: false}],
-  [Tap, {enable: false, interval: 0}]
+  [Tap, {enable: false, interval: 0}, null, 'doubletap']
 ];
 
 /**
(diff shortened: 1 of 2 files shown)


src/utils/events/constants.js
```diff
@@ -80,7 +80,7 @@ export const RECOGNIZERS = [
   [Swipe, {enable: false}],
   [Press, {enable: false}],
   [Tap, {event: 'doubletap', taps: 2, enable: false}],
-  [Tap, {enable: false, interval: 0}]
+  [Tap, {enable: false, interval: 0}, null, 'doubletap']
```
ibgreen: Unless this is obvious to any Hammer.js user, perhaps add a comment here about what this does (the same comment you made in this PR)?

ericsoco:
Turns out I was optimizing for a use case we don't really have -- this tinkering with recognizers was enabling us to have a) fast single-clicking and b) prevent single tap events from firing on a doubletap.

However, I'm not seeing any clear way, with recognizer configs or hammer-time.js, to enable that behavior without incurring a single click delay when recognizing doubletaps.

So, I'm clearing out these attempts at a workaround; the resulting behavior is:
- fast single click when no doubletap handlers are registered;
- a doubletap handler, when registered, works as expected;
- single clicks are fired on the way to a doubletap.

If we need doubletaps to _not_ fire single click events, we need to find a way to enable that without incurring a single click delay. `react-map-gl` adds a `doubletap` handler and we don't want to slow down clicks.

PR #658, opened 2017-05-24
Conditionally require hammer.js

ericsoco:
Conditionally require hammer.js to avoid failure on server, on `window` and `document` references.

I'm not happy with this solution, but https://github.com/hammerjs/hammer.js/issues/930 suggests there isn't much of a better option for now :/

### DIFF
--- a/src/utils/events/event-manager.js
+++ b/src/utils/events/event-manager.js
@@ -1,13 +1,15 @@
-import {Manager} from 'hammerjs';
+import WheelInput from './wheel-input';
+import MoveInput from './move-input';
 
-import {
+const Manager = (typeof window === 'undefined' || typeof document === 'undefined') ?
+  ManagerMock : require('hammerjs').Manager;
+const {
   BASIC_EVENT_ALIASES,
   EVENT_RECOGNIZER_MAP,
   RECOGNIZERS,
   GESTURE_EVENT_ALIASES
-} from './constants';
-import WheelInput from './wheel-input';
-import MoveInput from './move-input';
+} = (typeof window === 'undefined' || typeof document === 'undefined') ?
+  {} : require('./constants');
 
 /**
  * Single API for subscribing to events about both
@@ -148,4 +150,14 @@ export default class EventManager {
   _aliasEventHandler(eventAlias) {
     return event => this.manager.emit(eventAlias, event);
   }
+}
+
+function ManagerMock(m) {
+  const noop = () => {};
+  return {
+    on: noop,
+    off: noop,
+    destroy: noop,
+    emit: noop
+  };
 }
--- a/src/utils/events/event-manager.js
+++ b/src/utils/events/event-manager.js
@@ -1,15 +1,30 @@
 import WheelInput from './wheel-input';
 import MoveInput from './move-input';
+import {isBrowser} from '../../controllers/globals';
 
-const Manager = (typeof window === 'undefined' || typeof document === 'undefined') ?
-  ManagerMock : require('hammerjs').Manager;
+// Hammer.js directly references `document` and `window`,
+// which means that importing it in environments without
+// those objects throws errors. Therefore, instead of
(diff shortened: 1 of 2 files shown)


shaojingli:
@ericsoco

What I see from `npm run build` and `npm run test-dist`

# uncaughtException
not ok 56 TypeError: Cannot read property 'protocol' of undefined
  ---
    operator: error
    expected: |-
      undefined
    actual: |-
      [TypeError: Cannot read property 'protocol' of undefined]
    at: Test.die (/Users/shaojing/Workspace/Uber/deck.gl/node_modules/tape-catch/index.js:56:10)
    stack: |-
      TypeError: Cannot read property 'protocol' of undefined
          at Object.<anonymous> (/Users/shaojing/Workspace/Uber/deck.gl/node_modules/sinon/lib/sinon/util/fake_server.js:25:37)
          at Module._compile (module.js:571:32)
          at Module.Mp._compile (/Users/shaojing/Workspace/Uber/deck.gl/node_modules/reify/node/compile-hook.js:10:21)
          at Object.Module._extensions..js (module.js:580:10)
          at Module.load (module.js:488:32)
          at Module.Mp.load (/Users/shaojing/Workspace/Uber/deck.gl/node_modules/reify/node/runtime.js:16:23)
          at tryModuleLoad (module.js:447:12)
          at Function.Module._load (module.js:439:3)
          at Module.require (module.js:498:17)
          at require (internal/module.js:20:19)
  ...

1..56
# tests 56
# pass  55
# fail  1

shaojingli:
We have a `isBrowser` test in src/controllers/globals.js, should we use that?

src/utils/events/event-manager.js
```diff
@@ -1,13 +1,15 @@
-import {Manager} from 'hammerjs';
+import WheelInput from './wheel-input';
+import MoveInput from './move-input';
 
-import {
+const Manager = (typeof window === 'undefined' || typeof document === 'undefined') ?
+  ManagerMock : require('hammerjs').Manager;
```
shaojingli: We actually don't need that as well. We could just Manager set to null and have a if check when Manager object is created.

src/utils/events/event-manager.js
```diff
@@ -149,3 +151,13 @@ export default class EventManager {
     return event => this.manager.emit(eventAlias, event);
   }
 }
+
+function ManagerMock(m) {
```
ibgreen: JavaScripts "hoisting" of function names (allowing them to be used before declaration) is a mixed blessing. 

I'd move this up top, so that we can see how the mocking is done in one place.

src/utils/events/event-manager.js
```diff
@@ -1,13 +1,15 @@
-import {Manager} from 'hammerjs';
+import WheelInput from './wheel-input';
+import MoveInput from './move-input';
 
-import {
+const Manager = (typeof window === 'undefined' || typeof document === 'undefined') ?
```
ibgreen: We have a function in utils that is intended to do a safe check for whether we are running under node.js (a check that is not fooled e.g. by jsdom inclusion). Consider using that?

Also, add a short comment?

src/utils/events/event-manager.js
ericsoco: The mock should be there so that future consumers of EventManager can call its public API without fear of failure in some environments.

ericsoco: As for the `isBrowser` test, and `src/controllers` generally, that folder should be cleaned up to bring it more in-line with `react-map-gl`...and the [`globals.js`](https://github.com/uber/deck.gl/blob/master/src/controllers/globals.js) file there seems to be a near-duplicate of [`src/utils/globals.js`](https://github.com/uber/deck.gl/blob/master/src/lib/utils/globals.js)...thoughts on this @ibgreen @shaojingli?

For now, I can use that `isBrowser`.

PR #660, opened 2017-05-24
Fix the layer tests on node

shaojingli:
Fix the issue that headless GL context is not created properly on node


shaojingli:
@ibgreen yeah... you are right. It's not needed. The real problem is `npm run test-dist` runs 'node-dist.js` and it doesn't include './setup-gl.js'

ericsoco:
Just tested locally after rebasing on master, looks good. Pushed the rebase up. Merge at will

PR #662, opened 2017-05-24
Remove JSDom to align the behavior between node.js and node-dist.js

shaojingli:
Remove JSDom setup in test/node.js.


test/node.js
```diff
@@ -21,12 +21,6 @@
...
 
-// setup for jsDom
-require('./setup-jsdom-env');
-// Mock addEventListener on window, required for seer
-const {window} = require('../src/lib/utils/globals');
-window.addEventListener = () => {};
-
 // Enables import of glsl
 const fs = require('fs');
 require.extensions['.glsl'] = function readShader(module, filename) {
```
ibgreen: Do we still need this? We have moved to inline GLSL. Try to cut?

PR #667, opened 2017-05-30
Fix bug where `mousemove` event is fired twice during drag

Pessimistress:
This bug is causing major performance regression when interacting with the map.
Also fixed a bug in event aliasing (always use registered event types instead of source event type).

@ericsoco We are currently firing all three of `pointermove`, `mousemove` and `touchmove` events on any pointer move, which is technically incorrect. `touchmove` should not be fired on `mousemove`, for example. In the case of `pointermove`, we will need to do some poking to determine the source input device. Yet we do not need to make that differentiation in any of our built-in interactions. I propose we remove the `mouse*` and `touch*` aliases and only surface `pointer*` events to avoid confusion.

### DIFF
--- a/src/utils/events/event-manager.js
+++ b/src/utils/events/event-manager.js
@@ -153,8 +153,7 @@ export default class EventManager {
    * and pipe back out through same (Hammer) channel used by other events.
    */
   _onOtherEvent(event) {
-    const {srcEvent: {type}} = event;
-    this.manager.emit(type, event);
+    this.manager.emit(event.type, event);
   }
 
   /**
--- a/src/utils/events/move-input.js
+++ b/src/utils/events/move-input.js
@@ -1,4 +1,4 @@
-const MOVE_EVENTS = ['pointermove', 'touchmove', 'mousemove'];
+const MOUSE_EVENTS = ['mousedown', 'mousemove', 'mouseup'];
 
 /**
  * Hammer.js swallows 'move' events (for pointer/touch/mouse)
@@ -10,10 +10,11 @@ const MOVE_EVENTS = ['pointermove', 'touchmove', 'mousemove'];
  */
 export default class MoveInput {
 
-  constructor(element, callback, events = MOVE_EVENTS) {
+  constructor(element, callback, events = MOUSE_EVENTS) {
     this.element = element;
     this.callback = callback;
     this.events = events;
+    this.pressed = false;
 
     this.handler = this.handler.bind(this);
     this.events.forEach(event => element.addEventListener(event, this.handler));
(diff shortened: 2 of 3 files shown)


src/utils/events/move-input.js
```diff
@@ -1,4 +1,5 @@
-const MOVE_EVENTS = ['pointermove', 'touchmove', 'mousemove'];
+const MOUSE_EVENTS = ['mousedown', 'mousemove', 'mouseup'];
+const EVENT_TYPES = ['mousemove', 'pointermove'];
```
ericsoco: How about `MOVE_EVENT_TYPES`?

src/utils/events/move-input.js
```diff
@@ -24,9 +26,29 @@ export default class MoveInput {
...
+      if (event.button === 0) {
+        this.pressed = true;
+      }
+      break;
+    case 'mousemove':
+      if (event.which !== 1) {
+        this.pressed = false;
+      }
+      if (!this.pressed) {
+        // Drag events are emitted by hammer already, we just need the hover event
```
ericsoco: nit: "we just need the _move_ event" (for detecting hover)

src/utils/events/event-manager.js
```diff
@@ -153,8 +153,7 @@ export default class EventManager {
    * and pipe back out through same (Hammer) channel used by other events.
    */
   _onOtherEvent(event) {
-    const {srcEvent: {type}} = event;
-    this.manager.emit(type, event);
+    this.manager.emit(event.type, event);
```
ericsoco: Good catch, sorry!

src/utils/events/move-input.js
Pessimistress: Among all move events we have drag (button down) and hover (button up). The point here is that the move events with button down is already emitted. And this custom input class should only fire move events when the button is up.

src/utils/events/move-input.js
```diff
@@ -24,9 +26,29 @@ export default class MoveInput {
   }
 
   handler(event) {
```
ibgreen: This is the only place I am aware of where we use a noun to `name` a method. Usually we use `verbNoun` or `verb`.

src/utils/events/move-input.js
```diff
@@ -24,9 +26,29 @@ export default class MoveInput {
...
-      target: this.element
-    });
+    switch (event.type) {
+    case 'mousedown':
+      if (event.button === 0) {
+        this.pressed = true;
+      }
+      break;
+    case 'mousemove':
+      if (event.which !== 1) {
```
ibgreen: Do `mousedown` and `mousemove` events name this member differently: `event.button` vs. `event.which`?  I think that is worth a comment, for a reviewer it looks like a typo.

src/utils/events/move-input.js
ericsoco: The name is from hammer.js, which uses it as an abstract method on its input classes. So, while it's not idiomatic in our frameworks, it is so for Hammer...

src/utils/events/move-input.js
ericsoco: `button` only applies to events in which a button is pressed or released:
https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/button. And `which` is non-standard but better supported than `buttons`...IOW this part of JS is a mess that dates back to before there were many "good parts" :/

src/utils/events/move-input.js
ibgreen: > So, while it's not idiomatic in our frameworks, it is so for Hammer...

Thanks, that explains it, I should have realized this. Still, maintainers and reviewers might not have objected to a short comment indicating this.

ericsoco: No reason you should have known. I'm ambivalent about it, actually -- that's just the origin of the name, not an argument for or against it. I agree that it's not a great name so in a future pass through here perhaps we can change it.

PR #668, opened 2017-05-30
Conditionally register custom event handlers

Pessimistress:


ericsoco:
Is there a specific reason you're making this change, or just generally a good idea?

Also, as of this PR I believe we have some changes in deck.gl's `event-manager` that aren't in react-map-gl's, and v.v. Any reason you're landing these changes in deck.gl instead of react-map? (Maybe it's a good time to do another pass at manually synchronizing them?)

src/utils/events/event-manager.js
```diff
@@ -91,6 +94,13 @@ export default class EventManager {
           this.aliasedEventHandlers[event] = aliasedEventHandler;
         }
       }
+      if (!this.wheelInput && WheelInput.isSourceOf(event)) {
```
ibgreen: Nit: To keep code as clean as possible (reduce number of if statements) I would have preferred if the `isSourceOf` thing was internal to these classes - e.g. we'd always create these but they would only register events if needed.

In fact it seems to me that the classes already support a third parameter that does loop over events and only register what was asked for, would it be possible to use that?

PR #669, opened 2017-05-31
Event handling bug fixes

Pessimistress:
- Merge changes from react-map-gl
- Fix gl error when dragging outside of the canvas
- Fix drag events registration

### DIFF
--- a/src/react/deckgl.js
+++ b/src/react/deckgl.js
@@ -138,10 +138,10 @@ export default class DeckGL extends React.Component {
       events: {
         click: this._onClick,
         mousemove: this._onMouseMove,
-        dragstart: this._onDragEvent,
-        dragmove: this._onDragEvent,
-        dragend: this._onDragEvent,
-        dragcancel: this._onDragCancel
+        panstart: this._onDragEvent,
+        panmove: this._onDragEvent,
+        panend: this._onDragEvent,
+        pancancel: this._onDragEvent
       }
     });
   }
@@ -154,7 +154,8 @@ export default class DeckGL extends React.Component {
 
     if (center) {
       // processed pointer input, supports touch
-      [x, y] = center;
+      x = center.x;
+      y = center.y;
     } else if (srcEvent) {
       // fallback
       x = srcEvent.clientX;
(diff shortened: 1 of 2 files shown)


src/react/deckgl.js
```diff
@@ -138,23 +138,49 @@ export default class DeckGL extends React.Component {
...
+    if (center) {
+      // processed pointer input, supports touch
+      x = center.x;
+      y = center.y;
+    } else if (srcEvent) {
+      // fallback
+      x = srcEvent.clientX;
+      y = srcEvent.clientY;
+    }
+    if (!Number.isFinite(x)) {
```
ibgreen: Should this just be an `else` statement rather than and `if isFinite`?

If one of the two loops above can generated a bad `x`, maybe add a comment why this happens.

ericsoco: Looks like @Pessimistress just moved it in here from what was previously a check in the event handlers. I'm not sure what would have generated a bad value before, something in luma's EventProxy?

src/react/deckgl.js
```diff
@@ -138,23 +138,49 @@ export default class DeckGL extends React.Component {
...
     });
   }
 
+  // Get mouse position {x, y} relative to the container
+  _getPos(event) {
+    const {srcEvent, center, rootElement} = event;
+    let x;
+    let y;
+
+    if (center) {
```
ericsoco: Another vote for cracking Hammer open to do what we want it to (per our conv today)...it's annoying that our application layer needs to know about pointer vs. mouse events. Ideally we could just query the event for some standardized `x` and `y`.

src/react/deckgl.js
```diff
@@ -138,23 +138,49 @@ export default class DeckGL extends React.Component {
...
+        panstart: this._onDragEvent,
+        panmove: this._onDragEvent,
+        panend: this._onDragEvent,
+        pancancel: this._onDragEvent
       }
     });
   }
 
+  // Get mouse position {x, y} relative to the container
+  _getPos(event) {
```
ericsoco: Nice way to DRY out the event handlers.

src/react/deckgl.js
Pessimistress: `srcEvent` may not have a field `clientX` if it's not from a mouse. I'll add a comment.

PR #673, opened 2017-06-01
Add queryObjects api to DeckGL component

Pessimistress:
Given a bounding box and a list of layer ids (optional), returns an array of unique picking info objects.

As discussed before, the limitation of this API is that it doesn't return occluded objects.

### DIFF
--- a/docs/api-reference/deckgl.md
+++ b/docs/api-reference/deckgl.md
@@ -121,21 +121,38 @@ are affected.
 
 ### Methods
 
-##### `queryObjects(boundingBox, layerId)`
+##### `queryObject(options)`
+
+Get the closest pickable and visible object at screen coordinate.
+
+Parameters:
+- `options` (Object)
+  + `x` (Number) - x position in pixels
+  + `y` (Number) - y position in pixels
+  + `radius` (Number, optional) - radius of tolerance in pixels. Default `0`.
+  + `layerIds` (Array, optional) - a list of layer ids to query from.
+    If not specified, then all pickable and visible layers are queried.
+
+Returns: a single [`info`](/docs/get-started/interactivity.md#the-picking-info-object) object, or `null` if nothing is found.
+
+##### `queryObjects(options)`
 
 Get all pickable and visible objects within a bounding box.
 
 Parameters:
-- `boundingBox` (Object) - the bounding box in screen coordinates
-  + `x` (Number) - left of the box in pixels
-  + `y` (Number) - top of the box in pixels
-  + `width` (Number) - width of the box in pixels
-  + `height` (Number) - height of the box in pixels
-- `layerIds` (Array, optional) - a list of layer ids to query from.
-  If not specified, then all pickable and visible layers are queried.
+- `options` (Object)
+  + `x` (Number) - left of the bounding box in pixels
+  + `y` (Number) - top of the bouding box in pixels
+  + `width` (Number, optional) - width of the bouding box in pixels. Default `1`.
+  + `height` (Number, optional) - height of the bouding box in pixels. Default `1`.
+  + `layerIds` (Array, optional) - a list of layer ids to query from.
+    If not specified, then all pickable and visible layers are queried.
 
 Returns: an array of unique [`info`](/docs/get-started/interactivity.md#the-picking-info-object) objects
 
+Remarks:
+- This query methods are designed to quickly find objects by utilizing the picking buffer. They offer more flexibility for developers to handle events in addition to the built-in hover and click callbacks.
+- Note there is a limitation in the query methods: occluded objects are not returned. To improve the results, you may try setting the `layerIds` parameter to limit the query to fewer layers.
 
 ## Source
 [src/react/deckgl.js](https://github.com/uber/deck.gl/blob/4.0-release/src/react/deckgl.js)
--- a/docs/api-reference/deckgl.md
+++ b/docs/api-reference/deckgl.md
@@ -119,5 +119,23 @@ object for the topmost picked layer at the coordinate
 are affected.
 - `event` - the original [MouseEvent](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent) object
 
+### Methods
+
+##### `queryObjects(boundingBox, layerId)`
+
+Get all pickable and visible objects within a bounding box.
+
+Parameters:
+- `boundingBox` (Object) - the bounding box in screen coordinates
+  + `x` (Number) - left of the box in pixels
+  + `y` (Number) - top of the box in pixels
+  + `width` (Number) - width of the box in pixels
+  + `height` (Number) - height of the box in pixels
+- `layerIds` (Array, optional) - a list of layer ids to query from.
+  If not specified, then all pickable and visible layers are queried.
+
+Returns: an array of unique [`info`](/docs/get-started/interactivity.md#the-picking-info-object) objects
+
+
 ## Source
 [src/react/deckgl.js](https://github.com/uber/deck.gl/blob/4.0-release/src/react/deckgl.js)
--- a/examples/layer-browser/src/app.js
+++ b/examples/layer-browser/src/app.js
@@ -52,7 +52,8 @@ class App extends PureComponent {
         // rotationX: 0
       },
       hoveredItem: null,
-      clickedItem: null
+      clickedItem: null,
+      queriedItems: null
     };
 
     this._effects = [new ReflectionEffect()];
@@ -115,6 +116,7 @@ class App extends PureComponent {
     const {width, height} = this.state;
     const infos = this.refs.deckgl.queryObjects({x: 0, y: 0, width, height});
     console.log(infos); // eslint-disable-line
+    this.setState({queriedItems: infos});
   }
 
   _renderExampleLayer(example, settings, index) {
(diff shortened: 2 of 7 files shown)


docs/api-reference/deckgl.md
```diff
@@ -119,5 +119,23 @@ object for the topmost picked layer at the coordinate
 are affected.
 - `event` - the original [MouseEvent](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent) object
 
+### Methods
+
+##### `queryObjects(boundingBox, layerId)`
```
ibgreen: layerId => layerIds

Should we call this `fastQueryObjects` or similar to emphasize that it is imprecise?

docs/api-reference/deckgl.md
```diff
@@ -119,5 +119,23 @@ object for the topmost picked layer at the coordinate
...
+Parameters:
+- `boundingBox` (Object) - the bounding box in screen coordinates
+  + `x` (Number) - left of the box in pixels
+  + `y` (Number) - top of the box in pixels
+  + `width` (Number) - width of the box in pixels
+  + `height` (Number) - height of the box in pixels
+- `layerIds` (Array, optional) - a list of layer ids to query from.
+  If not specified, then all pickable and visible layers are queried.
+
+Returns: an array of unique [`info`](/docs/get-started/interactivity.md#the-picking-info-object) objects
```
ibgreen: We should explain that this method has limitations...

docs/api-reference/deckgl.md
```diff
@@ -119,5 +119,23 @@ object for the topmost picked layer at the coordinate
...
+
+##### `queryObjects(boundingBox, layerId)`
+
+Get all pickable and visible objects within a bounding box.
+
+Parameters:
+- `boundingBox` (Object) - the bounding box in screen coordinates
+  + `x` (Number) - left of the box in pixels
+  + `y` (Number) - top of the box in pixels
+  + `width` (Number) - width of the box in pixels
```
ibgreen: - Could `width` and `height` default to `1`
- Or should we have a separate `queryObject` that could have ({x, y, radius}) ?
- I am thinking the latter since `queryObject` is precise but `queryObjects` isn't.

- I generally try to go with flat named parameter lists. If we feel that the app is going to maintain boundingBoxes separately I guess the nesting could be worthwhile.

src/lib/draw-and-pick.js
```diff
@@ -55,6 +55,57 @@ export function drawLayers({layers, pass}) {
     ${visibleCount} visible, ${layers.length} total`);
 }
 
+export function queryLayers(gl, {
```
ibgreen: A short comment header would be nice. No need to list all the params but a sentence that explains what this function is supposed to do (in particular a reader would be curious to see what this does compared to all other similar functions is this file).

src/lib/draw-and-pick.js
```diff
@@ -212,6 +263,93 @@ function pickFromBuffer(gl, {
...
+          });
+        }
+      }
+      i += 4;
+    }
+  }
+
+  return Array.from(uniqueColors.values());
+}
+/* eslint-enable max-depth, max-statements */
```
ibgreen: Are these linter disables actually needed for above function? I typically find them reliable indicators that a function should be split so at least remove if not needed, and consider splitting if they are.

src/lib/draw-and-pick.js
```diff
@@ -212,6 +263,93 @@ function pickFromBuffer(gl, {
...
+
+// Returns an Uint8ClampedArray of picked pixels
```
ibgreen: Good, but also describe the purpose of the function.

src/lib/layer-manager.js
```diff
@@ -198,6 +207,22 @@ export default class LayerManager {
 
   // PRIVATE METHODS
 
+  _getPickingBuffer() {
+    const {gl} = this.context;
+
+    // Set up a frame buffer if needed
+    if (this.context.pickingFBO === null ||
+      gl.canvas.width !== this.context.pickingFBO.width ||
```
ibgreen: luma.gl v4 has `framebuffer.resize(...)` that can be called every frame, it only resizes attachment when needed.

`FramebufferObject` is deprecated and just an alias for `Framebuffer`

src/lib/draw-and-pick.js
```diff
@@ -212,6 +263,93 @@ function pickFromBuffer(gl, {
...
+          const pickedLayer = layers[pickedLayerIndex];
+          const pickedObjectIndex = pickedLayer.decodePickingColor(pickedColor);
+          closestResultToCenter = {pickedColor, pickedLayer, pickedObjectIndex};
+        }
+      }
+      i += 4;
+    }
+  }
+
+  return closestResultToCenter || {
```
ibgreen: Nit: I normally see this in pattern: the initialization of `closestResultToCenter` to the default value, rather than initializing it to null and setting it at the last minute. Technically that approach lowers the complexity count because of saving an if statement (hidden in the `||`)

src/lib/layer-manager.js
```diff
@@ -147,29 +147,38 @@ export default class LayerManager {
...
       layers: this.layers,
       mode,
       viewport: this.context.viewport,
-      pickingFBO: this.context.pickingFBO,
+      pickingFBO: this._getPickingBuffer(),
       lastPickedInfo: this.context.lastPickedInfo
     });
   }
 
+  queryLayer({x, y, width, height}, layerIds) {
```
ibgreen: Maybe comments above this and pickLayer so that the reader understands how and why they are different

examples/layer-browser/src/app.js
```diff
@@ -111,6 +111,12 @@ class App extends PureComponent {
     this.setState({clickedItem: info});
   }
 
+  _onQueryObjects() {
+    const {width, height} = this.state;
+    const infos = this.refs.deckgl.queryObjects({x: 0, y: 0, width, height});
+    console.log(infos); // eslint-disable-line
```
ericsoco: If the console is not open, clicking the "Query Rendered Objects" button will appear to do nothing. Probably better to pop open a div with `JSON.stringify(infos)` in it.

examples/layer-browser/src/app.js
```diff
@@ -222,6 +228,7 @@ class App extends PureComponent {
         { this._renderMap() }
         { !MAPBOX_ACCESS_TOKEN && this._renderNoTokenWarning() }
         <div id="control-panel">
+          <button onClick={this._onQueryObjects}>Query Rendered Objects</button>
```
ericsoco: "Query Objects" instead of "Query Rendered Objects"? Sounds like a Mapbox leftover...

src/lib/layer-manager.js
ericsoco: ^^ and that `layerIds` is an optional filter, else all layers will be queried

src/lib/draw-and-pick.js
```diff
@@ -212,6 +263,93 @@ function pickFromBuffer(gl, {
...
+}
+/* eslint-enable max-depth, max-statements */
+
+// Returns an Uint8ClampedArray of picked pixels
+function getPickedColors(gl, {
+  layers,
+  pickingFBO,
+  deviceRect: {x, y, width, height}
+}) {
   // TODO - just return glContextWithState once luma updates
```
ericsoco: I've worked around this TODO before as well -- @ibgreen , any context here? How low does this fruit hang?

ericsoco:
Do `queryFromBuffer()` and `pickFromBuffer()` (and their companions `queryLayers()` and `pickLayers()`) need to be separate functions? Is it possible to combine them and DRY this up a bit without sacrificing legibility?

They feel like the same process, but with each operating on a slightly different domain. Maybe I'm missing some idiosyncrasies of each...

ibgreen:
One final thing, before we land, please add to the "What's New" section as well. This will also ensure that we don't miss taking another look at this during our formal API review for 4.1.

Also, based on your clever image diff test PR from yesterday, I suspect that we now have the capability to write actual test cases for `queryObjects`. Not needed for this PR, but that would be pretty awesome, so maybe open a separate task on that?

examples/layer-browser/src/app.js
```diff
@@ -111,6 +111,12 @@ class App extends PureComponent {
     this.setState({clickedItem: info});
   }
 
+  _onQueryObjects() {
+    const {width, height} = this.state;
+    const infos = this.refs.deckgl.queryObjects({x: 0, y: 0, width, height});
```
shaojingli: I know this is just an example, but still, can we change the query boundary to be something like, {x: width / 4, y: height / 4, width / 4 * 3, height / 4 * 3} so that this will return only part of the objects current on screen? It might serve partially as a sanity test

src/lib/draw-and-pick.js
```diff
@@ -212,6 +263,93 @@ function pickFromBuffer(gl, {
...
+    pickedLayer: null,
+    pickedObjectIndex: -1
+  };
+}
+
+/**
+ * Query within a specified rectangle
+ * Returns array of unique objects in shape `{x, y, pickedColor, pickedLayer, pickedObjectIndex}`
+ */
+function queryFromBuffer(gl, {
```
shaojingli: I'm really bad at naming stuff... but we might consider change the name of queryFromBuffer() as framebuffer is not buffer... people with webgl knowledge will try to find which buffer this function executes on and figure out there is no buffer involved.

I propose queryFromPickingBuffer? or queryForRect

src/lib/layer-manager.js
```diff
@@ -198,6 +207,22 @@ export default class LayerManager {
...
   // PRIVATE METHODS
 
+  _getPickingBuffer() {
+    const {gl} = this.context;
+
+    // Set up a frame buffer if needed
+    if (this.context.pickingFBO === null ||
+      gl.canvas.width !== this.context.pickingFBO.width ||
+      gl.canvas.height !== this.context.pickingFBO.height) {
+      this.context.pickingFBO = new FramebufferObject(gl, {
```
shaojingli: As @ibgreen mentioned, let take the chance and update deck.gl with the new naming - Framebuffer

shaojingli:
@Pessimistress please remember to rebase this PR to latest master. I changed the history of master yesterday morning to properly merge @Firenze11 's PR.

examples/layer-browser/src/app.js
Pessimistress: You can zoom in  to query a smaller region.

PR #676, opened 2017-06-01
Flatten renderLayers

ibgreen:


shaojingli:
Just curious, under what circumstances does a user want to use a multi-dimensional array instead of a simple array to specify the sublayers of a composite layer?

ibgreen:
> Just curious, under what circumstances does a user want to use a multi-dimensional array instead of a simple array to specify the sublayers of a composite layer? 

I added example code in the docs. 

> Is it requested?

Requested by me :-) I use a lot of custom composite layers in my apps and conditionally generate several sublists of layers, so I need to do a lot of ugly array management.

React has a special set of utils for handling the parallel situation [React.Children](https://facebook.github.io/react/docs/react-api.html#react.children) and I was accustomed to that and growing increasingly envious of it...

src/lib/utils/flatten.js
```diff
@@ -19,18 +19,32 @@
...
+export function flatten(array, {
+  filter = () => true,
+  result = []
+} = {}) {
+  array = Array.isArray(array) ? array : [array];
+  return flattenArray(array, {filter, result});
+}
+
+// Deep flattens an array. Helper to `flatten`, see its parameters
+function flattenArray(array, {filter, result} = {}) {
```
Pessimistress: Does the construction and deconstruction of argument objects have perf impact in an incursive function? Given it's an internal helper we don't have to have a pretty interface.

ibgreen: > Does the construction and deconstruction of argument objects have perf impact in an incursive function? Given it's an internal helper we don't have to have a pretty interface.

Excellent suggestion. Fixed!

src/lib/utils/flatten.js
```diff
@@ -19,18 +19,32 @@
...
   let index = -1;
   while (++index < array.length) {
     const value = array[index];
     if (Array.isArray(value)) {
-      flatten(value, result);
-    } else {
+      flatten(value, {result, filter});
```
ibgreen: > flattenArray since all arguments are already valid?

Yes of course. Good catch!

PR #678, opened 2017-06-02
Fix modelMatrix in METER_OFFSET mode

Pessimistress:
`project_to_clipspace` bug introduced by https://github.com/uber/deck.gl/pull/607


Pessimistress:
We don't have a test case that uses modelMatrix.

PR #680, opened 2017-06-02
fix point cloud examples bug

gnavvy:
changed how the empty data is checked.

the filtering logics was moved to post-renderLayers() in https://github.com/uber/deck.gl/pull/676,
causing 0, generated by previous check logics, being used as a 'layer' that leads to the error.

### DIFF
--- a/examples/point-cloud-laz/app.js
+++ b/examples/point-cloud-laz/app.js
@@ -139,9 +139,14 @@ class Example extends PureComponent {
   }
 
   _renderLazPointCloudLayer() {
-    return this.state.points.length && new PointCloudLayer({
+    const {points} = this.state;
+    if (!points || points.length === 0) {
+      return null;
+    }
+
+    return new PointCloudLayer({
       id: 'laz-point-cloud-layer',
-      data: this.state.points,
+      data: points,
       projectionMode: COORDINATE_SYSTEM.IDENTITY,
       getPosition: d => d.position,
       getNormal: d => [0, 0.5, 0.2],
--- a/examples/point-cloud-laz/package.json
+++ b/examples/point-cloud-laz/package.json
@@ -7,7 +7,7 @@
     "start-local": "webpack-dev-server --env.local --progress --hot --open"
   },
   "dependencies": {
-    "deck.gl": "^4.0.0-rc.4",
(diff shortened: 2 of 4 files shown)


examples/point-cloud-laz/app.js
```diff
@@ -139,9 +139,14 @@ class Example extends PureComponent {
   }
 
   _renderLazPointCloudLayer() {
-    return this.state.points.length && new PointCloudLayer({
+    const {points} = this.state;
+    if (!points || points.length === 0) {
```
shaojingli: Yep... this is what we should've done

examples/point-cloud-laz/package.json
```diff
@@ -3,19 +3,19 @@
   "version": "0.0.1",
   "license": "MIT",
   "scripts": {
-    "start": "webpack-dev-server --progress --hot --port 3000 --open",
+    "start": "webpack-dev-server --progress --hot --open",
```
gnavvy: allow running multiple examples without port conflicts

examples/point-cloud-laz/app.js
ibgreen: With the support for falsy values in the layer array that I merged yesterday, you should be able to just do
```
return points && points.length && newPointCloudLayer({
```

gnavvy:
> Is it simply a version problem?
no, previous the layers were filtered before passing to deck.gl, so 0 was not an issue.
in https://github.com/uber/deck.gl/pull/676, the "pre-filtering" was moved to "post-renderLayers-filtering", which caused the error.

the version update is irrelevant to the error, but wanted to make them update to date.

gnavvy:
> With the support for falsy values in the layer array that I merged yesterday, you should be able to just do
```
return points && points.length && newPointCloudLayer({
```

I am considering to stop using such ^ "convenient" one-line falsy check. I have ran into issues in other places where we were expected a component would not be rendered after such check, but end up having a div showing **0** in the UI.

ibgreen:
> I have ran into issues in other places where we were expected the component would not be rendered after such check, but end up having a div showing **0** in the UI.

I've been bitten by that too in React but I can't see how that could happen here. We only accept `Layer` subclasses, there is no possibility for confusion like in react.

PR #692, opened 2017-06-07
Simplify activation of WheelInput and MoveInput.

ericsoco:
Add missing devDep.
Note that tests are failing, apparently due to a transpilation error, so this PR probably won't pass CI.

### DIFF
--- a/src/utils/events/move-input.js
+++ b/src/utils/events/move-input.js
@@ -25,12 +25,14 @@ export default class MoveInput {
     this.options.events.forEach(event => this.element.removeEventListener(event, this.handleEvent));
   }
 
-  set(options) {
-    Object.assign(this.options, options);
-  }
-
-  isSourceOf(eventName) {
-    return MOVE_EVENT_TYPES.indexOf(eventName) >= 0;
+  /**
+   * Enable this input (begin processing events)
+   * if the specified event type is among those handled by this input.
+   */
+  enableOnEventType(eventType) {
+    if (MOVE_EVENT_TYPES.indexOf(eventType) >= 0) {
+      this.options.enable = true;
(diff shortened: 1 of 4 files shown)


src/utils/events/move-input.js
```diff
@@ -25,12 +25,14 @@ export default class MoveInput {
...
-    Object.assign(this.options, options);
-  }
-
-  isSourceOf(eventName) {
-    return MOVE_EVENT_TYPES.indexOf(eventName) >= 0;
+  /**
+   * Enable this input (begin processing events)
+   * if the specified event type is among those handled by this input.
+   */
+  enableOnEventType(eventType) {
```
ibgreen: Should we drop the `On` and just call this `enableEventType`? Are we saying something important with the `On`?

Do we only need the ability to enable event types (not disable them)?

package.json
```diff
@@ -85,6 +85,7 @@
     "react-addons-test-utils": "^15.4.2",
     "react-dom": "^15.4.0",
     "reify": "^0.4.4",
+    "source-map-loader": "^0.2.1",
```
ibgreen: Didn't @shaojingli already add this? Maybe rebase?

ericsoco: It was added in a followup this morning, wasn't there last night.

src/utils/events/move-input.js
ericsoco: I can't find a good name for this function. We're not enabling for a specific event type, we're enabling the entire input if the passed event type is one that the input handles.

As for disabling, yes, I suppose we might need to (though I'd say it's fairly unlikely); I added @Pessimistress' `set()` method back in to support that.

Issue #656, opened 2017-05-23
Events refactor followups


ericsoco:
This issue tracks followups to the [event handling refactor](https://github.com/uber/deck.gl/pull/636):

- [x] Currently, only one handler for an aliased event (only an issue for `'click'` for the foreseeable future) can exist at a time. Local state should be expanded to support an array of handlers against each aliased event. (#693)
- [X] Allow overriding recognizers in constructor. (#696)
- [x] Unit tests. (#705)
- [x] Changelog / docs.
- [x] Produce consistent Event object fields with custom input (currently wheel and pointermove). (#703)

Once landed, the following events-related work remains:
- [x] Fix [picking-on-every-event bug](https://github.com/uber/deck.gl/issues/634).
- [x] Move event handling from `deckgl.js` React wrapper into `layer-manager` so that non-React deck.gl consumers get event handling out of the box. (#738)
- [ ] Move per-layer handler calls from `draw-and-pick.js` into `layer-manager.js` event handlers added in #738
- [x] This PR breaks drag event handling, as drag events are not supported by Hammer.js out-of-the-box. That's ok, because drag events are [currently implemented incorrectly anyway](https://github.com/uber/deck.gl/issues/648). So, drag event handling needs to be refactored. (#738, #741, #742)
- [ ] Avoid reflows on (pointermove) event (@Pessimistress).
- [ ] Silence `passive` event warnings in Chrome.

Pessimistress:
Importing from `deck.gl` breaks isomorphic rendering because hammer.js accesses `document`. Should we do a sanity check in EventManager?

ericsoco:
Probably, yes -- don't see any reason anything in EventManager needs to (or can be expected to) work on the server.

Also, for reference: https://github.com/hammerjs/hammer.js/issues/930

ericsoco:
See also:
- [ ] unit tests

Pessimistress:
There is a bug with `mousemove`:
- Both `pointermove` and `mousemove` events are fired
- When dragging, an additional move event is fired from hammer.js
This causes picking to be triggered 3 times each frame during drag.

shaojingli:
@Pessimistress can you do a pass on the remaining items listed in this issue tracker?

ibgreen:
event handling is moved to mjolnir.js, let's track it there if needed.

PR #693, opened 2017-06-07
Ensure multiple handlers can coexist for an aliased event, and can be…

ericsoco:
… removed independently. (#656)

### DIFF
--- a/src/utils/events/event-manager.js
+++ b/src/utils/events/event-manager.js
@@ -47,7 +47,7 @@ export default class EventManager {
     this.manager = new Manager(element, {recognizers: RECOGNIZERS})
       .on('hammer.input', this._onBasicInput);
 
-    this.aliasedEventHandlers = {};
+    this.aliasedEventHandlerCounts = {};
 
     // Handle events not handled by Hammer.js:
     // - mouse wheel
@@ -95,15 +95,15 @@ export default class EventManager {
       // Alias to a recognized gesture as necessary.
       const eventAlias = GESTURE_EVENT_ALIASES[event];
       if (eventAlias) {
-        if (!this.aliasedEventHandlers[event]) {
+        if (!this.aliasedEventHandlerCounts[event]) {
           // Alias the event type and register the alias with Hammer.
           const aliasedEventHandler = this._aliasEventHandler(event);
           this.manager.on(eventAlias, aliasedEventHandler);
-          this.aliasedEventHandlers[event] = [];
+          this.aliasedEventHandlerCounts[event] = 0;
         }
-        // Store the handler whose event type required aliasing
-        // for later removal.
-        this.aliasedEventHandlers[event].push(handler);
+        // Keep track of the number of aliased event handlers.
+        // (The original handler is added below.)
+        this.aliasedEventHandlerCounts[event]++;
       }
     }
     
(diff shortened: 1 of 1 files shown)


src/utils/events/event-manager.js
```diff
@@ -76,38 +76,46 @@ export default class EventManager {
...
+      this._onOne(event, handler);
+    } else {
+      // If `event` is a map, call `on()` for each entry.
+      for (const eventName in event) {
+        this._onOne(eventName, event[eventName]);
+      }
+    }
+  }
+
+  _onOne(event, handler) {
```
ibgreen: Even if we don't change the public interface, could we just call this `_addEventHandler`, or `_registerEvent`?

Would love to see a one line comment header above each internal method describing its purpose (I trust the external methods are documented in markdown).

src/utils/events/event-manager.js
```diff
@@ -76,38 +76,46 @@ export default class EventManager {
...
+    // Special handling for gestural events.
+    const recognizerEvent = EVENT_RECOGNIZER_MAP[event];
+    if (recognizerEvent) {
+      // Enable recognizer for this event.
+      this.manager.get(recognizerEvent).set({enable: true});
```
ibgreen: For readability, It might be good to split this into two lines with a temporary variable to make it more clear what kind of object `get` is returning (I would have to look it up in the code). `get` is a very generically named method, assuming it is set by Hammer and we can't change it?

src/utils/events/event-manager.js
```diff
@@ -76,38 +76,46 @@ export default class EventManager {
...
-        if (eventAlias && !this.aliasedEventHandlers[event]) {
+      this._onOne(event, handler);
+    } else {
+      // If `event` is a map, call `on()` for each entry.
+      for (const eventName in event) {
+        this._onOne(eventName, event[eventName]);
+      }
+    }
+  }
+
```
ibgreen: I prefer to see all public method first, followed by private (underscore ) methods, it should be easy to scan a class from top to bottom to see the public interface.

src/utils/events/event-manager.js
```diff
@@ -117,26 +125,40 @@ export default class EventManager {
...
     } else {
       // If `event` is a map, call `off()` for each entry.
       for (const eventName in event) {
-        this.off(eventName, event[eventName]);
+        this.offOne(eventName, event[eventName]);
       }
     }
   }
 
+  _offOne(event, handler) {
```
ibgreen: Naming, same suggestion as above.

src/utils/events/event-manager.js
```diff
@@ -76,38 +76,46 @@ export default class EventManager {
    */
   on(event, handler) {
```
ericsoco: `.on()/.off()` is a very common idiom for adding event handlers. [Node](https://nodejs.org/api/events.html#events_emitter_on_eventname_listener) and [jQuery](http://api.jquery.com/on/) both use it, and it's less verbose than e.g. `addEventListener`. I think it's fine as-is but if you feel strongly about it we can change it.

src/utils/events/event-manager.js
```diff
@@ -117,26 +125,35 @@ export default class EventManager {
...
+          // remove the aliased handler from Hammer.
+          // (The original handler is removed below.)
+          this.manager.off(eventAlias);
+          delete this.aliasedEventHandlerCounts[event];
+        }
+      }
+    }
+
+    // Deregister event handler.
+    this.manager.off(event, handler);
```
ericsoco: For posterity:

`aliasedEventHandlerCounts[event]` only gets decremented if it's truthy, and if the handler wasn't previously registered and aliased it won't be truthy.

also, i already have a reset in there:
`delete this.aliasedEventHandlerCounts[event];`

PR #700, opened 2017-06-09
Fast update

ibgreen:


ibgreen:
* Fix for spurious react render of identical layers when using react-map-gl 3.0.0-alpha.12 (avoids crashing deck.gl)
* Log tuning. 
* Some light code cleanup
* Move long comments in code into README.

src/lib/layer-manager.js
```diff
@@ -114,9 +90,14 @@ export default class LayerManager {
   }
 
   updateLayers({newLayers}) {
-    /* eslint-disable */
-    assert(this.context.viewport,
-      'LayerManager.updateLayers: viewport not set');
+    // TODO - something is generating state updates that cause rerender of the same
+    if (newLayers === this.lastRenderedLayers) {
```
shaojingli: I'm a bit confused over this newly added `lastRenderedLayers`... we probably should have a doc explaining the relations between `this.prevLayers`, `this.lastRenderedLayers`, `this.layers`, etc... (and also additional variables like `generatedLayers`, `oldLayers`, `newLayers`)

ibgreen: This is a hack to avoid a crash caused by react-map-gl v3.

Pessimistress: I am not able to reproduce this on master.

Issue #618, opened 2017-05-01
Internet Explorer 11 support


howtimeflies0:
*** ISSUES ***

- [ ] Sliders in interactive layer explorer doesn't work

Various data worker in our gallery examples do not work
- [ ] IE doesn't support template literals
- [ ] IE doesn't support shorthand property names

Things we need to do:
- [ ] Add docs in `getting-started` that explains that polyfills are needed when building for IE11. #845

ibgreen:
>  IE doesn't support template literals
>  IE doesn't support shorthand property names

Sounds like these haven't been transpiled? Maybe an easy fix?

kujon:
Would it be possible to squeeze this into 4.0 if it's really the case of transpilation? Happy to help w/ that.

ehallq:
Hi
I am currently developing a application with deck.gl and it works really good! But I really need IE11 support. What is the status of this issue? I can see that this is a milestone set for v4.1, and v4.1 was released 15 days ago. Should I be prepered to wait 1 week, 1 month or a year for IE11 to get support? I just want to be able to get some kind of estimate :)

ibgreen:
@ehallq There are no specific plans. But if the issues are small, we might be able to address them in a patch release to deck.gl 4.1 (best case within a couple of days). 

A great way to help us would be to let us know what problems you are seeing. Are they the same as listed in this issue? Did fix #702 not address them?

IE11 is not one of our primary environments. We mainly want to support IE11 (as the one exception to our focus on "evergreen" browsers) for the benefit external users such as yourself - to ensure we have a reasonable coverage of browsers for people building web apps on deck.gl. So any help will be appreciated!

jkchen777atgmail:
IE 11 has been used and supported on Windows only in Corp space for a
while. These are the basics for use of IE v11 as reference.

1. Enterprise Mode for Internet Explorer 11

https://docs.microsoft.com/en-us/internet-explorer/ie11-deploy-guide/enterprise-mode-overview-for-ie11

2. Baseline
https://docs.microsoft.com/en-us/internet-explorer/ie11-deploy-guide/system-requirements-and-language-support-for-ie11

kktam:
There seemed to be some problem with trans-piled output.

for example in the [Arc Example](https://github.com/uber/deck.gl/tree/master/examples/arc), the error is located at bundle.js 

```
line 74974
var pageLoadPromise = isBrowser ? new Promise(function (resolve, reject) {
  window.onload = function () {
[...]

PR #702, opened 2017-06-12
Transpile demo workers to es5

Pessimistress:
https://github.com/uber/deck.gl/issues/691

### DIFF
--- a/demo/package.json
+++ b/demo/package.json
@@ -9,7 +9,7 @@
   "scripts": {
     "start": "webpack-dev-server --config webpack/dev --progress --content-base ./src/static --port 3000 --open",
     "build-clean": "rm -rf ./dist && mkdir dist",
-    "build-static": "cp -r ./src/static/* dist && cd `git rev-parse --show-toplevel` && git checkout origin/blog -- blog && mv blog demo/dist/blog && cd -",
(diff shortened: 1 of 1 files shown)


shaojingli:
This will also solve #618 right?

kktam:
There seemed to be some problem with trans-piled output.

for example in the [Arc Example](https://github.com/uber/deck.gl/tree/master/examples/arc), the error is located at bundle.js 

```
line 74974
var pageLoadPromise = isBrowser ? new Promise(function (resolve, reject) {
  window.onload = function () {
    isPageLoaded = true;
    resolve(document);
  };
}) : Promise.resolve({});
```

```
SCRIPT5009: 'Promise' is undefined
File: bundle.js, Line: 74974, Column: 1
```

This error is captured in IE 11 Version 11.0.9600.18617

ibgreen:
> 'Promise' is undefined

@kktam  Thanks for reporting this. This should be solvable by the application by importing a polyfill like `babel-polyfill` or similar.

In the example folder.
`yarn add babel-polyfill` or 
`npm install babel-polyfill`

In `app.js`, add
```
require('babel-polyfill`);
```
before other code

While we do want to support IE11, I think we should handle this through documentation (i.e. have a section about how to add the polyfill), rather than make all apps pay the cost of a polyfill when all modern browsers already define `Promise` by default.

@shaojingli @Pessimistress

ibgreen:
Opened #845

PR #704, opened 2017-06-13
Fix 64-bit PathLayer

shaojingli:
This is a fix to the 64-bit PathLayer. The original `lineJoin` function in the 64-bit vertex shader actually uses 32-bit positions. This make the center vertices of 64-bit PathLayer accurate but not the vertices extending in the lateral direction.

The fix solves this by using 64-bit floating point numbers for currPosition, prevPosition and nextPosition to do `lineJoin`

### DIFF
--- a/src/layers/core/path-layer/path-layer-vertex-64.glsl.js
+++ b/src/layers/core/path-layer/path-layer-vertex-64.glsl.js
@@ -51,13 +51,23 @@ const float EPSILON = 0.001;
 float flipIfTrue(bool flag) {
   return -(float(flag) * 2. - 1.);
 }
-vec3 lineJoin(vec3 prevPoint, vec3 currPoint, vec3 nextPoint) {
+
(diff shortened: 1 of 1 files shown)


src/layers/core/path-layer/path-layer-vertex-64.glsl.js
```diff
@@ -187,14 +194,38 @@ void main() {
...
+  // Calculate prev position 64bit
+
+  vec2 tempPrevPosition64[2];
+  tempPrevPosition64[0] = sum_fp64(vec2(prevPosition.x, 0.0), instanceStartPositions64[0]);
+  tempPrevPosition64[1] = sum_fp64(vec2(prevPosition.y, 0.0), instanceStartPositions64[1]);
+
+  vec4 prevPosition64 = vec4(tempPrevPosition64[0].xy, tempPrevPosition64[1].xy);
+
+  vec2 projected_prev_position[2];
+  project_position_fp64(prevPosition64, projected_prev_position);
```
ibgreen: This line looks like it could cost us some performance.

Do we still have the render benchmark support? Could we collect before and after numbers?

Issue #648, opened 2017-05-19
Drag event handling is implemented incorrectly


ericsoco:
`drag-*` events are [only supposed to be fired by browsers on a selection or `draggable` element](https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API).

Those concepts don't apply to elements within a `canvas`; therefore, we should be implementing dragging using pointer/touch/mouse start, move, and end events instead of using `drag-*` events.

ericsoco:
This conversation should be considered as well:
https://github.com/uber/react-map-gl/pull/258

ericsoco:
Consensus here is to remove drag event handling from deck.gl and leave it up to applications. This means:
- [ ] Remove `MOTION_EVENTS` logic from `draw-and-pick::pickLayers.js`
- [ ] Remove drag event handlers from `deckgl.js`

Related is removing `mouse` event handling from `deckgl.js` and instead using only `pointer` events. That entails:
- [ ] Remove `mousemove` event handler from `deckgl.js` and add `pointerdown`, `pointermove`, and `pointerup` event handlers
- [ ] Wire up `onHover` callback to `pointermove` event

Then, we can provide an updated example of an application handling object-dragging logic:
- [ ] Add object dragging to `graph-layer` example: maintain picked object on `pointerdown` and update it across `pointermove` and `pointerup` events

/cc @ibgreen @shaojingli @Pessimistress

ericsoco:
Closed by #738, #741, #742.

Issue #634, opened 2017-05-08
Picking always runs


ericsoco:
deckgl.js [always adds input event handlers](https://github.com/uber/deck.gl/blob/master/src/react/deckgl.js#L129) to its `canvas`, even when no layers have specified mouse event callbacks. This means that picking runs on every mouse event (move and click and drag) right now.

We should only be adding those handlers when the application indicates it wants to respond to input events.

Planning to fix as part of the event handling refactor.
/cc @Pessimistress

shaojingli:
Is this still an issue? @ericsoco @Pessimistress

Pessimistress:
Yes. We do not do any checking before calling `pickLayers`. In a common scenario, if no layers are pickable, when mouse moves, we unnecessarily switch context to the picking frame buffer, renders nothing, then return an empty pixel. Worst case scenario, if all layers are pickable but no `onHover` callback is assigned, we still run the full picking render cycle.

ericsoco:
Yes, the [TODO is now in layer-manager](https://github.com/uber/deck.gl/blob/master/src/lib/layer-manager.js#L159-L162)

1chandu:
We fixed picking flow, now when no layers are pickable, we don't enter into picking flow.

PR #708, opened 2017-06-15
Move event management from deckgl.js React component into LayerManager.

ericsoco:
*** * * DO NOT MERGE, WIP * * ***

This PR tackles the EventManager followup items in #656. Among its goals:
- move event management into LayerManager
- fix to the picking-on-every-event bug (#634)
- begin to remove drag handling (#648)

More to come.


src/lib/layer-manager.js
```diff
@@ -200,7 +265,51 @@ export default class LayerManager {
...
+  //
+
+  /**
+   * Route click events to layers.
+   * `pickLayer` will call the `onClick` prop of any picked layer,
+   * and `onLayerClick` is called directly from here
+   * with any picking info generated by `pickLayer`.
+   */
+  _onClick(event) {
+    const pos = getPos(event);
```
ibgreen: So the main thing the user needs to specify is how to extract the position from the event. The rest of the logic here is independent of event handling system, right?

So exploring the idea of keeping event registration separate, we could potentially move out the event registration, keep these functions in here (maybe removing the `_` underscore prefixes) so that the can be called by the external event handlers, and add a `getPositition` argument to the `LayerManager` constructor ?

src/lib/layer-manager.js
```diff
@@ -200,7 +265,51 @@ export default class LayerManager {
...
+    if (!pos) {
+      return;
+    }
+    const radius = this.pickingRadius;
+    const selectedInfos = this.pickLayer({x: pos.x, y: pos.y, radius, mode: 'click'});
+    if (selectedInfos.length) {
+      const firstInfo = selectedInfos.find(info => info.index >= 0);
```
ibgreen: Should we just have a function `pickTopLayer` so we can drop the `find` here? @Pessimistress

src/lib/layer-manager.js
```diff
@@ -428,3 +537,33 @@ function layerName(layer) {
   }
   return !layer ? 'null layer' : 'invalid layer';
 }
+
+/**
+ * Get mouse position {x, y} relative to the container
+ */
+function getPos(event) {
```
ibgreen: Abbreviated names are typically avoided in this code base: `getPos` => `getPosition`

src/lib/layer-manager.js
```diff
@@ -428,3 +537,33 @@ function layerName(layer) {
...
+    // fallback
+    x = srcEvent.clientX;
+    y = srcEvent.clientY;
+  }
+  if (!Number.isFinite(x)) {
+    // srcEvent.clientX may be undefined if it's not from a mouse
+    return null;
+  }
+
+  const rect = rootElement.getBoundingClientRect();
```
ibgreen: I am not an expert on these things, but folks who seem to be have advised me in the past to avoid calling `getBoundingClientRect` on any regular basis, as it can trigger an expensive [reflow](https://gist.github.com/paulirish/5d52fb081b3570c81e3a) in the browser.

My understanding was that events have `offsetX` and `offsetY` or similar parameters that contain relative coordinates so that this can be handled in a more lightweight fashion.

src/react/deckgl.js
```diff
@@ -79,6 +69,7 @@ export default class DeckGL extends React.Component {
   }
 
   componentWillReceiveProps(nextProps) {
+    this._setLayerManagerProps(nextProps);
```
ibgreen: Feels a little misleading since `LayerManager` is a traditional JS class and doesn't really have props. Just call it `_updateLayerManager`?

src/lib/layer-manager.js
```diff
@@ -138,6 +167,42 @@ export default class LayerManager {
...
+      }
+    });
+    this.eventManager.on(events);
+  }
+
+  removeEventListeners(element, eventNames) {
+    // TODO
+  }
+
+  validateEventHandling() {
```
ibgreen: While I understand why you put this here, I'm a little torn about it. Can this be moved "deeper" into the layer manager life cycle?

src/react/deckgl.js
```diff
@@ -118,141 +129,19 @@ export default class DeckGL extends React.Component {
     this.props.onWebGLInitialized(gl);
 
     // Note: avoid React setState due GL animation loop / setState timing issue
-    this.layerManager = new LayerManager({gl});
+    this.layerManager = new LayerManager({gl, canvas});
```
ibgreen: FWIW: [WebGLRenderingContext.canvas](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/canvas)

ericsoco: There's no change here in the way we're using `canvas`, just passing it along to `LayerManager` for `EventManager` event registration. Is there some caveat we're not covering here (hence the link)?

src/lib/layer-manager.js
```diff
@@ -138,6 +167,42 @@ export default class LayerManager {
     return this;
   }
 
+  addEventListeners(eventNames) {
```
ericsoco: Hm...so then `EventManager` or an equivalent would be passed into `LayerManager`? This feels a bit heavy, though I can see how it may improve extensibility for event handling. I wonder though how useful that flexibility is, given it would slightly increase complexity of the API.

We could make it an optional param and assume most people will use the default, which we would set to `EventManager`. Is that what you're thinking?

src/lib/layer-manager.js
ericsoco: Yes -- after I wrote this I realized I needed to store `onLayerClick` and `onLayerHover` somewhere that `LayerManager` could get at them. That means now that all this logic can live within `LayerManager` instead of straddling across `deckgl.js` as well.

src/lib/layer-manager.js
ibgreen: > Hm...so then EventManager or an equivalent would be passed into LayerManager?

I had a different mode in mind. The external event handling would be independent and just call `layerManager.onClick` and `layerManager.onPointerMove` when events occurred to get help with picking and delivering the resulting infos to the deck.gl callback system in the right way.

src/react/deckgl.js
ibgreen: The link was just to show that `canvas` is already available as a member on the `gl` context that we are already passing as a parameter, so it might be unnecessary to add the `canvas` parameter.

But perhaps we want to handle the case where the event canvas is different from the canvas that the context was created on.

ericsoco: I was thinking that the DOM element handling events should not be coupled with the gl context. If event handling exists in its own module, alongside LayerManager (instead of within it), this will be moot anyway (it will accept an arbitrary DOM element just like EventManager does).

src/lib/layer-manager.js
ericsoco: I wrote a comment here earlier today and GH seems to have swallowed it.

IIRC I suggested we use `rootElement.offsetLeft`/`Top` instead of `getBoundingClientRect()` to avoid reflows. @Pessimistress what's the use case here, was this logic added to handle the fact that touch events don't have `clientX`/`Y` or to handle situations in which the `canvas` doesn't fill the viewport?

ericsoco:
The diff is a little weird now due to @Pessimistress' PR and my rebasing onto it.
[Here is a more legible version](https://github.com/uber/deck.gl/compare/b177245b46ca335ada106d57c93ff77947cd24b2) (that just diffs between the last commit and `master`).

src/lib/layer-event-manager.js
```diff
@@ -0,0 +1,111 @@
...
+   * and `onLayerHover` is called directly from here
+   * with any picking info generated by `pickLayer`.
+   */
+  _onPointerMove(event) {
+    const pos = event.offsetCenter;
+    // TODO: consider using this.eventManager.element size instead of layerManager.context
+    // but do so in a way that doesn't cause reflow (e.g. `offsetWidth/Height`).
+    // maybe the event object offers offsetCenter as a 0<>1 value as well?
+    // since it's already doing size calculations...
+    const {width, height} = this.layerManager.context.viewport;
```
ibgreen: Right or wrong, my expectation was that this file would contain code that was unique to doing event integration with the EventManager.

The code in _onPointerMode and _onClick (with the exception of the refernece to `oevent.offsetCenter`) line like it would be identical for any event system integration, and so might be better to keep inside LayerManager. 

Thus 
```
_onPointerMove(event) {
   this.layerManager.onPointerMode({event, pos: event.offsetCenter});
}
```

Doing it like this woud also mean that this class wouldn't need to rake for the `LayerManager`s internal state like the `context.viewport`.

src/lib/layer-event-manager.js
```diff
@@ -0,0 +1,111 @@
...
+    this.layerManager = layerManager;
+    this.eventManager = new EventManager(eventTarget);
+
+    this.options = {};
+    this.set(options);
+  }
+
+  set(options = {}) {
+    Object.keys(DEFAULT_OPTIONS).forEach(k => {
+      this.options[k] = options[k] !== 'undefined' ? options[k] : DEFAULT_OPTIONS[k];
```
Pessimistress: `undefined`?

src/lib/layer-event-manager.js
ericsoco: Makes sense, will do.

src/lib/layer-event-manager.js
ericsoco: crap, good catch.

ericsoco:
@ibgreen and I discussed where to draw the lines between modules here.

Currently (before this PR):
- `deckgl.js` listens for `<canvas>` events from `event-manager.js` and handles them by calling:
- the per-layer handlers (e.g. `onClick`), which are set on the layers themselves and called by `draw-and-pick.js`, via 'layer-manager.js';
- the top-level event handlers (e.g. `onLayerClick`), which belong to `deckgl.js`.

A cleaner implementation of all of this would involve:
- lifting per-layer handler calls out of `draw-and-pick.js` and instead calling them from `layer-manager.js` based on the picking info objects returned from `draw-and-pick.js::pickLayers`;
- allowing `layer-manager` to manage top-level event handlers as well;
- calling both per-layer and top-level handlers from a single function per event type, that lives on `layer-manager.js`;
- keeping the connection to `event-manager.js` for DOM event handling broken out separately from either `deckgl.js` or `layer-manager.js` so that applications can use their own DOM event handling if so desired.

In this scenario, `layer-event-manager.js` acts as a handler pass-through from `deckgl.js` to `layer-manager.js` for React apps, and can either be used or bypassed by non-React apps to connect DOM events to `layer-manager.js`'s handler functions that call per-layer and top-level event handlers.

I'll rework this PR a bit further toward this goal, but will refrain from lifting per-layer callbacks out of `draw-and-pick.js` until another PR.
/cc @ibgreen

Pessimistress:
Is the latest luma.gl still registering its own event handlers? If so, why are we not using those?

ibgreen:
> Is the latest luma.gl still registering its own event handlers? If so, why are we not using those?

luma.gl has a legacy system for registering event handlers, an optional `addEvents` function. It does not get automatically called, it is a utility that deck.gl can choose to use.

Due to the complexity of the event handling topic and the fact that we already have two repos in play I have intentionally kept luma.gl event handling out of the discussions.

A problem with the `addEvents` that I just ran into is that there is no `removeEvents`, which I would need to handle the new example website in a good way. So it could make sense to ask whether to invest more time in the old system, or whether to bring the new system into luma.gl as well.

ericsoco:
@Pessimistress we replaced a luma `addEvents()` call in `deckgl.js` with EventManager. AFAIK (and per @ibgreen above) luma's event handling isn't used anywhere else by deck.gl.

ericsoco:
@ibgreen [here's a diff](https://github.com/uber/deck.gl/compare/e6895c2a3d9279fc853c6be09697de4a2da69292) for the state of this PR as of the latest push, rebased onto master.

This commit does away with LayerEventManager and instead leaves layer event handling in control of LayerManager.

In the interest of closing this PR, I stopped short of fixing the picking-on-every-event bug (#634), since it will require adding/removing handlers to/from EventManager every time new top-level handlers are passed to `setEventParams()` or layers are updated with new per-layer handlers.

For the same reason, I'm also leaving for another PR the work to pull the per-layer handler calls out of `draw-and-pick.js`.

ericsoco:
Closing in favor of #738.

PR #710, opened 2017-06-15
EventManager consistency and bug fixes

Pessimistress:
- Consistent `event` object: https://github.com/uber/deck.gl/issues/703
- Do not call `preventDefault` on wheel input
- Simplify emit chain

### DIFF
--- a/src/react/deckgl.js
+++ b/src/react/deckgl.js
@@ -156,37 +156,9 @@ export default class DeckGL extends React.Component {
     });
   }
 
-  // Get mouse position {x, y} relative to the container
-  _getPos(event) {
-    const {srcEvent, center, rootElement} = event;
-    let x;
-    let y;
-
-    if (center) {
-      // processed pointer input, supports touch
-      x = center.x;
-      y = center.y;
-    } else if (srcEvent) {
-      // fallback
-      x = srcEvent.clientX;
-      y = srcEvent.clientY;
-    }
-    if (!Number.isFinite(x)) {
-      // srcEvent.clientX may be undefined if it's not from a mouse
-      return null;
-    }
-
-    const rect = rootElement.getBoundingClientRect();
-
-    return {
-      x: x - rect.left - rootElement.clientLeft,
-      y: y - rect.top - rootElement.clientTop
-    };
-  }
-
   // Route events to layers
   _onClick(event) {
-    const pos = this._getPos(event);
+    const pos = event.relativeCenter;
     if (!pos) {
       return;
     }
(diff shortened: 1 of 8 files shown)


src/utils/events/event-manager.js
```diff
@@ -117,50 +117,65 @@ export default class EventManager {
...
+      const {element} = this;
+      const {srcEvent} = event;
+
+      const center = event.center || {
+        x: srcEvent.clientX,
+        y: srcEvent.clientY
+      };
+
+      // Calculate center relative to the root element
+      const rect = element.getBoundingClientRect();
```
ericsoco: https://github.com/uber/deck.gl/pull/708/files#r122335442

Pessimistress: `offsetTop` and `offsetLeft` don't account for page scroll. What is the case where `getBoundingClientRect()` doesn't work?

src/utils/events/event-manager.js
```diff
@@ -117,50 +117,65 @@ export default class EventManager {
...
-    // Register event handler.
-    this.manager.on(event, handler);
+    const wrappedHandler = this._wrapEventHandler(event, handler);
+    // Alias to a recognized gesture as necessary.
+    const eventAlias = GESTURE_EVENT_ALIASES[event] || event;
+
+    // Save wrapped handler
+    this.eventHandlers.push({event, eventAlias, handler, wrappedHandler});
+
+    this.manager.on(eventAlias, wrappedHandler);
```
ericsoco: Nice simplification here, thanks!

src/react/deckgl.js
```diff
@@ -156,37 +156,9 @@ export default class DeckGL extends React.Component {
...
-    return {
-      x: x - rect.left - rootElement.clientLeft,
-      y: y - rect.top - rootElement.clientTop
-    };
-  }
-
   // Route events to layers
   _onClick(event) {
-    const pos = this._getPos(event);
+    const pos = event.relativeCenter;
```
ericsoco: Maybe `offsetCenter` to be more in line with existing JS naming?

src/utils/events/event-manager.js
```diff
@@ -117,50 +117,65 @@ export default class EventManager {
...
+      const {srcEvent} = event;
+
+      const center = event.center || {
+        x: srcEvent.clientX,
+        y: srcEvent.clientY
+      };
+
+      // Calculate center relative to the root element
+      const rect = element.getBoundingClientRect();
+      const relativeCenter = {
```
ericsoco: Maybe `offsetCenter` to be more in line with existing JS naming?

src/utils/events/event-manager.js
ericsoco: Not an issue of it not working, it's about perf -- `getBoundingClientRect()` causes a reflow. Would be nice to avoid calling it if we can.

Pessimistress: I'll make a note here and address in a separate PR.

PR #711, opened 2017-06-15
Fix using external buffer for layer attributes

Pessimistress:


examples/layer-browser/src/examples/core-layers.js
```diff
@@ -191,6 +191,52 @@ const PointCloudLayerExample = {
   }
 };
 
+// const PointCloudLayerExample = {
```
ibgreen: A comment on why we are leaving this commented out?

test/lib/attribute-manager.spec.js
```diff
@@ -133,6 +134,33 @@ test('AttributeManager.update - 0 numInstances', t => {
...
+
+  const dummyUpdate = () => t.fail('updater should not be called when external buffer is present');
+
+  attributeManager.add({
+    positions: {size: 2, update: dummyUpdate},
+    colors: {size: 3, type: GL.UNSIGNED_BYTE, update: dummyUpdate}
+  });
+
+  // First update, should autoalloc and update the value array
+  attributeManager.update({
```
ibgreen: Add a test case with incorrect array types and `t.throws`?

PR #714, opened 2017-06-17
Update demo, examples and configs to react-map-gl 3 and webpack 2

ibgreen:


demo/README.md
```diff
@@ -3,7 +3,7 @@ This is the source code of the demo website for deck.gl.
 ## Developing
 
     npm install
-    export MAPBOX_ACCESS_TOKEN=<insert_your_token> && npm start
+    export MapboxAccessToken=<insert_your_token> && npm start
```
Apercu: why changing the env name?

demo/.babelrc
```diff
@@ -0,0 +1,10 @@
+{
+  presets: [
```
Apercu: neat: can be a one liner in the separate `babelrc`

demo/README.md
Pessimistress: We used to support both `MAPBOX_ACCESS_TOKEN` and `MapboxAccessToken` in react-map-gl, v3 dropped the former

src/controllers/map-controls.js
```diff
@@ -59,9 +59,9 @@ const propTypes = {
 
   constraints: PropTypes.object,
 
-  perspectiveEnabled: PropTypes.bool,  // Enables perspective control event handling
+  dragToRotate: PropTypes.bool,  // Enables perspective control event handling
```
Pessimistress: Why do we have a copy of react-map-gl components in deck.gl?

demo/.babelrc
ibgreen: > neat: can be a one liner in the separate babelrc

Actually the separate babelrc still doesn't work, will remove.

src/controllers/map-controls.js
ibgreen: > Why do we have a copy of react-map-gl components in deck.gl?

I used to do integration experiments with non-mapbox maps, and I was also investigating building on a family of controllers that would follow similar principles.

It is another chapter in the never ending story of what should reasonably be common components between these two frameworks, but with no good place to put them.

We can remove it for now. Another PR though.

PR #715, opened 2017-06-17
Stateless picking

ibgreen:
@1chandu - debugged this yesterday, it was not a state management issue, just the `frameBuffer` parameter that had changed - this caused picking to render into the default framebuffer, clearing/overwriting the results of the normal render.


src/lib/draw-and-pick.js
```diff
@@ -350,25 +350,21 @@ function getPickedColors(gl, {
...
+    scissorTest: true,
+    scissor: [x, y, width, height],
+    blend: true,
+    blendFunc: [gl.ONE, gl.ZERO, gl.CONSTANT_ALPHA, gl.ZERO],
+    blendEquation: gl.FUNC_ADD
+    // TODO - Set clear color
   }, () => {
 
     // Clear the frame buffer
     gl.clear(GL.COLOR_BUFFER_BIT | GL.DEPTH_BUFFER_BIT);
```
shaojingli: Could we eventually move this gl call to luma.gl through withParamters function as well?

ibgreen: > Could we eventually move this gl call to luma.gl through withParamters function as well?

Yes makes sense. Although one could argue that clearing the framebuffer is an "action" (i.e not just changing a setting), so it might be better off as a separate luma function? Maybe something @1chandu can look at?

ibgreen:
Overwrote the branch by mistake, reopened as separate PR.

PR #720, opened 2017-06-19
Seer performance badges for attribute updates and render

ibgreen:
screen shot 2017-06-19 at 8 56 40 am

### DIFF
--- a/src/lib/layer-manager.js
+++ b/src/lib/layer-manager.js
@@ -25,7 +25,8 @@ import assert from 'assert';
 import {drawLayers, pickLayers, queryLayers} from './draw-and-pick';
 import {LIFECYCLE} from './constants';
 import {Viewport} from './viewports';
-import {setOverride, layerEditListener, initLayer, logLayer} from '../debug/seer-integration';
+import {setPropOverrides, layerEditListener, initLayerInSeer, updateLayerInSeer}
+  from '../debug/seer-integration';
 import {experimental} from 'luma.gl';
 import {Framebuffer} from 'luma.gl';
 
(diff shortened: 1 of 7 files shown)


src/lib/stats.js
```diff
@@ -0,0 +1,56 @@
...
+    formatted = `${ms.toFixed(1)}ms`;
+  } else if (ms < 1000) {
+    formatted = `${ms.toFixed(0)}ms`;
+  } else {
+    formatted = `${(ms / 1000).toFixed(2)}s`;
+  }
+  return formatted;
+}
+
+export function leftPad(string, length = 8) {
```
Apercu: haha, better be safe ;)

test/index.js
```diff
@@ -24,4 +24,4 @@ import './lib';
 import './utils';
 import './experimental';
 import './deprecated';
-import './debug/seer-integration';
+import './debug/seer-integration.spec';
```
Apercu: why changing this? we are already in the tests directory

src/lib/layer-manager.js
```diff
@@ -70,10 +71,11 @@ export default class LayerManager {
     Object.seal(this);
 
     /**
-     * Set an override on the specified property and update the layers
-     */
+     import {setPropOverrides, layerEditListener, initLayerInSeer, updateLayerInSeer}
```
Apercu: what's this comment?

test/index.js
ibgreen: > why changing this? we are already in the tests directory

Every other file in the test directory follows this `.spec` convention.

This convention is pretty common in the JS community. FWIW, it does have the advantage of easily distinguishing the two files e.g. in sublime tabs or when do command-open in the Chrome debugger.

src/lib/layer-manager.js
ibgreen: > what's this comment?

Yes looks strange, will remove.

ibgreen:
> Very nice, but isn't the formatting of the ms badge strange? shouldn't we only see the number?

Yes, we can play around with it, but I prefer to leave as-is for now.

I want to see the count of how many updates have happened (often more important than the number). Also wanted the user to have a chance of understanding what it means. If we just have two numbers, it will also be confusing to the user. 

If I could provide a tooltip for these badges, with additional data might be the best option...

test/index.js
Apercu: sure

PR #721, opened 2017-06-19
Lifecycle tuning.

ibgreen:
* Support using the same layer instances multiple times (mainly for the sublayer case).
* Don't regenerate sublayers if `shouldUpdate` returns false.
* Improve prop and data diffing to avoid disturbing/updating layers in more cases.
* Improved logging of why a layer is updated (i.e. where the diffing process found a difference).


src/lib/layer.js
```diff
@@ -455,6 +456,26 @@ export default class Layer {
     // End lifecycle method
   }
 
+  // Checks state of attributes and model
+  // TODO - is attribute manager needed? - Model should be enough.
+  getNeedsRedraw({clearRedrawFlags = false} = {}) {
```
Pessimistress: The `clearRedrawFlags` arg always feels like a potential source of bugs... Will it be better if we reset the flag when `drawLayer` is called?

src/lib/props.js
```diff
@@ -0,0 +1,113 @@
...
+export function compareProps({oldProps, newProps, ignoreProps = {}} = {}) {
+  assert(oldProps !== undefined && newProps !== undefined, 'compareProps args');
+
+  // props are immutable, so shallow equality => deep equality
+  if (oldProps === newProps) {
+    return null;
+  }
+
+  if (typeof newProps !== 'object' || newProps === null) {
+    return 'new props is not an object';
```
Pessimistress: The message may be a little confusing when neither `oldProps` or `newProps` is an object.
Context: I think `diffUpdateTriggers` also uses this function.

PR #722, opened 2017-06-19
feat(seer): remove layer on finalyze

Apercu:


src/debug/seer-integration.js
```diff
@@ -75,6 +75,14 @@ export const layerEditListener = cb => {
   });
 };
 
+export const removeLayer = id => {
+  if (!window.__SEER_INITIALIZED__ || !id) {
```
ibgreen: Idea: Maybe this "uglyness" (of looking at variables on the window) could be hidden in a function `isSeerInitialized()`? Initially here in deck.gl, and later provided by the seer API?

ibgreen: Nit: I personally prefer to avoid returns when it doesn't complicate the code too much.
```
if (isSeerInitialized() && id) {
  seer.delete...
}
```
Would look tighter to me here.

src/lib/layer.js
```diff
@@ -432,6 +432,7 @@ export default class Layer {
     // Call subclass lifecycle method
     this.finalizeState();
     // End lifecycle method
+    removeLayer(this.id);
```
ibgreen: The seer calls can look quite confusing here, as all the functions here already deal with layers in really subtle and interconnected ways. It would be good if one could see at a glance that these calls were related to the seer integration. 
```
removeLayerFromSeer(this.id);
```
Would make it so much easier to spot what is going on.

I already started renaming some of the other functions for the same reason.

src/debug/seer-integration.js
Apercu: Depends I guess, I usually do this to remove as much nesting as I can. In this case it might be better. I can refactor once I add the initialize method

PR #726, opened 2017-06-20
Adopt to new luma state managment api

1chandu:
- Changing gl.*** calls into setParameters
- Use withParameters for temporarily setting state for specific draws (draw-and-pick.js)
- Tested with:  'run test, run test-browser (3 errors, not new), examples/layer-browser and all website examples'.

### DIFF
--- a/src/layers/core/screen-grid-layer/screen-grid-layer.js
+++ b/src/layers/core/screen-grid-layer/screen-grid-layer.js
@@ -83,10 +83,13 @@ export default class ScreenGridLayer extends Layer {
   draw({uniforms}) {
     const {minColor, maxColor} = this.props;
     const {model, cellScale, maxCount} = this.state;
-    const {gl} = this.context;
-    gl.depthMask(true);
     uniforms = Object.assign({}, uniforms, {minColor, maxColor, cellScale, maxCount});
-    model.render(uniforms);
+    model.draw({
+      uniforms,
+      settings: {
+        depthMask: true
+      }
+    });
   }
 
   _getModel(gl) {
--- a/src/lib/draw-and-pick.js
+++ b/src/lib/draw-and-pick.js
@@ -366,38 +366,38 @@ function getPickedColors(gl, {
     // Clear the frame buffer
     gl.clear(GL.COLOR_BUFFER_BIT | GL.DEPTH_BUFFER_BIT);
 
-    // Save current blend settings
-    // const oldBlendMode = getBlendMode(gl);
     // Set blend mode for picking
     // always overwrite existing pixel with [r,g,b,layerIndex]
-    // gl.enable(gl.BLEND);
-    // gl.blendFuncSeparate(gl.ONE, gl.ZERO, gl.CONSTANT_ALPHA, gl.ZERO);
-    // gl.blendEquation(gl.FUNC_ADD);
+    const settings = {
+      blend: true,
+      blendFunc: [gl.ONE, gl.ZERO, gl.CONSTANT_ALPHA, gl.ZERO],
+      blendEquation: gl.FUNC_ADD
+    };
 
-    // Render all pickable layers in picking colors
-    layers.forEach((layer, layerIndex) => {
-      if (!layer.isComposite && layer.props.visible && layer.props.pickable) {
+    withParameters(gl, settings, () => {
+      // Render all pickable layers in picking colors
+      layers.forEach((layer, layerIndex) => {
+        if (!layer.isComposite && layer.props.visible && layer.props.pickable) {
 
-        // Encode layerIndex with alpha
-        gl.blendColor(0, 0, 0, (layerIndex + 1) / 255);
-
-        layer.drawLayer({
-          uniforms: Object.assign(
-            {renderPickingBuffer: 1, pickingEnabled: 1},
-            layer.context.uniforms,
-            getUniformsFromViewport(layer.context.viewport, layer.props),
-            {layerIndex}
-          )
-        });
-      }
+          // Encode layerIndex with alpha
+          const blendColorSettings = {blendColor: [0, 0, 0, (layerIndex + 1) / 255]};
+          withParameters(gl, blendColorSettings, () => {
+            layer.drawLayer({
+              uniforms: Object.assign(
+                {renderPickingBuffer: 1, pickingEnabled: 1},
+                layer.context.uniforms,
+                getUniformsFromViewport(layer.context.viewport, layer.props),
+                {layerIndex}
+              )
+            });
+          });
+        }
+      });
     });
 
     // Read color in the central pixel, to be mapped with picking colors
     const pickedColors = new Uint8Array(width * height * 4);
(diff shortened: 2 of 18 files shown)


src/lib/draw-and-pick.js
```diff
@@ -354,47 +354,47 @@ function getPickedColors(gl, {
   // Make sure we clear scissor test and fbo bindings in case of exceptions
   // We are only interested in one pixel, no need to render anything else
   // Note that the callback here is called synchronously.
-  return glContextWithState(gl, {
+  return withParameters(gl, {
     frameBuffer: pickingFBO,
     framebuffer: pickingFBO,
     scissorTest: {x, y, w: width, h: height}
```
ibgreen: Does this work? Didn't you decide to go with two values?

src/lib/draw-and-pick.js
```diff
@@ -354,47 +354,47 @@ function getPickedColors(gl, {
...
-        layer.drawLayer({
-          uniforms: Object.assign(
-            {renderPickingBuffer: 1, pickingEnabled: 1},
-            layer.context.uniforms,
-            getUniformsFromViewport(layer.context.viewport, layer.props),
-            {layerIndex}
-          )
-        });
-      }
+    const settings = {
```
ibgreen: Move these into the withParameters call, they should be set for all the draw calls...

src/lib/draw-and-pick.js
```diff
@@ -354,47 +354,47 @@ function getPickedColors(gl, {
...
+      // Render all pickable layers in picking colors
+      layers.forEach((layer, layerIndex) => {
+        if (!layer.isComposite && layer.props.visible && layer.props.pickable) {
+
+          // Encode layerIndex with alpha
+          const blendColorSettings = {blendColor: [0, 0, 0, (layerIndex + 1) / 255]};
+          withParameters(gl, blendColorSettings, () => {
+            layer.drawLayer({
+              uniforms: Object.assign(
+                {renderPickingBuffer: 1, pickingEnabled: 1},
```
ibgreen: For these temporary uniforms, it would be best to unset them with a setUniforms call. uniforms don't autoreset like settings. Maybe make a note about this for now.

src/lib/draw-and-pick.js
```diff
@@ -354,47 +354,47 @@ function getPickedColors(gl, {
...
+      blendEquation: gl.FUNC_ADD
+    };
+
+    withParameters(gl, settings, () => {
+      // Render all pickable layers in picking colors
+      layers.forEach((layer, layerIndex) => {
+        if (!layer.isComposite && layer.props.visible && layer.props.pickable) {
+
+          // Encode layerIndex with alpha
+          const blendColorSettings = {blendColor: [0, 0, 0, (layerIndex + 1) / 255]};
```
ibgreen: Add a `settings` parameter to layer.drawLayer and pass this in there, instead of using another nested `withParameters`.

Even just a `setParameters` call would work since we are already inside a withParameters, value will be restored on exit.

src/lib/layer.js
```diff
@@ -419,11 +419,15 @@ export default class Layer {
...
-    gl.polygonOffset(offset[0], offset[1]);
+    const offsets = getPolygonOffset && getPolygonOffset(uniforms) || [0, 0];
+    const polygonOffsetSettings = {
+      polygonOffset: offsets
+    };
 
     // Call subclass lifecycle method
-    this.draw({uniforms});
+    withParameters(gl, polygonOffsetSettings, () => {
+      this.draw({uniforms});
```
ibgreen: `this.draw({uniforms, settings: polygonOffsetSettings})`

src/react/deckgl.js
```diff
@@ -109,11 +109,11 @@ export default class DeckGL extends React.Component {
...
   _onRendererInitialized({gl, canvas}) {
-    gl.enable(GL.BLEND);
-    gl.blendFunc(GL.SRC_ALPHA, GL.ONE_MINUS_SRC_ALPHA);
-
-    // Enable polygon offset
-    gl.enable(GL.POLYGON_OFFSET_FILL);
+    setParameters(gl, {
+      blend: true,
+      blendFunc: [GL.SRC_ALPHA, GL.ONE_MINUS_SRC_ALPHA],
+      polygonOffsetFill: true
```
ibgreen: Should we just move this into the draw setting you modified above, in layer.js, rather than flip it on globally?

src/react/webgl-renderer.js
```diff
@@ -145,13 +145,7 @@ export default class WebGLRenderer extends React.Component {
...
     h = h * dpr;
 
-    if (x !== this.x || y !== this.y || w !== this.w || h !== this.h) {
-      gl.viewport(x, y, w, h);
-      this.x = x;
-      this.y = y;
-      this.w = w;
-      this.h = h;
-    }
+    setParameters(gl, {viewport: [x, y, w, h]});
```
ibgreen: Oh, you are relying on the internal caching, nice!

src/layers/core/screen-grid-layer/screen-grid-layer.js
```diff
@@ -79,10 +79,13 @@ export default class ScreenGridLayer extends Layer {
   draw({uniforms}) {
     const {minColor, maxColor} = this.props;
     const {model, cellScale, maxCount} = this.state;
-    const {gl} = this.context;
-    gl.depthMask(true);
     uniforms = Object.assign({}, uniforms, {minColor, maxColor, cellScale, maxCount});
-    model.render(uniforms);
+    model.draw({
```
shaojingli: @1chandu Is there a reason we start using model.draw() here? We always use model.render() in deck.gl before.

1chandu: model.draw() also takes, settings and wraps render within `withParameters` call.

src/lib/draw-and-pick.js
1chandu: yes, I am passing settings to withParameters() below.

src/lib/draw-and-pick.js
1chandu: Yeah agree, given every layer only updates blendColor, I can just use setParameters, `withParameters` may be too heavy and un-necessary here.

src/react/deckgl.js
1chandu: I am matching the old behavior, you suggested change can go into different PR?

src/lib/draw-and-pick.js
1chandu: @ib didn't get your question, what are `two values` ?

src/lib/layer.js
1chandu: Agree, will make this change.

src/lib/draw-and-pick.js
ibgreen: > didn't get your question, what are two values ?

`scissorTest` and `scissorBox` or something like that

1chandu: Oh yeah scissorTest and scissor, this is the old code, you already fixed it :), will reflect when I update the PR.

PR #727, opened 2017-06-21
fix(layer): correct shadercache import

Apercu:


Apercu:
Just a not, realized I will need to bump luma with a new alpha version and bump it there for this fix to make sense

PR #741, opened 2017-06-23
Remove the last vestiges of drag event handling from the framework

ericsoco:
drag events in graph-layer example to follow in another PR.

PR #742, opened 2017-06-23
Graph layer drag refactor

ericsoco:
Update graph example to maange drag state on its own, using mouse handlers and queryObject().

I branched off of #741, so once that's landed I'll just rebase this to remove the duplicate commit from #741.

(Also, looks like this is the first use of `queryLayers()` in our examples. Works like a charm, thx @Pessimistress 🙏 🙆

### DIFF
--- a/examples/graph/app.js
+++ b/examples/graph/app.js
@@ -146,15 +146,15 @@ class Root extends Component {
     });
   }
 
-  _onHover(el) {
-    if (el) {
-      this.setState({hovered: el});
+  _onHover(info) {
+    if (info) {
+      this.setState({hovered: info});
     }
   }
 
-  _onClick(el) {
-    if (el) {
-      this.setState({clicked: el});
+  _onClick(info) {
+    if (info) {
+      this.setState({clicked: info});
     } else {
       const {clicked} = this.state;
       if (clicked) {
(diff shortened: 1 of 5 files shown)


ericsoco:
Along with #741, this will close #648.

examples/graph/app.js
```diff
@@ -162,34 +163,51 @@ class Root extends Component {
     }
   }
 
-  _onDragMove(el) {
+  _onMouseDown(event) {
+    // Use DeckGL.queryObject() to find the object under the mouse,
+    // and store it for updating on mouse move.
+    const point = {x: event.clientX, y: event.clientY};
+    const el = this.deckGL.queryObject(point);
```
ibgreen: Nit: Please type out the name of variables (e.g. `el`). We have called this object `info` or `pickInfo` in other places, not sure if that makes sense any more `queryInfo` perhaps.

examples/graph/app.js
```diff
@@ -162,34 +163,51 @@ class Root extends Component {
...
     if (el) {
-      const {viewport} = this.state;
-      const {width, height} = viewport;
-      const x = el.x - width / 2;
-      const y = el.y - height / 2;
       this.setState({
         clicked: el,
         dragging: {
```
ibgreen: Do you still need this nesting `dragging:{ node: }`
Maybe just `dragNode` and `dragPoint`?

examples/graph/app.js
```diff
@@ -162,34 +163,51 @@ class Root extends Component {
...
-    if (el) {
-      this.setState({
-        dragging: null,
-        lastDragged: {
-          node: el.object
-        }
-      });
+  _onMouseMove(event) {
+    if (this.state.dragging) {
+      const point = {x: event.clientX, y: event.clientY};
```
ibgreen: In almost all other places in this framework stack we use arrays for points. 

```
const point = [event.clientX, event.clientY];
```

examples/graph/app.js
```diff
@@ -162,34 +163,51 @@ class Root extends Component {
...
+      this._updateDraggedElement(point);
     }
   }
 
+  _onMouseUp(event) {
+    this.setState({
+      dragging: null,
+      lastDragged: {
```
ibgreen: ditto

examples/graph/app.js
ericsoco: I can change it in the example, but note that picking uses objects with `x` and `y` props (supersets of points so not exactly apples-to-apples but that's where it came from in this example).

examples/graph/app.js
ericsoco: `{node, x, y}` is expected by the layout.

PR #744, opened 2017-06-23
feat(seer): finalize layer-manager and unbind events

Apercu:


src/lib/layer-manager.js
```diff
@@ -78,21 +82,33 @@ export default class LayerManager {
...
     Object.seal(this.context);
     Object.seal(this);
+  }
 
-    seerInitListener(() => {
-      this.layers.forEach(layer => {
-        initLayerInSeer(layer);
-        updateLayerInSeer(layer);
-      });
+  _initSeer() {
```
ibgreen: Nit: One line comment header would be an opportunity to explain e.g. when this is called and why

src/lib/layer-manager.js
```diff
@@ -78,21 +82,33 @@ export default class LayerManager {
...
+      updateLayerInSeer(layer);
     });
+  }
 
-    layerEditListener(payload => {
-      setPropOverrides(payload.itemKey, payload.valuePath.slice(1), payload.value);
-      const newLayers = this.layers.map(layer => new layer.constructor(layer.props));
-      this.updateLayers({newLayers});
-    });
+  _editSeer(payload) {
```
ibgreen: One line comment header

PR #754, opened 2017-07-05
Avoid deep comparison error in compareProps when oldProp is empty

Pessimistress:
Comparator `equals` from luma.gl's math classes fail when comparing to `null`.


ibgreen:
Looks good. We should probably make sure luma.gl math library doesn't crash on this. maybe open an issue and assign to @1chandu?

PR #755, opened 2017-07-06
Fix the fluctuation of the end cap for path layer 64bit

shaojingli:
The end cap of path layer is unstable in 64-bit mode
@ibgreen @Pessimistress

### DIFF
--- a/src/layers/core/path-layer/path-layer-vertex-64.glsl.js
+++ b/src/layers/core/path-layer/path-layer-vertex-64.glsl.js
@@ -135,8 +135,9 @@ vec3 lineJoin(vec2 prevPoint64[2], vec2 currPoint64[2], vec2 nextPoint64[2]) {
   offsetScale = mix(offsetScale, 1.0 / max(cosHalfA, 0.001), step(0.5, cornerPosition));
 
   // special treatment for start cap and end cap
-  float isStartCap = lenA < 1.0e-6 ? 1.0 : 0.0;
-  float isEndCap = lenB < 1.0e-6 ? 1.0 : 0.0;
+  // using a small number as the limit for determining if the lenA or lenB is 0
+  float isStartCap = step(lenA, 1.0e-5);
+  float isEndCap = step(lenB, 1.0e-5);
   float isCap = max(isStartCap, isEndCap);
 
   // 0: center, 1: side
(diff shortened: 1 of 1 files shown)


src/layers/core/path-layer/path-layer-vertex-64.glsl.js
```diff
@@ -63,20 +63,22 @@ vec3 lineJoin(vec2 prevPoint64[2], vec2 currPoint64[2], vec2 nextPoint64[2]) {
   vec2_sub_fp64(currPoint64, prevPoint64, deltaA64);
   vec2_sub_fp64(nextPoint64, currPoint64, deltaB64);
 
+  vec2 lengthA64 = vec2_length_fp64(deltaA64);
+  vec2 lengthB64 = vec2_length_fp64(deltaB64);
```
shaojingli: Decide to calculate the length of the segments around a joint in 64-bit because they are used to decide if any joint is start/end or not

src/layers/core/path-layer/path-layer-vertex-64.glsl.js
```diff
@@ -133,8 +135,8 @@ vec3 lineJoin(vec2 prevPoint64[2], vec2 currPoint64[2], vec2 nextPoint64[2]) {
   offsetScale = mix(offsetScale, 1.0 / max(cosHalfA, 0.001), step(0.5, cornerPosition));
 
   // special treatment for start cap and end cap
-  float isStartCap = step(0.0, -lenA);
-  float isEndCap = step(0.0, -lenB);
+  float isStartCap = lenA < 1.0e-6 ? 1.0 : 0.0;
```
shaojingli: Using 0.0 with the step function makes the startCap / endCap decision fluctuates in 64-bit mode. A small delta is chosen to make this decision more numerically stable. Is 1e-6 appropriate here? @Pessimistress

Pessimistress: just use step(lenA, 1.0e-6)

Pessimistress:
Per discussion with @shaojingli: when releasing v4 we intentionally removed all if statements and replaced with `mix`. The result has proved to be extremely hard to read and maintain. We should measure the perf gain of this approach to determine if it actually makes sense.

ibgreen:
@georgios-uber

georgios-uber:
LGTM

PR #756, opened 2017-07-06
Update canvas size to match with device framebuffer size.

### LINKED ISSUE #747: Picking broken on devicePixelRatio change
Repro steps:
- In the layer-browser example, use cmd + to increase `window.devicePixelRatio`
- Move mouse. `pickLayers` throws error for negative typed array length.

Cause:
We handle `devicePixelRatio` differently in rendering and picking.
- `WebGLRenderer` uses fixed pixel ratio that is determined when the app loads.
- The picking frame buffer is set to the same size as the WebGL canvas.
- `pickLayers` retrieves pixel ratio every time an event fires to adjust the mouse position, then clips the picking rectangle with the frame buffer size. The current clipping implementation results in negative size when the adjusted mouse position is outside the canvas.

Proposed fixes:
- There should be a single source of truth for pixel ratio, probably managed by `DeckGL` and passed down to both `WebGLRenderer` and `LayerManager`
- `pickLayers` should abort if picking rectangle size is invalid

@ibgreen @shaojingli

1chandu:
- Fixes #747 
- Update canvas size to match with device framebuffer for each frame, considering logical size and `devicePixelRation`
- Use canvas size to set viewport.

@ibgreen @Pessimistress  : not ready to merge, have some open questions

### DIFF
--- a/src/react/webgl-renderer.js
+++ b/src/react/webgl-renderer.js
@@ -25,18 +25,14 @@ import autobind from './autobind';
 import {createGLContext, setParameters} from 'luma.gl';
 /* global requestAnimationFrame, cancelAnimationFrame */
 
-const DEFAULT_PIXEL_RATIO =
-  (typeof window !== 'undefined' && window.devicePixelRatio) || 1;
-
 const propTypes = {
   id: PropTypes.string.isRequired,
 
   width: PropTypes.number.isRequired,
   height: PropTypes.number.isRequired,
+  useDevicePixelRatio: PropTypes.bool.isRequired,
   style: PropTypes.object,
 
-  pixelRatio: PropTypes.number,
-  viewport: PropTypes.object.isRequired,
   events: PropTypes.object,
   gl: PropTypes.object,
   glOptions: PropTypes.object,
@@ -52,7 +48,6 @@ const defaultProps = {
   gl: null,
   glOptions: {preserveDrawingBuffer: true},
   debug: false,
-  pixelRatio: DEFAULT_PIXEL_RATIO,
 
   onInitializationFailed: error => {
     throw error;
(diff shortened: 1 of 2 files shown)


src/lib/viewport-uniforms.js
```diff
@@ -155,6 +155,8 @@ export function getUniformsFromViewport(viewport, {
     modelViewMatrix: new Float32Array(modelViewMatrix),
 
     // Screen size
+    // _Q2# : Verify and remove multiplicaiton with devicePixelRatio (assuming
```
1chandu: @ibgreen @Pessimistress , is multiplication with `devicePixelRation` needed here? Who sets these viewport bounds?

src/react/webgl-renderer.js
```diff
@@ -131,9 +131,52 @@ export default class WebGLRenderer extends React.Component {
...
+      // canvas.style.width = `${cssSize.width}px`;
+      // canvas.style.height = `${cssSize.height}px`;
+    }
+  }
+
   // Updates WebGL viewport to latest props
   // for clean logging, only calls gl.viewport if props have changed
   _updateGLViewport() {
+    /*
     let {viewport: {x, y, width: w, height: h}} = this.props;
```
1chandu: @Pessimistress @ibgreen , how are these props set? Can we ignore this and alway use [0, 0, canvas.width, canvas.height] , given canvas width/height matches device framebuffer size? If need to use this.props, is it guranteed the size provided by props updated with `devicePixelRatio`?

src/react/webgl-renderer.js
```diff
@@ -131,9 +131,52 @@ export default class WebGLRenderer extends React.Component {
     }
   }
 
+  // Gets current size of canvas in css (logical/window) coordinates
+  _getCSSSize(canvas) {
```
1chandu: All these added methods are copy pasted from luma, we can import from `luma` once we finalize the fix.

src/react/webgl-renderer.js
```diff
@@ -131,9 +131,52 @@ export default class WebGLRenderer extends React.Component {
...
+  // Intention is that every pixel in the drawing buffer will have a 1-to-1 mapping with
+  // actual device pixels in the hardware framebuffer, allowing us to render at the full
+  // resolution of the device.
+  _calculateDrawingBufferSize(canvas, {useDevicePixelRatio = true}) {
+    const cssToDevicePixels = useDevicePixelRatio ? window.devicePixelRatio || 1 : 1;
+
+    // Lookup the size the browser is displaying the canvas in CSS pixels
+    // and compute a size needed to make our drawingbuffer match it in
+    // device pixels.
+    const cssSize = this._getCSSSize(canvas);
```
Pessimistress: Just use `this.props.width` and `this.props.height`.

src/react/webgl-renderer.js
Pessimistress: The size in props is not multiplied by `devicePixelRatio`. It is the `width` and `height` that users pass to the `DeckGL` component.

src/react/webgl-renderer.js
```diff
@@ -131,9 +131,52 @@ export default class WebGLRenderer extends React.Component {
...
+  // Resizes canvas width and height to match with device drawing buffer
+  _resizeDrawingBuffer(canvas, {useDevicePixelRatio = true}) {
+    // Resize the render buffer of the canvas to match canvas client size
+    // multiplying with dpr (Optionally can be turned off)
+    const newBufferSize = this._calculateDrawingBufferSize(canvas, {useDevicePixelRatio});
+    // Only update if the canvas size has not changed
+    if (newBufferSize.width !== canvas.width || newBufferSize.height !== canvas.height) {
+      // Make the canvas render buffer the same size as
+      canvas.width = newBufferSize.width;
+      canvas.height = newBufferSize.height;
```
Pessimistress: These are overwriting what's set in the `render()` method. Should remove the `width` and `height` assignment in `render` to avoid confusion.

Pessimistress:
Remove the `pixelRatio` prop if it's not used?

src/lib/viewport-uniforms.js
Pessimistress: Yes. `DeckGL` defined `viewport` from user provided props. The uniform is used by layers to project pixel size to clipspace, which requires the actual canvas size.

Pessimistress:
After this change we still have multiple copies of this line in the code:
`const devicePixelRatio = (window && window.devicePixelRatio) || 1;`
Can we consolidate it to either `DeckGL` or `LayerManager` and pass down to `WebGLRenderer`, `draw-and-pick` and `viewport-uniforms`?

src/react/webgl-renderer.js
Pessimistress: `clientWidth` and `clientHeight` have perf impacts. Since we already know the canvas' `width` and `height` styles (set in the `render` method) we don't need the work around here.

ibgreen: So this is one of trickiest aspects of the canvas API. Namely that `width` and `height` control the size of the drawing buffer, not the DOM element, and are set independently from the style.width, height (which are read back as` clientWidth` and `clientHeight`). We can't keep reading width and height since they will be multiplied by DPR, we'll keep multiplying the drawing buffer size. If there is a reflow issue with `clientWidth` and `clientHeight` we could use the `props.width` and `props.height` here as they are intended to be in logical coordinates.

1chandu:
@ibgreen @Pessimistress , created this #760 to track remaining issues.

src/react/webgl-renderer.js
```diff
@@ -131,31 +126,53 @@ export default class WebGLRenderer extends React.Component {
...
+  // resolution of the device.
+  _calculateDrawingBufferSize(canvas, {useDevicePixelRatio = true}) {
+    const cssToDevicePixels = useDevicePixelRatio ? window.devicePixelRatio || 1 : 1;
+
+    // Lookup the size the browser is displaying the canvas in CSS pixels
+    // and compute a size needed to make our drawingbuffer match it in
+    // device pixels.
+    return {
+      width: Math.floor(canvas.clientWidth * cssToDevicePixels),
+      height: Math.floor(canvas.clientHeight * cssToDevicePixels),
```
Pessimistress: Just use `this.props.width` and `this.props.height`?
`clientWidth` and `clientHeight` causes page reflow. We already know the canvas css size since we set it here in `render()`.

1chandu: What is `page reflow`, is it because clientWidth/clientHeight can be different than props.width/height?

Based on discussion with @ibgreen (also I verified changing window size), the width and height extracted from props and set to style object in `render()`, will be set to `canvas.clientWidth` and `canvas.clientHeight`, so they should identical.

Pessimistress: The browser attempts to recalculate the layout of the whole page when you access certain layout properties. It triggers a lot more action under the hood than just reading the `props` object.

PR #757, opened 2017-07-06
MapController clean up

Pessimistress:
- Remove `InteractiveMap` and old copy of `EventManager`
- Copied `mapControls` and `mapState` from react-map-gl. I assume the goal is to eventually extract all controllers to a separate package?
- Add example


examples/without-map/app.js
```diff
@@ -0,0 +1,74 @@
+/* global window,document,fetch */
+import React, {Component} from 'react';
+import {render} from 'react-dom';
+import {MapController} from 'deck.gl/dist/controllers';
```
ericsoco: Why not expose `MapController` as a top-level export of `deck.gl` so you can just
`import {MapController} from 'deck.gl';`
and eliminate the change to webpack config?

examples/webpack.config.local.js
```diff
@@ -26,6 +26,7 @@ const LOCAL_DEV_CONFIG = {
   resolve: {
     alias: {
       // Imports the deck.gl library from the src directory in this repo
+      'deck.gl/dist': SRC_DIR,
```
ibgreen: Good to add a comment explaining why we add this.

examples/without-map/app.js
ibgreen: > Why not expose MapController as a top-level export of deck.gl so you can just import {MapController} from 'deck.gl'; and eliminate the change to webpack config?

I assume we could make it an experimental export. On the downside, it will add to the bundle size of all applications, so we should have a decent use case.

src/controllers/index.js
```diff
@@ -1 +1,2 @@
-export {default as default} from './interactive-map.js';
+export {default as MapController} from './map-controller/map-controller';
+export {default as OrbitController} from './orbit-controller';
```
ibgreen: orbit-controller/orbit-controller?

PR #759, opened 2017-07-06
Remove gl-matrix

Pessimistress:


src/layers/deprecated/extruded-choropleth-layer-64/extruded-choropleth-layer-64.js
```diff
@@ -24,7 +24,7 @@ import {fp64ify} from '../../../lib/utils/fp64';
 import {GL, Model, Geometry} from 'luma.gl';
 import {flatten, log} from '../../../lib/utils';
 import earcut from 'earcut';
-import {vec3} from 'gl-matrix';
+import normalize from 'gl-vec3/normalize';
```
ibgreen: call it vec3_normalize for consistency, here and above?

PR #761, opened 2017-07-07
OrbitController clean up

Pessimistress:


src/controllers/orbit-controller/orbit-state.js
```diff
@@ -0,0 +1,377 @@
...
+  _applyConstraints(props) {
+    // Ensure zoom is within specified range
+    const {maxZoom, minZoom, zoom} = props;
+    props.zoom = zoom > maxZoom ? maxZoom : zoom;
+    props.zoom = zoom < minZoom ? minZoom : zoom;
+
+    return props;
+  }
+
+  /* Cast a ray into the screen center and take the average of all
```
ibgreen: A lot going on here. Should this go into its own file, together with all the helper functions?

A ray casting algorithm would be a nice addition to the library. Maybe this could become a starting point for that?

src/lib/viewports/orbit-viewport.js
```diff
@@ -0,0 +1,125 @@
+import Viewport, {createMat4} from './viewport';
+
+import mat4_multiply from 'gl-mat4/multiply';
```
ibgreen: ditto.

src/lib/viewports/orbit-viewport.js
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  unproject(xyz, {topLeft = false} = {}) {
+    const [x, y, z] = xyz;
+    const y2 = topLeft ? this.height - y : y;
+
+    return this.transformVector(this.pixelUnprojectionMatrix, [x, y2, z, 1]);
+  }
+
+  fitBounds([min, max]) {
```
ibgreen: Interesting parameter choice, array of arrays. Worth returning to in an API audit when we make this public. 

Essentially deciding on a convention for handling bounding boxes. E.g. THREE.js has a BoundingBox class.

src/controllers/orbit-controller/orbit-state.js
```diff
@@ -0,0 +1,377 @@
+import OrbitViewport from '../../lib/viewports/orbit-viewport';
+import vec3_add from 'gl-vec3/add';
```
Pessimistress: I'm still expecting to extract these viewport control components to their own package some day (and share between react-map-gl and deck.gl). The Vector3 class requires luma.gl as a dependency.

src/controllers/orbit-controller/orbit-state.js
Pessimistress: I made some preliminary experiments and still not entirely happy with the result of this function. It's very much a hack. I'll poke around some more to get this to a useful state.

src/controllers/orbit-controller/orbit-state.js
ibgreen: > The Vector3 class requires luma.gl as a dependency.

No worries, we will break out the math package as well. Most likely luma.gl will become a monorepo, publishing multiple modules, but we could make it into a separate repo.

PR #762, opened 2017-07-07
TagMap: add shouldUpdateState function

rivulet-zhang:
This worked well in the previous deck.gl version. But in the recently released beta version, the `tagMap` example doesn't call the `updateState` function when the viewport changes. Adding the `shouldUpdateState` function fixed the issue though.


ibgreen:
@rivulet-zhang Good catch

For performance reasons, in v4.1 we don't "disturb" the layers when only the viewport changed, as most layers don't need to do anything in this case. 

And yes, it does mean that the few layers that do calculations in "screen space" do need to override `shouldUpdateState`

I thought this was mentioned in the upgrade guide but it looks like it isn't, will open a ticket.

PR #764, opened 2017-07-08
Convert to luma.gl shadertools

ibgreen:


examples/plot/plot-layer/axes-layer.js
```diff
@@ -284,7 +283,7 @@ export default class AxesLayer extends Layer {
       }
     }
 
-    const labels = new Model({
+    const labels = new Model(gl, {
       gl,
```
shaojingli: remove `gl` here?

src/layers/core/arc-layer/arc-layer.js
```diff
@@ -42,17 +42,10 @@ const defaultProps = {
...
-      shaderCache: this.context.shaderCache
-    } : {
-      vs: arcVertex,
-      fs: arcFragment,
-      modules: [],
-      shaderCache: this.context.shaderCache
-    };
+    const {shaderCache} = this.context;
+    return enable64bitSupport(this.props) ?
+      {vs: vs64, fs, modules: ['project64'], shaderCache} :
```
shaojingli: so now fp64 modules are attached by default?

examples/plot/plot-layer/axes-layer.js
ibgreen: Good catch. Fixing in separate PR.

src/layers/core/arc-layer/arc-layer.js
ibgreen: There are no default modules anymore. Instead, shader modules now can specify `dependencies` on other modules and the luma.gl `assembleShaders` automatically resolves the resulting DAG (directed acyclic graph) and injects dependencies recursively and deduplicated in the right order.

PR #765, opened 2017-07-08
Fine tune shader module dependencies

ibgreen:


src/shaderlib/fp64/fp64.js
```diff
@@ -0,0 +1,44 @@
...
+  vs: `${fp64arithmeticShader}`,
+  fs: null
+};
+
+// Fragment shader fp64
+export const fp64fs = {
+  name: 'fp64-fs',
+  vs: null,
+  fs: fp64shader
+};
```
shaojingli: It seems to me that luma.gl's assembleShader() works very differently from deck.gl's assembleShaders(). So each module has name, vs, and fs field. How does this work? and I also don't quite understand how fp64fs works. We probably could discuss offline.

src/shader-utils/index.js
```diff
@@ -20,16 +20,22 @@
...
+import fp32 from '../shaderlib/fp32/fp32';
+import fp64, {fp64arithmetic, fp64fs} from '../shaderlib/fp64/fp64';
+import project from '../shaderlib/project/project';
+import project64 from '../shaderlib/project64/project64';
+import lighting from '../shaderlib/lighting/lighting';
 
 import {registerShaderModules} from 'luma.gl';
 
-registerShaderModules([fp64, project, project64, lighting], {});
+registerShaderModules([
```
shaojingli: Do we plan to move those modules to luma.gl as well?

src/shaderlib/fp64/fp64.js
ibgreen: It is actually very similar, the big difference is support for `dependencies` and more advanced `getUniforms`. I just opened https://github.com/uber/luma.gl/pull/258, with hopefully better documentation of the new shader module system.

About `fp64fs` - The complication here is that many shader modules are matched pairs of fragment and vertex shaders, but some like `fp64` can be useful as a general library either in vertex or fragment (e.g. your 64 bit Mandelbrot example could use `fp64fs`).

I admit it is a little clunky, I think we may be able to come up with a better API for this.

src/shader-utils/index.js
ibgreen: Yes, except the `project`/`project64` modules (those have web mercator functionality which I think belongs to deck.gl).

PR #767, opened 2017-07-10
Additional shadertools alignment

ibgreen:
* Updates all layers to just add `modules` parameter to `Model` instead of calling assembleShaders explicitly.
* Moves `getUniformsFromViewport` into `project` package.
* Some doc, test reorg.


src/index.js
```diff
@@ -49,14 +46,17 @@ export {default as PathLayer} from './layers/core/path-layer/path-layer';
...
 // Deprecated Layers
 export {default as ChoroplethLayer} from './layers/deprecated/choropleth-layer/choropleth-layer';
 export {default as ChoroplethLayer64} from './layers/deprecated/choropleth-layer-64/choropleth-layer-64';
 export {default as ExtrudedChoroplethLayer64} from './layers/deprecated/extruded-choropleth-layer-64/extruded-choropleth-layer-64';
 
-// React exports
-export {default as DeckGL} from './react/deckgl';
-export {default as default} from './react/deckgl';
+// Deprecated Exports
+export {assembleShaders} from './shader-utils'; // Just a forward for the luma.gl version (integrated with Model class)
```
shaojingli: We don't need to _deprecate_ `assembleShaders()` as it has never been exported before

src/lib/viewports/viewport.js
```diff
@@ -245,30 +201,72 @@ export default class Viewport {
...
+     * Note: Starts with the GL projection matrix and adds steps to the
+     *       scale and translate that matrix onto the window.
+     * Note: WebGL controls clip space to screen projection with gl.viewport
+     *       and does not need this step.
+     */
+
+    // matrix for conversion from location to screen coordinates
+    const m = createMat4();
+    mat4_scale(m, m, [this.width / 2, -this.height / 2, 1]);
+    mat4_translate(m, m, [1, -1, 0]);
```
1chandu: Is this translation supposed to move the center of the window to origin (before above scaling is applied to match with viewport bounds) ? Why [1, -1, 0], a comment will help.

shaojingli: The translation is to invert the y-axis. You know, y axis is different in the map space and screen space. This set of translation is copied from Mapbox and they do have a weird translation system for viewport. I can explain to you in details later

ibgreen:
> So in this PR, we merged explicit calls to assembleShaders to the constructor of the Model class of luma.gl. We probably want to consider carefully if we'd like to do that or not. The Model class of luma.gl is already pretty large, is it a good idea to continue adding functionality to it?

Good points. I'll hold the PR until we have settled this.

One of the complication is that the shader modules now come with JS function that can be used to set the uniforms for each module, and `assembleShaders` generates a composite setter function that calls these functions on each module using a composite props object. Someone needs to keep track of this generated function, the `Model` does it automatically. It would be more code if we kept `assembleShaders` separate and even more code if we didn't have `Model` manage it.

> It also makes debugging shaders harder as developers can no longer easily output complete shader strings from deck.gl.

They are still available but it requires a little more digging. We are also planning to surface them in the seer extension. We could do additional things to make it easier to get to them.

PR #770, opened 2017-07-11
Fix `getPolygonOffset` prop

Pessimistress:
`polygonOffset` parameter was not applied.

### DIFF
--- a/src/lib/layer.js
+++ b/src/lib/layer.js
@@ -25,7 +25,7 @@ import Stats from './stats';
 import {getDefaultProps, compareProps} from './props';
 import {log, count} from './utils';
 import {applyPropOverrides, removeLayerInSeer} from '../debug/seer-integration';
-import {GL} from 'luma.gl';
+import {GL, withParameters} from 'luma.gl';
 import assert from 'assert';
 
 const LOG_PRIORITY_UPDATE = 1;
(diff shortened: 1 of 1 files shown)


src/lib/layer.js
```diff
@@ -446,7 +446,10 @@ export default class Layer {
     settings.polygonOffset = offsets;
 
     // Call subclass lifecycle method
-    this.draw({moduleParameters, uniforms, settings});
+    withParameters(this.context.gl,
```
ibgreen: The original idea was that we would forward the settings to the layer's draw function, but we can certainly apply them here. 

We wanted to support `props.settings`, this is a nice way to do it. Won't allow overriding the layers hardcoded draw settings though, but maybe that is a good thing.

Not sure if we are actually supplying `props.settings `somewhere, maybe they are already supplied by layer-manager or draw-and-pick?)

Pessimistress: polygonOffset was a feature that layers used to get for free. Do we have a good reason to ask layers to do more work in their `draw` methods now?

Pessimistress: Also `settings` is not documented - do we have a good use case for it? If yes, maybe should call it `glParameters` to be clear?

ibgreen: There should be many use cases for a layer `settings` prop (`props.parameters` or `props.glParameters`).  An editing app that I work on now wants to control blend modes per layer, and this parameter would handle that nicely with our stock layers.

Exposing this gives a ton of flexibility to apps, at a very low API cost for deck.gl (luma.gl already does the heavy lifting of allowing parameters to be specified as keys in objects). 

No objections to changing to `glParameters` although it would be (partly) inconsistent with luma.gl, where the the new `Model.draw({uniforms, ..., settings})` function takes a `settings` field. That said, the global luma.gl **function** is called `setParameters`.

I'd probably vote for `parameters` and if so, maybe update `Model.draw({settings})` to take `Model.draw({parameters})` as well?

@1chandu @shaojingli

Issue #769, opened 2017-07-11
Perf regression with deck.gl@4.1.0-beta.2 and luma.gl@4.0.0-beta.3


Pessimistress:

shaojingli:
should we close this now? @Pessimistress

PR #774, opened 2017-07-11
Remove a redundant withParameters call in draw-and-pick

shaojingli:
@Pessimistress Could you try this branch to see if the performance regression is still there?
#769

### DIFF
--- a/src/lib/draw-and-pick.js
+++ b/src/lib/draw-and-pick.js
@@ -325,6 +325,8 @@ function getPickedColors(gl, {
   // Make sure we clear scissor test and fbo bindings in case of exceptions
   // We are only interested in one pixel, no need to render anything else
   // Note that the callback here is called synchronously.
+  // Set blend mode for picking
+  // always overwrite existing pixel with [r,g,b,layerIndex]
   return withParameters(gl, {
     framebuffer: pickingFBO,
     scissorTest: true,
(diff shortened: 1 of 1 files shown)


src/lib/draw-and-pick.js
```diff
@@ -338,34 +341,20 @@ function getPickedColors(gl, {
...
-
-    withParameters(gl, settings, () => {
-      // Render all pickable layers in picking colors
-      layers.forEach((layer, layerIndex) => {
-        if (!layer.isComposite && layer.props.visible && layer.props.pickable) {
-
-          // Encode layerIndex with alpha
-          setParameters(gl, {blendColor: [0, 0, 0, (layerIndex + 1) / 255]});
-          layer.drawLayer({
-            moduleParameters: Object.assign({}, layer.props, {
```
Pessimistress: Why remove?

src/lib/draw-and-pick.js
```diff
@@ -338,34 +341,20 @@ function getPickedColors(gl, {
...
-              viewport: layer.context.viewport
-            }),
-            uniforms: Object.assign(
-              {renderPickingBuffer: 1, pickingEnabled: 1},
-              layer.context.uniforms,
-              {layerIndex}
-            ),
-            settings: layer.props.settings || {}
```
Pessimistress: And this?

src/lib/draw-and-pick.js
```diff
@@ -338,34 +341,20 @@ function getPickedColors(gl, {
...
+    layers.forEach((layer, layerIndex) => {
+      if (!layer.isComposite && layer.props.visible && layer.props.pickable) {
+
+        // Encode layerIndex with alpha
+        setParameters(gl, {blendColor: [0, 0, 0, (layerIndex + 1) / 255]});
+        layer.drawLayer({
+          uniforms: Object.assign(
+            {renderPickingBuffer: 1, pickingEnabled: 1},
+            layer.context.uniforms,
+            getUniformsFromViewport(layer.context.viewport, layer.props),
```
Pessimistress: This function has been removed.

src/lib/draw-and-pick.js
shaojingli: the settings used in this withParameters() has been accommodated by the outer withParameters() call at line 328

Pessimistress: I was referring to moduleParameters, but it seems like a bad merge before.

ibgreen: FWIW, moduleParameters is part of the new shader module system

PR #775, opened 2017-07-11
Fix picking perf when dragging

Pessimistress:
- Update EventManager
- Add `isDown` property to pointer events
- Prevent picking when dragging


src/utils/events/event-manager.js
```diff
@@ -136,6 +142,8 @@ export default class EventManager {
    * Process the event deregistration for a single event + handler.
    */
   _removeEventHandler(event, handler) {
+    let success = false;
```
ericsoco: nit: call this `eventHandlerRemoved` to be a bit more explicit about what succeeded

src/utils/events/event-manager.js
```diff
@@ -163,7 +186,6 @@ export default class EventManager {
       };
 
       // Calculate center relative to the root element
-      // TODO/xiaoji - avoid using getBoundingClientRect for perf?
```
ericsoco: Did you decide this is not a perf concern?

src/utils/events/event-manager.js
```diff
@@ -191,7 +213,7 @@ export default class EventManager {
     const alias = BASIC_EVENT_ALIASES[srcEvent.type];
     if (alias) {
       // fire all events aliased to srcEvent.type
-      const emitEvent = Object.assign({}, event, {type: alias});
+      const emitEvent = Object.assign({}, event, {isDown: true, type: alias});
```
ericsoco: Looks like you're always setting `isDown: true` here, should the third `Object.assign()` param be the second?

src/utils/events/event-manager.js
Pessimistress: I don't find it prominent during profiling.

src/utils/events/event-manager.js
Pessimistress: we're overriding `event.type`.

ericsoco:
Worth noting that this PR fixes the more obvious perf problem during dragging (perf slows down the screen update while dragging), but not the more subtle perf problem that happens when running picking on hover (mousemove without mousedown). The latter is for a future PR.

src/utils/events/event-manager.js
ibgreen: May not be conclusive. It reportedly depends on the surrounding DOM, how deep we are in the DOM tree, styling etc.

Whether it is worth the trouble to avoid I don't know...

src/utils/events/event-manager.js
```diff
@@ -1,6 +1,6 @@
 import WheelInput from './wheel-input';
 import MoveInput from './move-input';
-import {isBrowser} from '../globals';
+import {isBrowser} from '../../lib/utils/globals';
```
ibgreen: Some of these utils should be moved up a level if they are no longer used only inside lib.

src/lib/layer-manager.js
```diff
@@ -594,6 +594,10 @@ export default class LayerManager {
    *                        {Object} srcEvent:             native JS Event object
    */
   _onPointerMove(event) {
+    if (event.isDown) {
+      // Is dragging
```
ibgreen: Can we add a little more to the comment  i.e. why does dragging mean that we don't need move events?

src/utils/globals.js
```diff
@@ -1,15 +0,0 @@
-// Purpose: include this in your module to avoids adding dependencies on
```
ibgreen: Why was this deleted?

src/lib/layer-manager.js
shaojingli: this comment means we should not fire pointermove event when mouse is down because pointermove leads to pickLayer getting called. Yes, we probably want to explain a bit more in the comments, or considering put a link to this PR

ericsoco: Well, more specifically `_onPointerMove` feeds into our hover handlers (`onHover` and `onLayerHover`), and dragging is not supported by the framework itself, it's the app's responsibility. But yeah, fleshing out the comment will help future detective work.

Pessimistress:
https://github.com/uber/deck.gl/issues/769

Issue #266, opened 2016-12-20
Cannot run exhibits.


davidkell:
**Problem**

Cannot run the webpack exhibits example.

**Steps to reproduce**

`git clone https://github.com/uber/deck.gl`
`cd deck.gl`
`cd exhibits`
`npm run webpack`

I am using yarn.
Machine: MacBook Pro 11" Retina with macOS Sierra

**Error**

```
ERROR in Cannot find module 'babel-core'
 @ multi main
```

I have tried to add babel-core as a dependency.
However I then run into a bunch of other errors.

Thanks

rajshah4:
I was able to run the app.js in exhibits by doing the following: (with maybe unnecessary step)
```
git clone https://github.com/uber/deck.gl
cd deck.gl
npm install
cd exhibits
npm run webpack
```

davidkell:
OK- followed these steps- get a new error.

Fails when it runs `webpack-dev-server --config config.js --progress --hot --port 3000`

Also- if we need to install deck.gl in the upper directory, how do we extract this exhibit into a new standalone project with webpack as the builder?

Thanks

rajshah4:
I don't know the answer for making this is standalone.  Here is what I am using as my app.js  . . . updated it with your mapbox key and hopefully you should see a map and the drawing layer: 
https://gist.github.com/rajshah4/e703768b3dd65060387ab168a8d7c000

davidkell:
That's interesting - that code you've linked is different to mine. Are you on a different branch?

rajshah4:
Hmm, I wrote this App.js about two weeks ago.  I haven't been closely following all the developments, but I know its enough for me to show others how deck.gl works.  Does this work for you?

davidkell:
Afraid not :( Do you have any idea what that web console log means? WOuld love to get this working!

rajshah4:
I don't know about the error, but feel free to contact me and I can try to help you troubleshoot this.

Apercu:
Could you please post your node, yarn and osx versions?

@rajshah4 Thanks for your help but it would be best if we could stick with the original example rather than introduce new overhead. Trimming the conversation.

davidkell:
node: v7.2.1
osx: v10.12.1
yarn: v0.17.10

This is the log--

```
bundle.js:81620 deck.gl 3.1.1 - set deck.log.priority=2 to trace attribute updates
[...]

PR #779, opened 2017-07-12
Add `project` as a default shader module.

1chandu:
Note: Requires luma.gl's #266.


Pessimistress:
Is the test failure due to using outdated luma.gl?
Can you bump the dependency version and remove the inclusion of `project` module in all core layers/examples?

1chandu:
Yeah the failures are due to an issue this change exposes in dependencyGraph parsing in luma, which will be fixed by (#266).
Ok, I will remove project module setting in layers/examples.

PR #781, opened 2017-07-12
Add the missing radius mix/max and scale control for GeoJSON layer

shaojingli:
@Pessimistress as we discussed yesterday.

### DIFF
--- a/docs/layers/geojson-layer.md
+++ b/docs/layers/geojson-layer.md
@@ -112,6 +112,23 @@ Type of joint. If `true`, draw round joints. Otherwise draw miter joints.
 The maximum extent of a joint in ratio to the stroke width.
 Only works if `lineJointRounded` is `false`.
 
+##### `pointRadiusScale` (Number, optional)
+
+- Default: `1`
+
+Global radius across all points.
+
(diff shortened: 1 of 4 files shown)


shaojingli:
@Pessimistress should we call the new props `pointRadius***` or `markerRadius***` or something else?

Pessimistress:
We don't use the word "marker" anywhere. I like "point" because it refers to the corresponding GeoJson feature type.

docs/layers/geojson-layer.md
```diff
@@ -112,6 +112,23 @@ Type of joint. If `true`, draw round joints. Otherwise draw miter joints.
 The maximum extent of a joint in ratio to the stroke width.
 Only works if `lineJointRounded` is `false`.
 
+##### `pointRadiusScale` (Number, optional)
+
+- Default: `1`
+
+Global radius across all points.
```
Pessimistress: Global radius multiplier?

shaojingli: I'm just copying from the scatterplot layer doc... I will change both

shaojingli:
I feel the same. Was asking because we use "marker" in our scatterplot layer doc. I will clean up

PR #783, opened 2017-07-12
Hex grid layer methods

heshan0131:
For easier subclassing:
- Add static methods to layer class
- Add method `getSubLayer ` to get SubLayer class
- Add method `getSubLayerProps` to get SubLayer props

### DIFF
--- a/src/layers/core/hexagon-layer/hexagon-layer.js
+++ b/src/layers/core/hexagon-layer/hexagon-layer.js
@@ -210,6 +210,8 @@ export default class HexagonLayer extends CompositeLayer {
   }
 
   getSubLayerProps() {
+    // for subclassing, override this method to return
+    // customized sub layer props
     const {id, radius, elevationScale, extruded, coverage, lightSettings, fp64} = this.props;
 
     // base layer props
@@ -218,6 +220,7 @@ export default class HexagonLayer extends CompositeLayer {
     // viewport props
     const {positionOrigin, projectionMode, modelMatrix} = this.props;
 
+    // return props to the sublayer constructor
     return {
       id: `${id}-hexagon-cell`,
       data: this.state.hexagons,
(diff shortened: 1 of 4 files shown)


shaojingli:
@heshan0131 can you give a brief explanation why we need this change and what it resolves? is it for people to extend Hexagon / Cell layer more easily?

heshan0131:
Yes.  It would be a lot easier to break layer class into smaller methods for Subclassing
For example, I am working on a Subclass of hexagon layer where I want to render a customized cell layer that uses customized shader. To do this, I only need to overide the getSubLayer methods, instead of rewrite the entire renderSubLayer call.

src/layers/core/hexagon-layer/hexagon-layer.js
```diff
@@ -238,7 +239,19 @@ export default class HexagonLayer extends CompositeLayer {
...
-    });
+    };
+  }
+
+  getSubLayer() {
+    return HexagonCellLayer;
+  }
+
+  renderLayers() {
+    const SubLayerComp = this.getSubLayer();
```
shaojingli: Can we change the name of this variable? `SubLayerComp` doesn't provide much info....

src/layers/core/hexagon-layer/hexagon-layer.js
```diff
@@ -238,7 +239,19 @@ export default class HexagonLayer extends CompositeLayer {
       getColor: this._onGetSublayerColor.bind(this),
       getElevation: this._onGetSublayerElevation.bind(this),
       updateTriggers: this.getUpdateTriggers()
-    });
+    };
+  }
+
+  getSubLayer() {
```
shaojingli: Also this, can we change it to `getSubLayerClass`

src/layers/core/hexagon-layer/hexagon-layer.js
```diff
@@ -217,7 +218,7 @@ export default class HexagonLayer extends CompositeLayer {
     // viewport props
     const {positionOrigin, projectionMode, modelMatrix} = this.props;
 
-    return new HexagonCellLayer({
+    return {
```
shaojingli: Add a comment saying `// return a bunch of props to the sublayer constructor specified in  getSubLayer()`

heshan0131:
rebased on updated

PR #786, opened 2017-07-14
Update doc for updateTriggers

Pessimistress:
Per feedback of
https://github.com/uber/deck.gl/issues/577#issuecomment-314934239

- Better explanation of `updateTriggers` in layer documentation
- Remove arbitrary trigger keys from examples to avoid confusion

### DIFF
--- a/docs/api-reference/base-layer.md
+++ b/docs/api-reference/base-layer.md
@@ -204,17 +204,49 @@ manually override it using this prop.
 
 ##### `updateTriggers` (Object, optional)
 
-This prop expects an object of which the keys matching the accessor names of a layer.
-If any values in this object are changed between props update, the attribute corresponding
-to the accessors named by the key will be invalidated.
+Accessors such as `getColor` and `getPosition` are called to retrieve colors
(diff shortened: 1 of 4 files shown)


docs/api-reference/base-layer.md
```diff
@@ -204,17 +204,49 @@ manually override it using this prop.
...
+  const layer = new ScatterplotLayer({
+    ... // Other props
+    getColor: d => d.male ? maleColor : femaleColor,
+    updateTriggers: {
+        getColor: [maleColor, femaleColor]
+    }
+  });
+```
+
+`updateTriggers` expect an object whose keys are names of accessor props of this layer,
```
ericsoco: expect**s**

ericsoco: lets remove mention of `object` for API clarity, even if it's still supported under the hood. Instead, `a value or array of values` should suffice.

Pessimistress: `updateTriggers` itself is still an object (accessor name to array mapping).

PR #787, opened 2017-07-15
ShaderModule: Remove supplying 'project' module, fix 'lighting' module dependency.

1chandu:
* Now that 'project' module is added by default, remove code supplying to `model` object.
* Add dependency `lighting` -> `project`, so modules are inserted in right order.
* Tested with : test-browser, layer-browser, examples.
NOTE: not ready to merge, requires luma.gl change (https://github.com/uber/luma.gl/pull/270)

### DIFF
--- a/docs/upgrade-guide.md
+++ b/docs/upgrade-guide.md
@@ -6,14 +6,6 @@ A highly compatible release, the biggest change is that deck.gl v4.1 brings in l
 
 ## Shader Assembly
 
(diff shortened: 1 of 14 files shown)


Pessimistress:
@1chandu Can you also check and make sure that the custom layers in `trips` and `plot` examples work?

1chandu:
@Pessimistress , yeah both `trips` and `plot` examples run fine with these changes.
@shaojingli , yeah we need bot this and luma.gl PR should be landed.

PR #794, opened 2017-07-19
Add Wind Layer (with luma v4 api).

1chandu:
- Update wind demo to use luma v4 api. (contributions from @gnavvy  and myself) (squashing all commits in`origin/wind-luma-v4` into 1.)
- This maintains parity with old wind demo (using raw gl), I will make another PR to cleanup/optimize the code.


examples/wind/src/layers/particle-layer/program-transform-feedback.js
```diff
@@ -0,0 +1,18 @@
+import {Program} from 'luma.gl';
+
+export default class ProgramTransformFeedback extends Program {
```
gnavvy: This file is not used anymore. shall we add TODOs for the cleanup PR.

examples/wind/src/layers/wind-layer/wind-layer.js
```diff
@@ -0,0 +1,228 @@
...
+    const {gl} = this.context;
+    const {dataTextureSize, originalBoundingBox} = this.props;
+
+    // FIXME - Layer API for async loading
+    const data = {};
+    const image = new Image(584, 253);
+    image.onload = () => {
+      data.img = image;
+    };
+    image.src = ELEVATION_DATA_IMAGE;
```
gnavvy: better to use the new loadTextures API, let's sync on how to integrate it with the createTextureNew() function in DelaunayInterpolation

1chandu: Sure, makes sense, we can do this as part of cleanup PR.

examples/wind/src/layers/particle-layer/program-transform-feedback.js
1chandu: Yeah, wanted to keep this file in this commit so we can retrieve if needed, we can remove this in following PR.

1chandu:
@shaojingli @Pessimistress Removed `data` and `yarn.lock`.

Issue #795, opened 2017-07-20
Invalid prop `viewport` of type `OrbitViewport` supplied to `DeckGL`


dazhi1011:
when I run the 'deck.gl-4.1.0-beta.6\examples\plot' example,there is an error like this:
**## "Failed prop type: Invalid prop `viewport` of type `OrbitViewport` supplied to `DeckGL`, expected instance of `Viewport`."**
but the `OrbitViewport` is extended from `Viewport`... any suggestions here?

shaojingli:
For some reason, the `OrbitViewport` is not recognized as a subclass of `Viewport` when running from the transpiled deck.gl. This shows up only when using `npm run start`, NOT when using `npm run start-local`. Any ideas, @Pessimistress ?

shaojingli:
This is due to we import `OrbitViewport` directly from `deck.gl/dist` and can be resolved by importing it from `deck.gl/dist-es6` instead. This is a quick fix and will break this example on older browsers. Eventually, we should export the `OrbitViewport`.

dazhi1011:
thx,fix it by changing the import path. @shaojingli

PR #796, opened 2017-07-20
Import Viewport from dist-es6 folder

shaojingli:
and also remove an unused file
#795


examples/plot/app.js
```diff
@@ -2,7 +2,7 @@
 import React, {Component} from 'react';
 import {render} from 'react-dom';
 import DeckGLOverlay from './deckgl-overlay.js';
-import {OrbitController} from 'deck.gl/dist/controllers';
+import {OrbitController} from 'deck.gl/dist-es6/controllers';
```
Pessimistress: This will not work for older browsers.
Instead, we need to make all controllers import `Viewport` from `deck.gl` instead of a relative path.

src/controllers/orbit-controller/orbit-viewport.js
```diff
@@ -1,4 +1,4 @@
-import Viewport, {createMat4} from '../../lib/viewports/viewport';
+import {Viewport} from 'deck.gl';
```
shaojingli: @Pessimistress is this what you meant in your comment?

PR #799, opened 2017-07-21
Wind: cleanup of texture creation code

1chandu:
- Remove lot of redundant/unused props in texture creation path.

### DIFF
--- a/examples/wind/src/layers/particle-layer/particle-layer.js
+++ b/examples/wind/src/layers/particle-layer/particle-layer.js
@@ -1,8 +1,10 @@
 import {Layer} from 'deck.gl';
-import {GL, Model, Geometry, Buffer, TransformFeedback, setParameters, loadTextures} from 'luma.gl';
+import {
+  GL, Model, Geometry, Buffer, TransformFeedback,
+  setParameters, loadTextures, Texture2D
+} from 'luma.gl';
 import ProgramTransformFeedback from './program-transform-feedback';
 
-import DelaunayInterpolation from '../delaunay-interpolation/delaunay-interpolation';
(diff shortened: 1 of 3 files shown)


examples/wind/src/layers/particle-layer/particle-layer.js
```diff
@@ -332,24 +334,21 @@ export default class ParticleLayer extends Layer {
   }
 
   createTexture(gl, opt) {
```
gnavvy: `createTexture(gl, opt = {}) {`
so we don't need the 2nd param in line 51 and 52

examples/wind/src/layers/wind-layer/wind-layer.js
```diff
@@ -166,24 +165,21 @@ export default class WindLayer extends Layer {
   }
 
   createTexture(gl, opt) {
```
gnavvy: same as above, let opt = {} by default.
but how do we control the texture size without passing down the width and height?

examples/wind/src/layers/particle-layer/particle-layer.js
1chandu: Yeah, if you look at the flow, they are never passed down to Texture2D constructor.

examples/wind/src/layers/wind-layer/wind-layer.js
1chandu: Same as above.

examples/wind/src/layers/particle-layer/particle-layer.js
shaojingli: Can we put this as comment in the code?

PR #804, opened 2017-07-24
Text-layer

rivulet-zhang:
So here comes a text layer for deck.gl. This version is probably more scalable than the existing label layer implementation as it does not require dynamically generating texture for labels. Instead, it uses a font atlas (small, around 50k) to iteratively render letters of the label. The layer is mainly built upon the icon layer and overrides its vertex shader and calculates the offset of each letter in the string.

Let me know if there is anything I need to adjust before merging.


rivulet-zhang:
@shaojingli The style file has been removed (will also remove the same file from the tagmap example). There are some known issues with style diffing, so set `preventStyleDiffing` to be `true` for now.

rivulet-zhang:
Hi @shaojingli I've made a few changes in the vertex shader to fix the icon rotation issue (See #834 ). You can feel free to merge it now.

examples/text-layer/text-layer/text-layer.js
```diff
@@ -0,0 +1,101 @@
+/* eslint-disable max-len */
```
ibgreen: We have an `examples/sample-layers` folder, might be a good place.

rivulet-zhang: Yep. Actually, it would be nice to have it as a core layer so that for other text-related layer dev (e.g., tagmap) we do not need to duplicate the code.

PR #805, opened 2017-07-24
Apply modelMatrix before project_position

Pessimistress:


src/shaderlib/project/project.glsl.js
```diff
@@ -73,45 +83,33 @@ vec4 project_scale(vec4 meters) {
...
-vec4 project_position(vec4 position) {
-  return vec4(project_position(position.xyz), position.w);
+vec2 project_position(vec2 position) {
+  vec4 projected_position = project_position(vec4(position, 0.0, 1.0));
+  return projected_position.xy;
 }
 
-//
-
+// TODO/xiaoji: this function no longer produces correct result, remove?
```
Pessimistress: With the changes in this PR this function no longer accepts raw positions, which defeats the original purpose of adding it. It also gives incorrect result in METER_OFFSETS mode. 

we don't use this function anywhere except in the plot layer example. As shown above, I am able to make the plot layer work without it. At this point it's too much trouble fixing it and it doesn't provide a lot of value.

src/shaderlib/project/project.glsl.js
```diff
@@ -33,38 +33,32 @@ uniform float projectionScale;
...
 
-uniform mat4 modelViewMatrix;
+uniform mat4 modelMatrix;
+uniform mat4 viewMatrix;
 uniform mat4 projectionMatrix;
-uniform mat4 projectionMatrixUncentered;
 
 //
 // Scaling offsets
 //
```
shaojingli: It's unclear to readers why the scalar version of `project_scale` multiplies meters by the `z` component of `projectionPixelsPerUnit`. We should change the name of this function to `project_scale_z` or at least put a comment here saying that "the scalar version of `project_scale` is for scaling the z component only"

src/shaderlib/project/project.glsl.js
```diff
@@ -81,37 +75,42 @@ vec2 project_mercator_(vec2 lnglat) {
...
+      project_mercator_(position.xy) * WORLD_SCALE * projectionScale,
+      project_scale(position.z),
+      position.w
+    );
   }
-  // Covers projectionMode == PROJECT_MERCATOR
-  return project_mercator_(position) * WORLD_SCALE * projectionScale;
+
+  // Apply model matrix
+  vec4 position_modelspace = modelMatrix * position;
```
shaojingli: Yes. This is reasonable. The `modelMatrix` should be multiplied before `project_scale` so it's applied on top of original, untransformed, `position` in the "meters" world.

src/shaderlib/project/project.glsl.js
shaojingli: Let's remove it

PR #806, opened 2017-07-25
Add UTM_OFFSETS mode

Pessimistress:


src/lib/viewports/web-mercator-viewport.js
```diff
@@ -216,12 +225,11 @@ export default class WebMercatorViewport extends Viewport {
    * @param {[Number,Number]|[Number,Number,Number]) xyz - array of meter deltas
    * @return {[Number,Number]|[Number,Number,Number]) - array of [lng,lat,z] deltas
    */
-  metersToLngLatDelta(xyz) {
+  metersToLngLatDelta(lngLatZ, xyz) {
```
shaojingli: is this lngLatZ the reference point for calculating the lng/lat delta?

src/shaderlib/project/project.glsl.js
```diff
@@ -46,6 +48,9 @@ float project_scale(float meters) {
 }
 
 vec2 project_scale(vec2 meters) {
+  if (projectionMode == PROJECT_UTM_OFFSETS) {
```
shaojingli: Though it's obvious right now, let's add a comment here stating that in UTM_OFFSETS mode, lat and lon are no longer independent in the scaling transformation

src/lib/viewports/web-mercator-viewport.js
Pessimistress: Yes. I'll update the JSDoc.

PR #827, opened 2017-08-07
Reorganize controls, export react controllers

ibgreen:
Another preparatory PR


examples/without-map/app.js
```diff
@@ -1,7 +1,7 @@
 /* global window,document,fetch */
 import React, {Component} from 'react';
 import {render} from 'react-dom';
-import {MapController} from 'deck.gl/dist/controllers';
+import {MapController} from 'deck.gl';
```
Pessimistress: Should we be updating the examples at this stage? This app will not work with `npm start`. 
We also discussed about not depending on alpha packages in examples. @shaojingli

src/react/controllers/map-controller.js
```diff
@@ -60,7 +60,7 @@ const propTypes = {
   // A map control instance to replace the default map controls
   // The object must expose one property: `events` as an array of subscribed
   // event names; and two methods: `setState(state)` and `handle(event)`
-  mapControls: PropTypes.shape({
+  controls: PropTypes.shape({
```
Pessimistress: Are you proposing we change the interface in react-map-gl as well?

examples/without-map/app.js
ibgreen: > Should we be updating the examples at this stage? This app will not work with npm start.

True. At the same time, importing from dist is very problematic in a multi-distribution package like deck.gl. Depending on which bundler is used, one can easily get the main library and these files from different trees (e.g. dist vs. dist-es6), which can lead to very subtle and unpleasant bugs.

> We also discussed about not depending on alpha packages in examples. @shaojingli

Hmm, I see. I understand the thinking, but having to hold off changing examples until we release is a little tricky. 

I really want to isolate the react wrappers into the react directory. I guess I could add one line forwarding export files in the old directory to make the old imports work.

src/react/controllers/map-controller.js
ibgreen: > Are you proposing we change the interface in react-map-gl as well?

No strong opinions on that yet. Just saw an opportunity to make the different controllers more alike.

PR #828, opened 2017-08-08
Use luma picking module in scatterplot layer.

1chandu:
- Add a new prop to Layer class to specify picking highlight color, by default its value is set to 'NULL_PICKING_COLOR' which will disable highlighting.
- Set highlighting for scatterplot layer of layer-browser example.
NOTE: Not ready to merge, also requires luma changes.

### DIFF
--- a/examples/layer-browser/src/examples/core-layers.js
+++ b/examples/layer-browser/src/examples/core-layers.js
@@ -168,6 +168,7 @@ const ScatterplotLayerExample = {
     getRadius: d => get(d, 'SPACES'),
     opacity: 0.5,
     pickable: true,
+    pickingHighlightColor: [1, 0, 0, 1],
     radiusScale: 30,
     radiusMinPixels: 1,
     radiusMaxPixels: 30
--- a/src/layers/core/scatterplot-layer/scatterplot-layer-fragment.glsl.js
+++ b/src/layers/core/scatterplot-layer/scatterplot-layer-fragment.glsl.js
@@ -35,6 +35,8 @@ void main(void) {
 
   if (distToCenter <= 1.0 && distToCenter >= innerUnitRadius) {
     gl_FragColor = vColor;
+    gl_FragColor = picking_filterHighlightColor(gl_FragColor);
+    gl_FragColor = picking_filterPickingColor(gl_FragColor);
   } else {
     discard;
   }
--- a/src/layers/core/scatterplot-layer/scatterplot-layer-vertex.glsl.js
+++ b/src/layers/core/scatterplot-layer/scatterplot-layer-vertex.glsl.js
@@ -60,9 +60,10 @@ void main(void) {
   vec3 vertex = positions * outerRadiusPixels;
   gl_Position = project_to_clipspace(vec4(center + vertex, 1.0));
 
+  // Set color to be rendered to picking fbo.
+  picking_setPickingColor(instancePickingColors);
+
   // Apply opacity to instance color, or return instance picking color
-  vec4 color = vec4(instanceColors.rgb, instanceColors.a * opacity) / 255.;
-  vec4 pickingColor = vec4(instancePickingColors / 255., 1.);
-  vColor = mix(color, pickingColor, renderPickingBuffer);
+  vColor = vec4(instanceColors.rgb, instanceColors.a * opacity) / 255.;
 }
 `;
--- a/src/lib/draw-and-pick.js
+++ b/src/lib/draw-and-pick.js
@@ -19,7 +19,7 @@
 // THE SOFTWARE.
 
 /* global window */
-import {GL, withParameters, setParameters} from 'luma.gl';
+import {GL, withParameters, setParameters, NULL_PICKING_COLOR} from 'luma.gl';
 import {log} from './utils';
 
 const EMPTY_PIXEL = new Uint8Array(4);
@@ -40,7 +40,7 @@ export function drawLayers({layers, pass}) {
           viewport: layer.context.viewport
         }),
         uniforms: Object.assign(
-          {renderPickingBuffer: 0, pickingEnabled: 0},
+          {renderPickingBuffer: 0, pickingEnabled: 0, picking_uActive: 0},
           layer.context.uniforms,
           {layerIndex}
         ),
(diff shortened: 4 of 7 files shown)


src/lib/draw-and-pick.js
```diff
@@ -196,6 +196,15 @@ export function pickLayers(gl, {
     if (info) {
       infos.set(info.layer.id, info);
     }
+
```
1chandu: @ibgreen , is there a better place to update these uniforms?

1chandu:
@ibgreen , highlighting in scatterplot layer (in layer-browser and in examples/scatterplot) works with these changes.

src/layers/core/scatterplot-layer/scatterplot-layer-fragment.glsl.js
```diff
@@ -35,6 +35,8 @@ void main(void) {
 
   if (distToCenter <= 1.0 && distToCenter >= innerUnitRadius) {
     gl_FragColor = vColor;
+    gl_FragColor = picking_filterHighlightColor(gl_FragColor);
```
ibgreen: Is it possible to replace this with
```
if (distToCenter >= 1.0 && distToCenter < innerUnitRadius) {
   discard;
   return;
}
```
It probably won't save any processing but will make the remaining code (which is the "main" path) look cleaner by removing indentation.

src/layers/core/scatterplot-layer/scatterplot-layer-vertex.glsl.js
```diff
@@ -60,9 +60,10 @@ void main(void) {
   vec3 vertex = positions * outerRadiusPixels;
   gl_Position = project_to_clipspace(vec4(center + vertex, 1.0));
 
+  // Set color to be rendered to picking fbo.
```
ibgreen: ".. and used to check for selection highlight.?"

src/lib/draw-and-pick.js
```diff
@@ -19,7 +19,7 @@
 // THE SOFTWARE.
 
 /* global window */
-import {GL, withParameters, setParameters} from 'luma.gl';
+import {GL, withParameters, setParameters, NULL_PICKING_COLOR} from 'luma.gl';
```
ibgreen: Not sure about this, i.e. that we should enforce a specific `NULL_PICKING_COLOR`, or even export it from luma.gl. 

The original system allowed the user of the picking module to specify it.

src/lib/draw-and-pick.js
```diff
@@ -40,7 +40,7 @@ export function drawLayers({layers, pass}) {
           viewport: layer.context.viewport
         }),
         uniforms: Object.assign(
-          {renderPickingBuffer: 0, pickingEnabled: 0},
+          {renderPickingBuffer: 0, pickingEnabled: 0, picking_uActive: 0},
```
ibgreen: Suggest splitting this into three lines and commenting on each of these how/where it is used.

src/lib/draw-and-pick.js
```diff
@@ -196,6 +196,15 @@ export function pickLayers(gl, {
     if (info) {
       infos.set(info.layer.id, info);
     }
+
+    // update pickingSelectedColor so picked model is highlighted.
+    if (layer.props.pickingHighlightColor !== NULL_PICKING_COLOR) {
```
ibgreen: This seems wrong - I assume that the highlight color is the color that should be used to render (highlight) a selected item, and this it is a real color (has nothing to do with picking colors). Why not simply use `null` to detect whether no highlighting is desired.

examples/layer-browser/src/examples/core-layers.js
```diff
@@ -168,6 +168,7 @@ const ScatterplotLayerExample = {
     getRadius: d => get(d, 'SPACES'),
     opacity: 0.5,
     pickable: true,
+    pickingHighlightColor: [1, 0, 0, 1],
```
ibgreen: Colors are 0-255 right? this seems like a very muted color?

I am assuming that the highlight color is the color that should be used to render (highlight) a selected item, and this it is a real color (has nothing to do with picking colors).

src/lib/layer.js
```diff
@@ -62,7 +62,10 @@ const defaultProps = {
   // Offset depth based on layer index to avoid z-fighting.
   // Negative values pull layer towards the camera
   // https://www.opengl.org/archives/resources/faq/technical/polygonoffset.htm
-  getPolygonOffset: ({layerIndex}) => [0, -layerIndex * 100]
+  getPolygonOffset: ({layerIndex}) => [0, -layerIndex * 100],
+
+  // Picking
+  pickingHighlightColor: NULL_PICKING_COLOR
```
ibgreen: Again suggest simply using `null` as the default highlight color (unless we want highlighting on by default, in which case e.g. a transparent blue might be nice).

src/lib/layer.js
```diff
@@ -608,7 +611,10 @@ export default class Layer {
     this.setUniforms({
       // apply gamma to opacity to make it visually "linear"
       opacity: Math.pow(this.props.opacity, 1 / 2.2),
-      ONE: 1.0
+      ONE: 1.0,
+      // Picking
+      pickingSelectedColor: NULL_PICKING_COLOR,
+      pickingHighlightColor: this.props.pickingHighlightColor
```
ibgreen: These need to be passed through `model.setModuleUniforms`.

src/lib/draw-and-pick.js
1chandu: The layer class sets every prop to a uniform (updateModuleSettings). So for `null` to work, we can't overload single prop to specify null and a vector, lets discuss this offline.

examples/layer-browser/src/examples/core-layers.js
1chandu: For highlight color, picking module doesn't scale the color from [0-255] to [0-1], we might need to fix that then we can specify colors in 0-255 range.

ibgreen: Got it. But I think that is worth a comment in the source code.

1chandu:
Most of these changes are covered by https://github.com/uber/deck.gl/pull/848, closing ...

PR #837, opened 2017-08-10
wind demo: fix particle/elevation rendering

1chandu:
- Port depth behavior from demo version. This matches rendering of particles/elevations with the demo.
- Increase field layer sample count, to show high wind speeds at `mount washington`.

### DIFF
--- a/examples/wind/src/layers/delaunay-cover-layer/delaunay-cover-layer.js
+++ b/examples/wind/src/layers/delaunay-cover-layer/delaunay-cover-layer.js
@@ -92,9 +92,10 @@ export default class DelaunayCoverLayer extends Layer {
         setParameters(gl, {
           blend: true,
           blendFunc: [gl.SRC_ALPHA, gl.ONE_MINUS_SRC_ALPHA],
-          blendEquation: gl.FUNC_ADD
+          blendEquation: gl.FUNC_ADD,
+          depthTest: true,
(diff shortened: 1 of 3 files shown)


examples/wind/src/layers/delaunay-cover-layer/delaunay-cover-layer.js
```diff
@@ -92,9 +92,10 @@ export default class DelaunayCoverLayer extends Layer {
         setParameters(gl, {
           blend: true,
           blendFunc: [gl.SRC_ALPHA, gl.ONE_MINUS_SRC_ALPHA],
-          blendEquation: gl.FUNC_ADD
+          blendEquation: gl.FUNC_ADD,
+          depthTest: true,
```
ibgreen: So there was a depth test issue after all... It has bit me in the past as well, annoying that we lose so much time on this. Maybe we should turn on depth testing by default after all. Maybe open an issue on this so we can discuss? There was a suggestion to change default blending as well today...

1chandu: Yeah more over, this issue is also present in demo version except if you run field layer first, which is setting this state and other layers depend on it, which made the issue even  complicated as nothing came out when comparing particle and elevation layers in both versions.

PR #838, opened 2017-08-10
RFC: First person viewport

ibgreen:


RFCs/first-person-mercator-viewport-rfc.md
```diff
@@ -0,0 +1,230 @@
...
+By moving some of the required properties into the base viewport, it would be possible to extend cartographic projection to all cameras.
+
+Note: There are subtleties around the positioning of the camera which is handled in the next proposal.
+
+
+### Viewport
+
+The current base viewport class would be extended with a lnglat property
+
+- lnglat anchor
```
1chandu: `lnglat anchor` is this camera position or camera center?

RFCs/first-person-mercator-viewport-rfc.md
```diff
@@ -0,0 +1,233 @@
...
+
+As an alternative to the `Perspective`/`Orthographic` camera class, a `FirstPerson`/`ThirdPerson` approach.
+
+
+### FirstPersonViewport
+
+extends `Viewport`
+
+- player position (lngLat anchor + offset)
+- player direction
```
gnavvy: is _player direction_ here referring to the _up_ vector (whether the player is standing straight) or the _face_ vector (the direction (s)he is looking at? 

also, we might want users to override the fov via the 1st person viewport as well.

RFCs/first-person-mercator-viewport-rfc.md
```diff
@@ -0,0 +1,233 @@
...
+
+
+### ThirdPersonViewport
+
+extends `Viewport`
+
+- player position
+- player direction
```
gnavvy: same question as above.

RFCs/first-person-mercator-viewport-rfc.md
```diff
@@ -0,0 +1,233 @@
...
+- camera direction, relative to player direction
```
gnavvy: could you explain what the camera direction here is used for? is it to say the camera can look at a position other than the player's position? (e.g. with an offset?)

gnavvy: also, what's your thoughts on _bearing_ (need a _up_ vector) for the camera in 3rd person view? not sure how value analytical value it adds though

RFCs/first-person-mercator-viewport-rfc.md
ibgreen: > lnglat anchor is this camera position or camera center?

The camera position is the lng lat anchor + any meter offset specified by position. Just like any object is positioned in a METER_OFFSETS layer. I will clarify.

RFCs/first-person-mercator-viewport-rfc.md
ibgreen: > is player direction here referring to the up vector (whether the player is standing straight) or the face vector (the direction (s)he is looking at?

I support both, as well as fov. Will clarify.

RFCs/first-person-mercator-viewport-rfc.md
ibgreen: > also, what's your thoughts on bearing (need a up vector) for the camera in 3rd person view? not sure how value analytical value it adds though.

I am also not sure how useful it is for our use cases. Maybe we can add it later if needed.

ibgreen: > could you explain what the camera direction here is used for? is it to say the camera can look at a position other than the player's position? (e.g. with an offset?)

The idea here was that an object has a direction, and in third person view, the camera has a direction relative to the object. When the object moves and turns, we might want to keep the relative direction of the camera constant. So the actual direction would be the sum of the two directions.

ibgreen:
> Since we are generalizing viewports: I've come across the need to supply a custom viewMatrix to any viewport.

> In the WebVR context, this means that I want to offset a given viewport with the leftViewMatrix and rightViewMatrix from VRFrameData. The way to do this today is to take the viewMatrix of the existing viewport and manually construct the required matrices.

Happy to look into this. But are the left and right matrices really just eye offset/direction matrices, i.e not the complete viewProjection matrices for each eye?

gnavvy:
@ibgreen Somehow some of the content in the first-person-mercator-viewport-rfc.md are not shown in this diff, will comment here directly.

```
### OrthographicViewport and PerspectiveViewport

These should be deprecated. Since they are not extensively used we could deprecate them directly in the next major release. That said, if needed, they can both be supplied as trivial subclasses of `FirstPersonViewport`.
```
These two viewports extensively used in non-geospatial layers, e.g.
OrthographicViewport: graph layers, svg-interoperability example, etc.
PerspectiveViewport: point-cloud-layer exmaples.

`FirstPersonViewport` does not fit in these cases.

Issue #847, opened 2017-08-15
Mouse pointer movement not registered when moving too fast


1chandu:
Steps to repro :
- Apply changes from #839
- In layer-browser, select Scatterplot layer , enable autoHighlight prop, zoom or pan, so one circle instance is partially clipped by border, move the mouse pointer slowly from with in this circle to outside of the canvas, you can see highlighting is disabled when mouse pointer moves away, repeat the same, but this time move the mouse fast, you can see highlighting is still on even after mouse pointer is outside of the canvas.

ibgreen:
@1chandu can you describe which browser events we are currently handling? How do we track mouse leaving the canvas?

Pessimistress:
This is because we do not handle the `mouseleave` event on the canvas. If the `mousemove` event doesn't fire while the mouse is over the void (due to frame rate), no more onHover is triggered.
I am adding a `mouseleave`/`pointerleave` event to the EventManager.

PR #848, opened 2017-08-15
Automatic/custom highlighting using picking shader module.

1chandu:
- Copy picking module from luma.gl, this is temporary and avoids dependencies as I make changes to this module, once all layers are modified to use picking module, I will update luma version and remove it from deck.gl.
- Add new props required for highlighting as described in RFC : https://github.com/uber/deck.gl/pull/839
- Update Scatterplot and Path layer shaders to use picking module.
- Verified rendering/performance using layer-browser example.
- Found one issue related to mouse leave event : #847

### DIFF
--- a/src/layers/core/path-layer/path-layer-fragment.glsl.js
+++ b/src/layers/core/path-layer/path-layer-fragment.glsl.js
@@ -41,5 +41,11 @@ void main(void) {
     discard;
   }
   gl_FragColor = vColor;
+
+  // use highlight color if this fragment belongs to the selected object.
+  gl_FragColor = picking_filterHighlightColor(gl_FragColor);
+
+  // use picking color if rendering to picking FBO.
+  gl_FragColor = picking_filterPickingColor(gl_FragColor);
 }
 `;
--- a/src/layers/core/path-layer/path-layer-vertex.glsl.js
+++ b/src/layers/core/path-layer/path-layer-vertex.glsl.js
@@ -175,5 +175,8 @@ void main() {
   pos = lineJoin(prevPosition, currPosition, nextPosition);
 
   gl_Position = project_to_clipspace(vec4(pos, 1.0));
+
+  // Set color to be rendered to picking fbo (also used to check for selection highlight).
+  picking_setPickingColor(instancePickingColors);
 }
 `;
--- a/src/layers/core/scatterplot-layer/scatterplot-layer-vertex.glsl.js
+++ b/src/layers/core/scatterplot-layer/scatterplot-layer-vertex.glsl.js
@@ -60,9 +60,10 @@ void main(void) {
   vec3 vertex = positions * outerRadiusPixels;
   gl_Position = project_to_clipspace(vec4(center + vertex, 1.0));
 
+  // Set color to be rendered to picking fbo (also used to check for selection highlight).
+  picking_setPickingColor(instancePickingColors);
+
   // Apply opacity to instance color, or return instance picking color
-  vec4 color = vec4(instanceColors.rgb, instanceColors.a * opacity) / 255.;
-  vec4 pickingColor = vec4(instancePickingColors / 255., 1.);
-  vColor = mix(color, pickingColor, renderPickingBuffer);
+  vColor = vec4(instanceColors.rgb, instanceColors.a * opacity) / 255.;
 }
 `;
--- a/src/lib/draw-and-pick.js
+++ b/src/lib/draw-and-pick.js
@@ -35,12 +35,26 @@ export function drawLayers({layers, pass}) {
       compositeCount++;
     } else if (layer.props.visible) {
 
+      const pickingSelectedColor = _selectedObjectPickingColor(layer);
+      if (pickingSelectedColor) {
+        layer.state.model.updateModuleSettings({
+          pickingSelectedColor,
+          pickingSelectedColorValid: true
+        });
+      }
+
       layer.drawLayer({
         moduleParameters: Object.assign({}, layer.props, {
           viewport: layer.context.viewport
         }),
         uniforms: Object.assign(
-          {renderPickingBuffer: 0, pickingEnabled: 0},
+          // TODO: Update all layers to use 'picking_uActive' (picking shader module)
+          // and then remove 'renderPickingBuffer' and 'pickingEnabled'.
+          {
+            renderPickingBuffer: 0,
+            pickingEnabled: 0,
+            picking_uActive: 0
+          },
           layer.context.uniforms,
           {layerIndex}
         ),
@@ -196,6 +210,17 @@ export function pickLayers(gl, {
     if (info) {
       infos.set(info.layer.id, info);
     }
+
+    const pickingSelectedColor = pickedColor;
+    const pickingSelectedColorValid = Boolean(
+      layer.props.autoHighlight &&
+      pickedLayer === layer &&
+      pickingSelectedColor !== EMPTY_PIXEL
+    );
+    layer.state.model.updateModuleSettings({
+      pickingSelectedColor,
+      pickingSelectedColorValid
+    });
   });
 
   infos.forEach(info => {
@@ -333,8 +358,8 @@ function getPickedColors(gl, {
     scissor: [x, y, width, height],
     blend: true,
     blendFunc: [gl.ONE, gl.ZERO, gl.CONSTANT_ALPHA, gl.ZERO],
-    blendEquation: gl.FUNC_ADD
-    // TODO - Set clear color
+    blendEquation: gl.FUNC_ADD,
+    clearColor: [0, 0, 0, 0]
   }, () => {
 
     // Clear the frame buffer
@@ -351,7 +376,13 @@ function getPickedColors(gl, {
             viewport: layer.context.viewport
           }),
           uniforms: Object.assign(
-            {renderPickingBuffer: 1, pickingEnabled: 1},
+            // TODO: Update all layers to use 'picking_uActive' (picking shader module)
+            // and then remove 'renderPickingBuffer' and 'pickingEnabled'.
+            {
+              renderPickingBuffer: 1,
+              pickingEnabled: 1,
+              picking_uActive: 1
+            },
             layer.context.uniforms,
             {layerIndex}
           ),
@@ -398,4 +429,16 @@ function getLayerPickingInfo({layer, info, mode}) {
     layer = layer.parentLayer;
   }
   return info;
+}
+
+/**
+ * Returns the picking color of currenlty selected object of the given 'layer'.
+ * @return {Array} - the picking color or null if layers selected object is invalid.
+ */
+function _selectedObjectPickingColor(layer) {
+  if (layer.props.highlightedObjectIndex < 0) {
+    return null;
+  }
+  // TODO: Add warning if 'highlightedObjectIndex' is > numberOfInstances of the model.
+  return layer.encodePickingColor(layer.props.highlightedObjectIndex);
 }
--- a/src/lib/layer.js
+++ b/src/lib/layer.js
@@ -62,7 +62,12 @@ const defaultProps = {
   // Offset depth based on layer index to avoid z-fighting.
   // Negative values pull layer towards the camera
   // https://www.opengl.org/archives/resources/faq/technical/polygonoffset.htm
-  getPolygonOffset: ({layerIndex}) => [0, -layerIndex * 100]
+  getPolygonOffset: ({layerIndex}) => [0, -layerIndex * 100],
+
+  // Selection/Highlighting
+  highlightedObjectIndex: -1,
+  autoHighlight: false,
+  highlightColor: [0, 0, 128, 128]
 };
 
 let counter = 0;
@@ -276,6 +281,7 @@ export default class Layer {
    * @return {Array} - the decoded color
    */
   encodePickingColor(i) {
+    assert(((((i + 1) >> 24) & 255) === 0), 'index out of picking color range');
     return [
       (i + 1) & 255,
       ((i + 1) >> 8) & 255,
@@ -385,6 +391,7 @@ export default class Layer {
     // Add any subclass attributes
     this.updateAttributes(this.props);
     this._updateBaseUniforms();
+    this._updateModuleSettings();
 
     const {model} = this.state;
     if (model) {
@@ -415,6 +422,7 @@ export default class Layer {
       // Run the attribute updaters
       this.updateAttributes(updateParams.props);
       this._updateBaseUniforms();
+      this._updateModuleSettings();
 
       if (this.state.model) {
         this.state.model.setInstanceCount(this.getNumInstances());
@@ -610,6 +618,14 @@ export default class Layer {
       opacity: Math.pow(this.props.opacity, 1 / 2.2),
       ONE: 1.0
     });
+  }
+
+  _updateModuleSettings() {
+    if (this.state.model) {
+      this.state.model.updateModuleSettings({
+        pickingHighlightColor: this.props.highlightColor
+      });
+    }
   }
 
   // DEPRECATED METHODS
--- a/src/shaderlib/picking/picking.js
+++ b/src/shaderlib/picking/picking.js
@@ -1,0 +1,97 @@
+const DEFAULT_HIGHLIGHT_COLOR = new Uint8Array([0, 64, 128, 64]);
+
+const DEFAULT_MODULE_OPTIONS = {
+  pickingSelectedColor: null, //  Set to a picking color to visually highlight that item
+  pickingHighlightColor: DEFAULT_HIGHLIGHT_COLOR, // Color of visual highlight of "selected" item
+  pickingThreshold: 1.0,
+  pickingActive: false, // Set to true when rendering to off-screen "picking" buffer
+  pickingValid: false
+};
+
+/* eslint-disable camelcase */
+function getUniforms(opts = DEFAULT_MODULE_OPTIONS) {
+  const uniforms = {};
+  uniforms.picking_uValid = opts.pickingValid ? 1 : 0;
+  if (opts.pickingSelectedColor !== undefined) {
+    if (opts.pickingSelectedColor) {
+      const selectedColor = [
+        opts.pickingSelectedColor[0],
+        opts.pickingSelectedColor[1],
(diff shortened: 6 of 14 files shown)


src/layers/core/path-layer/path-layer-fragment.glsl.js
```diff
@@ -41,5 +41,11 @@ void main(void) {
     discard;
   }
   gl_FragColor = vColor;
+
+  // use highlight color if this fragment belongs to the selected object.
+  gl_FragColor = picking_filterHighlightColor(gl_FragColor);
+
+  // use picking color if rendering to pikcing FBO.
```
ibgreen: Typo `pikcing FBO` please check in all shaders.

src/layers/core/path-layer/path-layer-vertex.glsl.js
```diff
@@ -175,5 +175,8 @@ void main() {
   pos = lineJoin(prevPosition, currPosition, nextPosition);
 
   gl_Position = project_to_clipspace(vec4(pos, 1.0));
+
+  // Set color to be rendered to picking fbo which is used to check for selection highlight.
```
ibgreen: Nit a little clearer if we change:
`Set color to be rendered to picking fbo which is used for ... `
=> 
`Set color to be rendered to picking fbo(also used for...)`
Please update in all placed

src/layers/core/scatterplot-layer/scatterplot-layer-vertex.glsl.js
```diff
@@ -60,9 +60,10 @@ void main(void) {
...
   gl_Position = project_to_clipspace(vec4(center + vertex, 1.0));
 
+  // Set color to be rendered to picking fbo which is used to check for selection highlight.
+  picking_setPickingColor(instancePickingColors);
+
   // Apply opacity to instance color, or return instance picking color
-  vec4 color = vec4(instanceColors.rgb, instanceColors.a * opacity) / 255.;
-  vec4 pickingColor = vec4(instancePickingColors / 255., 1.);
-  vColor = mix(color, pickingColor, renderPickingBuffer);
+  vColor = vec4(instanceColors.rgb, instanceColors.a * opacity) / 255.;
```
ibgreen: Do we have any perf measurements showing the impact of moving the check to the fragment shader?

src/lib/draw-and-pick.js
```diff
@@ -35,12 +35,26 @@ export function drawLayers({layers, pass}) {
       compositeCount++;
     } else if (layer.props.visible) {
 
+      const pickingSelectedColor = layer.selectedObjectPickingColor();
+      if (pickingSelectedColor) {
+        layer.state.model.updateModuleSettings({
+          pickingValid: true,
```
ibgreen: This parameter name (`pickingValid`) does not feel intuitive to me. Would love to see an API audit to discuss what it does, why it is needed, and what possible alternatives could be.

But first: If we are making changes to the API of the picking module, we also need to update its docs, maybe copy them into this directly temporarily?

A thought: Most modules have a `...Enabled` param. This is different from that?

src/lib/draw-and-pick.js
```diff
@@ -196,6 +210,17 @@ export function pickLayers(gl, {
     if (info) {
       infos.set(info.layer.id, info);
     }
+
+    const pickingSelectedColor = pickedColor;
+    const pickingValid = Boolean(
```
ibgreen: This is a lot of logic required to use the `picking` module. Makes me feel that the `picking` module API design was not successful. I wonder if we could do anything to simplify?

src/shaderlib/picking/picking.js
```diff
@@ -0,0 +1,95 @@
+const DEFAULT_MODULE_OPTIONS = {
```
ibgreen: Hard to see what changes you made here since this was copied.

src/lib/layer.js
```diff
@@ -284,6 +290,16 @@ export default class Layer {
   }
 
   /**
+   * Returns the picking color of the object at 'selectedObjectIndex' prop.
```
ibgreen: Who uses this? Why do we need it?

src/lib/layer.js
```diff
@@ -284,6 +290,16 @@ export default class Layer {
...
+   * @return {Array} - the picking color or null if 'selectedObjectIndex is invalid.
+   */
+  selectedObjectPickingColor() {
+    if (this.props.selectedObjectIndex < 0) {
+      return null;
```
ibgreen: Why return null in this case, rather than the null picking color?

src/lib/layer.js
```diff
@@ -276,6 +281,7 @@ export default class Layer {
    * @return {Array} - the decoded color
    */
   encodePickingColor(i) {
+    assert(((((i + 1) >> 24) & 255) === 0), `index: ${i} out of coding range`);
```
ibgreen: I usually avoid dynamic string expressions in asserts as the error strings will be generated even on successful calls.

src/lib/layer.js
```diff
@@ -62,7 +62,12 @@ const defaultProps = {
   // Offset depth based on layer index to avoid z-fighting.
   // Negative values pull layer towards the camera
   // https://www.opengl.org/archives/resources/faq/technical/polygonoffset.htm
-  getPolygonOffset: ({layerIndex}) => [0, -layerIndex * 100]
+  getPolygonOffset: ({layerIndex}) => [0, -layerIndex * 100],
+
+  // Selection/Highlighting
+  selectedObjectIndex: -1,
```
ibgreen: Note for API review: I am wondering if this should be `highlightedObjectIndex` for consistency. `selected` could mean various things...

src/lib/draw-and-pick.js
```diff
@@ -351,7 +376,13 @@ function getPickedColors(gl, {
             viewport: layer.context.viewport
           }),
           uniforms: Object.assign(
-            {renderPickingBuffer: 1, pickingEnabled: 1},
+            // TODO: Update all layers to use 'picking_uActive' (picking shader module)
+            // and then remove 'renderPickingBuffer' and 'pickingEnabled'.
+            {
+              renderPickingBuffer: 1,
+              pickingEnabled: 1,
+              picking_uActive: 1
```
ibgreen: Shouldn't `picking` module params be set with `updateModuleSettings`?

src/lib/draw-and-pick.js
```diff
@@ -333,8 +358,8 @@ function getPickedColors(gl, {
     scissor: [x, y, width, height],
     blend: true,
     blendFunc: [gl.ONE, gl.ZERO, gl.CONSTANT_ALPHA, gl.ZERO],
-    blendEquation: gl.FUNC_ADD
-    // TODO - Set clear color
+    blendEquation: gl.FUNC_ADD,
+    clearColor: [0, 0, 0, 0]
```
ibgreen: Should we use a constant for the NULL color?

src/lib/draw-and-pick.js
```diff
@@ -35,12 +35,26 @@ export function drawLayers({layers, pass}) {
...
           viewport: layer.context.viewport
         }),
         uniforms: Object.assign(
-          {renderPickingBuffer: 0, pickingEnabled: 0},
+          // TODO: Update all layers to use 'picking_uActive' (picking shader module)
+          // and then remove 'renderPickingBuffer' and 'pickingEnabled'.
+          {
+            renderPickingBuffer: 0,
+            pickingEnabled: 0,
+            picking_uActive: 0
```
ibgreen: See comment below on `picking` uniforms.

src/lib/draw-and-pick.js
1chandu: `renderPickingBuffer` and `pickingEnabled` do not belong to picking module, they belong to existing picking functionality in deck.gl. But I can set `picking_uActive` with `updateModuleSettings`, as commented `renderPickingBuffer` and `pickingEnabled` will eventually be removed.

ibgreen: > renderPickingBuffer and pickingEnabled do not belong to picking module

Very good point, we can deprecated them, but they will be used by existing custom layers so we need to keep supporting them at least until we make a major release.

Ideally we'd emit warnings when layers use these uniforms in their shaders. There are ways we can check what uniforms layer shaders define. But not needed for this PR, better open a separate issue: "deprecate old picking uniforms and updating to use new picking module in all sample layers".

src/lib/draw-and-pick.js
1chandu: Sure, will copy the docs to deck.gl and update them.
This uniform tells VS version of the module if the current picking color set is valid or not. It was added to fix luma instancing example for cases when mouse position goes out of canvas. In the deck.gl, when currently picked layer changes, we set pickingValid to false on old layer.
We can definitely overload pickingColor (set it to null picking color ([0, 0, 0, 0])) and I am ok with it, but this approach adds clarity to code with setting of an extra uniform, lets discuss this.

src/lib/draw-and-pick.js
1chandu: We have two types of picking now, custom and auto highlight, and when both set custom picking takes precedence as discussed in RFC. This logic needed to handle all possible corner cases.
And all this happens inside of deck.gl, core layers and applications using this feature only deal with setting the props, lets discuss if any additional changes needed in this regard.

src/lib/layer.js
1chandu: When a specific object is selected using `selectedObjectIndex` prop, this method converts it to picking color to be set as a uniform.

src/lib/layer.js
1chandu: I would prefer it to return null, and handle the logic of using `null picking color` (when color is null) by `draw-and-picking` methods, so we don't tie that logic with this method.

src/lib/layer.js
ibgreen: Please change `index: ${i} out of coding range` =>` index out of picking color range`

src/shaderlib/picking/picking.js
```diff
@@ -1,17 +1,15 @@
-const DEFAULT_HIGHLIGHT_COLOR = new Uint8Array([0, 64, 128, 64]);
-
 const DEFAULT_MODULE_OPTIONS = {
   pickingSelectedColor: null, //  Set to a picking color to visually highlight that item
-  pickingHighlightColor: DEFAULT_HIGHLIGHT_COLOR, // Color of visual highlight of "selected" item
```
1chandu: removed `pickingHighlightColor`, ideally this can only be set once and doesn't have to be updated as frequently as other uniforms. So when other uniforms are set, we don't overwrite this with default value.

src/lib/layer.js
ibgreen: Do we need this to be a public method on the class or should we make it private (underscore)? If public, we should add it to `Layer` docs.

src/lib/layer.js
ibgreen: OK I can accept that.

The counter argument would be that it feels like we have an if statement returning null here, and another if statement checking for null in the shader module. And `encodePickingColor` already handles `index = -1`. So a bunch of additional logic that ultimately cancels out. But maybe this is still easer for the reader to follow.

src/shaderlib/picking/shadertools-picking.md
```diff
@@ -0,0 +1,77 @@
...
+### getUniforms
+
+`getUniforms` takes an object takes a set of key/value pairs, returns an object with key/value pairs representing the uniforms that the `picking` module shaders need.
+
+`getUniforms(opts)`
+opts can contain following keys:
+* `pickingValid` (*boolean*) - When true current instance picking color is ignored, hence no instance is highlighted.
+* `pickingSelectedColor` (*array*) - Picking color of the currently selected instance.
+* `pickingHighlightColor` (*array*)- Color used to highlight the currently selected instance.
+* `pickingActive`=`false` (*boolean*) - Renders the picking colors instead of the normal colors. Normally only used with an off-screen framebuffer during picking. Default value is `false`.
```
ibgreen: If true, ...

src/shaderlib/picking/shadertools-picking.md
```diff
@@ -0,0 +1,77 @@
...
+
+Note that the selected item will be rendered using `pickingHighlightColor`, if blending is enabled for the draw, alpha channel can be used to control the blending result.
```
ibgreen: **DOC CHANGE** This blending comment is a little confusing. It seems alpha is just rendered as it is normally. Does this sentence add anything or can we just cut it?

src/shaderlib/picking/shadertools-picking.md
```diff
@@ -0,0 +1,77 @@
...
+## Vertex Shader Functions
+
+### `void picking_setPickingColor(vec3)`
+
+Sets the color that will be returned by the fragment shader if color based picking is enabled. Typically set from a `pickingColor` uniform or a `pickingColors` attribute (e.g. when using instanced rendering, to identify the actual instance that was picked).
+
+
+## Fragment Shader Functions
+
+### picking_filterPickingColor
```
ibgreen: **DOC CHANGE** Let's switch the order of these two functions, both more alphabetical and corresponds to the order in which they should be called.

src/shaderlib/picking/shadertools-picking.md
```diff
@@ -0,0 +1,77 @@
...
+
+## JavaScript Functions
+
+### getUniforms
+
+`getUniforms` takes an object takes a set of key/value pairs, returns an object with key/value pairs representing the uniforms that the `picking` module shaders need.
+
+`getUniforms(opts)`
+opts can contain following keys:
+* `pickingValid` (*boolean*) - When true current instance picking color is ignored, hence no instance is highlighted.
```
ibgreen: **REQUIRED CHANGE** So these uniforms control both picking (the core use case) and highlighting (the additional feature of this module). 

And this new flag only controls highlighting, right? If so, should it be `pickingSelectedColorValid`?

src/shaderlib/picking/shadertools-picking.md
```diff
@@ -0,0 +1,77 @@
+# picking (Shader Module)
+
+Provides support for color-coding-based picking and highlighting. In particular, supports picking a specific instance in an instanced draw call and highlighting an instance based on its picking color.
+
+Color based picking lets the application draw a primitive with a color that can later be used to index this specific primitive.
+
+Highlighting allows application to specify picking color of instance that need to be highlighted and the highlight color to be used.
```
ibgreen: **DOC CHANGES** Add: to specify *a* picking color corresponding to an...

Change `instance`=>`object`? (shader is not always instanced)

src/shaderlib/picking/shadertools-picking.md
```diff
@@ -0,0 +1,77 @@
+# picking (Shader Module)
+
+Provides support for color-coding-based picking and highlighting. In particular, supports picking a specific instance in an instanced draw call and highlighting an instance based on its picking color.
```
ibgreen: **DOC CHANGE**: ..., and correspondingly, supports picking and highlighting groups of primitives with the same picking color in non-instanced draw-calls.

src/shaderlib/picking/picking.js
```diff
@@ -0,0 +1,95 @@
...
+uniform float picking_uThreshold;
+uniform bool picking_uValid;
+
+varying vec4 picking_vRGBcolor_Aselected;
+
+const float COLOR_SCALE = 1. / 255.;
+
+bool isVertexPicked(vec3 vertexColor, vec3 pickedColor, bool pickingValid) {
+  return
+    pickingValid &&
```
ibgreen: NIT: Can we just access the uniform here? No need for a param?

src/shaderlib/picking/picking.js
```diff
@@ -0,0 +1,95 @@
+const DEFAULT_MODULE_OPTIONS = {
+  pickingSelectedColor: null, //  Set to a picking color to visually highlight that item
+  pickingThreshold: 1.0,
+  pickingActive: false // Set to true when rendering to off-screen "picking" buffer
+};
```
ibgreen: **REQUIRED CHANGE** So I think the idea is that on the first run, `getUniforms` is called with the DEFAULT_MODULE_OPTIONS. After you removed it, we no longer have a `highlightColor` in the defaults, so this uniform will be undefined, at least in luma.gl.

src/lib/layer.js
ibgreen: **REQUIRED CHANGE** I don't see apps calling this directly, so this method should be private (start with an underscore and be moved down to the section with private functions).

1chandu: This is not used by the apps but by draw-and-pick.js, so this is not really private, it is the same category as 'encodePickingColor()'.

src/shaderlib/picking/picking.js
1chandu: yeah, luma examples would have to set this at least once before first draw. Normally highlight color is fixed and doesn't get updated as frequently as other uniforms.

src/lib/layer.js
ibgreen: Then it feels to me like this should be a private util in draw-and-pick (which knows most about the picking module and picking colors anyway)?

As a general principle we should avoid adding non-essential functions to the public API, since people will start using them and it will be harder to update functionality in the future.

If we do keep it here, we kind of need to document it in the public docs: Layer.md

src/shaderlib/picking/picking.js
ibgreen: Merging this back to luma will be a breaking change then? We'll want to be avoid that if possible.

src/shaderlib/picking/shadertools-picking.md
1chandu: agree, will change `pickingValid` to `pickingSelectedColorValid`.

src/shaderlib/picking/shadertools-picking.md
1chandu: An application writer might not realize that highlight color's alpha channel can affect highlighting, so wanted to add here explicitly as it directly relates to the feature.

ibgreen: But what does that mean. We allow him to specify an RGBA value. We'll render it. And it will be blended like any other color.

What we should probably do is be very clear in the parameter documentation above that the `pickingHighlightColor` parameter expects/supports an RGBA color. In my view that would be sufficient.

PR #849, opened 2017-08-15
Update dependencies to use luma 4.0.1 version

1chandu:


1chandu:
I thought the updates to package.json files within examples needed to run the examples as standalone(copy example folder to a different place and run without deck.gl source).

ibgreen:
OK I see. It would be more appropriate to update the packages to use `deck.gl 4.1.1` but we can't do that until we publish.

The reason that it is more appropriate is that for most users it is a technicality that luma.gl is a dependency of deck.gl and they should not have to worry about manually tracking its versions in addition to deck.gl versions.

Issue #843, opened 2017-08-14
`useDevicePixelRatio`: Address all cases for retina displays


1chandu:
#865 added a new prop `useDevicePixelRatio` to DeckGL component, which will be passed to `AnimationLoop` and `LayerManager` (for picking/querying).

Remaining: We are still using `window.devicePixelRatio` in following places:
 - [x] wheel-input.js
 - [x] layer:screenToDevicePixels()
 - [x] examples
Make sure above usages are in sync with DeckGL prop.

ibgreen:
Adding comments from discussion in PR #850

> Is it necessary to draw the picking buffer in high DPI? Mouse coordinates are always in screen pixels.

So what we are discussing now is to make it easy for the app to select whether to draw in High-DPI or not (since it has a significant perf impact), and maybe even add a checkbox to the examples enabling the user to try with and without.

I would expect that to be a new prop on the `DeckGL` (`useHighDPI`, `useDevicePixelRation`, `useRetina` or similar) and it would affect both normal render and picking.

Once that is done, a second reasonable question is whether we could use different values for picking and main rendering. For pixel perfect picking an app might want them to be the same, but many apps want to use a picking radius for better fat finger sensitivity and then it might not matter anyway.

Pessimistress:
> many apps want to use a picking radius for better fat finger sensitivity and then it might not matter anyway.

Speaking of pickingRadius, it already has a significant hit on perf as we process `R^2` pixels. With retina display that goes up to `4R^2`.

1chandu:
#865 added a new prop to DeckGL component, which will be passed to `WebGLRenderer` and `LayerManager` (for picking/querying).
Remaining:
We are still using `window.devicePixelRatio` in following places:
 - wheel-input.js
 - layer:screenToDevicePixels()
 - examples
Make sure above usages are in sync with DeckGL prop.

ibgreen:
I think layer:screenToDevicePixels can be deprecated now. It was used briefly for GL.LINES but that is not supported any more. and now we have a uniform instead.

shaojingli:
Bump this to P1

1chandu:
Resolved by #1006.

PR #850, opened 2017-08-16
A pass on refreshing picking and event handling docs

ibgreen:
Some cleanup of the overall picking/interactivity/event-handling documentation complex.

- I tried to keep new features out of this pass so that it could be merged to current 4.1-release and website.
- We should discuss how to add documentation for in-progress features such as auto highlighting, use of new picking module, and a presumably public EventManager class.

### DIFF
--- a/docs/advanced/performance.md
+++ b/docs/advanced/performance.md
@@ -4,31 +4,15 @@
 ## General Expectations
 
 There are mainly two aspects that developers usually consider regarding the
-performance of any computer programs: the time and the memory consumption, both
-of which obviously depends on the specs of the hardware deck.gl is ultimately
-running on.
+performance of any computer programs: the time and the memory consumption, both of which obviously depends on the specs of the hardware deck.gl is ultimately running on.
 
 On 2015 MacBook Pros with dual graphics cards, most basic layers
 (like `ScatterplotLayer`) renders fluidly at 60 FPS during pan and zoom
-operations up to about 1M (one million) data items, with framerates dropping
-into low double digits (10-20FPS) when the data sets approach 10M items.
+operations up to about 1M (one million) data items, with framerates dropping into low double digits (10-20FPS) when the data sets approach 10M items.
 
-Even if interactivity is not an issue, browser limitations on how
-big chunks of contiguous memory can be allocated (e.g. Chrome caps
-individual allocations at 1GB) will cause most layers to crash
-during WebGL buffer generation somewhere between 10M and 100M items.
-You would need to break up your data into chunks and use multiple
-deck.gl layers to get past this limit.
+Even if interactivity is not an issue, browser limitations on how big chunks of contiguous memory can be allocated (e.g. Chrome caps individual allocations at 1GB) will cause most layers to crash during WebGL buffer generation somewhere between 10M and 100M items. You would need to break up your data into chunks and use multiple deck.gl layers to get past this limit.
 
-Modern phones (recent iPhones and higher-end Android phones) are
-surprisingly capable in terms of rendering performance, but are considerably
-more sensitive to memory pressure than laptops, resulting in browser restarts
-or page reloads.
-
-Remarks:
-* The layer browser example has a couple of performance tests that you can use
-  to get FPS readings for 1M and 10M data sets on your hardware.
-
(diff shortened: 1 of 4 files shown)


docs/advanced/picking.md
```diff
@@ -51,42 +38,34 @@ system, which most layers use.
...
-      value[i++] = pickingColor[2];
-    }
+```js
+calculatePickingColors(attribute) {
+  const {data} = this.props;
+  const {value, size} = attribute;
+
+  let i = 0;
+  for (const object of data) {
+    const pickingColor = this.encodePickingColor(i);
```
1chandu: > const pickingColor = this.encodePickingColor(i);

Nit: should we use a separate index to pass to encodePickingColor for clarity, may be like:
```
let i = 0;
let objectIndex = 0;
 for (const object of data) {
    const pickingColor = this.encodePickingColor(objectIndex++);
    value[i++] = ...
```

docs/advanced/performance.md
```diff
@@ -68,26 +52,40 @@ Thus it is possible to render a scatterplot layer with 10M items with reasonable
...
+## Profiling Ideas
+
+Some profiling techniques:
+* Using the `seer` chrome extension you can also get GPU timings for your layers.
+* The `layer-browser` example (in the `examples` folder has a couple of performance tests that you can use to get FPS readings on your hardware for Scatterplot layers with 1M and 10M points.
+
+## Common Issues
+
+A couple of particular things to watch out for that tend to have a big impact on performance:
+* Be careful when enabling Retina/High DPI rendering. It generetes 4x the number of pixels (fragments) and can have a big performance impact that depends on which computer or monitor is being used.
```
Pessimistress: Unrelated: is it necessary to draw the picking buffer in high DPI? Mouse coordinates are always in screen pixels.

docs/get-started/interactivity.md
```diff
@@ -2,18 +2,64 @@
...
+The picking engine returns "picking info" objects which contains a variety of fields describing what layer and object was picked.
+
+| Key      | Value |
+| ---      | --- |
+| `layer`  | The layer that the picked object belongs to. Only layers with the `pickable` prop set to true can be picked. |
+| `index`  | The index of the object in the layer that was picked. |
+| `object` | The object that was picked. This is typically an entry in the layer's `props.data` array, but can vary from layer to layer. |
+| `x`      | Mouse position x relative to the viewport. |
+| `y`      | Mouse position y relative to the viewport. |
+| `lngLat` | Mouse position in geospatial coordinates. Only applies if `layer.props.projectionMode` is a geospatial mode such as `COORDINATE_SYSTEM.LNGLAT`. |
```
Pessimistress: With the addition of new viewports, maybe we should extend this field to be anything that `viewport.unproject` produces?

docs/advanced/picking.md
ibgreen: > Nit: should we use a separate index to pass to encodePickingColor for clarity, may be like:

Good suggestion. Now I don't think we should use two variables to represent the same thing, that seems confusing. I could just change `i` to `objectIndex` though.

docs/advanced/performance.md
ibgreen: > Unrelated: is it necessary to draw the picking buffer in high DPI? Mouse coordinates are always in screen pixels.

So what we are discussing now is to make it easy for the app to select whether to draw in High-DPI or not (since it has a significant perf impact), and maybe even add a checkbox to the examples enabling the user to try with and without.

I would expect that to be a new prop on the `DeckGL` (`useHighDPI`, `useDevicePixelRation`, `useRetina` or similar) and it would affect both normal render and picking.

Once that is done, a second reasonable question is whether we could use different values for picking and main rendering. For pixel perfect picking an app might want them to be the same, but many apps want to use a picking radius for better fat finger sensitivity and then it might not matter anyway. 

I will add these comments to the open ticket #843.

docs/get-started/interactivity.md
ibgreen: > With the addition of new viewports, maybe we should extend this field to be anything that viewport.unproject produces?

Good point. Should also check how this works with METER_OFFSET viewports. E.g. would it be valuable to get both the lngLat anchor and the offset as separate values, rather than a combined lngLat.

Issue #621, opened 2017-05-01
reducing dist size


contra:
not sure if you're worried about this, but the dist for deck.gl is approaching 2mb

some info:

<img width="810" alt="screen shot 2017-05-01 at 7 09 56 pm" src="https://cloud.githubusercontent.com/assets/425716/25598246/df2d3d0c-2ea1-11e7-9a3e-c308176b1fc4.png">

ericsoco:
Another thing to consider as we grow the layer catalog (and another argument for considering "primitive" layers as a distinction separate from geo/non-geo. The fewer layers we include in the core distribution, the more we can keep the core distro lightweight.

Thanks for the note, @contra.

ibgreen:
@contra  Good feedback - how much of an issue is this for you and other users?

As background, the distro has three complete copies of the tree, the original `src` (which is pure ES6 and can run untranspiled in modern browser and node 7, the `dist-es6` folder for tree-shaking and the vanilla transpiled `dist` folder)

Short term:
* We should probably strip comments from the dist/dist-es6 folders. This looks like a quick fix: https://www.npmjs.com/package/babel-plugin-remove-comments

Longer term:
* We are doing what we can to support tree shaking. When it works the app will only bundle what it needs and the size of the distro is mainly a development concern. Because of this we aren't planning to go to extremes splitting things up in tiny separate npm modules.
* That said, as @shaojing said, we will likely start publishing additional layer packages as separate npm modules in the next release (probably using a monorepo/lerna approach)

PS That looks like a nice analysis tool. Could you provide a link? Maybe we should add a dev script that runs it?

contra:
@ibgreen Debugging our bundle size on a webpack app and found a few offenders - deck.gl + mapbox-gl are ~3MB even after tree shaking in webpack 2.

The tree shaking is actually pretty awful in webpack 2 in my experience with it so far - it doesn't work very well on anything since most people put their compiled code on NPM (which uses module.exports, not usable w/ tree shaking).

Tool is https://github.com/th0r/webpack-bundle-analyzer

I'll try to mess around with aliasing `deck.gl` to `node_modules/deck.gl/dist-es6` and see if that improves the tree shaking - I think the package.json points to `dist/index.js` which uses the non-shakeable common.js exports.

ibgreen:
> deck.gl + mapbox-gl are ~3MB even after tree shaking in webpack 2
Is that for minified code? 

[...]

PR #859, opened 2017-08-18
Reducing distribution size RFC

ibgreen:
Since we are in RFC creation mode for deck.gl, did a write-up on current thoughts around reducing distribution size.

#621

@georgios-uber  @romaxa


ibgreen:
Landed so that RFC is available on master. It links to this PR where reviews can be added.

PR #860, opened 2017-08-18
EventManager Improvements

Pessimistress:
- Add `pointerleave` event (https://github.com/uber/deck.gl/issues/847)
- General code consistency
- Tests

### DIFF
--- a/src/controls/events/key-input.js
+++ b/src/controls/events/key-input.js
@@ -18,6 +18,7 @@ export default class KeyInput {
 
     this.handleEvent = this.handleEvent.bind(this);
 
+    // Use mock in test environment
     const parent = isBrowser ? window : element;
     this.events.forEach(event => parent.addEventListener(event, this.handleEvent));
   }
@@ -31,7 +32,7 @@ export default class KeyInput {
    * Enable this input (begin processing events)
    * if the specified event type is among those handled by this input.
    */
-  toggleIfEventSupported(eventType, enabled) {
+  enableEventType(eventType, enabled) {
     if (eventType === DOWN_EVENT_TYPE) {
       this.enableDownEvent = enabled;
     }
--- a/src/controls/events/key-input.js
+++ b/src/controls/events/key-input.js
@@ -1,34 +1,30 @@
 /* global window */
-const KEY_EVENTS = ['keydown', 'keyup'];
-// const EVENT_TYPE = 'KeyboardEvent';
+import {isBrowser} from '../../utils/globals';
 
-/**
- * Hammer.js swallows 'move' events (for pointer/touch/mouse)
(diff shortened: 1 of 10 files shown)


src/controls/events/key-input.js
```diff
@@ -1,71 +1,62 @@
...
+    this.enableUpEvent = this.options.enable;
+
+    this.events = KEY_EVENTS.concat(options.events || []);
 
     this.handleEvent = this.handleEvent.bind(this);
-    this.options.events.forEach(event => window.addEventListener(event, this.handleEvent));
-  }
 
-  destroy() {
-    this.options.events.forEach(event => window.removeEventListener(event, this.handleEvent));
```
ibgreen: Cool, I'll test this on my branch and see if it still works.

src/controls/events/move-input.js
```diff
@@ -16,65 +17,79 @@ export default class MoveInput {
...
 
   /**
    * Enable this input (begin processing events)
    * if the specified event type is among those handled by this input.
    */
   toggleIfEventSupported(eventType, enabled) {
-    if (EVENT_TYPE === eventType) {
-      this.options.enable = enabled;
+    if (eventType === MOVE_EVENT_TYPE) {
+      this.enableMoveEvent = enabled;
```
ibgreen: This makes more sense, I couldn't see how the old system could be made to fit...

src/controls/generic-controls.js
```diff
@@ -289,25 +289,31 @@ export default class GenericControls {
     }
 
     const KEY_BINDINGS = {
-      KeyW: 'moveForward',
+      w: 'moveForward',
+      W: 'moveForward',
       ArrowUp: 'moveForward',
```
ibgreen: Don't the arrow keys need to be replaced with numbers or similar as well? 

Good if we can keep them working, they are really important to the controller use case...

src/controls/events/move-input.js
```diff
@@ -16,65 +17,79 @@ export default class MoveInput {
...
-  set(options) {
-    Object.assign(this.options, options);
+    this.events.forEach(event => this.element.removeEventListener(event, this.handleEvent));
   }
 
   /**
    * Enable this input (begin processing events)
    * if the specified event type is among those handled by this input.
    */
   toggleIfEventSupported(eventType, enabled) {
```
ibgreen: For API Audit: Something about the name of this method bothers me. Wonder why we didn't just call it `enableEvent` or `enableEventType`? 

It doesn't even seem to toggle, as it takes an enabled flag...

src/controls/generic-controls.js
Pessimistress: `event.key` produces slightly different values from `event.code` but is better supported cross browsers.

PR #865, opened 2017-08-21
Add useDevicePixelRatio prop to DeckGL component.

1chandu:
- `useDevicePixelRatio` prop will be single point of truth to use retina resolution of the device or not, will affect picking and normal rendering.
- Verified using layer-browser example, by setting `useDevicePixelRatio` to true and false.

### DIFF
--- a/docs/api-reference/deckgl.md
+++ b/docs/api-reference/deckgl.md
@@ -75,9 +75,13 @@ This is helpful when rendered objects are difficult to target, for example
 irregularly shaped icons, small moving circles or interaction by touch.
 Default `0`.
 
-##### `pixelRatio` (Number, optional)
+##### `useDevicePixelRatio` (Boolean, optional)
 
-Will use device ratio by default.
+Default value is true.
+
+When true, device's full resolution will be used for rendering, this value can change per frame, like when moving windows between screens or when changing zoom level of the browser.
+
+Note: Set it false unless it is required, as it affects your application performance.
 
 ##### `gl` (Object, optional)
 
--- a/docs/whats-new.md
+++ b/docs/whats-new.md
@@ -1,5 +1,9 @@
 # deck.gl v4.2
 
+## useDevicePixelRatio
+
+This new prop for DeckGL component can be used to toggle usage of full resolution of retina/HD displays.
+
 ## Automatic and custom highlighting
 
 Three new props (`highlightColor`, `highlightedObjectIndex` and `autoHighlight`) are added to `Layer` class to support object highlighting.
--- a/examples/wind/src/wind-demo.js
+++ b/examples/wind/src/wind-demo.js
@@ -110,7 +110,8 @@ export default class WindDemo extends Component {
       <DeckGL
         glOptions={{webgl2: true}}
         {...viewport}
-        layers={layers} />
+        layers={layers}
+        useDevicePixelRatio={true} />
     );
   }
 }
--- a/src/lib/layer-manager.js
+++ b/src/lib/layer-manager.js
@@ -75,6 +75,7 @@ export default class LayerManager {
       viewport: null,
       viewportChanged: true,
       pickingFBO: null,
+      useDevicePixelRatio: true,
       lastPickedInfo: {
         index: -1,
         layerId: null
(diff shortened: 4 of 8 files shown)


src/react/deckgl.js
```diff
@@ -55,7 +56,8 @@ const defaultProps = {
   onWebGLInitialized: noop,
   onAfterRender: noop,
   onLayerClick: null,
-  onLayerHover: null
+  onLayerHover: null,
+  useDevicePixelRatio: false
```
1chandu: @ibgreen ,  this will change existing apps behavior to use non-retina resolution, do you think I should use `true` for default, then set to false only for mobile devices?

docs/api-reference/deckgl.md
```diff
@@ -75,9 +75,13 @@ This is helpful when rendered objects are difficult to target, for example
 irregularly shaped icons, small moving circles or interaction by touch.
 Default `0`.
 
-##### `pixelRatio` (Number, optional)
```
1chandu: This is not used anywhere.

src/react/deckgl.js
ibgreen: Good question. My take: Unless the default changed between 4.0 and 4.1 (in which case we could consider it a bug), let's keep it unchanged for now (`true`) and decide in review based on performance data. 

This PR is intended for a patch release after all, and is already pushing it by adding a new prop.

docs/api-reference/deckgl.md
ibgreen: > `pixelRation` is not used anywhere.

Maybe do a quick investigation and figure out when this was dropped. There is no mention in what's new or upgrade guides about this...

docs/api-reference/deckgl.md
```diff
@@ -75,9 +75,13 @@ This is helpful when rendered objects are difficult to target, for example
...
+##### `useDevicePixelRatio` (Boolean, optional)
 
-Will use device ratio by default.
+Default value is false.
+
+When true, device's full resolution will be used for rendering.
```
ibgreen: I think it would be valuable to add a note here (or in the remarks) that DPR is resampled every frame.

docs/api-reference/deckgl.md
```diff
@@ -75,9 +75,13 @@ This is helpful when rendered objects are difficult to target, for example
...
+
+Note: Only effects devices with retina/HD displays. Set it false (default) unless it is required, as it affects your application performance.
```
ibgreen: Add: "DPR can also change when zooming in the browser and when moving windows between screens."?

For the remarks section?

docs/whats-new.md
```diff
@@ -1,5 +1,9 @@
 # deck.gl v4.2
 
+## useDevicePixelRatio
+
+This new prop for DeckGL component can be used to toggle usage of full resolution of retina/HD displays.
```
ibgreen: Note: Enabling retina/high DPI rendering can have a big impact on performance.

Other comments:
* We probably need a paragraph in the `advanced/performance.md` doc that talks about this if not already there.

examples/wind/src/wind-demo.js
```diff
@@ -110,7 +110,8 @@ export default class WindDemo extends Component {
       <DeckGL
         glOptions={{webgl2: true}}
         {...viewport}
-        layers={layers} />
+        layers={layers}
+        useDevicePixelRatio={true} />
```
ibgreen: For the API review: I wonder if there is a better name... `useDevicePixels`?

src/lib/layer-manager.js
```diff
@@ -39,7 +39,7 @@ const LOG_PRIORITY_LIFECYCLE = 2;
 const LOG_PRIORITY_LIFECYCLE_MINOR = 4;
 
 export default class LayerManager {
-  constructor({gl}) {
+  constructor({gl, useDevicePixelRatio}) {
```
ibgreen: > constructor({gl, useDevicePixelRatio}) {

Why is this the only arg that goes into the constructor? 
Can this only be changed on construction? (If so, that limitation should be documented).
Shouldn't this be set in some `LayerManager.setOptions` method that can be called whenever props change?

docs/api-reference/deckgl.md
1chandu: It exists in 2.0 release, dropped in 3.0, so pretty old.

ibgreen: OK. 3.0 was the first release we promoted to the public so that should be OK to just remove it then. I suspect are no apps that use 2.0.

Issue #866, opened 2017-08-21
Maximum call stack size exceeded with 1.3M data points


fanez:
I'm trying to load a file with 1.3M points into hexagon layer (using the 3d-heatmap example) but I get this error in Chrome:

```
deck.gl deck.gl error during update of <HexagonLayer:'heatmap'> RangeError: Maximum call stack size exceeded
    at BinSorter.getMaxCount (bin-sorter.js:104)
    at new BinSorter (bin-sorter.js:45)
    at HexagonLayer.getSortedBins (hexagon-layer.js:172)
    at HexagonLayer.updateState (hexagon-layer.js:129)
    at HexagonLayer.updateLayer (layer.js:537)
    at LayerManager._updateLayer (layer-manager.js:701)
    at _loop (layer-manager.js:514)
    at LayerManager._matchSublayers (layer-manager.js:562)
    at LayerManager._updateLayers (layer-manager.js:465)
    at LayerManager.updateLayers (layer-manager.js:257)
```

ibgreen:
This should be a one line fix, just replace the apply with a loop.

ibgreen:
Fixed on master, will be published in patch release soon.

Issue #793, opened 2017-07-19
grid-aggregator fails on large number of coordinate points. 


shawel:
grid-aggregator.js fails on large number of points when trying to find max & min lat longs because of 
 the following two lines
```
const latMin = Math.min.apply(null, allLat);
const latMax = Math.max.apply(null, allLat);
```
Number of arguments on apply is not consistently implemented across browsers and it fails on some arbitrary limit. It is best to switch to something like  `max =Math.max(max, latlng) `  & it would help with performance to calculate both at once.

gnavvy:
Hi @shawel,

> Number of arguments on apply is not consistently implemented across browsers and it fails on some arbitrary limit. 
Thanks for reporting the issue.
Could you provide your dev env info so we can test after we made the fix?

ibgreen:
@heshan0131 fyi

ibgreen:
Fixed on master. Will be published in patch release soon.

PR #868, opened 2017-08-21
Fix Maximum call stack size exceeded error in grid aggregator

heshan0131:


heshan0131:
#866 #793

src/layers/core/hexagon-layer/hexagon-aggregator.js
```diff
@@ -33,10 +33,14 @@ export function pointToHexbin({data, radius, getPosition}, viewport) {
   // get hexagon radius in mercator world unit
   const radiusInPixel = getRadiusInPixel(radius, viewport);
 
+  const screenPoints = [];
+
   // add world space coordinates to points
-  const screenPoints = data.map(pt => Object.assign({
-    screenCoord: viewport.projectFlat(getPosition(pt))
-  }, pt));
+  for (let i = 0; i < data.length; i++) {
```
ibgreen: Nit: I think it is the `apply` that is the issue, not the `map` call. 

The `map` might actually be faster than the repeated `push`es as it can allocate the return array to the right size from the start.

1chandu:
@heshan0131 , are there any pending changes for this, can we merge?

heshan0131:
@ib, yes it can be merged

Issue #800, opened 2017-07-21
s2layer sample not working


minhdanh:
I start the examples on my machine. All the layers display on the map when I click on that layer in the layer browser, except for s2Layer. It doesn't show anything. Is this a bug or something?
Here is the screen shot:

![screenshot 2017-07-21 11 59 22](https://user-images.githubusercontent.com/4865412/28449712-47809a08-6e0c-11e7-8040-19e6cbea167f.png)

1chandu:
Fixed by #869

minhdanh:
Thank you

PR #869, opened 2017-08-21
Fix S2Layer base class and props.

1chandu:
Fix for #800

### DIFF
--- a/examples/layer-browser/src/examples/sample-layers.js
+++ b/examples/layer-browser/src/examples/sample-layers.js
@@ -10,7 +10,8 @@ const S2LayerExample = {
     opacity: 0.6,
     getS2Token: f => f.token,
     getPath: f => f.path,
-    getFillColor: f => [f.value * 256, (1 - f.value) * 256, (1 - f.value) * 128],
+    getStrokeColor: f => [255, 255, 128],
+    getFillColor: f => [255, 255, 128],
     getStrokeWidth: f => 10,
     pickable: true
   }
--- a/examples/sample-layers/s2-layer/s2-layer.js
+++ b/examples/sample-layers/s2-layer/s2-layer.js
@@ -38,7 +38,7 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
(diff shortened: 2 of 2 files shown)


examples/layer-browser/src/examples/sample-layers.js
```diff
@@ -10,7 +10,8 @@ const S2LayerExample = {
     opacity: 0.6,
     getS2Token: f => f.token,
     getPath: f => f.path,
-    getFillColor: f => [f.value * 256, (1 - f.value) * 256, (1 - f.value) * 128],
+    getStrokeColor: f => [255, 255, 128],
```
1chandu: >     getStrokeColor: f => [255, 255, 128],
>     getFillColor: f => [255, 255, 128],

When color-picker.js calls these methods, it doesn't pass any parameters, `f` is going to be undefined, kept them in same style as `getStrokeWidth` below, but may be we should just change them to `callback: () => [255, 255, 128]` ?

examples/sample-layers/s2-layer/s2-layer.js
```diff
@@ -64,13 +64,17 @@ const defaultProps = {
   getFillColor: f => f.fillColor || defaultFillColor
 };
 
-export default class S2Layer extends Layer {
+export default class S2Layer extends CompositeLayer {
+  constructor(props) {
+    super(props);
+  }
+
   initializeState() {
```
ibgreen: I think this can be cut on compositeLayers.

1chandu: good catch, I added it when debugging, forgot to remove.

ibgreen: I was referring to  `initializeState` but I guess the constructor could also be cut.

examples/layer-browser/src/examples/sample-layers.js
ibgreen: Not sure what happened to my comments, but we should keep the feature dependent fill color or we lose the ability to show neighbouring S2 cells, which is a key part of the demo.

We need to fix this in another place

1chandu: Fixed here : #870

PR #872, opened 2017-08-23
A cleaner version of tagMap

rivulet-zhang:
Here are a few improvements to the `tagmap` example:
- Remove the image and the mapbox style file from the repo to keep the size down
- Use d3-request instead of axios for consistency
- Use `componentWillMount` instead of `componentDidMount` to avoid the resizing issue during initialization
- Add a performance monitoring widget
- Remove the code for the tagmap algorithm and import it as an independent library to make the repo cleaner
- Use text-layer (#804)  instead of the label-util function. (the text-layer code can be removed once it is integrated as a core layer in deck.gl)


examples/tagmap/package.json
```diff
@@ -5,16 +5,17 @@
...
     "d3-scale": "^1.0.6",
     "deck.gl": "^4.1.0",
     "luma.gl": "^4.0.1",
-    "rbush": "^2.0.1",
     "react": "^15.4.1",
     "react-dom": "^15.4.1",
-    "react-map-gl": "^3.0.0"
+    "react-map-gl": "^3.0.0",
+    "stats.js": "^0.17.0",
+    "tagmap.js": "^1.1.0"
```
shaojingli: So we added this additional dependency. Is this tagmap.js doing what functions in source files deleted in PR do?

examples/tagmap/tagmap-layer/text-layer/multi-icon-layer/multi-icon-layer-vertex-64.glsl.js
```diff
@@ -0,0 +1,116 @@
...
+
+varying float vColorMode;
+varying vec4 vColor;
+varying vec2 vTextureCoords;
+
+// rotation should factor in the aspect ratio
+vec2 rotate_by_angle(vec2 vertex, float angle, float aspectRatio) {
+  float angle_radian = angle * PI / 180.0;
+  float cos_angle = cos(angle_radian);
+  float sin_angle = sqrt(1.0 - cos_angle * cos_angle);
```
shaojingli: This is actually a bug from my previous version of code. Calculating sin like this will make it always positive... limiting the rotating angle to be always in 0-180. let's just use `sin_angle = sin(angle_radian)`

examples/tagmap/package.json
rivulet-zhang: Yes, the `tagmap.js` does what the `tagmapper` directory did in the previous version.

rivulet-zhang:
@shaojingli The PR is rebased and you can feel free to merge it.

Issue #873, opened 2017-08-23
New `parameters` prop not documented and not working for `CompositeLayers`


ibgreen:
Looks like we do not have any documentation for the parameter prop on DeckGL layers.
* parameters is a prop on Layer class, so I would expect it to be in api-reference/base-layer.md, but it is not mentioned there.
* The docs need to link to the luma.gl `setParameters` API as all the parameters are described there.
* We renamed `settings` to `parameters` late in the cycle , so we have to double check and update the docs as well in case something wasn't updated.

* `parameters` is not mentioned in [What's New](https://uber.github.io/deck.gl/#/documentation/overview/whats-new). This needs to be addressed. 

Also I can't get `parameters` to work on composite layers (`GeoJsonLayer`), probably the parameters are just not being forwarded.

PR #874, opened 2017-08-23
Layer.parameters: Update docs and forward prop in GeojsonLayer

ibgreen:


ibgreen:
See #873

ibgreen:
@georgios-uber

georgios-uber:
Great

georgios-uber:
Great

1chandu:
@ibgreen looks good, thanks for taking care of this.

One thing I notices when looking into layer.js , `this.props` is never used, but then I saw , `layer.props.parameters` are passed back to `drawLayer` method, which was bit confusing. May be we should only pass any extra `parameters` to this method and then merge them with this.parameters.

PR #875, opened 2017-08-23
Add depthTest:false to ScreenGridLayer

### LINKED ISSUE #602: ScreenGrid layer affects rendering of other layers in layer browser
A number of layers render incorrectly when the ScreenGridLayer is turned on in the layer browser:

HexagonCellLayer + ScreenGridLayer

PointCloudLayer + ScreenGridLayer

Not sure if this is a regression or a long-standing never-before-noticed bug.

ibgreen:
Fixes #602 - removes the artifacts in layer-browser. We can discuss if this should be the default parameter setting in `ScreenGridLayer`.


ibgreen:
Changed my mind, updated the default settings in the layer + made sure app can override.

PR #876, opened 2017-08-23
Update docs for `getShaders()`

### LINKED ISSUE #867: `getShaders()` interaction with new luma.gl shader modules
In 4.1 we have added a new field to the object returned by `getShaders()`: `modules`.

This means that it is often no longer correct to just return a new shader object when overriding `getShaders()`

Instead one either needs to copy the original object and selectively overwrite fields, or copy the modules list.

We should update the docs in `docs/advanced/subclassed-layers.md` to make this clear, and review all subclassed layers (including `sample-layers`) to make sure this is done correctly in all places.

1chandu:
Fixes #867

### DIFF
--- a/docs/advanced/subclassed-layers.md
+++ b/docs/advanced/subclassed-layers.md
@@ -59,16 +59,21 @@ export default MultiColorPathLayer extends PathLayer {
 
 You can replace the shaders used in a layer by overriding the `getShaders()`
 method. Every core layer calls this method during initialization. It
-returns the shaders used by the layer in an object:
+returns the shaders and modules used by the layer in an object:
 - `vs`: string, GLSL source of the vertex shader
 - `fs`: string, GLSL source of the fragment shader
+- `modules`: Array, list of shader modules to be used
 
 Read about [writing your own shaders](/docs/advanced/writing-shaders.md).
 
-When you are implementing your own custom layers, it is encouraged that you
-also define a `getShaders()` function to retreive your shaders.
+When you are implementing your own custom layers, and want to change the shaders
+it is encouraged that you also define a `getShaders()` function and selectively
+overwrite required shader(s) with custom shaders.
 This makes it much easier for others to subclass your layer and make small
 changes to the shaders.
+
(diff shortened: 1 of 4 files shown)


docs/advanced/subclassed-layers.md
```diff
@@ -59,17 +59,22 @@ export default MultiColorPathLayer extends PathLayer {
...
 
-When you are implementing your own custom layers, it is encouraged that you
-also define a `getShaders()` function to retreive your shaders.
+When you are implementing your own custom layers, and want to change the shaders
+it is encouraged that you also define a `getShaders()` function and selectively
+overwrite required shader(s) with custom shaders.
 This makes it much easier for others to subclass your layer and make small
 changes to the shaders.
 
+Note: When overwriting `getShaders()` you pass down any unmodified shader(s)
```
ibgreen: `you pass` => ` you should pass` ...  `See code example below.`

docs/advanced/subclassed-layers.md
```diff
@@ -94,10 +99,10 @@ export default RoundedRectangleLayer extends ScatterplotLayer {
   }
 
   getShaders() {
-    return {
-      vs: super.getShaders().vs,
-      fs: fragmentShader
-    }
+    const shaders = Object.assign({}, super.getShaders(), {
```
ibgreen: Add a comment - 
```
// use object.assign to make sure we don't overwrite existing fields like `vs`, `modules`...
```

docs/advanced/subclassed-layers.md
```diff
@@ -59,17 +59,22 @@ export default MultiColorPathLayer extends PathLayer {
 
 You can replace the shaders used in a layer by overriding the `getShaders()`
```
ibgreen: Could we add any links to luma.gl shader modules docs (or maybe they are still pretty thin?)

1chandu: luma.gl's shader module documentation talks more about `assemberShaders` and `Model` class,  not about shader modules in general, probably not a good link.

1chandu:
Verified sample-layers, most extend Layer. Fixed in few examples.

PR #877, opened 2017-08-23
EventManager docs + cleanup

ericsoco:
- Update EventManager docs
- Centralize constants into constants.js
- Move constants that need to be conditionally required into their own file, to avoid conditional requires where possible

### DIFF
--- a/docs/api-reference/event-manager.md
+++ b/docs/api-reference/event-manager.md
@@ -2,9 +2,14 @@
 
 Provides a unified API for subscribing to events about both basic input events (e.g. 'mousemove', 'touchstart', 'wheel') and gestural input (e.g. 'click', 'tap', 'panstart').
 
+Delegates gesture-related event registration and handling to Hammer.js. Includes shims for handling event not supported by Hammer.js, such as keyboard input, mouse move, and wheel input.
+
 ## Usage
 
 TBA
+
+(What is the intent of this section? Install/import instructions?)
+
 
 
 ## Methods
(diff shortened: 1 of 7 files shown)


docs/api-reference/event-manager.md
```diff
@@ -2,10 +2,15 @@
 
 Provides a unified API for subscribing to events about both basic input events (e.g. 'mousemove', 'touchstart', 'wheel') and gestural input (e.g. 'click', 'tap', 'panstart').
 
+Delegates gesture-related event registration and handling to Hammer.js. Includes shims for handling event not supported by Hammer.js, such as keyboard input, mouse move, and wheel input.
+
 ## Usage
 
 TBA
 
+(What is the intent of this section? Install/import instructions?)
```
ibgreen: It is intended for a couple of examples of how to use this class. 

Create an event manager
```js
const manager = new EventManager({
```

Register events
...

docs/api-reference/event-manager.md
```diff
@@ -50,11 +55,86 @@ Register an event handler function to be called on `event`.
 * `[handler]` {Function} - The function to be called on `event`.
 
 
+
 ## Supported Events and Gestures
 
-TBA
+As of the time of this writing, the following event and gesture names are supported. Refer to [`constants.js`](../../src/controls/events/event-manager.js) for the most current support.
```
ibgreen: I think we can drop this, we should just keep it up to date in PR reviews, that is how we handle other code.

docs/api-reference/event-manager.md
```diff
@@ -2,10 +2,15 @@
 
 Provides a unified API for subscribing to events about both basic input events (e.g. 'mousemove', 'touchstart', 'wheel') and gestural input (e.g. 'click', 'tap', 'panstart').
 
+Delegates gesture-related event registration and handling to Hammer.js. Includes shims for handling event not supported by Hammer.js, such as keyboard input, mouse move, and wheel input.
```
ericsoco: No, I agree -- we should leave ourselves wiggle room. Will edit.

ericsoco:
@Pessimistress merging now but please review the docs and add/correct anything I missed.

PR #893, opened 2017-08-28
Clear hovered object when pointer leaves the canvas

Pessimistress:
https://github.com/uber/deck.gl/issues/847

### DIFF
--- a/src/controllers/events/move-input.js
+++ b/src/controllers/events/move-input.js
@@ -43,13 +43,11 @@ export default class MoveInput {
     }
     if (eventType === LEAVE_EVENT_TYPE) {
       this.enableLeaveEvent = enabled;
-      console.log(this.enableLeaveEvent); // eslint-disable-line
     }
   }
 
   handleEvent(event) {
     if (this.enableLeaveEvent) {
-      console.log(this.enableLeaveEvent, event); // eslint-disable-line
       if (event.type === 'mouseleave') {
         this.callback({
           type: LEAVE_EVENT_TYPE,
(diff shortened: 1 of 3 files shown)


src/controllers/events/move-input.js
```diff
@@ -43,13 +43,11 @@ export default class MoveInput {
     }
     if (eventType === LEAVE_EVENT_TYPE) {
       this.enableLeaveEvent = enabled;
```
ibgreen: Not related to this PR: 

I'm starting to think that we should maybe use a map with keys for the enabling of events, instead of creating duplicate `enable...` variables for all the events and gestures. 

But maybe the mapping is not one-to-one?

src/lib/draw-and-pick.js
```diff
@@ -265,6 +265,17 @@ function getClosestFromPickingBuffer(gl, {
   deviceY,
   deviceRadius
 }) {
+  let closestResultToCenter = {
+    pickedColor: EMPTY_PIXEL,
+    pickedLayer: null,
+    pickedObjectIndex: -1
+  };
```
ibgreen: While I realize you just moved this around, shouldn't this be an `EMPTY_PICK_INFO` constant at the top of the file?

And aren't there more fields in a pickInfo object [(doc link)](https://github.com/uber/deck.gl/blob/master/docs/get-started/interactivity.md#the-picking-info-object) that should be initialized to default values? (if there are no reasonable defaults, setting to `null` is better than leaving as `undefined` in my view as it shows that this was an explicit, intentional choice).

Pessimistress: This object is not the `pickInfo` object. It is only used by the other functions in draw-and-pick to create `pickInfo`.

ivanaxel:
cool

daumann:
@Pessimistress Has this been fixed? In the example pages, the hovered elements still do not clear (tooltip stays at last hovered item without cleanup on mouseleave - the same goes for autoHighlighted items).

PR #897, opened 2017-08-30
Experimental:  Add AnimationMapController for viewport animations.

1chandu:
- Clone `MapController` to `experimental/react/controllers` as `AnimationMapController`, add viewport animation support.
- `AnimationMapController` takes new prop, `aniamteViewport`, when true, a change viewport results in animation.
- `AnimationMapController` updates all its children recursively and sets their viewport prop to animated viewport, right before they get rendered.
- Other props added to control animation are : `viewportAnimationDuration`, and `viewportAnimationEasingFunc`.
- Temporarily clone `3d-heatmap` and `geojson` examples to examples/wip/viewport-animation to test viewport animation.

**Remaining work** 
- Some times map doesn't show up when animating, need to debug this issue.
- Adding `animationStyle` prop that control relative change between different viewport props (includes adding features like, zoom out before changing location etc)
- Documentation, will add docs once RFC is discussed with the team.


1chandu:
ezgif com-optimize

ivanaxel:
added

1chandu:
@ibgreen code is not ready to merge, as I left few comments and need cleanup, but if you have any feedback I can address that and make this PR ready to land.

ivanaxel:
cool

src/react/controllers/map-controller.js
```diff
@@ -109,18 +113,122 @@ export default class MapController extends PureComponent {
...
-  _onInteractiveStateChange({isDragging = false}) {
+  _onInteractiveStateChange(interactiveState) {
+    const {isDragging = false} = interactiveState;
     if (isDragging !== this.state.isDragging) {
       this.setState({isDragging});
     }
   }
 
+  _extractViewportFromProps(props) {
+    return {
```
ibgreen: More of a discussion point: A question is if this component should just take a `viewport` or `viewportState` prop so we don't have to unpack and pack these params again and again. 

This would also allow us to use the various `State` classes directly as input.

We could customize `shouldComponentUpdate` to compare the members of that field.

src/react/controllers/map-controller.js
```diff
@@ -109,18 +113,122 @@ export default class MapController extends PureComponent {
...
+    if (this.props.animateViewport !== true) {
+      return;
+    }
+    const startViewport = this._extractViewportFromProps(this.props);
+    const endViewport = this._extractViewportFromProps(nextProps);
+    if (this._viewportChanged(startViewport, endViewport)) {
+      if (this.state.animationInterval) {
+        clearInterval(this.state.animationInterval);
+      }
+      const animationInterval = setInterval(() => this._updateViewport(), 50);
```
ibgreen: Perhaps the interval management can be isolated into some functions?

src/react/controllers/map-controller.js
```diff
@@ -109,18 +113,122 @@ export default class MapController extends PureComponent {
...
+        animationT: 0.0,
+        animationStartViewport: startViewport,
+        animationEndViewport: endViewport,
+        animationInterval,
+        animatedViewport: startViewport
+      });
+    }
+  }
+
+  _viewportChanged(startViewport, endViewport) {
```
ibgreen: A one-line comment here.  It is only looking for animated props, should the function name hint about that?
`didViewportAnimatePropsChange`?

src/react/controllers/map-controller.js
```diff
@@ -109,18 +113,122 @@ export default class MapController extends PureComponent {
...
+        endViewport[p] !== undefined &&
+        startViewport[p] !== endViewport[p]) {
+          // console.log(`viewport changed : ${p}: ${startViewport[p]}->${endViewport[p]}`);
+        return true;
+      }
+    }
+    return false;
+  }
+
+  _updateViewport() {
```
ibgreen: One line comment header. 

Name: `_handleViewportAnimationTimer` or `_onViewportAnimationTimer` ?

src/react/controllers/map-controller.js
```diff
@@ -109,18 +113,122 @@ export default class MapController extends PureComponent {
...
+      animatedViewport: null
+    });
+  }
+
+  _recursiveCloneChildren(children, viewport) {
+    return Children.map(children, child => {
+      if (!isValidElement(child)) {
+        return child;
+      }
+      // TODO: we need to filter chidren and only update those that require
```
ibgreen: One technique could be to look for viewport props on the children. If they don't have the props you are animating, no need to clone them...

src/react/controllers/map-controller.js
1chandu: I like this idea, this way it is also easy if a child component contains `viewportState` prop, and only update then.

src/react/controllers/map-controller.js
1chandu: If we decided to group all viewport props into `viewportState`, this gets easier, we can just check if a child contains this prop, instead of checking 7/8 individual props.

examples/wip/viewport-animation/3d-heatmap/app.js
```diff
@@ -0,0 +1,111 @@
...
+    newViewport.pitch = (this.state.viewport.pitch === 0) ? 60.0 : 0.0;
+    newViewport.bearing = (this.state.viewport.bearing === 0) ? -60.0 : 0.0;
+    this.setState({
+      viewport: {...this.state.viewport, ...newViewport}
+    });
+  }
+
+  render() {
+    const {viewport, data} = this.state;
+    // const eventCanvasStyle = {
```
ibgreen: Should we delete these commented lines, or add a TODO?

examples/wip/viewport-animation/3d-heatmap/app.js
```diff
@@ -0,0 +1,111 @@
...
+    //   width: viewport.width,
+    //   height: viewport.height,
+    //   position: 'relative'
+    // };
+    return (
+      <MapController
+      /*
```
ibgreen: And these?

examples/wip/viewport-animation/3d-heatmap/README.md
```diff
@@ -0,0 +1,36 @@
+This is a minimal standalone version of the 3DHeatmap example
```
ibgreen: I'd prefer a flat directory structure for the examples in `wip`, it will make it easier to see what is being built for the new release at a glance. E.g.:
```
examples/wip/viewport-animation-3d-heatmap
examples/wip/viewport-animation-geojson
```

examples/wip/viewport-animation/geojson/README.md
```diff
@@ -0,0 +1,13 @@
+This is a minimal standalone version of the GeoJsonLayer example
```
ibgreen: Update this README (or delete it for now)

src/experimental/react/controllers/map-controller.js
```diff
@@ -0,0 +1,271 @@
+/* global setInterval, clearInterval */
```
ibgreen: Let's call this `animated-map-controller.js` / `AnimatedMapController` for now.

Then you can also export it as an additional experimental export at the end of `src/index.js` so that I can use it.

src/experimental/react/controllers/map-controller.js
```diff
@@ -0,0 +1,271 @@
...
+  scrollZoom: true,
+  dragPan: true,
+  dragRotate: true,
+  doubleClickZoom: true,
+  touchZoomRotate: true,
+
+  getCursor: getDefaultCursor
+});
+
+export default class MapController extends PureComponent {
```
ibgreen: `AnimatedMapController`

src/experimental/react/controllers/map-controller.js
```diff
@@ -0,0 +1,271 @@
...
+        ref: 'eventCanvas',
+        style: eventCanvasStyle
+      },
+        childrenWithProps
+      )
+    );
+  }
+}
+
+MapController.displayName = 'MapController';
```
ibgreen: Animated...

PR #898, opened 2017-08-31
Support dashed line in PathLayer

Pessimistress:
screen shot 2017-08-30 at 5 29 01 pm

### DIFF
--- a/docs/layers/path-layer.md
+++ b/docs/layers/path-layer.md
@@ -78,6 +78,12 @@ Only works if `rounded` is `false`.
 
 Whether the layer should be rendered in high-precision 64-bit mode
 
+##### `justified` (Boolean, optional)
+
+- Default: `false`
+
+Only effective if `getDashArray` is specified. If `true`, adjust gaps for the dashes to align at both ends.
+
 ### Data Accessors
 
 ##### `getPath` (Function, optional)
@@ -105,6 +111,15 @@ If the color alpha (the fourth component) is not provided,
 
 Method called to determine the width to draw each path with.
 Unit is meters.
+
+##### `getDashArray` (Function, optional)
+
+- Default: `null`
+
+Method called to get the dash array to draw each path with.
+Returns an array of two numbers: `[dashSize, gapSize]` relative to the width of the path. Returns `[0, 0]` to draw the path in solid line.
+
+If this accessor is not specified, all paths are drawn as solid lines.
 
 ## Source
 
--- a/src/layers/core/path-layer/path-layer-vertex-64.glsl.js
+++ b/src/layers/core/path-layer/path-layer-vertex-64.glsl.js
@@ -32,6 +32,7 @@ attribute vec3 instanceRightDeltas;
 attribute float instanceStrokeWidths;
 attribute vec4 instanceColors;
 attribute vec3 instancePickingColors;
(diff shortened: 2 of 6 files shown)


docs/layers/path-layer.md
```diff
@@ -106,6 +112,15 @@ If the color alpha (the fourth component) is not provided,
 Method called to determine the width to draw each path with.
 Unit is meters.
 
+##### `getDashArray` (Function, optional)
```
ibgreen: Fine for now. For the API audit, thinking that this name could be clearer and might be better as `getDashSizeAndGap`, or maybe `getDashParameters`, it is more extensible.

src/layers/core/path-layer/path-layer-vertex-64.glsl.js
```diff
@@ -163,6 +167,14 @@ vec3 lineJoin(vec2 prevPoint64[2], vec2 currPoint64[2], vec2 nextPoint64[2]) {
 
   vCornerOffset = offsetVec * offsetDirection * offsetScale;
 
+  // Generate variables for dash calculation
+  vDashArray = instanceDashArrays;
+  vPathLength = L / width;
```
ibgreen: I am not a fan of abbreviated identifiers (of which there are many in this shader), single letter ones even less so.

src/layers/core/path-layer/path-layer-vertex.glsl.js
```diff
@@ -150,6 +154,14 @@ vec3 lineJoin(vec3 prevPoint, vec3 currPoint, vec3 nextPoint) {
 
   vCornerOffset = offsetVec * offsetDirection * offsetScale;
 
+  // Generate variables for dash calculation
```
ibgreen: FP64 layers still involve tedious copying I see. Makes me wonder if we should have a `-vertex-common.glsl.js`. There must be some way to consolidate this, I'll try to think of something.

ibgreen:
@georgios-uber

docs/layers/path-layer.md
Pessimistress: I was thinking the `stroke-dasharray` style in SVG.

ibgreen: Sure. If there is good precedent, then why not. Without that context it sounded odd. 

Taking a look at the [`stroke-dasharray` style](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-dasharray) I see that unlike ours, that style actually supports actual arrays of multiple dashes and gaps. So not 100% obvious but if our other names are consistently based on SVG styles it may be the right choice.

ericsoco: True, and you can do a lot of nice tricks with that, e.g. [animating a path](https://bl.ocks.org/mbostock/5649592). We should consider future support for the full `stroke-dasharray` path as a future feature.

ibgreen: > you can do a lot of nice tricks with that, e.g. animating a path.

Cool! While that animation might take some work to achieve, it struck me that a simple `gapOffset` property/uniform could allow us to easily animate a ["marching ants"](https://ideastocreations.blogspot.com/2016/06/ae-building-selection-box-shaped-by-2.html)  selection.

Issue #834, opened 2017-08-09
Issue with icon rotation in IconLayer


rivulet-zhang:
It seems that rotating an icon generates distortion. It's mainly an issue with the aspect ratio. See the animation below in which a tall and thin rectangle is rotated around its center.
<img width="300" alt="screen shot 2017-07-11 at 4 08 35 pm" src="https://user-images.githubusercontent.com/10911507/29138118-a18afefe-7d10-11e7-880a-2fed344cefaf.gif">
The correct result should be something like this:
<img width="300" alt="screen shot 2017-07-11 at 4 08 35 pm" src="https://user-images.githubusercontent.com/10911507/29138143-b7e37c80-7d10-11e7-967c-cd81de5b4abd.gif">

It looks like in the vertex shader, the vertex is rotated before scaling, which stretches the texture wrongly.

```javascript
void main(void) {
  ...
  vertex = rotate_by_angle(vertex, instanceAngles) * iconSize_clipspace *
    sizeScale * instanceScale;
  ...
}
```
Here is a solution that correctly handles the rotation. Basically we want to scale before rotation. Then in the rotation function, we need to factor in the aspect ratio since the vertex contains fraction relative to the viewport size rather than the absolute size in pixels. The rotation looks like an ellipse rather than a circle.
```javascript
vec2 rotate_by_angle(vec2 vertex, float angle, float aspect_ratio) {
  ...
  float aspect_ratio_reverse = aspect_ratio == 0.0 ? aspect_ratio : 1.0 / aspect_ratio;
  mat2 rotationMatrix = mat2(cos_angle, - aspect_ratio_reverse * sin_angle, aspect_ratio * sin_angle, cos_angle);
  ...
}

void main(void) {
  ...
  float aspect_ratio = viewportSize.x == 0.0 ? viewportSize.x : viewportSize.y / viewportSize.x;
  vertex *= iconSize_clipspace;
  vertex = rotate_by_angle(vertex, instanceAngles, aspect_ratio) * sizeScale * instanceScale;
  ...
}
```

I did not create a PR for this for now since there might be better ways to handle this. Let me know if you think this is a reasonable solution and would want me to create a PR.

ibgreen:
@rivulet-zhang - looks like you have found (and solved) an important issue. 

I am trying to understand why this has not been an issue in other contexts. Is the extreme aspect ratio issue in your example the result of an very non-square screen? And the aspect ratio issue disappears if the screen is resized to be square?

[...]

Issue #895, opened 2017-08-30
IconLayer renders icons at small sizeScales with significant loss of quality


tangential:
I'm trying to render a small triangle pointing to the direction of travel on a map.

Here's the asset I'm currently using:
![direction-pointer-alt](https://user-images.githubusercontent.com/836037/29850292-49a6507c-8ce1-11e7-8442-533ecf167fc6.png)
Dimension: 7 x 10

When rendered on a map, I see that there is significant loss of quality, along with significant distortion (tracked in issue [834](https://github.com/uber/deck.gl/issues/834)). 
![screenshot 2017-08-29 17 38 32](https://user-images.githubusercontent.com/836037/29850325-940c067a-8ce1-11e7-8067-c018f1703677.png)

Here's how I'm instantiating the layer:
```
    new IconLayer({
      id: 'direction-pointer',
      data: arrows,
      fp64: true,
      getColor: (d, i) => {
        const opacity = d.index < currentIndex ? 255 : 0;
        return [0, 0, 0, opacity];
      },
      iconAtlas: assetUrl('/static/direction-pointer.png'),
      iconMapping:  {
        marker: {
          x: 0,
          y: 0,
          width: 7,
          height: 10,
          mask: true
        }
      },
      sizeScale: 10
    });
```

Pessimistress:
@1chandu @ibgreen Could this be related to the power-of-two texture change in luma v4?

1chandu:
Luma v4 uses WebGL2 by default,  NPOT (non power-of-two textures) textures are supported and shouldn't cause any issue.

I will take a look at this issue, after I wrap my current tasks, should be in next couple of days.

1chandu:
Need to re-test this with #834 fix (@rivulet-zhang working on it).

rivulet-zhang:
@1chandu #900 handles the distortion caused by rotation. As a separate issue, the scaling of the icon seems to be inaccurate, which could relate to this issue. Say I am rendering an icon with `getSize: d => s1` and `sizeScale = s2`, the resulting icon actually has a much smaller size as opposed to `s1 * s2`. This exists in the original version and #900 does not handle this for now. Let me know if you want me to look into this.

ibgreen:
> @1chandu @ibgreen Could this be related to the power-of-two texture change in luma v4?

Kind of, I was more thinking that this could be related to a change in default texture filtering parameters. We need to make sure we are using the best texture filtering.

ibgreen:
[...]

PR #900, opened 2017-08-31
Fix the icon rotation issue

rivulet-zhang:
A different (and better) fix for #834 as opposed to the way originally proposed. Manipulate the vertex scaling and rotation in "pixel" values before converting back to "fraction" in the screen space.  No need to calculate the aspect ratio and factor it in the rotation function. To do this, it is better to pass the anchor position into the shader as pixel values instead of normalizing it (see the change in `icon-layer.js`). The `devicePixelRatio` is used to avoid hardcoding in the previous implementation.

### DIFF
--- a/src/layers/core/icon-layer/icon-layer-vertex-64.glsl.js
+++ b/src/layers/core/icon-layer/icon-layer-vertex-64.glsl.js
@@ -36,6 +36,7 @@ attribute vec2 instanceOffsets;
 uniform vec2 viewportSize;
 uniform float sizeScale;
 uniform vec2 iconsTextureDim;
+uniform float devicePixelRatio;
 
 varying float vColorMode;
 varying vec4 vColor;
@@ -51,15 +52,13 @@ vec2 rotate_by_angle(vec2 vertex, float angle) {
 
 void main(void) {
   vec2 iconSize = instanceIconFrames.zw;
-  vec2 iconSize_clipspace = iconSize / viewportSize * 2.0;
   // scale icon height to match instanceSize
   float instanceScale = iconSize.y == 0.0 ? 0.0 : instanceSizes / iconSize.y;
 
-  // The vertex variable is in clip space and should not go through project_to_clipspace call
-  vec2 vertex = (positions / 2.0 + instanceOffsets);
-  vertex = rotate_by_angle(vertex, instanceAngles) * iconSize_clipspace *
-    sizeScale * instanceScale;
-
+  // scale and rotate vertex in "pixel" value and convert back to fraction in clipspace
+  vec2 vertex = positions / 2.0 * iconSize + instanceOffsets;
+  vertex = rotate_by_angle(vertex, instanceAngles) / viewportSize * sizeScale *
+    instanceScale * devicePixelRatio;
   vertex.y *= -1.0;
 
   vec4 instancePositions64xy = vec4(
--- a/src/layers/core/icon-layer/icon-layer-vertex-64.glsl.js
+++ b/src/layers/core/icon-layer/icon-layer-vertex-64.glsl.js
@@ -36,6 +36,7 @@ attribute vec2 instanceOffsets;
 uniform vec2 viewportSize;
 uniform float sizeScale;
 uniform vec2 iconsTextureDim;
+// devicePixelRatio is set up as uniform but not declared in the shader function
 uniform float devicePixelRatio;
 
 varying float vColorMode;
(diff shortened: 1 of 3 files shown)


shaojingli:
Can you review this @Pessimistress ?

src/layers/core/icon-layer/icon-layer-vertex-64.glsl.js
```diff
@@ -36,6 +36,7 @@ attribute vec2 instanceOffsets;
 uniform vec2 viewportSize;
 uniform float sizeScale;
 uniform vec2 iconsTextureDim;
+uniform float devicePixelRatio;
```
1chandu: @rivulet-zhang , I don't see any code that sets up`devicePixelRatio` uniform, is this done inside any base class?
Also we recently added code so that apps provide `devicePixelRatio` as a prop to `DeckGL` component (#865), if this uniform setting depends on that code, we can't pick this PR for 4.1 patch release as #865 is not in current 4.1.1 patch. If yes, we have to remove dependency on #865.

src/layers/core/icon-layer/icon-layer-vertex.glsl.js
```diff
@@ -50,15 +51,16 @@ vec2 rotate_by_angle(vec2 vertex, float angle) {
...
   float instanceScale = iconSize.y == 0.0 ? 0.0 : instanceSizes / iconSize.y;
-  vec3 center = project_position(instancePositions);
-  vec2 vertex = (positions / 2.0 + instanceOffsets);
-  vertex = rotate_by_angle(vertex, instanceAngles) * iconSize_clipspace *
-    sizeScale * instanceScale;
+
+  // scale and rotate vertex in "pixel" value and convert back to fraction in clipspace
+  vec2 vertex = positions / 2.0 * iconSize + instanceOffsets;
+  vertex = rotate_by_angle(vertex, instanceAngles) / viewportSize * sizeScale *
+    instanceScale * devicePixelRatio;
```
Pessimistress: There are other layers using the `viewportSize` uniform to calculate geometry size where they are specified in pixels. In fact I think that's the only reason we have it. I would argue that we should adjust `viewportSize` by `devicePixelRatio` before passing it in, instead of fixing it in individual shaders.

ibgreen: > we should adjust viewportSize by devicePixelRatio before passing it in, instead of fixing it in individual shaders.

I have a [PR](https://github.com/uber/deck.gl/pull/902) open on the `project` module. It formalizes and will document the uniforms provided by the project module, which will now include `project_uViewportSize` and `project_uDevicePixelRatio`. The intention is that the latter will be set correctly depending on the application's choice of `useDevicePixelRatio`.

So with these changes, at least layers will not need to inject these uniforms themselves, which is error prone due to the `devicePixelRatio` logic.

I am actually adding a new method `project_pixel_to_clipspace` that will calculate the clipspace offset of a given number of pixels so that it can be added directly to `gl_Position` after `project_to_clipspace`.

Another reason why this function is better than multiplying these directly, is that when using first person viewport, there is no implicit subject distance, so we need to choose a focal plane where the pixel distance is given.

The reason for this it that since we don't consider the `w` coordinate when we add to clipspace, "pixel" values are scaled with distance. Which is normally a desirable effect. 

So by transitioning layers to this function, we will get much cleaner shader code, better control to play with the logic centrally, and solve both DPR and first person view issues.

ibgreen: Lightly edited previous post for clarity.

src/layers/core/icon-layer/icon-layer-vertex-64.glsl.js
1chandu: so this is setup as part of project module, should be good.

rivulet-zhang:
@1chandu @Pessimistress Changed the default texture filtering method to improve texture rendering quality as suggested by @ibgreen

ibgreen:
@rivulet-zhang Did it make a difference? Any before/after screen shots?

rivulet-zhang:
@ibgreen The quality did improve, especially when the icon is pretty small. See comparison below:

Before:
screen shot 2017-08-31 at 6 28 38 pm
After:
screen shot 2017-08-31 at 6 29 19 pm

ibgreen:
> The quality did improve, especially when the icon is pretty small

@rivulet-zhang that looks like a huge improvements to my eyes. 

@tangential

shaojingli:
This PR fixed issue #834 #895

PR #904, opened 2017-09-01
Rename coordinateSystem props

ibgreen:
This is a follow up to the recent extensive rewrite of the "Coordinate Systems" article. 

One small but important step towards making these thing simpler to explain and use, is to have the prop names match the type (i.e. `COORDINATE_SYSTEM`) and the concepts used in the docs.

* `projectionMode` => `coordinateSystem`
* `projectionOrigin` => `coordinateOrigin`

Now when the docs talk about coordinate systems it will be much easier for the user to make the right associations.

* Backwards compatibility is supported.
* What's New and Upgrade Guide are updated
* All layers and examples are updated
* All tests pass

### DIFF
--- a/dev-docs/RFCs/v5.0/first-person-mercator-viewport-rfc.md
+++ b/dev-docs/RFCs/v5.0/first-person-mercator-viewport-rfc.md
@@ -2,10 +2,13 @@
 
 * **Author**: Ib Green
 * **Date**: August 10, 2017
-* **Status**: **Approved**
+* **Status**: **Pre-Approved**
 
 Notes:
 * Please add comments as reviews to the [PR](https://github.com/uber/deck.gl/pull/838)
+
+References:
+* Initial [Infovis Viewport RFC]() for deck.gl v4.
 
 
 ## Motivation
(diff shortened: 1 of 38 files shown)


shaojingli:
> `projectionOrigin => coordinateOrigin`

I prefer to call it `coordinateReference` or `referencePoint`. `coordinateOrigin` sound always like (0,0,0) to me.

Pessimistress:
Personally, I think `projectionMode` is more clear. It doesn't make sense to rename a prop to match a doc.
If the goal is to make this system more generic, which I think will be extremely valuable, more important than the word `projection` or `coordinate` is to generalize `projectionOrigin` to `projectionParameters`. [An example geo projection spec](http://spatialreference.org/ref/esri/usa-contiguous-albers-equal-area-conic/html/)

ibgreen:
> Personally, I think projectionMode is more clear. It doesn't make sense to rename a prop to match a doc.

Whoops, had no idea that this would not be a slam-dunk. Let me know if I need to do a write-up on this before landing...

I personally don't see the connection to "projection". To me this prop is simply stating what coordinate system the positions are specified in, how they will be interpreted.  

My thinking was that why should the user be confused by the word "projection" when he just needs to chose coordinate system.  "Projection" is also something that is selectable in deck.gl and it is handled by `Viewports`.

ibgreen:
> generalize projectionOrigin to projectionParameters. An example geo projection spec.

Maybe I'll think differently when I see a real proposal, but right now I think that trying to support parametrized projections is going a bridge too far. No matter how complicated we make our support for parameters, we are not going to cover every projection. 

What I have been thinking is to offer the user to provide a simple JS function that maps any position (from any other coordinates) to one of our supported coordinate systems. This function would be called during `position` attribute generation. I.e we'd still show a mercator projected world but be able to correctly position coordinates specified in other projections.

ibgreen:
> I prefer to call it coordinateReference or referencePoint. coordinateOrigin sound always like (0,0,0) to me.

I wanted the two props to have the same first word so it is clear that they are related. Other than that we should pick the best option.

I personally tend to call these point "anchors" but I am not sure that is the best way to "teach" deck.gl. 

The way I am trying to describe meter offsets is as a local coordinate system for which you must specify where the origin is located, thus `coordinateOrigin`.

Pessimistress:
> What I have been thinking is to offer the user to provide a simple JS function that maps any position (from any other coordinates) to one of our supported coordinate systems.

The perf is not going to work for some of our internal use cases. As a fully pluggable projection mode / coordinate system, preferably I will provide:
- A js function that takes projection parameters returns a collection of uniforms
- vertext shader functions to replace `project` and `project_scale`

ibgreen:
> A js function that takes projection parameters returns a collection of uniforms
> vertext shader functions to replace project and project_scale

OK got it, that is a pretty neat idea.

For what it is worth, the shadertools module system was designed with a focus on interfaces, with an intention of allowing modules to be replaced with similar modules implementing the same interface. It would take a little more work on shadertools, but it is almost there. This was mostly intended for replacing e.g. `lighting` but would most likely work for projections as well.

Also shadertools already supports a generic `getUniforms` system that allows each module to opt in and extract what uniforms it needs from a common JS object, so modules that have the same GLSL interface could still listen for different JS props.

This should allow apps to plug in a replacement `project` module as you suggest, that could pick up any uniforms it wanted from the layers' props, e.g. your suggested `projectionParams`. 

As for the **current PR**, I'd prefer to name the core props as simply as possible (deck.gl is already intimidating for most users as it stands), and let such a system add what props it needs when we introduce it.

ibgreen:
OK looks like I have approval to land. 

These changes have been added to the what's new/upgrade guide, this make sure they'll be visible for us to consider again in the API audit before release, should we want to.

PR #907, opened 2017-09-05
Move deprecated code into `src/deprecated`. Use explicit imports.

ibgreen:


dev-docs/RFCs/README.md
```diff
@@ -2,16 +2,16 @@
 
 Implementation of non-trivial new deck.gl features should typically be started off with the creation of an RFC (Request for Comments) to make sure we have a complete story. It also allow the bigger team (as well as the community) to comment and contribute insights.
 
-| RFC Status   | Description |
```
ibgreen: Just some cleanup. Not really part of this PR, but doesn't deserve its own commit.

PR #915, opened 2017-09-06
Use luma.gl `AnimationLoop` instead of `WebGLRenderer` React class

ibgreen:
Rationale:
* Collects the react integration of deck.gl into a single file, simplifying porting of deck.gl to non-React environments.
* increases code sharing with luma.gl, for a smaller combined bundle.
* fixes and improvements to `AnimationLoop` now benefit both luma and deck.

Concerns
* Backwards compatibility edge cases. Some undocumented props supported by the WebGLRenderer (onBeforeRender etc) are no longer supported.


src/react/deckgl.js
```diff
@@ -42,6 +40,7 @@ const propTypes = {
   height: PropTypes.number.isRequired,
   layers: PropTypes.arrayOf(PropTypes.instanceOf(Layer)).isRequired,
   effects: PropTypes.arrayOf(PropTypes.instanceOf(Effect)),
+  glOptions: PropTypes.object,
```
1chandu: Nit: Should we add a default prop value for `glOptions: PropTypes.object`, {} ?

src/react/deckgl.js
```diff
@@ -76,17 +76,41 @@ export default class DeckGL extends React.Component {
...
+    const canvas = this.refs.overlay;
+
+    this.animationLoop = new AnimationLoop({
+      width,
+      height,
+      useDevicePixelRatio,
+      onCreateContext: (opts) =>
+        gl || createGLContext(Object.assign({}, glOptions, {canvas, debug})),
+      onInitialize: this._onRendererInitialized,
+      onRender: this._onRenderFrame
```
Pessimistress: Where is `this.props.onAfterRender` used?

src/react/deckgl.js
ibgreen: > Nit: Should we add a default prop value for glOptions: PropTypes.object, {} ?

Done

src/react/deckgl.js
ibgreen: > Where is this.props.onAfterRender used?

Appears to be used in the testing framework. Added a comment

PR #916, opened 2017-09-06
New controllers

ibgreen:
This PR continues the generalization of controllers.

The vision is that there should be only one react component: 
```
<ViewportController >
```
that can take a state class and do the right thing

This PR is a partial step, it leaves things in a "half finished" state:
- temporarily breaking `npm run start-local` in some non-geospatial examples.
- more work is needed on `FirstPersonState` to handle mouse events.
- `OrbitState` should be updated to work better with `ThirdPersonViewport`.
- All `Viewports` should be able to extract at least some information from each `State` class

### DIFF
--- a/src/react/controllers/viewport-controller.js
+++ b/src/react/controllers/viewport-controller.js
@@ -1,0 +1,132 @@
+import {PureComponent, createElement} from 'react';
+import PropTypes from 'prop-types';
+
+import EventManager from '../../controllers/events/event-manager';
+import Controls from '../../controllers/controls';
+import CURSOR from './cursors';
+
+const propTypes = {
+  StateClass: PropTypes.func,
+  state: PropTypes.object,
+
+  /** Viewport constraints */
+  // Max zoom level
+  maxZoom: PropTypes.number,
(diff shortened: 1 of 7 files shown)


src/react/controllers/viewport-controller.js
```diff
@@ -0,0 +1,132 @@
...
+      width,
+      height,
+      position: 'relative',
+      cursor: getCursor(this.state)
+    };
+
+    return (
+      createElement('div', {
+        key: 'map-controls',
+        ref: 'eventCanvas',
```
Pessimistress: String refs are being removed in React 16. See [this thread on react-map-gl](https://github.com/uber/react-map-gl/pull/316).

ibgreen: > String refs are being removed in React 16. See this thread on react-map-gl.

OK. We are using them in more places, maybe just open a separate PR to remove them all?

ibgreen: Opened https://github.com/uber/deck.gl/issues/917

Issue #456, opened 2017-03-23
Reduce the props transfer clutter


shaojingli:
With the explicitly props transfer requirement, all props of the underlying layer of a composite layer needs to be set explicitly in renderLayers(). This created cluttered code and it's error-prone, as shown in #455. Some healthy refactor of the composite layer code is needed to mitigate this issue

PR #921, opened 2017-09-07
Remove the Property Forwarding Clutter

ibgreen:


ibgreen:
Fixes #456

ibgreen:
Includes a fix to draw-and-pick on multi-model layers and a light reorganization of seer-integration.

PR #922, opened 2017-09-07
Completing unification of controllers

ibgreen:
* Only a single `Controller` class now remains, capable of accepting any `...State`. 
* A generic `ViewportController` react class wraps it.
* `MapController` remains for backwards compatibility for now, but is essentially identical to `ViewportController`


src/controllers/controls.js
```diff
@@ -18,11 +18,15 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
```
Pessimistress: Does this mean we no longer provide a base class for non-mapbox controls?

ibgreen: The only difference between the two was in panToRotate. Both implementations are present here, in this incarnation, it autodetects the state type.

Pessimistress: What about orbit controller?

ibgreen: > What about orbit controller?

It should be in there. Controls is essentially a unification of `OrbitControls` and `MapControls` that also works as `FirstPersonControls`. It is still possible to subclass controller and add one to the `ViewportController` but not necessary in typical cases.

Let me go through all the examples using `OrbitController` and make sure they still work, will update.

PR #923, opened 2017-09-08
Multiviewport picking (Phase 1, non-performant)

ibgreen:
* Moves multiviewport handling into `layer-manager.js` and `draw-and-pick`
* Fixes multiviewport picking.

Next step:
* Ensure only a single framebuffer readback is made, instead of one per viewport.

Note this PR includes the changes in #922, will rebase once that lands.


ibgreen:
Landed as part of a different PR.

PR #924, opened 2017-09-08
Use EventManager from mjolnir.js

Pessimistress:


package.json
```diff
@@ -44,7 +44,7 @@
   "dependencies": {
     "d3-hexbin": "^0.2.1",
     "earcut": "^2.0.6",
-    "hammerjs": "^2.0.8",
+    "mjolnir.js": ">=0.1.0-beta.1",
```
ibgreen: Nit: Didn't catch this until know, but if we are using `0.x` releases we can probably dispense with the `beta`. It is kind of implicit.

PR #927, opened 2017-09-11
Viewport doc overhaul

ibgreen:


ibgreen:
I am halfway into a major overhaul of viewport docs and trying to write something about controllers as well. Plus reformatting and fixing some other things along the way. 

It's a lot of changes, if you guys want to review these things in more detail I could break into separate PRs. Otherwise I suggest we do a full doc review before we move to beta.

docs/advanced/viewports.md
```diff
@@ -30,11 +33,148 @@ Like many things in deck.gl, Viewports can be positioned using a lng/lat "anchor
...
+It is common in 3D applications to render a 3D scene multiple times, with different cameras:
+* To show views from multiple viewpoints (cameras), e.g. in a split screen setup.
+* To show a detail view (e.g, first person), and an overlaid, smaller "map" view (e.g. third person or top down, zoomed out to show where the primary viewpoint is).
+* To support stereoscopic rendering, where left and right views are needed, providing the necessary parallax between left and right eye.
+* For rendering into offscreen framebuffers, which can then be used for e.g. advanced visual effects, screen shot solutions, overlays onto DOM elements outside of the primary deck.gl canvas (e.g. a video).
+
+
+### The `DeckGL.viewports` property
+
+Allows the main `DeckGL` component to accept a list of Viewports (and/or viewport descriptors). This will be in addition to, or preferably instead of, the current `viewport` prop.
```
1chandu: > the current `viewport` prop.
There is no `viewport` prop, all the viewport parameters (pitch, bearing, etc) are separate props. Or you will be adding `viewport` prop too?

ibgreen: > There is no viewport prop, all the viewport parameters (pitch, bearing, etc) are separate props. Or you will be adding viewport prop too?

Currently we support both direct params and `viewport`, see https://github.com/uber/deck.gl/blob/viewport-docs/src/react/deckgl.js#L48. The viewport was added a while ago to support non-geospatial use cases.

PR #931, opened 2017-09-13
Fix picking buffer clear color.

1chandu:
- Clear color should be set before drawing any layers for picking. When none of the layers are eligible for picking, we still get correct picking results (alpha channel will be 0).
- Return early from picking flow, when layer count is 0.
- Fixes an issue with `point-cloud-laz` example.
- Tested with `test-browser` and `layer-browser`.

### DIFF
--- a/src/lib/draw-and-pick.js
+++ b/src/lib/draw-and-pick.js
@@ -275,8 +275,14 @@ function getClosestFromPickingBuffer(gl, {
     pickedObjectIndex: -1
   };
 
-  if (deviceX < 0 || deviceY < 0 || deviceX > pickingFBO.width || deviceY > pickingFBO.height) {
-    // x, y out of bounds
+  if (
+    deviceX < 0 ||
+    deviceY < 0 ||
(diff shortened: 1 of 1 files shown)


src/lib/draw-and-pick.js
```diff
@@ -275,8 +275,14 @@ function getClosestFromPickingBuffer(gl, {
...
   };
 
-  if (deviceX < 0 || deviceY < 0 || deviceX > pickingFBO.width || deviceY > pickingFBO.height) {
-    // x, y out of bounds
+  if (
+    deviceX < 0 ||
+    deviceY < 0 ||
+    deviceX > pickingFBO.width ||
+    deviceY > pickingFBO.height ||
+    layers.length <= 0
```
gnavvy: layers.length <= 0 is more of a sanity check.

the getClosestFromPickingBuffer function will be called via layerManager.pickLayer(), thus will be called even if there no layer being added to deck.gl (due to data loading in progress or error).

PR #942, opened 2017-09-22
Make React DeckGL component a thin wrapper over "pure JS" counterpart

ibgreen:
Moving the "pure" JavaScript Deck component into the src folder. Change the React DeckGL component to a thin wrapper.

This is a big step towards making the React components completely optional. 

Obviously more work is needed to do this for all React components and polish the API.


src/index.js
```diff
@@ -60,6 +60,10 @@ export {default as default} from './react/deckgl';
 export {default as ViewportController} from './react/controllers/viewport-controller';
 export {default as MapController} from './react/controllers/map-controller';
 
+// Experimental Pure JS (non-React) bindings
+import {default as DeckGLJS} from './pure-js/deck';
```
gnavvy: What's rationale before changing these names?
I guess exporting as DeckGLJS is to distinguish from the original DeckGL from ./react/deckgl.js.
but why changing DeckGL to Deck in the new ./pure-js/deck.js?

src/pure-js/deck.js
```diff
@@ -76,31 +80,27 @@ export default class DeckGL {
     this.needsRedraw = true;
     this.layerManager = null;
     this.effectManager = null;
-    autobind(this);
 
-    const {width, height, gl, glOptions, debug} = props;
+    // Bind methods
+    this._onRendererInitialized = this._onRendererInitialized.bind(this);
+    this._onRenderFrame = this._onRenderFrame.bind(this);
```
gnavvy: Since we are removing autobind, shall we move towards using the array function style instead of these explicit bindings? will need transform plugin in the .babelrc though.

src/index.js
ibgreen: > What's rationale before changing these names?
> I guess exporting as DeckGLJS is to distinguish from the original DeckGL from ./react/deckgl.js.

Yes. I don't like having two things called exactly the same name (goes for both global symbols and filenames). It tends to confuse both humans and computers.

> but why changing DeckGL to Deck in the new ./pure-js/deck.js?

Good question. I could call them `DeckGLJS` and `deckgljs.js`, I just didn't think those names are good, and would be changed asap. 

We need to come up with a naming scheme to separate the PureJS components and the React components in a logical way.

src/pure-js/deck.js
ibgreen: > will need transform plugin in the .babelrc though.

Currently deck.gl uses pure ES6 and runs without transpilation in Node 6+ and modern browsers. If we start adding ES7+ features this will break.

PR #944, opened 2017-09-22
Viewport animation RFC updates

1chandu:
CC @gnavvy @jianhuang01


dev-docs/RFCs/v5.0/viewport-animation-rfc.md
```diff
@@ -60,11 +60,26 @@ viewportAnimationDuration: (Number) Animation duration in milliseconds, default
...
+
+4. **animationFreeze** {bool, optional} : When set to `true` and if `animationDuration` not equal to 0, any following viewport updates are ignored until animation resulting this update is finished. This prop has no effect if `animationDuration` is 0. When it is set to `false` or not set, any following viewport updates will stop the current animation and update viewport according to other props.
+
+5. **animationCompleteOnUpdate** {bool, false} : This prop can be used to control how to end current animation, when new animation requested before it is completed, following table describes in more details.
+
+| ----------------------------------- Current Viewport ----------------------------------- | --- Viewport Update --- | Result |
```
1chandu: This is an attempt to group multiple columns into one header, is there a better option?

1chandu: Please tap `View` on top right for better formatted view.

dev-docs/RFCs/v5.0/viewport-animation-rfc.md
```diff
@@ -60,11 +60,23 @@ viewportAnimationDuration: (Number) Animation duration in milliseconds, default
...
+
+4. **animationFreeze** {bool, optional} : When set to `true` and if `animationDuration` not equal to 0, any following viewport updates are ignored until animation resulting this update is finished. This prop has no effect if `animationDuration` is 0. When it is set to `false` or not set, any following viewport updates will stop the current animation and update viewport according to other props.
+
+5. **animationCompleteOnUpdate** {bool, false} : This prop can be used to control how to end current animation, when new animation requested before it is completed, following table describes in more details.
```
jianhuang01: animationCompleteOnUpdate is kind of confusing, understand that it is for user to define how we connect two animations or end animation, but hard to remember all the settings, recommend we go the following approach.
1. In the first case ideally two animations should be always smoothly blended into a single animation, adding a new API to introduce frame dropping is not good. Animation blending is not easy, don't need implement for now.
2. In the second case we can define an API animationCompleteOnCancel(bool), which define how we end the animation when animation cancelling happens, jump to destination when bool is true.
3. If user wants to introduce "frame dropping" to connect two animations, they can always call animationCompleteOnCancel(true) and start a new animation.

dev-docs/RFCs/v5.0/viewport-animation-rfc.md
```diff
@@ -60,11 +60,23 @@ viewportAnimationDuration: (Number) Animation duration in milliseconds, default
 
 1. **animationDuration** {Number, default: 0} : Animation duration in milliseconds, default value disables animation.
 
-2. **easingFunction** {Function, default: t => t} : Easing function that can be used to achieve effects like "Ease-In-Cubic", "Ease-Out-Cubic", etc. Default value performs Linear easing. (list of sample easing functions: http://easings.net/)
+2. **animaitonEasingFunction** {Function, default: t => t} : Easing function that can be used to achieve effects like "Ease-In-Cubic", "Ease-Out-Cubic", etc. Default value performs Linear easing. (list of sample easing functions: http://easings.net/)
```
ibgreen: Prop names are getting long, maybe just `animationEasing`?

Typo in `animaitonEasing`

dev-docs/RFCs/v5.0/viewport-animation-rfc.md
```diff
@@ -60,11 +60,23 @@ viewportAnimationDuration: (Number) Animation duration in milliseconds, default
...
+3. **animationFunction** {Function, default: `viewportLinearAnimation`} : Function that gets called for each animation step to calculated animated viewport. It takes start, end viewports and animation step, returns animated viewport. We provide couple of utility functions, `viewportLinearAnimation` and `viewportFlyToAnimation`. By default `viewportLinearAnimation` is used where all viewport props are linearly animated. `viewportFlyToAnimation` animates viewports similar to MapBox `flyTo` API, this is pretty useful when camera center changes. But a user can provide any function for this prop to perform custom viewport animations.
```
ibgreen: A more specific name `animationInterpolator`?

dev-docs/RFCs/v5.0/viewport-animation-rfc.md
```diff
@@ -60,11 +60,23 @@ viewportAnimationDuration: (Number) Animation duration in milliseconds, default
...
+
+4. **animationFreeze** {bool, optional} : When set to `true` and if `animationDuration` not equal to 0, any following viewport updates are ignored until animation resulting this update is finished. This prop has no effect if `animationDuration` is 0. When it is set to `false` or not set, any following viewport updates will stop the current animation and update viewport according to other props.
```
ibgreen: We discussed supporting three modes:
`animationEvents`:
* ANIMATION_EVENTS.IGNORE, 
* ANIMATION_EVENTS.SNAP_TO_END
* ANIMATION_EVENTS.BREAK

dev-docs/RFCs/v5.0/viewport-animation-rfc.md
ibgreen: > **animationCompleteOnUpdate**

That is a long name. Could we call this something like `onAnimationComplete` ?

dev-docs/RFCs/v5.0/viewport-animation-rfc.md
```diff
@@ -60,11 +60,23 @@ viewportAnimationDuration: (Number) Animation duration in milliseconds, default
...
+
+5. **animationCompleteOnUpdate** {bool, false} : This prop can be used to control how to end current animation, when new animation requested before it is completed, following table describes in more details.
+
+| Current viewport animating | Current Viewport start value| Current Viewport end value| Current Viewport value | Current Viewport AnimationCompleteOnUpdate | Viewport update animation requested | Viewport update value | Result |
+| ----- | --- | --------- | ------------- | ------------------------- | ------------------- | ----- | ------ |
+| YES   | VX1 | VX2       |  VXt          |   **TRUE**                    | YES                 |  VY   | New animation starts between **VX2** and VY |
```
ibgreen: This seems to overlap the event scenarios, maybe a single table is possible?

dev-docs/RFCs/v5.0/viewport-animation-rfc.md
```diff
@@ -60,11 +60,23 @@ viewportAnimationDuration: (Number) Animation duration in milliseconds, default
...
+| YES   | VX1 | VX2       |  VXt          |   **NO**                    | YES                 |  VY   | New animation starts between **VXt** and VY |
+| NO   | VX | VX       |  VX2          |   **TRUE/FALSE**               | YES                 |  VY   | New animation starts between **VX** and VY |
+| YES   | VX1 | VX2       |  VXt          |   **TRUE/FALSE**               | NO                 |  VY   | Stop current animation and *jump* from VXt to VY |
+
+6. **onAnimationStop** {Function, optional} : This callback will be fired when requested animation is stopped. The object ({animationStep}) will be passed as an argument, which can be used to whether animation was interrupted or completed. This prop can be used to generate continuous animations, that loop animating between set of viewports.
```
ibgreen: Should we add `onAnimationStart`, `onAnimationUdpate` for completeness?

dev-docs/RFCs/v5.0/viewport-animation-rfc.md
1chandu: `animationFreeze` and `animationCompleteOnUpdate` provide all three scenarios. But providing above 3 constants and hiding controls behind these seems much simpler and less confusing. It also addresses other concerns on names.

dev-docs/RFCs/v5.0/viewport-animation-rfc.md
1chandu: `onAnimationStop` has a very good use case to kick of new animation right after current one to have continuous loop like animations, but I can't think of any for `onAnimationStart` and `onAnimationUpdate`. May be we can add them in next update and focus on testing current features in an application?

dev-docs/RFCs/v5.0/viewport-animation-rfc.md
1chandu: - AFAIK `Animation blending` is lot of effort with little gain, if there are any use cases we can always add it later.
- Yeah, I agree, these prop names are confusing, I remove `4` and `5` and just define 3 constants as Ib suggested below.

dev-docs/RFCs/v5.0/viewport-animation-rfc.md
```diff
@@ -27,13 +27,13 @@ When rendering deck.gl layers on top of Mapbox map.
 ```js
       <AnimationMapController
```
Pessimistress: This render structure should not be required. `AnimationMapController` should replace `MapController` in both deck.gl and react-map-gl, and the old structure:
```
<MapGL {...viewport}>
    <DeckGL {...viewport} />
</MapGL>
```
will suffice unless deck.gl is used without map.

ibgreen: While I am supportive of the idea to merge `Controller` into `DeckGL`, for the animation to be able to affect the map, the order will need to be reversed:
```
<DeckGL ...>
   <MapGL ...>
</DeckGL>
```
Unless we duplicate the entire animation functionality into react-map-gl of course, which is still being discussed.

This is due to animation being done by modifying child props, which is required to avoid initial "flashes" in animation.

Pessimistress: It is one of the top priorities for react-map-gl to support viewport animation, so that is definitely happening.

Pessimistress:
Similar to the attribute animation RFC: I think it's worth discussing whether this feature should be called "animation" or "transition" (CSS analogy). Animations are actively invoked by the application and are expected to support keyframes, cancellation etc. Transitions are passively invoked by setting other props.

PR #957, opened 2017-09-27
SegmentLayer - a Path layer that uses a shadow buffer to create outlines between paths at different heights

ibgreen:
screen shot 2017-09-26 at 7 21 39 pm

### DIFF
--- a/dev-docs/RFCs/vNext/composite-layer-customization-rfc.md
+++ b/dev-docs/RFCs/vNext/composite-layer-customization-rfc.md
@@ -70,7 +70,7 @@ const GeoJsonLayer extends CompositeLayer {
   ...
   renderLayers() {
   	return [
-  	  new this.props.subLayers.PointLayer({
+  	  new this.props.subLayerProps.PointLayer({
   	  }),
   	];
   }
--- a/dev-docs/RFCs/vNext/layer-shader-customization.md
+++ b/dev-docs/RFCs/vNext/layer-shader-customization.md
@@ -1,0 +1,12 @@
+# RFC: Layer Shader Customization
+
+* **Authors**: Ib Green
+* **Date**: Sep 2017
+* **Status**: Draft
+
+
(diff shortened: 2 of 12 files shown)


ibgreen:
@1chandu - `segment-layer.js` in this PR is where the framebuffer fixes you are working on will be used.

gnavvy:
Not in this PR but would it be feasible to merge this layer back to the path layer, as from the visuals the segment layer looks a path layer with enhanced visual effects? Or are targeting at making such effects applicable to all existing primitive layers like the filter effects in SVG?

shaojingli:
This is some great work. One thing that we could consider here is that if we'd like to rewrite the path layer in the future (both 32-bit and 64-bit). How easy is it to port it to the new path layer?

ibgreen:
> This is some great work. 

It is a port of @george-uber's THREE.js implementation, he should have the credit.

> One thing that we could consider here is that if we'd like to rewrite the path layer in the future (both 32-bit and 64-bit). How easy is it to port it to the new path layer?

> Not in this PR but would it be feasible to merge this layer back to the path layer, as from the visuals the segment layer looks a path layer with enhanced visual effects? Or are targeting at making such effects applicable to all existing primitive layers like the filter effects in SVG?

Long term I would very much like to make this into a generic feature for all layers. For now you can see in the PR that the outline shadowmap piece is already organized as a reusable shader module, so that is a start.

But there are a number of issues remaining:
- we are still flushing out some Framebuffer handling bugs in luma.gl, 
- there are problems with current implementation in perspective mode (only works with pitch=0), that is why I am keeping this as a sample layer for now.

Note that shader module injection system improvements could making adding this to other layers much easier. Have started to put down my ideas, I am afraid it is [Yet Another Half-Finished RFC](https://github.com/uber/deck.gl/blob/master/dev-docs/RFCs/vNext/shader-module-injection-rfc.md).

PR #959, opened 2017-09-27
Update docs with missing layer props.

### LINKED ISSUE #906: fp64 prop is not explicitly mentioned in the Scatterplot layer doc
Howdy - thanks for the awesome library!  I am using the scatterplot and whenever I zoom the dot 'jiggles' around.  I tried to fiddle with the various settings on the map & the layer but to no avail.  I am guessing its re-computing the layout every time the viewport changes and maybe there's some floating point math that's not precise enough or something?

dot

In the above animated gif the *only* thing I am doing (which is unfortunately non-obvious in the picture) is mouse-wheel zooming out slowly. Every time the mouse wheel 'ticks' and changes the zoom level the dot jiggles and changes shape. 

This behavior is very obvious when you're zoomed close in, but is also present at farther out zoom levels.

It's also reproducible on [this example](http://uber.github.io/deck.gl/#/documentation/layer-catalog/scatterplot-layer) from the docs if you choose one of those circles and zoom way in.  Eventually it'll start jiggling around as well.

dot

Any recommendation for a way to fix this?

1chandu:
Fixes #906 and some other missing props.

### DIFF
--- a/docs/layers/grid-cell-layer.md
+++ b/docs/layers/grid-cell-layer.md
@@ -47,6 +47,13 @@ Inherits from all [Base Layer](/docs/api-reference/base-layer.md) properties.
 
 Size of each grid cell in meters
 
+##### `coverage` (Number, optional)
+
+- Default: `1`
+
+Cell size scale factor. The size of cell is calculated by
+`cellSize * coverage`.
+
 ##### `elevationScale` (Number, optional)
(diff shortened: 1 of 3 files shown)


ibgreen:
@heshan0131 please review the coverage prop text.

PR #961, opened 2017-09-28
Attribute Transition RFC

Pessimistress:


dev-docs/RFCs/v5.0/attribute-animation-rfc.md
```diff
@@ -9,41 +9,128 @@ Notes:
...
+});
 ```
-while
+
+When creating layers, animation can be enabled by supplying an `transition` prop. Animation parameters are defined per attribute by using attribute names or accessor names as keys, similar to that of `updateTriggers`:
 ```js
-   new Layer({getColor: [255, 0, 0]}) // would set a generic vertex attribute (one value shared by all verts, not allocating any array/buffers at all)
+new Layer({
+  transition: {
+    getPositions: {
```
ibgreen: Should we also support just a number for folks who don't fiddle with easings? Makes the API prettier.
```
new Layer({
  transition: {
    getPositions: 600,
    getColors: {
      duration: 300
    }
  }
}
```

dev-docs/RFCs/v5.0/attribute-animation-rfc.md
```diff
@@ -9,41 +9,128 @@ Notes:
...
+      duration: 600
+    },
+    getColors: {
+      duration: 300,
+      easing: d3.easeCubicInOut
```
ibgreen: The easing just modifying the "time" parameter so it doesn't need to be visible in GLSL.

dev-docs/RFCs/v5.0/attribute-animation-rfc.md
```diff
@@ -9,41 +9,128 @@ Notes:
...
+| --------- | -------- | ----------- |
+| duration  | Number   | Duration of the transition animation, in milliseconds |
+| easing    | Function | Easing function that maps a value from [0, 1] to [0, 1], see http://easings.net/ |
+| onComplete | Function   | Callback when the transition is done |
+
+
+### Implementation
+
+- Layers will get the feature (almost) free:
+    + The animation states will be tracked in the attribute manager.
```
ibgreen: This will increase "memory pressure" up to 2x?

dev-docs/RFCs/v5.0/attribute-animation-rfc.md
```diff
@@ -9,41 +9,128 @@ Notes:
...
+    + The output is passed to the vertex shader for the target attribute
+- In non-WebGL2 enabled browsers, we either drop the feature, or fall back to interpolation on the CPU.
+
+### TransformFeedback vs VertexShader
+
+VertexShader pros:
+- Available in all supported browsers
+
+VertexShader cons:
+- Must rewrite existing layers’ vertex shaders to double each animatable attribute and perform interpolation before other calculations. The cost of a heavier shader and double attribute upload applies even if transition is disabled on a layer.
```
ibgreen: >  double attribute upload applies even if transition is disabled 

Not necessarily. One can set an attribute to a single value (constant) that applies to all vertices (which you mention in the ogther file in this PR) We could use that technique for attributes that are not used.

dev-docs/RFCs/v5.0/attribute-animation-rfc.md
```diff
@@ -9,41 +9,128 @@ Notes:
...
+**B** The data array is changed with increased size
+
+Exit (disappearance of rendered geometries) may happen when:
+**C** The layer is removed or becomes invisible
+**D** The data array is changed with decreased size
+
+Without adding additional complexity, the most straightforward behavior is no enter/exit animation - geometries appear and disappear immediately. The application can create their own enter/exit animations by adding invisible objects (e.g. instead of removing objects from the data array, change their colors to transparent).
+
+To support enter/exit animations, we may consider one of the following:
+- Add a field `voidValue` to attribute definitions. This could be a value or an transform to be used for enter animation to transition from, or exit animation to transition to.
```
ibgreen: This is cool!

dev-docs/RFCs/v5.0/attribute-animation-rfc.md
```diff
@@ -9,41 +9,128 @@ Notes:
...
+    getColors: {
+      duration: 300,
+      enter: feature => feature.properties.fill.concat(0),
+      exit: feature => feature.properties.fill.concat(0)
+    }
+  }
+});
 ```
 
+To properly support scenario **B** and **D**, we need a way to diff data arrays to determine which objects are added/removed.
```
ibgreen: Fine for now, although I'd like to discuss these scenarios a bit more further on.  There are complications when updating non-instanced layers like a path layer, a single path can change length messing up the interpolation of all the following verts if special care is not taken. These problems relate to much of the thinking behind my RFC on partial updates.

dev-docs/RFCs/v5.0/attribute-animation-rfc.md
```diff
@@ -9,41 +9,128 @@ Notes:
...
+### APIs
 
-## Proposal: Layer accessors can be constant values
-
-> NOTE: Copied from feature scratch-pad, but related to attribute animation, would allow simple interpolation.
-
-In cases where there is a direct correspondence between an accessor and a vertex attributes  would allow for a nice optimization where a vertex attribute could simply be set to a single value (using a so called “generic vertex attribute” from luma.gl)
-
-Allowing an accessor to be set to a value instead of a function could be a nice way to enable generic vertex attributes. I.e.
+In deck.gl layer classes, certain attributes can declare themselves to be "animatable", which allows for smooth interpolation when their values are updated. This will be desirable for attributes such as sizes, widths, positions and colors. Animation will be an opt-in feature for backward compatibility:
```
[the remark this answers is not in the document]
Pessimistress: I guess I'm thinking we'll reuse this flag in case we support other animation types later?

dev-docs/RFCs/v5.0/attribute-animation-rfc.md
Pessimistress: Admittedly it would be nice to have these, though we are circumventing these scenarios by calling the feature "transition" (i.e. instant enter and exit)

PR #964, opened 2017-09-29
Big draw and pick refactor

ibgreen:
* Split `draw-and-pick` into `draw-layers` and `pick-layers`
* Keep push drawing and picking code into these files from `deck` and `layer-manager`.

PR looks bigger than it is due to file split, not much I can do about that.

### DIFF
--- a/src/core/lib/pick-layers.js
+++ b/src/core/lib/pick-layers.js
@@ -18,167 +18,21 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-/* global window */
-import {GL, withParameters, setParameters} from 'luma.gl';
 import {log} from './utils';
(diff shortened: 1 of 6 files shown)


src/core/lib/pick-layers.js
```diff
@@ -18,192 +18,40 @@
...
-    });
-  });
-
-  return Array.from(uniqueInfos.values());
-}
 
 /* eslint-disable max-depth, max-statements */
 // Pick the closest object at the given (x,y) coordinate
-export function pickLayers(gl, {
+export function pickObject(gl, {
```
1chandu: This is a name change to a public method, is the documentation change planned for next PR?

ibgreen: > This is a name change to a public method, is the documentation change planned for next PR?

It is exported from this file for use in other files, but it is not documented, and as far as I know it is not exposed from index.js (which is in a sense the "source of truth" for what is considered public).

Issue #962, opened 2017-09-28
[FP64] Layer-browser: GridLayer and GridCellLayer position changes on fp64 prop toggle.


1chandu:
Seen in `master` and and latest prod `4.1.2`.

1chandu:
Resolved by #968

PR #968, opened 2017-09-29
 [Hexagon/Grid] add no render when elevation < 0.0 to vertex-64

heshan0131:
- Add checks in grid fp64 to hide cell when filtered by elevation (same operation as the none fp64, this is a patch to #954 
- Fix grid cell layer position shifts on fp64 #962


1chandu:
LGTM, @ibgreen @shaojingli , this change will also go into 4.1.3 patch release.

PR #969, opened 2017-09-29
Support initWebGLParameters prop to ensure examples on master work

ibgreen:


src/react/deckgl.js
```diff
@@ -53,6 +54,7 @@ const defaultProps = {
   gl: null,
   effects: [],
   onWebGLInitialized: noop,
+  initWebGLParameters: true,
```
1chandu: @ibgreen is it supposed to be `true` by default?

PR #971, opened 2017-09-29
Fix bug where layer does not update on updateTriggers change

Pessimistress:
https://github.com/uber/deck.gl/issues/950

Include `diffUpdateTriggers` result in `changeFlags.propsOrDataChanged`.

### DIFF
--- a/src/core/lib/layer.js
+++ b/src/core/lib/layer.js
@@ -495,15 +495,17 @@ export default class Layer {
 
     const propsChanged = Boolean(propsChangedReason);
     const dataChanged = Boolean(dataChangedReason);
-    const propsOrDataChanged = propsChanged || dataChanged;
     const viewportChanged = context.viewportChanged;
-    const somethingChanged = propsChanged || dataChanged || viewportChanged;
 
+    let updateTriggersChanged = false;
     // Check update triggers to determine if any attributes need regeneration
     // Note - if data has changed, all attributes will need regeneration, so skip this step
     if (!dataChanged) {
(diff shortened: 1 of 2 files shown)


src/core/lib/layer.js
```diff
@@ -495,16 +495,18 @@ export default class Layer {
...
 
+    let updateTriggersChanged = false;
     // Check update triggers to determine if any attributes need regeneration
     // Note - if data has changed, all attributes will need regeneration, so skip this step
     if (!dataChanged) {
-      this._diffUpdateTriggers(oldProps, newProps);
+      updateTriggersChanged = this._diffUpdateTriggers(oldProps, newProps);
     }
 
+    const propsOrDataChanged = propsChanged || dataChanged || updateTriggersChanged;
```
ibgreen: Q: Do we need to set `propsOrDataChanged`? If we set `propsOrDataChanged` to `true`,  I believe that we will also cause layers to update props. Not a terrible thing, but still unnecessary work in cases where it was just an `updateTrigger` that fired?

Would it make sense to just use the new flag `changeFlags.updateTriggersChanged` and trigger attribute invalidation based on that?

Pessimistress: `diffUpdateTrigger` already invalidates the attribute, the problem is that it doesn't recalculate them.
Invalidated attributes are updated in `updateState`, which is blocked by `propsOrDataChanged`.

ibgreen:
@georgios-uber @javidhsueh

PR #977, opened 2017-10-04
Move uniform updates from draw() to updateState()

ibgreen:
Main motivations:
* Based on measurements by @twojtasz `setUniforms` have started to show up as a noticeable contributor to time spent in some apps. Moving updates from `draw` to `updateState` should help reduce the number of calls to `setUniforms`.
* More importantly, in some scenarios it is important to be able to override as many rendering parameters as possible in the layer's draw method (e.g. when creating layer subclasses or doing picking).

### DIFF
--- a/src/core-layers/arc-layer/arc-layer.js
+++ b/src/core-layers/arc-layer/arc-layer.js
@@ -85,20 +85,25 @@ export default class ArcLayer extends Layer {
 
   updateState({props, oldProps, changeFlags}) {
     super.updateState({props, oldProps, changeFlags});
+
     // Re-generate model if geometry changed
     if (props.fp64 !== oldProps.fp64) {
       const {gl} = this.context;
       this.setState({model: this._getModel(gl)});
     }
     this.updateAttribute({props, oldProps, changeFlags});
+
+    // Update props if needed
+    if (changeFlags.propsChanged) {
+      const {strokeWidth} = this.props;
+      this.state.model.setUniforms({
+        strokeWidth
+      });
+    }
   }
 
-  draw({uniforms}) {
-    const {strokeWidth} = this.props;
-
-    this.state.model.render(Object.assign({}, uniforms, {
-      strokeWidth
-    }));
+  draw(opts) {
+    this.state.model.draw(opts);
   }
 
   _getModel(gl) {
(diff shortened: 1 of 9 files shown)


src/core-layers/arc-layer/arc-layer.js
```diff
@@ -85,20 +85,25 @@ export default class ArcLayer extends Layer {
...
       const {gl} = this.context;
       this.setState({model: this._getModel(gl)});
     }
     this.updateAttribute({props, oldProps, changeFlags});
-  }
 
-  draw({uniforms}) {
-    const {strokeWidth} = this.props;
+    // Update props if needed
+    if (changeFlags.propsChanged) {
```
Pessimistress: Why not compare `props.strokeWidth` and `oldProps.strokeWidth`? `propsChanged` will always be true with inline functions in most of our use cases.

src/core-layers/icon-layer/icon-layer.js
```diff
@@ -128,49 +117,40 @@ export default class IconLayer extends Layer {
 
     if (oldProps.iconAtlas !== iconAtlas) {
 
-      if (iconAtlas instanceof Texture2D) {
-        iconAtlas.setParameters({
+      const setIconsTexture = texture => {
```
Pessimistress: Define outside?

src/core-layers/arc-layer/arc-layer.js
ibgreen: > I think luma.gl's model.setUniforms should do the shallow comparison and only update the changed uniform. In fact, the same problem was identified before implementing the state management in luma and I thought it was already there.

The priority (and the hard problem) was gl context state caching, that is what was done. `Program` state caching is a second opportunity, and I agree that the time is right to look at this now @1chandu 

> Why not compare props.strokeWidth and oldProps.strokeWidth? propsChanged will always be true with inline functions in most of our use cases.

- Because I agree with you that updating luma.gl to do caching is the right fix for that, otherwise the clutter here will become intolerable. Layer code should be minimal and clean. 
- And for inline functions we should get the prop types system in place so we can filter them out. Since we don't recalculate props when they change there is no reason to set the `propsChanged` flag.

I currently still think the uniforms should move out of the render function. But maybe we should do a proof of concept for property transition/animation before landing this change?

1chandu:
Closing this now, we will resume this work and is being tracked by : #1048

Issue #979, opened 2017-10-05
Picking and highlighting broken on master


1chandu:
- When mouse point moved away from a selectable object to no object area, 'onHover' message on `layer-browser` isn't cleared out.
- Auto highlighting is not working.

ibgreen:
I just noticed an issue on line https://github.com/uber/deck.gl/blob/master/src/core/lib/draw-layers.js#L200

Reusing parameters here brings in previous layers parameters. We must use a fresh variable each iteration. This was a bug I introduced in my split of draw and pick into two files.

Not sure if your bug is related.

1chandu:
Turns out picking and highlighting are two different problems introduced with picking refactor, preparing a PR.

1chandu:
Resolved in #982

PR #980, opened 2017-10-05
Skip picking flow when no layer is pickable

1chandu:
Verified with `layer-browser`.
Found a regression on master for picking/highlighting feature (#979), will re-test this after it is resolved.

### DIFF
--- a/src/core/lib/layer.js
+++ b/src/core/lib/layer.js
@@ -177,6 +177,11 @@ export default class Layer {
     }
   }
 
+  // Returns true if the layer is pickable and visible.
+  isPickable() {
+    return this.props.pickable && this.props.visible;
+  }
+
   // Calls attribute manager to update any WebGL attributes, can be redefined
   updateAttributes(props) {
     const {attributeManager, model} = this.state;
--- a/src/core/lib/pick-layers.js
+++ b/src/core/lib/pick-layers.js
@@ -20,6 +20,7 @@
 
 import {log} from './utils';
 import {drawPickingBuffer, getPixelRatio} from './draw-layers';
+import assert from 'assert';
 
 const EMPTY_PIXEL = new Uint8Array(4);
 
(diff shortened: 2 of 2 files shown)


src/core/lib/layer.js
```diff
@@ -116,6 +116,10 @@ export default class Layer {
     return this.state.stats;
   }
 
+  get isPickable() {
+    return this.props.pickable;
```
Pessimistress: visible && pickable ?

ibgreen: >  get isPickable() {

1) In general I feel that getters (the `get` keyword) should be used sparingly and with rigorous consistency, as they can cause as much confusion as they can help (looks like a member/field but is actually a function and magically changes value). 

Let's drop the `get` for now unless we articulate a strategy.

2) Alternative is to drop the `get` and just define an `isPickable` method. 

Do we need another method just to wrap `layer.props.pickable`?  If there was more complex logic to determine "pickability" it would seem more strongly justified. I am certainly open to the idea, but if we introduce such a method/accessor we should really go through the code (possibly in another PR) and make sure it is used everywhere, otherwise we are just making the code more inconsistent. 

We might also consider if there are other similar accessors we should add, `isVisible()`, etc.

src/core/lib/pick-layers.js
```diff
@@ -268,9 +268,11 @@ function getClosestFromPickingBuffer(gl, {
     pickedObjectIndex: -1
   };
 
+  const pickableLayers = layers.filter(layer => layer.isPickable);
```
ibgreen: By the way, does this cover all picking cases? This file is really confusing with functions calling functions calling draw.... I.e. do all roads end here, going through this if?

src/core/lib/layer.js
1chandu: changed `getter` to member function, added `visible` check. Apparently we were going through all picking logic even when layers are not visible.

src/core/lib/pick-layers.js
1chandu: I moved `isPickable()` check into `getPickedColors` which is shared by all picking flows, good catch.

src/core/lib/layer.js
ibgreen: Question (not for this PR): How should handle a layer that is visible but not pickable. Should we render it into the picking buffer using null color to make sure we can't pick what is behind it? It would be a small extension to the picking module.

Right now I believe that users can pick "through" occluding, non-pickable layers.

PR #981, opened 2017-10-05
Attribute Transition: Initial Implementation

Pessimistress:
- New attribute API (`transition` flag)
- New base layer API (`transitions` prop mapping from accessor name to settings: `duration`, `easing`, `onStart`, `onEnd`, `onInterrupt`)
- Enable transition on applicable attributes of instanced core-layers
- Add example app

TODO:
- Tests
- Update layer documentations
- Composite layers need to map transition key names @heshan0131


examples/attribute-transition/README.md
```diff
@@ -0,0 +1,7 @@
+<div align="center">
+   <img width="150" heigth="150" src="https://webpack.js.org/assets/icon-square-big.svg" />
+</div>
+
+## Attribute Transition
+
+Demonstrate the abilities to interpolate attribute arrays.
```
ibgreen: To reduce confusion for external users who tend to go directly to master, we are creating new examples in the `examples/wip` folder.

Something to think about: 
* While we want simple examples, when we have integrated a new feature so cleanly into the API it would be natural to start using it / showing it off in other (existing or new) examples to make them more compelling. Since the API is so simple it shouldn't make those examples less accessible. It would also show that this feature handles more corner cases, I am still uncertain about exactly where the limitations land in practice. 
* Also, we have more animation features coming in, do they all get separate examples, or could we keep building on this one?

examples/attribute-transition/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "scripts": {
+    "start-local": "webpack-dev-server --env.local --progress --hot --port 3000 --open",
+    "start": "webpack-dev-server --progress --hot --port 3000 --open"
```
ibgreen: Put `start` first, to avoid tripping up external users?

src/core/lib/attribute-transition-manager.js
```diff
@@ -0,0 +1,367 @@
...
+  constructor({id, gl}, opts) {
+    this.id = id;
+    this.gl = gl;
+    this.opts = opts;
+
+    this.attributeTransitions = {};
+
+    this.needsRedraw = false;
+    this.model = null;
+    this.transformFeedback = new TransformFeedback(gl, {});
```
ibgreen: Empty braces no longer needed `new TransformFeedback(gl);`

src/core/lib/attribute-transition-manager.js
```diff
@@ -0,0 +1,367 @@
...
+const TRANSITION_STATE = {
+  NONE: 0,
+  PENDING: 1,
+  STARTED: 2,
+  ENDED: 3
+};
+
+const noop = () => {};
+
+class ProgramTransformFeedback extends Program {
```
ibgreen: This should to be sorted with luma.gl before we land. Looks like a copy of a hack from the Wind example which predates the current [luma.gl version](https://github.com/uber/luma.gl/blob/master/src/webgl/program.js#L64)

src/core/lib/attribute-transition-manager.js
```diff
@@ -0,0 +1,367 @@
...
+    // const currentValues = new (value.constructor)(newBufferSize);
+    // No entrance transition
+    const currentValues = value.slice();
+    if (buffer) {
+      // Transfer old buffer data to the new one
+      const oldBufferData = new Float32Array(bufferSize);
+
+      buffer.getData({
+        dstData: oldBufferData,
+        srcByteOffset: 0
```
ibgreen: Nit: srcByteOffset defaults to 0. Do we gain any readability by listing it here?

dev-docs/RFCs/v5.0/attribute-transition-rfc.md
```diff
@@ -49,7 +49,9 @@ new Layer({
 | --------- | -------- | ----------- |
```
1chandu: List default values where applicable?

src/core/lib/attribute-transition-manager.js
```diff
@@ -0,0 +1,367 @@
...
+      });
+
+      const len = Math.min(bufferSize, newBufferSize);
```
1chandu: `buffer.getData()` can handle this case, it checks available sizes of `src` and `dst` and copies the minimum of it, unless you are doing anything different.

src/core/lib/attribute-transition-manager.js
```diff
@@ -0,0 +1,367 @@
...
+    const {attribute, buffer} = transition;
+    const {value, size} = attribute;
+
+    const transitionSettings = this._getTransitionSettings(transition);
+
+    const needsNewBuffer = !buffer || transition.bufferSize !== value.length;
+
+    let fromState;
+    if (transitionSettings) {
+      fromState = this._getCurrentAttributeState(transition);
```
1chandu: Nit: `let fromState` is used only at line 336, so remove this if block at line 314 and make the assignment directly at line 336.

src/core/lib/attribute-transition-manager.js
```diff
@@ -0,0 +1,367 @@
...
+    }
+    const toState = new Buffer(this.gl, {size, data: value});
+
+    if (needsNewBuffer) {
+      if (buffer) {
+        buffer.unbind();
+        buffer._deleteHandle();
```
1chandu: Looks like we need a public method in luma:buffer class.

1chandu: Actually you can just call `buffer.delete()`, it is implemented at `Resource` level.

src/core/lib/attribute-transition-manager.js
```diff
@@ -0,0 +1,367 @@
...
+      transition.bufferSize = value.length;
+    }
+
+    if (transitionSettings) {
+      Object.assign(transition, transitionSettings);
+      transition.fromState = fromState;
+      transition.toState = toState;
+
+      // Reset transition state
+      const oldState = transition.state;
```
1chandu: Nit: don't need `oldState`.

 if (transition.state === TRANSITION_STATE.STARTED) {
    transition.onInterrupt(transition);
  }
  transition.state = TRANSITION_STATE.PENDING;

src/core/lib/attribute-transition-manager.js
Pessimistress: The old transform feedback buffer is overwritten between here and then.

src/core/lib/attribute-transition-manager.js
```diff
@@ -0,0 +1,341 @@
...
+  }
+
+  // get current values of an attribute, clipped/padded to the size of the new buffer
+  _getCurrentAttributeState(transition) {
+    const {attribute, buffer} = transition;
+    const {value, type, size} = attribute;
+
+    if (buffer) {
+      // No entrance transition
+      let oldBufferData = new Float32Array(value);
```
1chandu: Lines 248 to 252, can be replaced with just one line:
`oldBufferData = buffer.getData()`
`buffer.getData()` should allocate correct type and sized array , copy contents and return.

src/core/lib/attribute-manager.js
```diff
@@ -620,4 +634,31 @@ export default class AttributeManager {
...
+   * @params {Object} opts - transition options
+   * Returns updated attributes
+   */
+  setTransitionOptions(opts) {
+    if (!opts) {
+      this.attributeTransition = null;
+    } else if (this.attributeTransition) {
+      this.attributeTransition.setOptions(opts);
+    } else if (this.isTransitionSupported) {
+      this.attributeTransition = new AttributeTransitionManager(this, opts);
```
1chandu: Nit: can we change the name from `attributeTransition` to `transitionManger` or `attributeTranstionManger` for readability?

src/core/lib/attribute-transition-manager.js
```diff
@@ -0,0 +1,341 @@
...
+
+    const transitionSettings = this._getTransitionSettings(transition);
+
+    const needsNewBuffer = !buffer || transition.bufferSize !== value.length;
+
+    let fromState;
+    if (transitionSettings) {
+      fromState = this._getCurrentAttributeState(transition);
+    }
+    const toState = new Buffer(this.gl, {size, data: value});
```
1chandu: `toState` -> `toBuffer` , given `fromState` is an object of 3 different values?

src/core/lib/attribute-transition-manager.js
```diff
@@ -0,0 +1,341 @@
...
+    const {value, type, size} = attribute;
+
+    if (buffer) {
+      // No entrance transition
+      let oldBufferData = new Float32Array(value);
+      buffer.getData({dstData: oldBufferData});
+      if (!(value instanceof Float32Array)) {
+        oldBufferData = new (value.constructor)(oldBufferData);
+      }
+      return {size, type, value: oldBufferData};
```
1chandu: Why do we have to extract the size, type and value from the buffer, why not just return the buffer and it can be set directly to model attribute (as a buffer).

Pessimistress: The new attribute might not be the same size.

src/core/lib/attribute-transition-manager.js
Pessimistress: New attribute value may be a different length. We need to convert the old buffer to at least the same length as the new data.

Pessimistress:
@ibgreen @1chandu @shaojingli  If there're no more comments, I'd like to get this PR approved. As discussed, I will not merge it before the 4.2 release.

examples/wip/attribute-transition/app.js
```diff
@@ -0,0 +1,131 @@
...
+      data: this._pointGenerator.points,
+      getPosition: d => d.position,
+      getColor: d => d.color,
+      getRadius: d => d.radius,
+      updateTriggers: {
+        getPosition: positionsUpdated,
+        getRadius: radiusUpdated,
+        getColor: colorsUpdated
+      },
+      transition: {
```
ibgreen: Should we call it `transitions` (ref. `updateTriggers`)

An interesting question is how should this "deep object" i.e. `props.transitions` be diffed?  My suggestion is that we specify that we ignore it completely during diffing (for max perf) and only consult it when needed. That means that changing a value here will not trigger any change to an ongoing transition.

Also we will likely want to add prop transitions to this object as well, so that we have a unified approach.

src/core-layers/arc-layer/arc-layer.js
```diff
@@ -54,9 +54,9 @@ export default class ArcLayer extends Layer {
 
     /* eslint-disable max-len */
     attributeManager.addInstanced({
-      instancePositions: {size: 4, accessor: ['getSourcePosition', 'getTargetPosition'], update: this.calculateInstancePositions},
-      instanceSourceColors: {size: 4, type: GL.UNSIGNED_BYTE, accessor: 'getSourceColor', update: this.calculateInstanceSourceColors},
-      instanceTargetColors: {size: 4, type: GL.UNSIGNED_BYTE, accessor: 'getTargetColor', update: this.calculateInstanceTargetColors}
+      instancePositions: {size: 4, accessor: ['getSourcePosition', 'getTargetPosition'], update: this.calculateInstancePositions, transition: true},
```
ibgreen: It seems to me that `transition: true` is the 95% case. Would it make sense to make `transition: true` the default so we only have to set it to false for attributes where we want to prevent it? 

I've been keen on keeping the attribute tables as compact as possible, for max readability/comprehensibility, so I'd want to hear a strong argument for why making `true` the default would be a bad idea.

src/core/lib/attribute-manager.js
```diff
@@ -137,8 +139,9 @@ export default class AttributeManager {
    * @param {Object} [props]
    * @param {String} [props.id] - identifier (for debugging)
    */
-  constructor({id = 'attribute-manager'} = {}) {
+  constructor({id = 'attribute-manager', gl, transition} = {}) {
```
ibgreen: Nit: The convention for classes in luma.gl that take gl context is to take it as first parameter. `AttributeManager(gl, {...})`
I guess following that pattern would risk some backwards compatibility. `Model` does handle both forms.

ibgreen: Note: Now that attribute manager has access to the gl context, we could have it create and manage webgl Buffers for normal attributes

src/core/lib/layer.js
```diff
@@ -136,13 +136,26 @@ export default class Layer {
     if (changeFlags.dataChanged) {
       this.invalidateAttribute('all');
     }
+    if (changeFlags.propsChanged) {
```
ibgreen: I think this can be handled more efficiently, we could just extract it from the props object that is already supplied to AttributeManager when we trigger an accessor.

As stated above, my current thinking is that `props.transition` should be ignore and never trigger a prop update (as otherwise it will always trigger an update or require an expensive deep equal or apps to declare `attribute` as global constants.)

Given that we have an open PR on not triggering a change on inline functions, we may want a table that describes how different props are diffed.

src/core/lib/layer.js
```diff
@@ -357,7 +370,11 @@ export default class Layer {
 
     // Initialize state only once
     this.setState({
-      attributeManager: new AttributeManager({id: this.props.id}),
+      attributeManager: new AttributeManager({
+        id: this.props.id,
+        gl: this.context.gl,
+        animation: this.props.animation
```
ibgreen: What is this `animation` prop, that is only read once? Is this actually supposed to be `prop.transition`? 

Regardless I think we can avoid supplying it in the constructor per other comments.

src/core/lib/attribute-transition-manager.js
```diff
@@ -0,0 +1,353 @@
+import {GL, Program, Model, Geometry, Buffer, TransformFeedback} from 'luma.gl';
```
ibgreen: A simple test suite for this one? At least a few place holder test cases.

src/core/lib/attribute-transition-manager.js
```diff
@@ -0,0 +1,353 @@
...
+  }
+
+  /* Public methods */
+  setOptions(opts) {
+    this.opts = opts || {};
+  }
+
+  // Called when attribute manager updates
+  // Extracts the list of attributes that need transition
+  update(attributes) {
```
ibgreen: I liked the split into a separate transition manager class, however the two classes are tightly coupled here by both working on the same internal attribute array, which is a pretty complex object. We may want to address this, perhaps by making each object in that array into an internal object with methods.

src/core-layers/arc-layer/arc-layer.js
Pessimistress: It is still unclear what we'd do to interpolate some attributes (e.g. polygon layer). It is safer to have layers opt in this feature so that existing layers will not be broken.

src/core/lib/layer.js
Pessimistress: Yes it should be `transition`.

src/core-layers/arc-layer/arc-layer.js
ibgreen: Got it. Sounds like for instanced attributes, the default could be true. But for "custom geometry" attributes, it should not. Would that be too subtle, or would it pretty much cover our use case 100%?

Pessimistress: Path layer is instanced (one instance per line segment) but not transition enabled (not sure how to deal with changes in vertex count).
`instancePickingColors` is not transition enabled either.

ibgreen: OK. Those are pretty special cases, so explicitly overriding for those might be OK.

But, since there are complications, I'll let it rest for now.

1chandu:
By the way, can you add a TODO placeholder in `whats-new.md`, more details can be added later along with other documentation TODOs, doesn't have to be in this PR.

src/core/lib/attribute-manager.js
```diff
@@ -310,13 +320,17 @@ export default class AttributeManager {
...
-    const {attributes} = this;
+    const {attributes, attributeTransitionManger} = this;
     const changedAttributes = {};
     for (const attributeName in attributes) {
       const attribute = attributes[attributeName];
       if (attribute.changed) {
         attribute.changed = attribute.changed && !clearChangedFlags;
-        changedAttributes[attributeName] = attribute;
+
+        // If there is transition, let the transition manager handle the update
```
ibgreen: Seems a bit subtle to hide this logic in here. Maybe refactor this (split getChangedAttributes?) to make this more explicit.

src/core/lib/attribute-manager.js
```diff
@@ -674,4 +688,15 @@ export default class AttributeManager {
...
     }
   }
+
+  /**
+   * Update attribute transition to the current timestamp
+   * Returns updated attributes if any, otherwise `null`
+   */
+  updateTransition() {
+    const {attributeTransitionManger} = this;
+    const transitionUpdated = attributeTransitionManger.run();
```
ibgreen: `run()`? Maybe not the best if this just updates to current time? (run feels like we are starting a loop) 

`setCurrentTime()`?

src/core/lib/attribute-transition-manager.js
```diff
@@ -0,0 +1,373 @@
...
+      [GL.RASTERIZER_DISCARD]: true
+    };
+
+    model.draw({uniforms, parameters});
+
+    transformFeedback.end();
+  }
+
+  // Create a model for the transform feedback
+  _createModel() {
```
ibgreen: Could this be broken into a separate file and be made more reusable? Could be a "seed" for the GPGPU lib.
  
Regardless, would be nice to get all the shader and webgl logic out of this long file.

src/core/lib/layer.js
```diff
@@ -139,6 +139,16 @@ export default class Layer {
   // App can destroy WebGL resources here
   finalizeState() {}
 
+  // Update attribute transition
+  updateTransition() {
+    const {model, attributeManager} = this.state;
```
ibgreen: Use `getModels()`.

Pessimistress: `updateAttributes` does not use `getModels` either. We do not automatically update attributes for multi-model layers.

ibgreen: mainly want to avoid accessing state.model directly. Use the first model from `getModels`, and assert if there are more, or add a `getModel` method (maybe already there?).

Pessimistress: will do in another PR as it modifies existing functionality.

PR #982, opened 2017-10-05
Fix picking and highlighting regressions.

1chandu:
Fix for #979

Verified picking and highlighting with `layer-browser` and running website (examples).

### DIFF
--- a/src/core/lib/draw-layers.js
+++ b/src/core/lib/draw-layers.js
@@ -197,10 +197,10 @@ function drawLayersInViewport(gl, {
       );
 
       // Blend parameters must not be overriden
-      let layerParameters = Object.assign({viewport: glViewport}, layer.props.parameters || {});
+      const layerParameters = Object.assign({viewport: glViewport}, layer.props.parameters || {});
 
       if (drawPickingColors) {
-        layerParameters = Object.assign(layerParameters, {
+        Object.assign(layerParameters, {
           blendColor: [0, 0, 0, (layerIndex + 1) / 255]
         });
       }
--- a/src/core/lib/draw-layers.js
+++ b/src/core/lib/draw-layers.js
@@ -197,10 +197,10 @@ function drawLayersInViewport(gl, {
       );
 
       // Blend parameters must not be overriden
-      parameters = Object.assign({viewport: glViewport}, layer.props.parameters || {}, parameters);
+      let layerParameters = Object.assign({viewport: glViewport}, layer.props.parameters || {});
 
       if (drawPickingColors) {
-        parameters = Object.assign(parameters, {
+        layerParameters = Object.assign(layerParameters, {
           blendColor: [0, 0, 0, (layerIndex + 1) / 255]
         });
       }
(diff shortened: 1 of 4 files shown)


src/core/lib/draw-layers.js
```diff
@@ -197,18 +197,18 @@ function drawLayersInViewport(gl, {
       );
 
       // Blend parameters must not be overriden
-      parameters = Object.assign({viewport: glViewport}, layer.props.parameters || {}, parameters);
+      let layerParameters = Object.assign({viewport: glViewport}, layer.props.parameters || {});
```
1chandu: @ibgreen not using previous layer's params (`parameters` -> `let layerParameters`) didn't fix this issue but made this change based on your comment in the issue.

src/core/lib/draw-layers.js
```diff
@@ -197,18 +197,18 @@ function drawLayersInViewport(gl, {
...
 
       if (drawPickingColors) {
-        parameters = Object.assign(parameters, {
+        layerParameters = Object.assign(layerParameters, {
```
ibgreen: Let's drop the assignment, Object.assign modifies its first argument.
```
Object.assign(layerParameters, {
```

src/core/lib/draw-layers.js
```diff
@@ -239,11 +239,11 @@ function updateLayerHighlightColor(layer) {
...
 
-  if (layer.state.model) {
-    const pickingSelectedColorValid = layer.props.highlightedObjectIndex >= 0;
-    const pickingSelectedColor = pickingSelectedColorValid ?
-      layer.encodePickingColor(layer.props.highlightedObjectIndex) :
-      layer.nullPickingColor();
+  // Update picking module settings if highlightedObjectIndex is set.
+  // This will overwrite any settings from auto highlighting.
+  const pickingSelectedColorValid = layer.props.highlightedObjectIndex >= 0;
+  if (layer.state.model && pickingSelectedColorValid) {
```
ibgreen: Concern: What happens when we transition from valid to invalid? Won't the module settings need to be updated? This change seems to restrict updates to the valid case?

src/core/lib/layer-manager.js
```diff
@@ -658,8 +658,8 @@ export default class LayerManager {
     const radius = this._pickingRadius;
     const selectedInfos = this.pickObject({x: pos.x, y: pos.y, radius, mode: 'hover'});
 
-    const firstInfo = selectedInfos.find(info => info.index >= 0);
-    if (firstInfo && this._onLayerHover) {
+    const firstInfo = selectedInfos.find(info => info.index >= 0) || null;
```
ibgreen: Maybe a comment about the `null`, 
```
// Note: API docs state that info must be `null`. 
```

And why is this only in the hover case, not in the click case?

Share code between the two cases?

src/core/lib/draw-layers.js
ibgreen: > not using previous layer's params (parameters -> let layerParameters) didn't fix this issue but made this change based on your comment in the issue.

Yes good. Perhaps there is one more bug, but your change here is correct.

src/core/lib/draw-layers.js
1chandu: During picking cycle, we update `pickingSelectedColorValid` to true/false based on valid/invalid transition. Here we overwrite that if `highlightedObjectIndex` is set.

PR #984, opened 2017-10-05
add viewMatrix back so users can access it in their own vertex shader

jianhuang01:
When I was porting packed point cloud shaders from ATG to a customized deck.gl layer, found that there is no way to access viewMatrix, here add it back.

### DIFF
--- a/src/core/deprecated/shaderlib/project/project-deprecated.glsl.js
+++ b/src/core/deprecated/shaderlib/project/project-deprecated.glsl.js
@@ -21,6 +21,7 @@
 export default `\
 // Backwards compatibility
 uniform mat4 modelMatrix;
+uniform mat4 viewMatrix;
 uniform mat4 projectionMatrix;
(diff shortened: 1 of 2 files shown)


src/core/deprecated/shaderlib/project/project-deprecated.glsl.js
```diff
@@ -21,6 +21,7 @@
 export default `\
 // Backwards compatibility
 uniform mat4 modelMatrix;
+uniform mat4 viewMatrix;
```
1chandu: this looks like a deprecated file, do we need to change this?

jianhuang01: It is still available although the name has "deprecated" and this is temp fix, will implement a better solution later on as I proposed in my RFC, right now need focus on the packed-point-cloud rendering project, and make it unblocked.

PR #989, opened 2017-10-06
Retina option in layer browser

ibgreen:
Also fixes separation slider for non METER_OFFSET layers

### DIFF
--- a/examples/layer-browser/src/app.js
+++ b/examples/layer-browser/src/app.js
@@ -54,19 +54,23 @@ class App extends PureComponent {
         ScatterplotLayer: true
       },
       settings: {
-        // immutable: false,
-        // Effects are experimental for now. Will be enabled in the future
-        // effects: false,
         multiview: false,
-        separation: 0,
+        useDevicePixelRatio: false,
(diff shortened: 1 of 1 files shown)


ibgreen:
Retina mode is broken on master, this should help debug it.

src/core/shaderlib/project/project.glsl.js
```diff
@@ -75,7 +75,7 @@ vec2 project_mercator_(vec2 lnglat) {
 vec4 project_position(vec4 position) {
   // TODO - why not simply subtract center and fall through?
   if (project_uCoordinateSystem == COORDINATE_SYSTEM_LNG_LAT) {
-    return vec4(
+    return project_uModelMatrix * vec4(
```
1chandu: Is this change tied to the example change or should be ported to next patch? if so we have to separate this.

ibgreen:
> Is this change tied to the example change or should be ported to next patch? if so we have to separate this.

I don't think we should include it in a patch unless someone asks for it. But point taken, broke out #993

Issue #987, opened 2017-10-05
Class name cleanup


shaojingli:
Clean up names in deck.gl so that file names / class names / export, import names are consistent.

ibgreen:
We are generally pretty consistent. Any particular files that stick out as needing attention?

shaojingli:
For example, in `src/core/controllers/`, we are actually defining various `Controls` classes, `MapState` and `OrbitState` classes. We need to rename either the folder or the classes

The class defined in `ViewportController.js` is `Controller`

ibgreen:
Yes, I see what you mean. 
* From a previous discussion, since `controls` refers to things like buttons etc, we wanted to avoid that word and call these something else (like controllers). 
* And since the naming was still up in the air and changing, it was hard to be rigorous.

My RFC contains one attempt to settle on names, something like:
* `ViewState` (collects all position fields from MapState, OrbitState, FirstPersonState).
* `Controller` -> base class for `MapController`, `ThirdPersonController`, `FirstPersonController` (collects all `ViewState` update logic from MapState, OrbitState, FirstPersonState). They all generate a stream of `ViewState`s.
* `Viewport` -> base class for `MapViewport`, `ThirdPersonViewport`, `FirstPersonViewport` - they all consume `ViewState`s and display them in different ways.

The master component could either be `DeckGL` or a `ViewStateController`.

Pessimistress:
@ibgreen Can we also rename the `Controls` class to something else? `Controls` vs. `Controller` is extremely confusing even in our internal discussions.

jianhuang01:
#990

PR #990, opened 2017-10-06
Class name cleanup

jianhuang01:
#987


src/core/controllers/viewport-controls.js
```diff
@@ -35,7 +35,7 @@ const EVENT_TYPES = {
   KEYBOARD: ['keydown', 'keyup']
 };
 
-export default class Controls {
+export default class ViewportControls {
```
ibgreen: @Pessimistress has requested that avoid the use of the word controls as it is being used e.g. in react-map-gl to talk about other things.

jianhuang01: It is fine, we can discuss a good name and change it later on, here we just make the class name, file name and import name are all the same.

PR #996, opened 2017-10-07
Multi model fixes, OrbitController fix in plot layer

ibgreen:
https://github.com/uber/deck.gl/issues/991

### DIFF
--- a/src/core/lib/draw-layers.js
+++ b/src/core/lib/draw-layers.js
@@ -176,11 +176,6 @@ function drawLayersInViewport(gl, {
 
       visibleCount++;
 
-      if (!drawPickingColors) {
-        updateLayerHighlightColor(layer);
-        // TODO - Disable during picking
-      }
-
       const moduleParameters = Object.assign({}, layer.props, {
         viewport: layer.context.viewport
       });
@@ -195,6 +190,9 @@ function drawLayersInViewport(gl, {
       const layerParameters = Object.assign({viewport: glViewport}, layer.props.parameters || {});
 
       if (drawPickingColors) {
+        // TODO - Disable during picking
+        Object.assign(moduleParameters, getLayerHighlightColor(layer));
+
         Object.assign(layerParameters, {
           blendColor: [0, 0, 0, (layerIndex + 1) / 255]
         });
@@ -231,24 +229,20 @@ function getViewportFromDescriptor(viewportOrDescriptor) {
  * Returns the picking color of currenlty selected object of the given 'layer'.
  * @return {Array} - the picking color or null if layers selected object is invalid.
  */
-function updateLayerHighlightColor(layer) {
+function getLayerHighlightColor(layer) {
   // TODO - inefficient to update settings every render?
   // TODO: Add warning if 'highlightedObjectIndex' is > numberOfInstances of the model.
 
   // Update picking module settings if highlightedObjectIndex is set.
   // This will overwrite any settings from auto highlighting.
   const pickingSelectedColorValid = layer.props.highlightedObjectIndex >= 0;
-  // Note: Autohighlighting only works for single model layers
-  if (layer.state.model && pickingSelectedColorValid) {
+  if (pickingSelectedColorValid) {
     const pickingSelectedColor = layer.encodePickingColor(layer.props.highlightedObjectIndex);
 
-    const pickingParameters = {
+    return {
       pickingSelectedColor,
       pickingSelectedColorValid
     };
-
-    for (const model of layer.getModels()) {
-      model.updateModuleSettings(pickingParameters);
-    }
   }
+  return null;
 }
--- a/src/core/lib/draw-layers.js
+++ b/src/core/lib/draw-layers.js
@@ -181,9 +181,9 @@ function drawLayersInViewport(gl, {
         // TODO - Disable during picking
       }
 
-      if (layer.state.model) {
+      for (const model of layer.getModels()) {
         // Update project module parameters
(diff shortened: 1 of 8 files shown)


src/core/lib/pick-layers.js
```diff
@@ -123,7 +123,7 @@ export function pickObject(gl, {
       pickedLayer === layer &&
       pickingSelectedColor !== EMPTY_PIXEL
     );
-    // TODO - handle multi model layers?
+    // Note: Auto highlighting only works for single model layers
```
Pessimistress: Why does it only work for single model layers?

src/core/lib/layer.js
```diff
@@ -610,29 +616,41 @@ export default class Layer {
   }
 
   _updateBaseUniforms() {
-    this.setUniforms({
```
Pessimistress: Is there a reason to not call `setUniforms` here?

src/core/lib/pick-layers.js
ibgreen: > Why does it only work for single model layers?

In general it is not clear to me that our picking system works at all for multi model layers. I believe we assign the alpha picking identifier per layer, not per model, so there is no way to distinguish which model was picked, and AFAIK `pickInfo` contains no provisions for communicating it either.

src/core/lib/layer.js
ibgreen: > Is there a reason to not call setUniforms here?

Having a `setUniforms` method on the layer has been pointed out as point of confusion when writing new layers... should the layer call, or override `Layer.setUniforms`, or call `model.setUniforms`, or even `Program.setUniforms`, or should I use `Layer.draw(uniforms)` etc. 

In my mind, `Layer.setUniforms` is not a well defined function as it stands (when should it be called? what uniforms does it handle?). It does not have an underscore so it looks like a public method, but it is not documented as far as I can tell. So this was intended as a first step towards phasing it out.

src/core/lib/layer.js
```diff
@@ -610,29 +616,41 @@ export default class Layer {
...
-        pickingHighlightColor: this.props.highlightColor
-      });
+  _updateBaseModuleSettings() {
+    const settings = {
+      pickingHighlightColor: this.props.highlightColor
+    };
+    this.updateModuleSettings(settings);
+  }
+
+  updateModuleSettings(settings) {
```
shaojingli: Is `updateModuleSettings` an external API or a private method? Also, what are the reasons to have two similar methods here (`_updateBaseModuleSettings` and `updateModuleSettings`)?

src/core/lib/pick-layers.js
Pessimistress: True there may be more work required to make picking work in a multi-model layer, but it's definitely doable.
- A layer owns the picking color encoding/decoding process, therefore it could use the encoding to differentiate which model is picked.
- From the user's perspective, `pickInfo` only need to contain an object from the `data` array, and should never expose implementation details such as source model.

In practice, most multi-model layers I've written use multiple primitive models to render a more complex mesh. Models in this case share most attributes such as `instancePosition` and `instancePickingColor`, and picking already works out of the box. The code in question here makes it impossible for these layers to leverage in the auto highlight system.

src/core/lib/layer.js
Pessimistress: `_updateBaseModuleSettings` sets default module settings based on props, similar to what `_updateBaseUniforms` do. Per @ibgreen 's comment above, we may not want to keep `updateModuleSettings` and `setUniforms` as public methods, if we prefer external code directly accessing `getModels`.

src/core/lib/pick-layers.js
1chandu: +1, picking/highlighting system should work with multi models , as long as the layer encodes/decodes `instancingPickingColor` differently for each instance of the model.

> The code in question here makes it impossible for these layers to leverage the auto highlight system.

All this code does, checks layer's props and returns moduleSetting needed for auto-highlighing, this method doesn't have to access layer.state.mode, we can move this into layer class (as a public method, takes `drawPickingColors`  and return the moduelSettings, which can be merged with with `moduleParameters` in above 'drawLayersInViewport' method. So this code doesn't have to access `layer.state.model`.

ibgreen: > as long as the layer encodes/decodes instancingPickingColor differently for each instance of the model.

I believe that this is not what this PR does. The assumption here is that in a multi model layer all `Model`s have the same sets of "indices", corresponding to the same objects in the same data array, and should be highlighted at the same time. (i.e. the different models do not show different parts of the data array).

I think this is a valid choice, and I do not have strong opinions either way, but it should probably be mentioned somewhere in the docs.

src/core/lib/draw-layers.js
```diff
@@ -231,24 +229,20 @@ function getViewportFromDescriptor(viewportOrDescriptor) {
  * Returns the picking color of currenlty selected object of the given 'layer'.
  * @return {Array} - the picking color or null if layers selected object is invalid.
  */
-function updateLayerHighlightColor(layer) {
+function getLayerHighlightColor(layer) {
```
ibgreen: With this change, the name is no longer as instructive: `getPickingModuleParameters`?

Pessimistress: In this particular case, we are indeed only populating module settings when needed.

Generally speaking, I prefer sending all overridable moduleSettings/uniforms/parameters to `draw`. Currently we have multiple `withParameters` wrapped around the draw call and it's very difficult to track the context when debugging.

PR #1002, opened 2017-10-10
Fixes to drawlayers

ibgreen:
Make sure blending modes are correctly overridden, use correct parameters when rendering.

### DIFF
--- a/src/core/lib/draw-layers.js
+++ b/src/core/lib/draw-layers.js
@@ -186,8 +186,13 @@ function drawLayersInViewport(gl, {
         {layerIndex}
       );
 
+      // All parameter resolving is done here instead of the layer
       // Blend parameters must not be overriden
-      const layerParameters = Object.assign({viewport: glViewport}, layer.props.parameters || {});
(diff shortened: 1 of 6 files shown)


src/core/lib/draw-layers.js
```diff
@@ -186,8 +186,13 @@ function drawLayersInViewport(gl, {
         {layerIndex}
       );
 
+      // All parameter resolving is done here instead of the layer
       // Blend parameters must not be overriden
-      const layerParameters = Object.assign({viewport: glViewport}, layer.props.parameters || {});
+      const layerParameters = Object.assign({}, layer.props.parameters || {}, parameters);
+
+      Object.assign(layerParameters, {
```
Pessimistress: Why not merge with the line above?

ibgreen: > Why not merge with the line above?

Readability, I wanted to show more clearly the sequence of things. I was going back and forth, both ways are OK.

Since you are about to do another pass I'll merge and let you change it if you want to.

PR #1003, opened 2017-10-10
Publish work-in-progress layers in new `deck.gl-layers` module

ibgreen:
…nd MeshLayer


ibgreen:
Tested with layer-browser. All new layers, should not break anything.

ibgreen:
screen shot 2017-10-09 at 7 44 31 pm
@georgios-uber  Initial support for direction arrow rendering in PathLayer (not yet perfect but a start).

Pessimistress:
Shall we define a process how a layer will make it into the library?
The TextLayer model (start from example then graduate to core layers) worked pretty well IMO.
I do not have a problem with the MeshLayer, but the value of the other two layers is not very clear.

ibgreen:
> Shall we define a process how a layer will make it into the library?

Yes. My suggestion is that we start by moving such layers from sample-layers to experimental layers and start iterative improvements. At the end we either move a much improved layer back to sample layers or we decide on a good layer catalog to publish it in.

> The TextLayer model (start from example then graduate to core layers) worked pretty well IMO.
I do not have a problem with the MeshLayer, but the value of the other two layers is not very clear.

* The `PathOutlineLayer` in my opinion should be a default part of the `PathLayer`. Since that is a planar layer it seems natural to want to be able to handle relative relations and overlap. I just didn't want to propose merging it into the PathLayer at the moment due to the status of the PathLayer itself.

* The `PathMarkerLayer` is something I see applications implement on their own again and again. The interesting question here is if the layer can become sufficiently generic to be worth publishing.

ibgreen:
In response to concerns about bundling work-in-progress layers into alpha releases:
* Now publishes `experimental-layers` into a new `deck.gl-layers` npm module, in anticipation of longer-term consensus "monorepo" setup. 
* `layer-browser` has been updated to install new `deck.gl-layers` module and import experimental layers from it, and `npm run start-local` updated to consume from source.

src/experimental-layers/package.json
```diff
@@ -0,0 +1,90 @@
...
+  },
+  "main": "dist/index.js",
+  "module": "dist-es6/index.js",
+  "files": [
+    "dist",
+    "dist-es6",
+    "src"
+  ],
+  "scripts": {
+    "start": "(cd examples/layer-browser && (path-exists node_modules || npm i) && npm run start-local)",
```
Pessimistress: This was for the root. Just remove?

src/experimental-layers/package.json
```diff
@@ -0,0 +1,90 @@
...
+    "build-clean": "rm -fr dist dist-es6 && mkdir -p dist dist-es6",
+    "build-es6": "rm -fr dist-es6 && babel src --out-dir dist-es6 --source-maps inline",
+    "build-es5": "rm -fr dist && babel src --out-dir dist --plugins=transform-es2015-modules-commonjs --source-maps inline",
+    "build": "npm run build-clean && npm run build-es6 && npm run build-es5",
+    "cover": "NODE_ENV=test nyc --reporter html --reporter cobertura --reporter=lcov npm run test-cover",
+    "lint": "eslint src test examples && npm run lint-yarn",
+    "lint-yarn": "!(grep -q unpm.u yarn.lock) || (echo 'Please rebuild yarn file using public npmrc' && false)",
+    "publish-prod": "npm run build && npm run test && npm run test-dist && npm publish",
+    "publish-beta": "npm run build && npm run test && npm run test-dist && npm publish --tag beta",
+    "test": "npm run lint && npm run build && npm run test-node",
```
Pessimistress: Do we intend to have separate test set up for each module?

src/experimental-layers/src/mesh-layer/mesh-layer-fragment.glsl.js
```diff
@@ -0,0 +1,30 @@
...
+uniform bool hasTexture;
+uniform sampler2D sampler;
+uniform vec4 color;
+
+varying vec2 vTexCoord;
+varying vec4 vColor;
+varying float vLightWeight;
+
+void main(void) {
+  gl_FragColor = vec4(0., 0., 0., 1.);
```
Pessimistress: ?

src/experimental-layers/src/mesh-layer/mesh-layer-vertex.glsl.js
```diff
@@ -0,0 +1,40 @@
...
+  vec3 pos = positions;
+  pos = project_scale(pos * sizeScale);
+  pos = vec3(rotationMatrix * pos.xy, pos.z);
+  gl_Position = project_to_clipspace(vec4(instancePos + pos, 1.0));
+
+  picking_setPickingColor(instancePickingColors / 255.0);
+
+  vTexCoord = texCoords;
+  vColor = instanceColors;
+  vLightWeight = getLightWeight(pos, normals);
```
Pessimistress: Should apply `rotationMatrix` to `normals` too.

src/experimental-layers/src/mesh-layer/mesh-layer.js
```diff
@@ -0,0 +1,240 @@
...
+      id: this.props.id,
+      geometry: this.props.mesh,
+      isInstanced: true
+    }));
+  }
+
+  loadTexture(src) {
+    const {gl} = this.context;
+    const {model} = this.state;
+    getTexture(gl, src).then(texture => model.setUniforms({sampler1: texture}));
```
Pessimistress: The layer will not rerender when the texture loads because `needsRedraw` is not marked.

src/experimental-layers/src/mesh-layer/mesh-layer.js
```diff
@@ -0,0 +1,240 @@
...
+  }
+
+  calculateInstancePositions(attribute) {
+    const {data, getPosition} = this.props;
+    const {value, size} = attribute;
+    let i = 0;
+    for (const point of data) {
+      const position = getPosition(point);
+      value[i + 0] = position[0] || 0;
+      value[i + 1] = position[1] || 0;
```
Pessimistress: lng and lat should not need fallbacks.

Pessimistress: Use `get` for consistency?

src/experimental-layers/src/path-marker-layer/path-marker-layer.md
```diff
@@ -0,0 +1,26 @@
+# PathMarkerLayer
+
+Create small markers along a path (defaults to arrows showing "direction").
+
+
+// TODO - Move to Nebula
```
Pessimistress: Remove

src/experimental-layers/src/mesh-layer/mesh-layer-vertex.glsl.js
ibgreen: > Should apply rotationMatrix to normals too.

Good point left a TODO for now

Issue #903, opened 2017-08-31
PointCloudLayer fp64 + COORDINATE_SYSTEM.LNGLAT mode is not working


cmsdkfz:
I just created a simple point cloud layer as below, if I turn off 64fp, it renders OK, but if I set 64fp: true, nothing is rendered.

function generateRandomPoints() {
  var points = [];
  for (var i = 0; i < 10000; i++) {
    var diff1 = Math.random() / 1000;
    var diff2 = Math.random() / 1000;
    points.push({
      position: [-122.41669 + diff1, 37.7853 + diff2, 0],
      normal: [0, 0, 1]
    });
  }
  return points;
}

function generatePointLayer(points) {
  return new PointCloudLayer({
    id: 'point-cloud-layer',
    data: points,
    projectionMode: COORDINATE_SYSTEM.LNGLAT,
    getPosition: d => d.position,
    getNormal: d => d.normal,
    radiusPixels: 2,
    fp64: true
  });
}

Pessimistress:
@cmsdkfz Can you share more information about your browser, OS and GPU?

cmsdkfz:
Chrome Version 60.0.3112.101 (Official Build) (64-bit)
macOS sierra
AMD Radeon R9 M370X 2048 MB
Intel Iris Pro 1536 MB

1chandu:
@cmsdkfz, @ibgreen @Pessimistress I was able to reproduce the problem, working on the fix.

For some reason happens only when coordinate system is `LNGLAT`.

ibgreen:
`fp64` is only needed and used in `LNGLAT` mode. For `METER_OFFSETS` 32-bit precision is sufficient.

1chandu:
The 64 bit version for vertex shaders needs fixes.
Removing usage of unset uniform 'screenSize`, partially fixed the problem, points get rendered but their position is off by a small offset and size of the point do no match when compared non fp64 rendering, needs more digging.

ibgreen:
There might be two things going on here:
1) the original breakage of this layer
2) my refactor of the project shader module in a recent PR.

The intention was that the refactored `project` module would still set all the old uniforms for backwards compatibility. Maybe I forgot `screenSize`, we could add it back.

I also added a `project_pixel` function to the project module, and used it in the 32 bit shader.
* Can it also be used in the 64 bit?
* Is it giving incorrect results in 32 bits?

[...]

PR #1004, opened 2017-10-10
Fix point missing issue for 64 bits point cloud layer vertex shader

jianhuang01:
Reproduced issue in #903 and fix it here.

### DIFF
--- a/src/core-layers/point-cloud-layer/point-cloud-layer-vertex-64.glsl.js
+++ b/src/core-layers/point-cloud-layer/point-cloud-layer-vertex-64.glsl.js
@@ -31,7 +31,7 @@ attribute vec3 instancePickingColors;
 
 uniform float opacity;
 uniform float radiusPixels;
-uniform vec2 screenSize;
+uniform vec2 viewportSize;
 
 varying vec4 vColor;
 varying vec2 unitPosition;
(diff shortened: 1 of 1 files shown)


src/core-layers/point-cloud-layer/point-cloud-layer-vertex-64.glsl.js
```diff
@@ -47,17 +47,14 @@ void main(void) {
...
-  vertex_pos_modelspace[1] = sum_fp64(vertex_pos_localspace[1], projected_coord_xy[1]);
-  vertex_pos_modelspace[2] = sum_fp64(vertex_pos_localspace[2],
-    vec2(project_scale(instancePositions.z), 0.0));
+  vertex_pos_modelspace[0] = projected_coord_xy[0];
+  vertex_pos_modelspace[1] = projected_coord_xy[1];
+  vertex_pos_modelspace[2] = vec2(project_scale(instancePositions.z), 0.0);
   vertex_pos_modelspace[3] = vec2(1.0, 0.0);
 
   gl_Position = project_to_clipspace_fp64(vertex_pos_modelspace);
+  gl_Position += vec4(positions.xy * radiusPixels / viewportSize * 2.0, 0.0, 0.0);
```
ibgreen: Nice catch!

I do recommend we use the `project_to_pixel`s function for this (I think that is already done in the 32 bit layer). It handles a number of cases like devicePixelRatio, adjusting `w` to match desired  `focalDistance` etc, and since the project module already calculates the necessary uniforms we could drop them from the point cloud layer JS code.

jianhuang01: Thanks, have fixed it. Thought that function was not injected to the 64 bit glsl shader, it actually works.

ibgreen: I think the 64 bit project module imports the 32 bit one.

jianhuang01:
Tested with "npm run test-browser", " layer-browser" and "point-cloud-laz" example.

PR #1006, opened 2017-10-10
Fix useDevicePixelRatio prop usage

1chandu:
Fix for #843 
- Turn 'useDevicePixelRatio' to true, to match previous behavior.
- Add an option to toggle 'useDevicePixelRatio' to wind example. (This is useful in testing as layer-browser rendering doesn't change much when toggling this option.)

Verified with: `test-browser`, `layer-browser` (picking/highlighing/toggling useDevicePixelRatio) and `examples`.

### DIFF
--- a/examples/wind/src/app.js
+++ b/examples/wind/src/app.js
@@ -36,7 +36,8 @@ class Root extends Component {
         time: 0,
         showParticles: false,
         showWindDemo: false,
-        showElevation: false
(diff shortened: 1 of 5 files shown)


1chandu:
>So animationLoop.setViewParameters was the actual fix. Good catch.

Yeah, that and also passing correct value to AnimationLoop constructor.

Issue #1000, opened 2017-10-09
queryVisibleObject returns empty array in the layer-browser example


shaojingli:
the `queryVisibleObject` button on the top right in the layer-browser example returns no result

1chandu:
Over to Jian.

PR #1007, opened 2017-10-10
Fix issue that visible object query in layer-brower returns empty

jianhuang01:
As said in #1000, the query visible object button doesn't work in layer-browser, fix the bug in pick-layers.js where a for loop is not valid because of wrong condition check.

Tested with "npm run test-browser" and "layer-browser" example.

### DIFF
--- a/src/core/lib/pick-layers.js
+++ b/src/core/lib/pick-layers.js
@@ -374,23 +374,25 @@ function getUniquesFromPickingBuffer(gl, {
   const uniqueColors = new Map();
 
   // Traverse all pixels in picking results and get unique colors
-  for (let i = 0; i < pickedColors && pickedColors.length; i += 4) {
-    // Decode picked layer from color
-    const pickedLayerIndex = pickedColors[i + 3] - 1;
+  if (pickedColors) {
+    for (let i = 0; i < pickedColors.length; i += 4) {
+      // Decode picked layer from color
+      const pickedLayerIndex = pickedColors[i + 3] - 1;
 
-    if (pickedLayerIndex >= 0) {
-      const pickedColor = pickedColors.slice(i, i + 4);
-      const colorKey = pickedColor.join(',');
-      if (!uniqueColors.has(colorKey)) {
-        const pickedLayer = layers[pickedLayerIndex];
-        if (pickedLayer) { // eslint-disable-line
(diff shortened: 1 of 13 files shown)


src/core/lib/pick-layers.js
```diff
@@ -374,23 +374,25 @@ function getUniquesFromPickingBuffer(gl, {
...
-          });
-        } else {
-          log.error(0, 'Picked non-existent layer. Is picking buffer corrupt?');
+  if (pickedColors) {
+    for (let i = 0; i < pickedColors.length; i += 4) {
+      // Decode picked layer from color
+      const pickedLayerIndex = pickedColors[i + 3] - 1;
+
+      if (pickedLayerIndex >= 0) {
+        const pickedColor = pickedColors.slice(i, i + 4);
```
ibgreen: Not related to your change: We are generating a lot of temporary `pickedColor` arrays here. We could just build the key directly
```
const colorKey = `${pickedColor[i]}-${pickedColor[i]}-${pickedColor[i]}-${pickedColor[i]}`
```
And only generate the color array when find uniques
```
uniqueColors.set(colorKey, {
              pickedColor: pickedColors.slice(i, i + 4),
```

src/core/lib/pick-layers.js
```diff
@@ -374,23 +374,25 @@ function getUniquesFromPickingBuffer(gl, {
...
-            pickedObjectIndex: pickedLayer.decodePickingColor(pickedColor)
-          });
-        } else {
-          log.error(0, 'Picked non-existent layer. Is picking buffer corrupt?');
+  if (pickedColors) {
+    for (let i = 0; i < pickedColors.length; i += 4) {
+      // Decode picked layer from color
+      const pickedLayerIndex = pickedColors[i + 3] - 1;
+
+      if (pickedLayerIndex >= 0) {
```
ibgreen: Since these are JS arrays, it is OK to index with negative values. You could drop one if statement and just start with
```
const pickedLayer = layers[pickedLayerIndex];
if (pickedLayer) {
```
This will also reduce the nesting warnings that you are handling with eslint-disable

jianhuang01: We need "pickedLayerIndex >=0" check here, otherwise it will go to a layer which is not intended.

PR #1011, opened 2017-10-10
Improve Extruded Polygon Tesselation

Pessimistress:
- Directly filling into a typed array. `lodash.flattenDeep` will no longer be a dependency once the old choropleth layer is deprecated.
- Correct normal calculation. The tesselator was using the wrong up vector, and incorrectly assuming projection mode.

Benchmark under Chrome 61:
```
current - flat x 71.94 ops/sec ±1.57% (54 runs sampled)
current - extruded x 2.14 ops/sec ±2.30% (10 runs sampled)
current - wireframe x 8.05 ops/sec ±3.58% (24 runs sampled)

new - extruded x 19.65 ops/sec ±3.44% (36 runs sampled)
new - wireframe x 71.20 ops/sec ±4.48% (50 runs sampled)
```
Tested: node, browser, layer-browser, geojson example


Pessimistress:
@shaojingli @jianhuang01 I recommend cherry picking this into 4.1-release for the normal calculation fix.

PR #1014, opened 2017-10-10
Upgrade layers to new luma.gl API

Pessimistress:
Remove console warnings about inline attributes

### DIFF
--- a/examples/plot/plot-layer/surface-layer.js
+++ b/examples/plot/plot-layer/surface-layer.js
@@ -82,7 +82,8 @@ export default class SurfaceLayer extends Layer {
       vs: surfaceVertex,
       fs: fragmentShader,
       geometry: new Geometry({
-        drawMode: GL.TRIANGLES
+        drawMode: GL.TRIANGLES,
+        attributes: {}
       }),
       vertexCount: 0,
       isIndexed: true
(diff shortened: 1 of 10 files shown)


examples/plot/plot-layer/surface-layer.js
```diff
@@ -82,7 +82,8 @@ export default class SurfaceLayer extends Layer {
       vs: surfaceVertex,
       fs: fragmentShader,
       geometry: new Geometry({
-        drawMode: GL.TRIANGLES
+        drawMode: GL.TRIANGLES,
+        attributes: {}
```
Pessimistress: This seems to be a luma.gl bug (always issue a warning if `attributes` does not exit).

1chandu: Yeah, attributes can be set by attributeManger directly. We should check for possible attribute names in `opts` instead of `opts.attributes` to generate deprecated message.

ibgreen: Supporting attributes outside of the `attributes` field became a little messy when subclassing `Geometry` so I added this warning. Was about to disable it, but this works too.

src/core-layers/solid-polygon-layer/solid-polygon-layer.js
```diff
@@ -152,7 +152,8 @@ export default class SolidPolygonLayer extends Layer {
     return new Model(gl, Object.assign({}, this.getShaders(), {
       id: this.props.id,
       geometry: new Geometry({
-        drawMode: this.props.wireframe ? GL.LINES : GL.TRIANGLES
+        drawMode: this.props.wireframe ? GL.LINES : GL.TRIANGLES,
+        attributes: {}
```
ibgreen: @1chandu - Nit: this should not be needed. We should have a default value for attributes.

PR #1016, opened 2017-10-10
Pass useDevicePixelRatio to picking flow.

1chandu:
Verified with : `test-browser`, `examples`, `layer-browser` (toggle useDevicePixelRatio and verify picking/highlighting).

### DIFF
--- a/src/core/lib/draw-layers.js
+++ b/src/core/lib/draw-layers.js
@@ -21,12 +21,15 @@
 /* global window */
 import {GL, withParameters, setParameters} from 'luma.gl';
 import {log} from './utils';
+import assert from 'assert';
 
 let renderCount = 0;
 
 // TODO - Exported for pick-layers.js - Move to util?
-export const getPixelRatio = ({useDevicePixelRatio}) =>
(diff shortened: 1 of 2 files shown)


src/core/lib/draw-layers.js
```diff
@@ -21,12 +21,15 @@
...
 import {log} from './utils';
+import assert from 'assert';
 
 let renderCount = 0;
 
 // TODO - Exported for pick-layers.js - Move to util?
-export const getPixelRatio = ({useDevicePixelRatio}) =>
-  useDevicePixelRatio && typeof window !== 'undefined' ? window.devicePixelRatio : 1;
+export const getPixelRatio = ({useDevicePixelRatio}) => {
+  assert(typeof useDevicePixelRatio === 'boolean', 'Invalid useDevicePixelRatio');
```
ibgreen: Nit; To avoid bloating the library with string constants for error message, I usually only do an assert without a string. The developer can still find the exception pretty quickly in the debugger.

Ultimately we should have a production transform that strips asserts, then it wouldn't matter.

Issue #1017, opened 2017-10-11
Picking matches wrong layer when layer list starts with non-pickable layers


ibgreen:
https://github.com/uber/deck.gl/blob/master/src/core/lib/pick-layers.js#L266

Picking filters down the list of layers before picking but resolves the resulting layers indices in the original list, matching the wrong layers.

PR #1018, opened 2017-10-11
Fix picking when rendereing pickable and non-pickable layers

1chandu:
- When picking multiple layers, starting with non-pickable layers, if we only render pickable layers, resulting layerIndex is used on original layers, causing wrong layer being picked. Used pickableLayers to only detect early return, but always render all layers.


1chandu:
Fixes #1017

Issue #1015, opened 2017-10-10
Crash if `shouldUpdateState` returns true on viewport active


Pessimistress:
## Issue
If `Layer.shouldUpdateState` returns `true` with `changeFlags.viewportChanged`, the app will crash.

## Repro steps
Modify `Layer.shouldUpdateState` to
```
    return changeFlags.propsOrDataChanged || changeFlags.viewportChanged;
```
Run any example app.

## Cause

This issue is Introduced by the multi-viewport feature. At every render call, `LayerManager._activateViewport` calls `updateLayer` with the following `updateParams`:
```
{changeFlags: {viewportChanged: true}}
```
If `Layer.shouldUpdateState` returns `true`, it will invoke `Layer.updateState` and `Layer.updateAttributes` which expect `updateParams` to have non-empty `oldProps` and `props` fields, crashing the app.

The default `shouldUpdateState` looks for `changeFlags.somethingChanged` or `changeFlags.propsOrDataChanged`, both undefined in this case, therefore suppressing the issue.

@ibgreen

jianhuang01:
Ib is working on this and have a PR. Forward to him.

Issue #83, opened 2016-09-26
Performance Ideas


ibgreen:
- [ ] Picking performance improvement: It might be helpful to allow the application to control what type of picking a layer supports (e.g. hover vs. click vs. none) - as hovering over a large layer can cause a noticeable lag.
- [ ] Picking performance improvement: It might be helpful to have a mode that disables picking just during pan and zoom operations.
Layer rendering, memory use and buffer creation time:
- [ ] Experiment with packing colors as Uint8s instead of Float32s
- [ ] Layer rendering: Experiment with creating single interleaved buffers with  offsets and strides (harder as it likely requires luma.gl changes)
- [ ] Rewrite layer handle matching of many layers more efficiently. Use a map to match layer ids O(N) rather than repeated linear array searches O(N2)

k:
Are there any plans for when these performance improvements will be done?

Pessimistress:
@k Reusing layers should be available in 4.1.0-beta. It is still experimental, so testing is welcome.

k:
@Pessimistress great thanks. I'll check it out soon.

ibgreen:
Surprisingly, given though this is 10 years old, most items are still promising optimization opportunities. 
But given this it is not being worked on, lets close it for now.

PR #1019, opened 2017-10-11
Fix multi viewport update state

ibgreen:


ibgreen:
Fixes #1015

src/core/lib/layer-manager.js
```diff
@@ -351,7 +351,9 @@ export default class LayerManager {
       // TODO - reimplement viewport change detection (single viewport optimization)
       // TODO - don't set viewportChanged during setViewports?
       if (this.context.viewportChanged) {
-        this._updateLayerStates({viewportChanged: true});
+        for (const layer of this.layers) {
+          this._updateLayer(layer, {viewportChanged: true, somethingChanged: true});
```
1chandu: `somethingChanged`, is this supposed to be removed?

ibgreen: somethingChanged is set to indicated that any flag changed. I don't think any layer uses it.

Pessimistress: It's used by the default `shouldUpdateState` of CompositeLayer.

src/core/lib/layer-manager.js
```diff
@@ -362,9 +364,6 @@ export default class LayerManager {
 
   // Walk the layers and update their states
   _updateLayerStates(changeFlags) {
```
Pessimistress: Remove?

src/core/lib/layer-manager.js
1chandu: Ok, looks to me `somethingChanged` is referring to any change other than props or data and `Layer.diffProps` considers all cases and sets the `changeFlags`. So instead of setting these flags manually here should we always call `diffProps` , in future if any new flags added and diffProps changed this might cause an issue.

ibgreen: > It's used by the default `shouldUpdateState` of `CompositeLayer`.

I can't find a `shouldUpdateState` in `CompositeLayer`: https://github.com/uber/deck.gl/blob/master/src/core/lib/composite-layer.js#L24

ibgreen: > Ok, looks to me somethingChanged is referring to any change other than props or data and Layer.diffProps considers all cases and sets the changeFlags. So instead of setting these flags manually here should we always call diffProps , in future if any new flags added and diffProps changed this might cause an issue.

Good discussion. In this case the purpose was to avoid calling `diffProps` so I can't really call it again. 

I think the entire changeFlags system could do with a review:
1) it is not clear to me that we need a `somethingChanged` flag: i.e. if nothing has changed, why are we even calling these functions.
2) The flags that we do have should be better documented.
3) A better implementation would probably be to separate the "dirtying of change flags" from "updates" and then after we call `updateState` we clear all the flags. This would for instance allow us to stop calling `updateState` when setting a layer to `visible: false` while still being sure we only update what's needed when `visible` become `true` again.

ibgreen:
Comments addressed, performance discussion recorded in #83

PR #1021, opened 2017-10-12
Pass useDevicePixelRatio to picking flow

1chandu:
Verified with: `test-browser`, `examples`, `layer-browser` (Query Visible Object with/without retina).


1chandu:
> refactor this into these functions taking an opt object and just passing it on

Good idea 👍

PR #1024, opened 2017-10-12
Add new project64 function that removes raw fp64 math from layer shaders

ibgreen:


shaojingli:
@ibgreen , could you specify which tests have been run for this PR in the description?

ibgreen:
I have tested with layer-browser and the normal CI test suite.

But I couldn't test the point cloud layer since the example uses METER_OFFSET mode which disables fp64.

At this point I mostly want to float the idea of adding a function of this type. If we think it is a good idea, I think @jianhuang01 should probably be the one to actually complete the PR polishing the function and updating all layers.

src/core-layers/point-cloud-layer/point-cloud-layer-vertex-64.glsl.js
```diff
@@ -39,29 +39,23 @@ void main(void) {
...
-  vec2 projected_coord_xy[2];
-  project_position_fp64(instancePositions64xy, projected_coord_xy);
+  vec3 vertex = vec3(positions.xy * radiusPixels, 0.);
 
-  vec2 vertex_pos_modelspace[4];
-  vertex_pos_modelspace[0] = projected_coord_xy[0];
-  vertex_pos_modelspace[1] = projected_coord_xy[1];
-  vertex_pos_modelspace[2] = vec2(project_scale(instancePositions.z), 0.0);
-  vertex_pos_modelspace[3] = vec2(1.0, 0.0);
+  gl_Position = project_model_position_and_offset_to_clipspace_fp64(
```
jianhuang01: I need implement project_model_position_and_offset_to_clipspace_fp64, right?

jianhuang01:
@ibgreen Thanks for the PR, definitely a good direction.

src/core-layers/point-cloud-layer/point-cloud-layer-vertex-64.glsl.js
ibgreen: >I need implement project_model_position_and_offset_to_clipspace_fp64, right

The new GLSL function itself is already implemented in this PR: src/core/shaderlib/project64/project64.glsl.js.

* However, the layers need to be adapted to use it. I only updated two layers.

* Also, I think you should carefully consider the name/parameters of this API, I am not convinced that the name I gave it is very good.

* Finally, if we can add a similar function to the 32 bit project module, it could be possible to unify our 32 and 64 bit vertex shaders. That would reduce effort to develop new layers, the cost to maintain old layers and deck.gl code/bundle size.

1chandu:
Closing this as discussed in meeting, will use #1049 to track/resume this work.

ibgreen:
I will move the new functions into a new optional shader module on top of `project64` and put that into experimental-layers. 

This way mesh layer and other experimental layers can use this while discussion continue on whether this should be part of the core project64 modules.

PR #1028, opened 2017-10-13
Fix point missing issue for 64 bits point cloud layer vertex shader

jianhuang01:
To resolve issue #903, need this PR for release branch.


jianhuang01:
Tested with "npm run test-browser" and a modified example which can render lat lon based points with 64 bits precision.

jianhuang01:
64bits-points

Red dots are 64bits precision points fixed by this PR, while the yellow dots are rendered by scatterplot layer.

jianhuang01:
@1chandu 32bits shader is fine, already tested.

PR #1030, opened 2017-10-14
Add 3rd person viewport support

gnavvy:
First attempt on adding the 3rd person viewport support.
Now testing in the examples/wip/first-person-map (I know, will add a separate example later)

screen shot 2017-10-13 at 5 54 31 pm

### DIFF
--- a/examples/wip/first-person-map/app.js
+++ b/examples/wip/first-person-map/app.js
@@ -8,6 +8,7 @@ import {
   FirstPersonState,
   // Viewport classes provides various views on the state
   FirstPersonViewport,
+  ThirdPersonViewport,
   WebMercatorViewport,
   PolygonLayer,
   PointCloudLayer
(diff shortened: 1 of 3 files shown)


examples/wip/first-person-map/app.js
```diff
@@ -22,8 +23,10 @@ import {StaticMap} from 'react-map-gl';
 
 // Source data CSV
 const DATA_URL = {
-  BUILDINGS: 'https://raw.githubusercontent.com/uber-common/deck.gl-data/master/examples/trips/buildings.json',  // eslint-disable-line
-  TRIPS: 'https://raw.githubusercontent.com/uber-common/deck.gl-data/master/examples/trips/trips.json'  // eslint-disable-line
+  BUILDINGS:
+    'https://raw.githubusercontent.com/uber-common/deck.gl-data/master/examples/trips/buildings.json', // eslint-disable-line
```
ibgreen: Why this change? Should we drop the `eslint-disable-line` since this was only intended to allow us to break the length limit? If we are reformatting to avoid breaking line max, then drop.

examples/wip/first-person-map/app.js
```diff
@@ -167,7 +176,7 @@ class Root extends Component {
         id: 'trips',
         data: trips,
         getPath: d => d.segments,
-        getColor: d => d.vendor === 0 ? [253, 128, 93] : [23, 184, 190],
+        getColor: d => (d.vendor === 0 ? [253, 128, 93] : [23, 184, 190]),
```
ibgreen: Why the parenthesis?

examples/wip/first-person-map/app.js
```diff
@@ -187,23 +196,27 @@ class Root extends Component {
...
         radiusPixels: 20
       }),
       new PointCloudLayer({
         id: 'ref-point',
-        data: [{
-          position: [-1, 0, 2],
-          color: [255, 0, 0, 255],
-          normal: [1, 0, 0]
-        }],
+        data: [
```
ibgreen: I guess this could be a constant.

examples/wip/first-person-map/app.js
gnavvy: just saw these comments, those are from the prettier plugins with no functional changes.
I'll disable prettier for deck.gl before we decide to adopt it.

PR #1031, opened 2017-10-14
Fix plot layer state bug

Pessimistress:
Fix problem where the axes are rendered until mouse interaction.

### DIFF
--- a/examples/plot/plot-layer/surface-layer.js
+++ b/examples/plot/plot-layer/surface-layer.js
@@ -1,24 +1,21 @@
 import {Layer} from 'deck.gl';
 import {GL, Model, Geometry} from 'luma.gl';
 
-import {scaleLinear} from 'd3-scale';
 import surfaceVertex from './surface-vertex.glsl';
 import fragmentShader from './fragment.glsl';
 
 const DEFAULT_COLOR = [0, 0, 0, 255];
-const DEFAULT_GET_SCALE = () => scaleLinear();
 
 const defaultProps = {
   data: [],
   getPosition: () => [0, 0, 0],
   getColor: () => DEFAULT_COLOR,
-  getXScale: DEFAULT_GET_SCALE,
-  getYScale: DEFAULT_GET_SCALE,
-  getZScale: DEFAULT_GET_SCALE,
+  xScale: null,
+  yScale: null,
+  zScale: null,
   uCount: 100,
   vCount: 100,
-  lightStrength: 0.1,
-  onUpdate: () => {}
+  lightStrength: 0.1
 };
 
 /*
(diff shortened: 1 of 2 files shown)


examples/plot/plot-layer/surface-layer.js
```diff
@@ -199,51 +188,21 @@ export default class SurfaceLayer extends Layer {
...
-      const startIndex = j * 4;
-      if (!value[startIndex + 3]) {
-        value[startIndex] = xScale(value[startIndex]);
-        value[startIndex + 1] = zScale(value[startIndex + 1]);
-        value[startIndex + 2] = yScale(value[startIndex + 2]);
-      }
-    }
-
     attribute.value = value;
-    this.props.onUpdate({xScale, yScale, zScale});
```
Firenze11: Just to make sure I understand this solution correctly, is there a reason that `calculatePositions` in `SurfaceLayer` is not called when the `SurfaceLayer` is initiated, such that `xScale` is not computed and therefore not passed to `PlotLayer` (which results in `AxisLayer` not being rendered)? If that is the case, `x`, `y`, `z` positions in `SurfaceLayer` would not be computed either (and would only be computed until user's interaction). Is that a bug that needs to be fixed as well?

Pessimistress: `updateState` is called during initiation.

Firenze11: I see, then `this.props.onUpdate` should be called too. Why the `AxesLayer` was not shown then? (I was assuming the `AxesLayer` was not shown because its `xScale` etc. was not computed.

Pessimistress: The `onUpdate` callback is removed in this PR. We used to use this callback to set the state of the parent layer, and then use the new state to draw the axes layer. However, currently there is no way to trigger `CompositeLayer.renderLayers()` from within. Therefore the axes layer was not displayed until the user manipulates the viewport.

@ibgreen I was half expecting `needsRedraw` to serve this purpose. The catch being composite layers' "draw" calls happen in `updateLayers`, while regular layers' draw calls happen in `drawLayers`.

Firenze11: I see, thanks for the explanation!

ibgreen: > I was half expecting needsRedraw to serve this purpose. The catch being composite layers' "draw" calls happen in updateLayers, while regular layers' draw calls happen in drawLayers.

Yes, I believe we still have some work to do here. I also feel that we don't have a convincing definition of exactly what should trigger composite layer updates / rerenders of sub layers. E.g. `shouldUpdateState` might be implemented differently by different sublayers, but right now we ignore that.

PR #1032, opened 2017-10-14
Fix PolygonLayer issue with Data Accessors

georgios-uber:


src/core-layers/polygon-layer/polygon-layer.js
```diff
@@ -160,8 +160,8 @@ export default class PolygonLayer extends CompositeLayer {
         miterLimit: lineMiterLimit,
 
         getPath: x => x.path,
-        getColor: getLineColor,
-        getWidth: getLineWidth,
+        getColor: x => getLineColor(x.object),
```
ibgreen: @georgios-uber As far as I can tell this would make the layer look for `<object>.object.fillColor` which is inconsistent with the docs that says we are looking for `<object>.fillColor` and would be a big change. 

https://uber.github.io/deck.gl/#/documentation/layer-catalog/polygon-layer?section=render-options

Maybe something in the docs is unclear?

georgios-uber: This layer takes paths as input, not the original user-provided objects.
`const {paths} = this.state;`

This is generated in line 74 of this file.

The current implementation violates the docs (because fillColor is ignored) and this fixes it.

jianhuang01:
@ibgreen @georgios-uber Do we need port it to release branch? Can anyone specify what's the issue and how to test it in LayerBrowser? When I debug it in LayerBrowser, didn't see any color or width property in paths

georgios-uber:
@jianhuang01 It's a bug fix. Do you usually port them to release branch?

jianhuang01:
@georgios-uber yeah, we need port bug fixes to release branch and publish patch release if they are existing there. But I don't know how to test and verify this on our release branch. Need more details.

PR #1033, opened 2017-10-14
Optimize prop diffing when using inline functions as accessors.

ibgreen:
* Add AttributeManager.getAccessors() - use these a simple prop types system (determine which function props can be shallow equaled)
* Change prop type diffing to ignore changes in function-valued props (if the prop is in the `propTypes`) 
* Move `shallow-equal.js` to react "package" (not used by core)
* Move `compare-arrays.js` to experimental (not used by core, but may be needed for prop-types)


src/core/lib/layer.js
```diff
@@ -497,11 +497,19 @@ export default class Layer {
   }
 
   diffProps(oldProps, newProps, context) {
+    // Until we have proper prop types, autodetect accessors by querying the attribute manager
+    const {attributeManager} = this.state || {};
+    const propTypes = attributeManager ? attributeManager.getAccessors() : {};
```
Pessimistress: I assume it is called `propTypes` so that we can expand this object to cover more than functions in the future. If that is true, then it is really strange that `propTypes` comes from a function called `getAccessors` and from the attribute manager.

The logic implemented in this PR will perfectly apply if we just pass the defaultProps object as `propTypes` here. Is it necessary to deduce the list of accessors over and over again?

src/core/lib/attribute-manager.js
```diff
@@ -318,6 +318,21 @@ export default class AttributeManager {
...
+  /**
+   * Returns object containing all accessors as keys
+   * @return {Object} - accessors object
+   */
+  getAccessors() {
+    const accessors = {};
+    for (const attributeName in this.attributes) {
+      const attribute = this.attributes[attributeName];
+      if (attribute.accessor) {
+        accessors[attribute.accessor] = true;
```
Pessimistress: `attribute.accessor` may be an array.

ibgreen: > attribute.accessor may be an array.

Yep good catch. Using the `updateTriggers` member now, it already contains what I need (and handles this case).

src/core/lib/layer.js
ibgreen: > I assume it is called propTypes so that we can expand this object to cover more than functions in the future. If that is true, then it is really strange that propTypes comes from a function called getAccessors and from the attribute manager.

Changed from `propTypes` to `shallowCompareProps` for now to remove any confusion.

> The logic implemented in this PR will perfectly apply if we just pass the defaultProps object as propTypes here. Is it necessary to deduce the list of accessors over and over again?

The problem here is that @heshan0131 is testing for function changes in the aggregating layers. I cannot just make any function compare shallowly equal until we have resolved that.

PR #1034, opened 2017-10-15
Fix blending in ScatterplotLayer example

ibgreen:
Micro fix for blending in our scatterplot layer example.  Before and after:
screen shot 2017-10-15 at 8 53 35 am
screen shot 2017-10-15 at 8 53 25 am

### DIFF
--- a/examples/layer-browser/src/examples/core-layers.js
+++ b/examples/layer-browser/src/examples/core-layers.js
@@ -171,7 +171,7 @@ const ScatterplotLayerExample = {
     getPosition: d => get(d, 'COORDINATES'),
     getColor: d => [255, 128, 0],
     getRadius: d => get(d, 'SPACES'),
-    opacity: 0.5,
+    opacity: 1,
     pickable: true,
     radiusScale: 30,
     radiusMinPixels: 1,


examples/layer-browser/src/examples/core-layers.js
```diff
@@ -175,7 +177,11 @@ const ScatterplotLayerExample = {
     pickable: true,
     radiusScale: 30,
     radiusMinPixels: 1,
-    radiusMaxPixels: 30
+    radiusMaxPixels: 30,
+    parameters: {
+      blend: true,
+      blendEquation: GL.MAX
```
Pessimistress: The weird blend effect was caused by improper blend functions of the alpha channel.
But this is also incorrect.
First, we should change the default blending mode instead of in the example.
Second, it should be
```
gl.blendFuncSeparate(gl.SRC_ALPHA, gl.ONE_MINUS_SRC_ALPHA, gl.ONE, gl.ONE_MINUS_SRC_ALPHA);
gl.blendEquationSeparate(gl.FUNC_ADD, gl.FUNC_ADD);
```

jianhuang01: I think Ib's intension is to remove the cascaded blending effects of circles, make all the circles melted together, doubt that can be resolved by a regular alpha blending.

Pessimistress: Because we use this app for debugging, it is important that it reflects the default behavior of deck.gl. If we believe the default behavior does not provide the best user experience, then let's change the defaults. We've seen hidden customization in examples causing more confusion for the users.

Pessimistress: Also `gl.MAX` is [WebGL2 only](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/blendEquation) so this will not work in Safari.

ibgreen: > Also gl.MAX is WebGL2 only so this will not work in Safari.

FWIW, this is available as a fairly commonly supported [extension](https://developer.mozilla.org/en-US/docs/Web/API/EXT_blend_minmax) under WebGL1. luma.gl transparently uses the extension when its available under WebGL1. It would be my guess that it is available in Safari.

ibgreen: Just tested on desktop Safari, it works.

ibgreen: > It should be
> ```
gl.blendFuncSeparate(gl.SRC_ALPHA, gl.ONE_MINUS_SRC_ALPHA, gl.ONE, gl.ONE_MINUS_SRC_ALPHA);
gl.blendEquationSeparate(gl.FUNC_ADD, gl.FUNC_ADD);
```

We added recently a new `DeckGL` prop `initWebGLParameters` that defaults to false (it is supposed to be set to default true in next major release). This allows us to play with changing the default parameter set already now without breaking apps that don't opt in.

I tried adding the recommended changes to `initWebGLParameters`. It changed the color quite a bit, but did not solve the overdraw and Z fighting issues.

<img width="691" alt="screen shot 2017-10-16 at 3 44 27 pm" src="https://user-images.githubusercontent.com/7025232/31638556-1f7c7f6a-b289-11e7-8a50-e03403cf0358.png">

ibgreen: Updated formatting of previous post

Pessimistress: I don't see why overlapping is an issue, if you don't like it you can just set layer opacity to 1.
Z fighting is something we should definitely resolve, and hiding it from our primary debugging portal sounds like the wrong approach.

ibgreen:
> and hiding it from our primary debugging portal sounds like the wrong approach.

I get where you are coming from and you make a good point, and I am OK with closing this PR. 

Before I do, on reflection, isn't a bigger issue that we are not testing gl parameters in our "debugging portal". Several apps are now relying on this feature but tests are not available in our manual test rig. Shouldn't we instead have a representative spread of parameters in the examples?

Also, currently we don't have much of a story to offer users who has been raising the blending issue. There are several tricks available but we haven't shared the information. Until we implement a system for dealing with z fighting in instanced layers, or write an article about blending techniques, things will remain in limbo. Having an example in the layer browser would at least make it discoverable.

ibgreen:
@Pessimistress Removed custom `blendEquation` and set default ScatterplotLayer example opacity to 1 per review feedback. Can you approve and merge?

jianhuang01:
scatterplot

@ibgreen @Pessimistress By changing opacity to 1.0, here is the effect, just want to make sure this is the one you want.

ibgreen:
@jianhuang01 Exactly, setting it to 1 removes the z-fighting for the default case. 

But you must have made the blendEquation change to, as the colors are different.

PR #1035, opened 2017-10-16
Move prop diffing from layer.js to props.js

ibgreen:
Consolidates prop diffing and moves prop related code from `layer.js` to `props.js`. 
* Will make layer code easier to work with, and to prepare for features like more sophisticated prop diffing, prop types and prop animations/transitions.
* simple interface, the new `diffProps` function just returns an object with changeFlags.
* and a callback is supplied by the layer to handle any detected updateTriggers.

Tested in layer-browser.

### DIFF
--- a/src/core/lib/layer.js
+++ b/src/core/lib/layer.js
@@ -22,7 +22,7 @@
 import {COORDINATE_SYSTEM, LIFECYCLE} from './constants';
 import AttributeManager from './attribute-manager';
 import Stats from './stats';
-import {getDefaultProps, compareProps} from './props';
+import {getDefaultProps, diffProps} from './props';
 import {log, count} from './utils';
 import {applyPropOverrides, removeLayerInSeer} from './seer-integration';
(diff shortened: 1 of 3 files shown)


src/core/lib/layer.js
```diff
@@ -535,72 +506,47 @@ export default class Layer {
       propsOrDataChanged,
       viewportChanged,
       somethingChanged,
-      reason: dataChangedReason || propsChangedReason || 'Viewport changed'
+      reason: dataChanged || propsChanged || 'Viewport changed'
```
Pessimistress: Are we assuming that viewport is always changed?

Pessimistress: Should also use `updateTriggersChanged`?

ibgreen: > Are we assuming that viewport is always changed?
> Should also use updateTriggersChanged?

Fixed. This code will be removed in a followup diff but I made an update that makes sure this PR leaves things in a good state, in case the next PR gets held up in review.

PR #1036, opened 2017-10-16
Remove excessive nesting in picking code

ibgreen:
In pick-layers, instead of picking functions starting with rendering and reading the picking buffer, and then extracting data from it, the rendering was deeply nested.

Every time we change rendering parameters there were bugs introduced by forgetting to forward them to rendering at some stage.

This change is also needed for an "improved redraw logging" PR which is staged after this.


1chandu:
Undid  `redrawReason` change and tested with layer-browser and examples.

PR #1039, opened 2017-10-16
Fix the normal direction of the hexagon layer and lighting module

shaojingli:
The normals of our layers (hexagon cell and extruded polygon) has y axis as the vertical axis, so our lighting modules apply a transform to correct it. 

`vec3 normals_worldspace_vec3 = normals_worldspace.xzy;`

Now, the extruded polygons has z-axis as vertical (https://github.com/uber/deck.gl/pull/1011), so we need to fix the hexagon cell layer and lighting module to make the lighting correct.

@1chandu @Pessimistress @heshan0131

### DIFF
--- a/src/core-layers/hexagon-cell-layer/hexagon-cell-layer-vertex-64.glsl.js
+++ b/src/core-layers/hexagon-cell-layer/hexagon-cell-layer-vertex-64.glsl.js
@@ -49,10 +49,10 @@ void main(void) {
   mat2 rotationMatrix = mat2(cos(angle), -sin(angle), sin(angle), cos(angle));
 
   vec2 rPos = rotationMatrix * positions.xz;
-  vec2 rNorm = rotationMatrix * normals.xz;
+  vec2 rNorm = rotationMatrix * normals.xz; // the hexagon cells has y axis as the vertical axis
 
   vec3 rotatedPositions = vec3(rPos.x, positions.y, rPos.y);
-  vec3 rotatedNormals = vec3(rNorm.x, normals.y, rNorm.y);
+  vec3 rotatedNormals = vec3(rNorm.x, rNorm.y, normals.y);
 
   // calculate elevation, if 3d not enabled set to 0
   // cylindar gemoetry height are between -0.5 to 0.5, transform it to between 0, 1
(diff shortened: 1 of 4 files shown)


src/core-layers/hexagon-cell-layer/hexagon-cell-layer-vertex-64.glsl.js
```diff
@@ -49,10 +49,10 @@ void main(void) {
   mat2 rotationMatrix = mat2(cos(angle), -sin(angle), sin(angle), cos(angle));
 
   vec2 rPos = rotationMatrix * positions.xz;
-  vec2 rNorm = rotationMatrix * normals.xz;
+  vec2 rNorm = rotationMatrix * normals.xz; // the hexagon cells has y axis as the vertical axis
```
ibgreen: My sense is that we should change the geometry to be consistent with a geospatial coordinate system which we clearly document to be Z = up, and adjust the shader accordingly. Don't mind if land this as a quick fix, but in that case it would be good to open an issue to make sure we eventually do it right.

src/core-layers/hexagon-cell-layer/hexagon-cell-layer-vertex-64.glsl.js
```diff
@@ -62,11 +62,11 @@ void main(void) {
     elevation = project_scale(instancePositions.z * (positions.y + 0.5) *
       ELEVATION_SCALE * elevationScale);
   }
-  
-  // if ahpha == 0.0 or z < 0.0, do not render element  
+
+  // if ahpha == 0.0 or z < 0.0, do not render element
```
ibgreen: Nit: Typo, alpha...

src/core-layers/hexagon-cell-layer/hexagon-cell-layer-vertex-64.glsl.js
Pessimistress: +1. At lease this should be rolled into the rotation matrix.

1chandu:
@shaojingli , should we land this? Do we need additional changes after this?

shaojingli:
@1chandu we should land this for now.

PR #1046, opened 2017-10-17
PolygonLayer- add elevationScale, fix updateTriggers.getElevation doesn't trigger updateGeometry

heshan0131:
- bug fix: Recalculate geometry when updateTriggers.getElevation has changed
- add elevationScale

### DIFF
--- a/examples/layer-browser/src/examples/core-layers.js
+++ b/examples/layer-browser/src/examples/core-layers.js
@@ -124,11 +124,12 @@ const PolygonLayerExample = {
     getLineColor: f => [0, 0, 0, 255],
     getLineDashArray: f => [20, 0],
     getWidth: f => 20,
-    getHeight: f => Math.random() * 1000,
+    getElevation: f => Math.random() * 1000,
     opacity: 0.8,
     pickable: true,
     lineDashJustified: true,
-    lightSettings: LIGHT_SETTINGS
+    lightSettings: LIGHT_SETTINGS,
+    elevationScale: 0.6
   }
 };
 
--- a/src/core-layers/solid-polygon-layer/solid-polygon-layer.js
+++ b/src/core-layers/solid-polygon-layer/solid-polygon-layer.js
@@ -21,6 +21,7 @@
 import {COORDINATE_SYSTEM, Layer, experimental} from '../../core';
 const {enable64bitSupport, get} = experimental;
 import {GL, Model, Geometry} from 'luma.gl';
+import {compareProps} from '../../core/lib/props';
 
 // Polygon geometry generation is managed by the polygon tesselator
 import {PolygonTesselator} from './polygon-tesselator';
@@ -36,6 +37,9 @@ const defaultProps = {
   // Whether to draw a GL.LINES wireframe of the polygon
   wireframe: false,
   fp64: false,
+
+  // elevation multiplier
+  elevationScale: 1,
 
   // Accessor for polygon geometry
   getPolygon: f => get(f, 'polygon') || get(f, 'geometry.coordinates'),
@@ -101,11 +105,12 @@ export default class SolidPolygonLayer extends Layer {
   }
 
   draw({uniforms}) {
-    const {extruded, lightSettings} = this.props;
+    const {extruded, lightSettings, elevationScale} = this.props;
     const {viewport} = this.context;
 
     this.state.model.render(Object.assign({}, uniforms, {
       extruded: extruded ? 1.0 : 0.0,
+      elevationScale,
       pixelsPerUnit: viewport.getDistanceScales().pixelsPerDegree
     },
     lightSettings));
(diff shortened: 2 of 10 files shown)


src/core-layers/solid-polygon-layer/solid-polygon-layer.js
```diff
@@ -127,9 +132,17 @@ export default class SolidPolygonLayer extends Layer {
     const geometryConfigChanged = props.extruded !== oldProps.extruded ||
       props.wireframe !== oldProps.wireframe || props.fp64 !== oldProps.fp64;
 
-     // When the geometry config  or the data is changed,
-     // tessellator needs to be invoked
-    if (changeFlags.dataChanged || geometryConfigChanged) {
+    // check if updateTriggers.getElevation has been triggered
+    const getElevationTriggered = changeFlags.updateTriggersChanged &&
```
ibgreen: While I think I understand why you needed to do this (you have an accessor that affects positions), you'll probably agree with me when I say that I don't like the amount of code needed to handle this case here. 

We should take a second look at the `PolygonTesselator` it can handle this - to make sure we don't need it.

examples/layer-browser/src/examples/core-layers.js
```diff
@@ -117,11 +117,17 @@ const PolygonLayerExample = {
...
-    getHeight: f => Math.random() * 1000,
+    getElevation: f => Math.random() * 1000,
     opacity: 0.8,
     pickable: true,
-    lightSettings: LIGHT_SETTINGS
-  }
+    lightSettings: LIGHT_SETTINGS,
+    elevationScale: 0.6
+  },
+  getUpdateTriggers: settings => ({
```
ibgreen: Is this intended to create a manual test case for the `getElevation` updateTrigger?

Maybe we should add a separate button to trigger updateTriggers rather than hooking it to some arbitrary setting? 

The prop handling in the layer browser is already very hacky. I would like us to at least think twice before adding more hacks.

src/core-layers/solid-polygon-layer/solid-polygon-layer.js
heshan0131: Yes,  this is what I had to do here to have polygon tessellation triggered by `updateTriggers.getElevation`. This bug affects both the Polygon layer and the GeojsonLayer. The frustration here is `changeFlags.updateTriggersChanged` only returns true or false, so there is no way to tell which accessor has changed. If there is way to tell whether updateTriggers.getElevation has changed in one line, it would be mostly ideal.

examples/layer-browser/src/examples/core-layers.js
heshan0131: I agree this is hacky at a glance. At the moment, our layer browser doesn't provide any mechanism to allow tests on `updateTriggers`. It really increases the risks that `updateTriggers` are not handled correctly in layers. Though I agree with you this probably should not be included in this PR, will remove it and leave it for further discussion

src/core-layers/solid-polygon-layer/solid-polygon-layer.js
Pessimistress: I don't think this is the right approach.
With the current polygon layer setup, `calculatePositions` is already being called when `getElevation` update trigger changes, where we call `this.state.polygonTesselator.positions()`. What needs to happen is that we move the position calculation from the tesselator's constructor into this method.

ibgreen: The tesselator is also a bit tricky as all other attributes depend on the positions. They need to be precalculated for the other attribute calculators to work. The model doesn't integrate smoothly with our normal attribute manager model. 

I've made a couple of attempts to create a generic Tesselator/GeometryManager for layers with "custom geometries" to handle this (and a couple of other related issues) in a unified way but haven't had time to get it ready, unfortunately.

I guess we'll need to make a temporary fix here for now, proceed as makes sense. I'll approve the PR.

Pessimistress: We never use the `positions` array in other attribute calculation. Each attribute is calculated independently from the source data and accessors.

ibgreen: To clarify - in a custom geometry the number of vertices per object is not constant. This is a calculation that depends on the "position" accessor (perhaps getPolygon or getPath) that affects all other attributes.

I tend to think of the number of verts that gets discovered during "position" calculation, an example is the PolygonLayer where we use something like the earcut library, although I think we are make some short cuts there.

In general one will need to either be able to calculate the number of verts in every attribute updater, or build an intermediary data structure to hold verts and indices and make it available to attribute updates. This is one of the problems I was tackling.

Pessimistress: The accessor that will change all attributes is `getPolygon`, whose update trigger we are not tracking, and that should be a bug.
The specific issue that this PR is trying to fix is regarding `getElevation`, which only affects `positions`, and should not need to reconstruct the tesselator.

ibgreen: Of course. The main thing that concerns me is the amount of logic this is adding to the layers.  I was trying to explain a solution I have been trying to build on the side. Have already approved this PR.

Issue #1049, opened 2017-10-17
Add new project64 function that removes raw fp64 math from layer shaders


1chandu:
This is to track and resume the work partially done in #1024, which is closed without merging. We will use this issue to track resuming/completing this work.

ibgreen:
Opened alternative PR #1050

1chandu:
Resolved by #1050.

PR #1050, opened 2017-10-17
experimental `project64utils`

ibgreen:
Alternative approach to #1024 that puts this code into a separate module while work on the shader module master plan continues.


ibgreen:
See #1049

PR #1051, opened 2017-10-18
More rigorous changeFlags handling.

ibgreen:
This PR is an initial effort towards making our layer life cycle more rigorous, this one focuses on change flag handling

* changeFlags handling now goes through `layer.setChangeFlags` and `layer.clearChangeFlags`
    * The idea here is that this will enable us to decouple setting of changeFlags and actual updates, allowing changeFlags to accumulate but postponing the update until we need to do it. Think layers with `visible: false`
* `LayerManager` is now no longer responsible for sending changeFlags to layer.updateLayer. It sets the changeFlags and then calls updateLayer. This simplifies logic in `LayerManager` which is where simplification really counts.
* Drops `Layer.invalidateAttribute` which added nothing and just created one more method for layer writers to be confused by.

### DIFF
--- a/src/core/lib/composite-layer.js
+++ b/src/core/lib/composite-layer.js
@@ -35,11 +35,6 @@ export default class CompositeLayer extends Layer {
   initializeState() {
   }
 
-  // No-op for the invalidateAttribute function as the composite
-  // layer has no AttributeManager
-  invalidateAttribute() {
-  }
-
   // called to augment the info object that is bubbled up from a sublayer
   // override Layer.getPickingInfo() because decoding / setting uniform do
   // not apply to a composite layer.
@@ -71,11 +66,19 @@ export default class CompositeLayer extends Layer {
     };
   }
 
+  // Called by layer manager to render sublayers
   _renderLayers(updateParams) {
-    if (this.state.oldSubLayers && !this.shouldUpdateState(updateParams)) {
-      log.log(2, `Composite layer reused sublayers ${this}`, this.state.oldSubLayers);
-      return this.state.oldSubLayers;
-    }
+    // TODO - won't updateLayer also be called? Avoid "double diffing"
+    // const {oldProps, props} = updateParams;
+    // this.diffProps(oldProps, props);
+
+    // TODO - Bug: Composite layers need to call the shouldUpdateState method
+    // of all its cached sublayers to ensure they all get properly updated
+    // it is not a "reuse all or none"...
+    // if (this.state.oldSubLayers && !this.shouldUpdateState(updateParams)) {
+    //   log.log(2, `Composite layer reused sublayers ${this}`, this.state.oldSubLayers);
+    //   return this.state.oldSubLayers;
+    // }
     const subLayers = this.renderLayers();
     this.state.oldSubLayers = subLayers;
     log.log(2, `Composite layer rendered new sublayers ${this}`, this.state.oldSubLayers);
--- a/src/core/lib/composite-layer.js
+++ b/src/core/lib/composite-layer.js
@@ -69,16 +69,13 @@ export default class CompositeLayer extends Layer {
   // Called by layer manager to render sublayers
   _renderLayers(updateParams) {
     // TODO - won't updateLayer also be called? Avoid "double diffing"
-    // const {oldProps, props} = updateParams;
-    // this.diffProps(oldProps, props);
+    const {oldProps, props} = updateParams;
+    this.diffProps(oldProps, props);
 
-    // TODO - Bug: Composite layers need to call the shouldUpdateState method
-    // of all its cached sublayers to ensure they all get properly updated
-    // it is not a "reuse all or none"...
-    // if (this.state.oldSubLayers && !this.shouldUpdateState(updateParams)) {
-    //   log.log(2, `Composite layer reused sublayers ${this}`, this.state.oldSubLayers);
-    //   return this.state.oldSubLayers;
-    // }
+    if (this.state.oldSubLayers && !this.shouldUpdateState(updateParams)) {
+      log.log(2, `Composite layer reused sublayers ${this}`, this.state.oldSubLayers);
+      return this.state.oldSubLayers;
+    }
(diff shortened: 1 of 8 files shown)


src/core/lib/layer.js
```diff
@@ -480,10 +487,67 @@ export default class Layer {
...
+      changeFlags.dataChanged = flags.dataChanged;
+      changeFlags.propsOrDataChanged = flags.dataChanged;
+      changeFlags.somethingChanged = flags.dataChanged;
+      log.log(LOG_PRIORITY_UPDATE,
+        `dataChanged: ${flags.dataChanged} in ${this.id}`);
+    }
+    if (flags.updateTriggersChanged && !changeFlags.dataChanged) {
+      changeFlags.updateTriggersChanged = flags.updateTriggersChanged;
+      changeFlags.propsOrDataChanged = flags.updateTriggersChanged;
+      changeFlags.somethingChanged = flags.updateTriggersChanged;
```
Pessimistress: This will potentially override `propsOrDataChanged` and `somethingChanged`. Depending on the sources and order that this function is called, the output is somewhat unpredictable. (e.g. sometimes `'data changed'`, sometimes `true`)
Will it be better to assign these two fields outside of the if statements from the final `changeFlags` state?

src/core/lib/layer.js
```diff
@@ -388,17 +381,29 @@ export default class Layer {
...
-  // Called by layer manager when existing layer is getting new props
-  updateLayer(updateParams) {
-    // Check for deprecated method
-    if (this.shouldUpdate) {
-      log.deprecated('shouldUpdate', 'shouldUpdateState');
-    }
+  // Called by layer manager
+  // if this layer is new (not matched with an existing layer) oldProps will be empty object
+  updateLayer({oldProps = {}, oldContext = {}}) {
+    this.oldProps = oldProps;
```
Pessimistress: Is this used anywhere? Shouldn't layer manager be the only source of truth?

ibgreen: > Is this used anywhere? Shouldn't layer manager be the only source of truth?

Assuming you are talking about `oldProps`, I believe I understand your point. 

This is my thinking: if we want to enable a delayed call to `updateState` which is part of what this PR paves the way for, we will need to be able to serve `oldProps` to the function at that time, when the layer manager may no longer be involved.

My intention is that oldProps should be set on the layer by the layer manager during layer matching and then it would never change again. Then the layer manager should not need to keep sending it in.  I am wary of doing to much in one PR though.

src/core/lib/layer.js
ibgreen: > This will potentially override propsOrDataChanged and somethingChanged. Depending on the sources and order that this function is called, the output is somewhat unpredictable. (e.g. sometimes 'data changed', sometimes true)

My direction here (this is a set of linked PRs in the interest of keeping things in reviewable pieces) is to drop `reason` completely and always have the actual flags be strings or `false`. It simplifies the code across the board assuming all change detecting code follows the convention, and as far as I can tell there is no value gained in keeping separate track of a boolean and a string.

> Will it be better to assign these two fields outside of the if statements from the final changeFlags state?

Well the whole point of this PR is to disconnect changeFlags from updateState. `setChangeFlags` can now be called repeatedly and must be able to gradually flip on more flags until an update clears it. Since this function the only interface to modifying changeFlags makes it natural to manage them here, why burden the caller with that?

src/core/lib/layer.js
```diff
@@ -480,10 +487,67 @@ export default class Layer {
...
+      changeFlags.updateTriggersChanged = flags.updateTriggersChanged;
+      changeFlags.propsOrDataChanged = flags.updateTriggersChanged;
+      changeFlags.somethingChanged = flags.updateTriggersChanged;
+      log.log(LOG_PRIORITY_UPDATE,
+        `updateTriggersChanged: ${flags.updateTriggersChanged} in ${this.id}`);
+    }
+    if (flags.propsChanged && !changeFlags.propsChanged) {
+      changeFlags.propsChanged = true;
+      changeFlags.propsOrDataChanged = true;
+      changeFlags.somethingChanged = true;
```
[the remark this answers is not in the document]
ibgreen: > Use flags.propsChanged for reason?

Yes, good catch

src/core/lib/composite-layer.js
```diff
@@ -71,11 +66,19 @@ export default class CompositeLayer extends Layer {
...
-      log.log(2, `Composite layer reused sublayers ${this}`, this.state.oldSubLayers);
-      return this.state.oldSubLayers;
-    }
+    // TODO - won't updateLayer also be called? Avoid "double diffing"
+    // const {oldProps, props} = updateParams;
+    // this.diffProps(oldProps, props);
+
+    // TODO - Bug: Composite layers need to call the shouldUpdateState method
+    // of all its cached sublayers to ensure they all get properly updated
+    // it is not a "reuse all or none"...
```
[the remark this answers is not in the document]
ibgreen: The problem I think I am seeing is  that we are calling `this.shouldUpdateState` rather than `subLayer.shouldUpdateState`. How can the composite layer decide if the rendered layer needs to update?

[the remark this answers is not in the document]
Pessimistress: `this.shouldUpdateState` only determines if a new set of sub layers will be generated. LayerManager will call `shoudlUpdateState` on each sublayer.

src/core/lib/layer.js
Pessimistress: The flags should be flipped in `setChangedFlags`, but outside of the `if` statements.
Of course, if we drop `reason`, then it doesn't matter. Though I do believe that each PR should work without assuming a future feature change.

src/core/lib/composite-layer.js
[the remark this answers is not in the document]
ibgreen: OK I will drop this change. I am quite convinced there are very serious bugs in our composite layer handling, but makes sense to address them in a separate PR.

ibgreen:
Merging per discussion

PR #1055, opened 2017-10-18
temporary fix for orthographic-viewport dependent apps

gnavvy:
This is a temporary fix to bring OrthographicViewport-dependent apps to a good state.
Later the OrthographicViewport will extend the ThirdPersonViewport instead of the base Viewport once all functions are fully tested in new ThirdPersonViewport.

To test:
Use either the `svg-interoperability` or the `graph` example in the /examples folder.
Before: nothing/partially visible in the viewport.
After: visual elements are properly rendered.
(There's a bug in the graph example that the hovered info is y-flipped, will fix in a later PR)

### DIFF
--- a/examples/point-cloud-ply/app.js
+++ b/examples/point-cloud-ply/app.js
@@ -14,7 +14,6 @@ class Example extends PureComponent {
     super(props);
 
     this._onResize = this._onResize.bind(this);
-    this._onInitialized = this._onInitialized.bind(this);
     this._onViewportChange = this._onViewportChange.bind(this);
     this._onUpdate = this._onUpdate.bind(this);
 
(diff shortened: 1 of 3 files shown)


gnavvy:
will do. (did not find an example in deck.gl that uses the PerspectiveViewport for testing, will fix first and create a simple example later)

examples/point-cloud-ply/app.js
```diff
@@ -14,7 +14,6 @@ class Example extends PureComponent {
     super(props);
 
     this._onResize = this._onResize.bind(this);
-    this._onInitialized = this._onInitialized.bind(this);
```
gnavvy: bug: _onInitialized is no longer used after switching to the new orbit-controller

PR #1057, opened 2017-10-18
refactoring on ortho- and perspective-viewports related files

gnavvy:
refactoring on ortho- and perspective-viewports related files, exports, and docs.
tested w/ the svg-interoperability, graph and the GeoJSON examples.
(layer-browser is broken due to other issues)


docs/api-reference/orthographic-viewport.md
```diff
@@ -1,3 +1,48 @@
-# OrthographicViewport Class (Deprecated)
+# OrthographicViewport Class
```
gnavvy: no change in content for the documents, just moved them back from deprecation.
will rewrite them once the APIs are fixed.

src/index.js
```diff
@@ -32,10 +32,16 @@ export {default as Layer} from './core/lib/layer';
...
+export {
+  Viewport,
+  FirstPersonViewport,
+  ThirdPersonViewport,
+  WebMercatorViewport,
+  PerspectiveViewport,
+  OrthographicViewport,
+  // TODO: orbit-viewport to be merged w/ third-person-viewport
+  OrbitViewport
+} from './core';
```
gnavvy: Seems we are exporting using explicit path in both /index.js and /core/index.js.
I can do around of cleanup if no objection.

ibgreen:
I think the docs would greatly benefit from a chart or table that describes how the various viewports are related and when each one should be used. 

We are offering quite a few Viewports now, adding these back might make it easier for some users but for others we have just added more choices to think about.

PR #1058, opened 2017-10-18
Apply default values for transition props

1chandu:
Default props of transition params were not always applied, make them part of `ViewportController` `defaultProps`.

Verified with : 
test-browser
layer-browser ( interact with map to trigger viewport change, default values get applied here)
all 3 viewport transition examples.

### DIFF
--- a/src/react/experimental/transition-manager.js
+++ b/src/react/experimental/transition-manager.js
@@ -28,7 +28,7 @@ const DEFAULT_PROPS = {
 export default class TransitionManager {
   constructor(props, onTransitionUpdate) {
 
-    this.props = Object.assign({}, DEFAULT_PROPS, props);
+    this.props = Object.assign({}, props);
     this.onTransitionUpdate = onTransitionUpdate;
     this.transitionContext = {
       time: 0,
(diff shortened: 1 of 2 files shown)


src/react/experimental/transition-manager.js
```diff
@@ -15,7 +15,7 @@ export const TRANSITION_EVENTS = {
   IGNORE: 3
 };
 
-const DEFAULT_PROPS = {
+export const DEFAULT_TRANSITION_PROPS = {
```
ibgreen: Alternatively you could assign `TransitionManager.defaultProps = defaultProps` at the end.

src/react/viewport-controller.js
```diff
@@ -90,7 +90,7 @@ const propTypes = {
 
 const getDefaultCursor = ({isDragging}) => isDragging ? CURSOR.GRABBING : CURSOR.GRAB;
 
-const defaultProps = {
+const defaultProps = Object.assign({}, DEFAULT_TRANSITION_PROPS, {
```
ibgreen: Would be `const defaultProps = Object.assign({}, TransitionManager.defaultProps, {`

src/react/experimental/transition-manager.js
1chandu: Agree, avoids extra import and looks more clean, updated with suggested change.

Issue #976, opened 2017-10-03
Move TextLayer into core layers in deck.gl


shaojingli:
#### To Do List 
  - [x] Add label and assign to milestone 
  - [x] Coding
  - [x] Doc update (check if code and text-layer.md are in sync @gnavvy by 03/09)
  - [x] Whats new update 
  - [x] Testing (move from exp to core folder and then test on rendering-test)

ibgreen:
Just as a clarification: This was moved to experimental-layers as a staging ground for audit and further improvements. It is not yet part of core-layers.

rivulet-zhang:
Thanks @ibgreen It is better to reopen it to track the status and we can safely close it once the text layer is moved to the core layers from the experimental layers. (The text layer has been moved from the example folder to the experimental layers in this PR #1062)

rivulet-zhang:
Just to clarify, the text layer implementation in the `example` folder (`text-layer`, `tagmap`) is still the old version. The new version is only located in the `experimental-layers` for review. Once the new version is reviewed and moved to the core layers, we can just safely delete the old version in the `example` folder.

1chandu:
@gnavvy 
Teams feedback: This needs to go through teams review and then depending on time we should make into 4.2 or 5.0 release.

ibgreen:
We have started using the `TextLayer` in apps and worked through the initial set of minor bugs, it looks promising. At the point it is mainly an API Audit that is required.

gnavvy:
Will do in the following steps:
- [ ] 1. Move Text-Layer code and doc from experimental-layers to core-layers
- [ ] 2. Publish patch release of deck.gl to have Text-layer exported from deck.gl instead of deck.gl-layers
- [ ] 3. Move Text-layer related examples from experimental to core, update paths accordingly
- [ ] 4. API Audit
- [ ] 5. Publish document on the v5.2 website

PR #1062, opened 2017-10-19
Move text layer to experimental layers

rivulet-zhang:
A few changes to the original text layer in the example folder:

- New features added: including `getTextAnchor`, `getAlignmentBaseline`, `getPixelOffset`. (Details in the doc)
- Dynamically generate font atlas once the layer is initialized. No need to read static json and bitmap files. This change avoids the headache of adding additional webpack loaders or plugins to deal with those static assets.
- Use `project_pixel_to_clipspace` in the shader to deal with conversion between pixels and clipspace coordinates. (Will create another PR to make this change to the icon layer)
- An example added in the layer-browser. Run `yarn start-local` to see the result.

Once this is officially added to the core layers, we can remove the duplicated code in the text-layer and tagmap examples.

### DIFF
--- a/src/experimental-layers/src/text-layer/text-layer.js
+++ b/src/experimental-layers/src/text-layer/text-layer.js
@@ -1,0 +1,155 @@
+// Copyright (c) 2015 - 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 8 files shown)


rivulet-zhang:
More comments:
- Right now the font family is predefined and is invisible to the user, but we can definitely change to allow users to customize it if needed.
- The `font-atlas.js` is basically a duplicate of `label-utils.js` in the label-layer. The only change is to add some padding to the letter to improve the quality of the font texture. May make sense to have a generic "helper" function somewhere in the repo to handle the generation of text-related texture.

src/experimental-layers/src/text-layer/text-layer.js
```diff
@@ -0,0 +1,155 @@
...
+      getSize,
+      getAngle,
+      getTextAnchor,
+      getAlignmentBaseline,
+      getPixelOffset,
+      fp64
+    } = this.props;
+
+    return [
+      new MultiIconLayer(
```
shaojingli: is it the plan that we remove `MultiIconLayer` altogether and merge its content to the text layer? since its usage is pretty much limited to render texts?

rivulet-zhang: Discussed this with @ibgreen. We think to keep it separate for now in case there are other use cases for multiple icons. Or we can also move it to somewhere that contains a collection of "helper" classes/functions for deck.gl (but not in the 4.2 release).

shaojingli: okay

ibgreen: We have a growing number of these internal layers in deck.gl (MultiIconLayer, SolidPolygonLayer, HexagonCellLayer etc). Not sure how to handle them, what directory to put them in etc. Would be a good API Audit topic.

rivulet-zhang:
#976

PR #1063, opened 2017-10-19
Transition manager improvements

Pessimistress:
Consolidating changes and feedback from https://github.com/uber/react-map-gl/pull/383

- Fix incorrect timer, replace `setInterval` with `requestAnimationFrame`
- Always use the transition settings from when the transition is triggered (removes the burden from app: set and forget)
- Use `onViewportChange` for transition update callback
- Do not trigger transition on viewport size change
- Do not compare viewports if there's no transition
- Remove nested function definitions in utils

### DIFF
--- a/src/react/experimental/viewport-transition-utils.js
+++ b/src/react/experimental/viewport-transition-utils.js
@@ -4,8 +4,28 @@ import {projectFlat, unprojectFlat} from 'viewport-mercator-project';
 import {Vector2} from 'math.gl';
 
 const EPSILON = 0.01;
-export const VIEWPORT_PROPS = ['longitude', 'latitude', 'zoom', 'bearing', 'pitch',
+const VIEWPORT_PROPS = ['longitude', 'latitude', 'zoom', 'bearing', 'pitch',
   'position', 'width', 'height'];
+const VIEWPORT_INTERPOLATION_PROPS =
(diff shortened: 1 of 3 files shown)


src/react/experimental/viewport-transition-utils.js
```diff
@@ -17,6 +37,23 @@ export function extractViewportFrom(props) {
   return viewport;
 }
 
+/* eslint-disable max-depth */
+export function areViewportsEqual(startViewport, endViewport) {
+  for (const p of VIEWPORT_INTERPOLATION_PROPS) {
```
1chandu: This will skip viewport updates when only non interpolation props (width or height) changed. `VIEWPORT_INTERPOLATION_PROPS` should be replaced with `VIEWPORT_PROPS`.

Pessimistress: This is intentional We do not want to trigger transition when viewport size changes.

1chandu: Yes, but we we ignoring width/height changes during transitions and map is not resized.

Pessimistress: Addressed in https://github.com/uber/deck.gl/pull/1063/commits/a6b4600ec7fcccd45639b9b51f3fabe91255435c

Issue #1043, opened 2017-10-17
Viewport does not update without changing `props.layers`


Pessimistress:
`LayerManager.updateLayers` does not update layer context if `layers` array remains the same:
https://github.com/uber/deck.gl/blob/master/src/core/lib/layer-manager.js#L206

Then in `drawLayersInViewport`, instead of the provided viewport, the viewport from layer context is used:
https://github.com/uber/deck.gl/blob/master/src/core/lib/draw-layers.js#L183

This issue does not surface in our React examples, as we reconstruct the layers whenever the viewport updates. You may reproduce the problem by opening this document in the browser, and then commenting out this line:
https://github.com/Pessimistress/deck.gl-runkit/blob/382b5ded83d2b493ecb96dcde9f74d66cc1bdadc/src/template.html#L38

@ibgreen @1chandu

1chandu:
Need to be verified.

Pessimistress:
Confirmed: fixed in alpha.22

PR #1072, opened 2017-10-24
Fix layer context update

### LINKED ISSUE #1069: Layer's context is not updated every time hence missing updates from layer-manager
- Every layer shares the same context that is owned by layer-manger.
- When viewport transitions are enabled, some how this link is broken hence any viewport updates done on layer-manger's context are not reflected in context object referenced by layer objects.
- I am seeing this issue only when I enable viewport transitions, it is exposing an issue that is otherwise hidden.

@ibgreen @Pessimistress @shaojingli

1chandu:
More details : #1069 
Fixes #1069
Related to #1043

Verified with `first-person-map` and `layer-browser` examples.

### DIFF
--- a/src/core/lib/layer-manager.js
+++ b/src/core/lib/layer-manager.js
@@ -92,7 +92,6 @@ export default class LayerManager {
     seerInitListener(this._initSeer);
     layerEditListener(this._editSeer);
 
-    Object.seal(this.context);
     Object.seal(this);
   }
 
@@ -143,7 +142,7 @@ export default class LayerManager {
       this.updateLayers({newLayers: parameters.layers});
     }
 
-    this.context = Object.assign({}, this.context, parameters);
+    Object.assign(this.context, parameters);
   }
 
   setViewports(viewports) {


src/core/lib/layer-manager.js
```diff
@@ -202,6 +204,17 @@ export default class LayerManager {
...
+    // Update context object if it is changed.
+    if (this._isContextDirty) {
+      for (const layer of newLayers) {
+        layer.context = this.context;
+      }
+      this._isContextDirty = false;
+    }
+
     // TODO - something is generating state updates that cause rerender of the same
     if (newLayers === this.lastRenderedLayers) {
```
Pessimistress: This will never be true since `filter` constructs a new array.

1chandu: Good catch, fixed.

src/core/lib/layer-manager.js
```diff
@@ -203,32 +205,38 @@ export default class LayerManager {
...
-      throw error;
+      this.layers = generatedLayers;
+      // Throw first error found, if any
+      if (error) {
+        throw error;
+      }
     }
+
+    // Update layers context
+    for (const layer of this.layers) {
```
1chandu: Only change here is moving this for loop out as it applies to both cases (newLayers and context updated), and looping over this.layers instead of newLayers, as they shouldn't differ.

ibgreen: I think `context` needs to be set during the _updateLayers cycle. That cycle calls layer life cycle callbacks and context must be set by then. So moving this out of the loop will at minimum do duplicate work.

1chandu:
I had put some details in #1069, copy pasting here : 
Every time DeckGL receives new props, it calls `setProps()` which in-turn calls #1. `layerManager.setParameters()` and then #2. `layerManager.updateLayers()`, if all this triggered with-in the ViewportController level (i.e. Application render is not called and layers have not changed), #1. creates a new context object in LayerManger (https://github.com/uber/deck.gl/blob/a85c7b85a8bfc3718f0fc023a1960627131f4659/src/core/lib/layer-manager.js#L146) , but #2 doesn't update this new object on the layers. Any further viewport updates are done on LayerManger.context are not reflected on layer.context as they now refer to different objects.

Alternate solution for this could be not seal the 'context' object and replace this line https://github.com/uber/deck.gl/blob/a85c7b85a8bfc3718f0fc023a1960627131f4659/src/core/lib/layer-manager.js#L146 with
`Object.assign(this.context, parameters)`

ibgreen:
@1chandu - Thanks for that explanation, yes it does seem that just updating context in place would solve the problem.

The one complication I see is that the `updateState` function was designed to take an `oldContext` parameter to let the app look for context changes, **however** it doesn't seem to be correctly supported, and it is an edge case, doubt that any layers depend on it.

So, I think we can make the change you suggested in the last comment. And just open a ticket reminding us that we need to sort out the `oldContext` story. Either we delete it or we figure out how to properly implement it.

ibgreen:
On a side note:
* I think the `LayerManager` should no longer expose any methods except `constructor`, `setProps` and `finalize`. This would likely make handling of oldContext easier.
* And we should consider whether we need both the `DeckGLJS` and the `LayerManager` or whether just one of them would do.

1chandu:
>The one complication I see is that the updateState function was designed to take an oldContext parameter to let the app look for context changes, however it doesn't seem to be correctly supported, and it is an edge case, doubt that any layers depend on it.

Opened https://github.com/uber/deck.gl/issues/1077 to track cleaning up oldContext

Agree that oldContext can be removed if not being used by anyone, but change in this PR should not affect it. All its doing is copying context into oldContext before it updates viewport details, so it should still work as is.

Issue #1012, opened 2017-10-10
projectionMode and coordinateSystem perform differently for COORDINATE_SYSTEM.IDENTITY


gnavvy:
According to the Upgrading from deck.gl v4.1 to v4.2, the old prop projectionMode will be deprecated and replaced with the new coordinateSystem. However, the two props behave differently, at least for the COORDINATE_SYSTEM.IDENTITY.

To reproduce, use the svg-interoperability example, switch the prop in line 118 between projectionMode and coordinateSystem, and check the visual result.

4.1.x using projectionMode: COORDINATE_SYSTEM.IDENTITY
![screen shot 2017-10-10 at 11 02 51 am](https://user-images.githubusercontent.com/144743/31402502-015cc3e2-adab-11e7-8c1f-791597669bf9.png)
^ Expected result.

4.1.x using coordinateSystem: COORDINATE_SYSTEM.IDENTITY
<img width="886" alt="screen shot 2017-10-10 at 11 03 10 am" src="https://user-images.githubusercontent.com/144743/31402440-d3249824-adaa-11e7-8e1c-21915728e584.png">
(Not only there's an offset, but also the projection is off - polar being distorted)

4.2.x using either projectionMode/coordinateSystem: COORDINATE_SYSTEM.IDENTITY
<img width="546" alt="screen shot 2017-10-10 at 11 03 47 am" src="https://user-images.githubusercontent.com/144743/31402442-d4bf0bce-adaa-11e7-9457-bd5498aede55.png">

Users using v4.1.x with the old API projectionMode should be not affected, but it is breaking for v4.2.x.

jianhuang01:
@gnavvy Would you please paste the code change for line 118? I don't know how to switch between projectionMode and coordinateSystem. Thanks!

ibgreen:
@gnavvy Thanks for bringing this up, this should be sorted asap.

The motivation for this change is that for geospatial projections, the "overlaid" meter coordinate system's positive y axis should point to north, and I found that it was possible to unify some of the geospatial matrix creation logic by 

This is my current assessment:
* For `COORDINATE_SYSTEM.IDENTITY`, we should clearly not be flipping Y coordinates. Identity means identity.
* For `COORDINATE_SYSTEM.METER_OFFSETS` backwards compatibility we should check if the layer is using `projectionMode` instead of `coordinateSystem` and in that case disable the flip. 

Both code and documentation and perhaps a test case should be added.

gnavvy:
[...]

PR #1073, opened 2017-10-24
Fix OrthograhicViewport bug and examples

jianhuang01:
Per issue #1012, found that the OrthographicViewport's top and bottom are inverted. Here change them back and fixed examples.

Tested with graph & svn-interoperability examples.


jianhuang01:
abandon this review cause didn't choose the right way to resolve conflicts.

PR #1078, opened 2017-10-26
Solidify viewport descriptor (aka `View`) support

ibgreen:
This is a proposal, for exposition.

The basic idea is that deck will take a list of "view descriptors" in preference to viewports. 
* This will allows deck to auto calculate the actual viewport sizes without passing in exact width and height to every viewport.
* Percentage widths and heights  (and x, y) can also be supported for viewports, significantly simplifying complex viewport layouts in apps.
* Going forward, we can add new parameters to each view, such as layer filters etc, while keeping our Viewport classes laser focused on projection related tasks.
* Also the view descriptors can take `viewState` objects, closing the loop on that line of development.

### DIFF
--- a/src/core/lib/layer-manager.js
+++ b/src/core/lib/layer-manager.js
@@ -462,11 +462,6 @@ export default class LayerManager {
     return this;
   }
 
-  // Get a viewport from a view descriptor (which can be a plain viewport)
-  _getViewportFromDescriptor(viewportOrDescriptor) {
-    return viewportOrDescriptor.viewport ? viewportOrDescriptor.viewport : viewportOrDescriptor;
-  }
-
   _getPickingBuffer() {
     const {gl} = this.context;
     // Create a frame buffer if not already available
--- a/src/core/lib/layer-manager.js
+++ b/src/core/lib/layer-manager.js
@@ -58,26 +58,23 @@ export default class LayerManager {
 
     this.prevLayers = [];
     this.layers = [];
-    this.oldContext = {};
+    // List of view descriptors, gets re-evaluated when width/height changes
+    this.viewDescriptors = [];
+    this.viewDescriptorsChanged = true;
+    this.width = 100;
+    this.height = 100;
+
+    // Generated viewports
+    this.viewports = [];
+
     this.screenCleared = false;
     this._needsRedraw = 'Initial render';
 
-    this._eventManager = null;
-    this._pickingRadius = 0;
-    this._onLayerClick = null;
-    this._onLayerHover = null;
-    this._onClick = this._onClick.bind(this);
-    this._onPointerMove = this._onPointerMove.bind(this);
-    this._onPointerLeave = this._onPointerLeave.bind(this);
-    this._pickAndCallback = this._pickAndCallback.bind(this);
-
-    this._initSeer = this._initSeer.bind(this);
-    this._editSeer = this._editSeer.bind(this);
-
+    // TODO - everything in this object is exposed to the layers
+    // We should only have information here that is API stable
     this.context = {
       gl,
       uniforms: {},
-      viewports: [],
       viewport: null,
       viewportChanged: true,
       pickingFBO: null,
@@ -88,7 +85,21 @@ export default class LayerManager {
       },
       shaderCache: new ShaderCache({gl})
     };
+    this.oldContext = {};
 
+    // Event handling
+    this._eventManager = null;
+    this._pickingRadius = 0;
+    this._onLayerClick = null;
+    this._onLayerHover = null;
+    this._onClick = this._onClick.bind(this);
+    this._onPointerMove = this._onPointerMove.bind(this);
+    this._onPointerLeave = this._onPointerLeave.bind(this);
+    this._pickAndCallback = this._pickAndCallback.bind(this);
+
+    // Seer integration
+    this._initSeer = this._initSeer.bind(this);
+    this._editSeer = this._editSeer.bind(this);
     seerInitListener(this._initSeer);
     layerEditListener(this._editSeer);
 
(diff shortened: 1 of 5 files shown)


src/core/lib/layer-manager.js
```diff
@@ -145,23 +171,21 @@ export default class LayerManager {
...
+
+    this.viewDescriptorsChanged =
+      this.viewDescriptorsChanged ||
+      this._diffViews(viewDescriptors, this.viewDescriptors);
+
+    // Try to not actually rebuild the viewports until `getViewports` is called
+    if (this.viewDescriptorsChanged) {
+      this.viewDescriptors = viewDescriptors;
+      this._rebuildViewportsFromViews({viewDescriptors: this.viewDescriptors});
+      this.viewDescriptorsChanged = false;
```
1chandu: > this.viewDescriptorsChanged = false;

Can be removed as it is done inside `_rebuildViewportsFromViews()`

src/core/lib/layer-manager.js
```diff
@@ -145,23 +171,21 @@ export default class LayerManager {
...
-      assert(viewport instanceof Viewport, 'Invalid viewport');
-
-      this.context.viewports = viewports;
-      this._activateViewport(viewport);
+    // Ensure viewports are wrapped in descriptors
+    const viewDescriptors = flatten(viewports, {filter: Boolean})
+      .map(viewport => viewport instanceof Viewport ? {viewport} : viewport);
+
+    this.viewDescriptorsChanged =
+      this.viewDescriptorsChanged ||
```
1chandu: >this.viewDescriptorsChanged = this.viewDescriptorsChanged || ...

`this.viewDescriptorsChanged` is will be false here always, can be removed from this equation.

src/core/lib/layer-manager.js
```diff
@@ -318,19 +350,87 @@ export default class LayerManager {
...
+    if (newViews.length !== oldViews.length) {
+      return true;
+    }
+
+    return newViews.some(
+      (_, i) => this._diffView(newViews[i], oldViews[i])
+    );
+  }
+
+  // TODO: this will be true always if descriptors are used as we create new object
```
1chandu: > // TODO: this will be true always if descriptors are used as we create new object

Remove.

src/core/lib/layer-manager.js
```diff
@@ -318,19 +350,87 @@ export default class LayerManager {
...
+  _diffView(newView, oldView) {
+    // `View` hiearchy supports an `equals` method
+    if (newView.viewport) {
+      return !oldView.viewport || !newView.viewport.equals(oldView.viewport);
+    }
+    return newView !== oldView;
```
1chandu: > return newView !== oldView;

This is comparing viewDescriptor objects, we should do deep equal on these objects?

src/core/lib/layer-manager.js
```diff
@@ -318,19 +350,87 @@ export default class LayerManager {
...
+  }
+
+  // Support for relative viewport dimensions (e.g {y: '50%', height: '50%'})
+  _getViewDimensions({viewDescriptor, width, height}) {
+    const parsePercent = (value, max) => value;
```
1chandu: Is this intentionally left incomplete ?, Add a TODO for whats remaining and re-enable the code.

src/core/lib/layer-manager.js
```diff
@@ -145,23 +171,21 @@ export default class LayerManager {
...
-      this._activateViewport(viewport);
+    // Ensure viewports are wrapped in descriptors
+    const viewDescriptors = flatten(viewports, {filter: Boolean})
+      .map(viewport => viewport instanceof Viewport ? {viewport} : viewport);
+
+    this.viewDescriptorsChanged =
+      this.viewDescriptorsChanged ||
+      this._diffViews(viewDescriptors, this.viewDescriptors);
+
+    // Try to not actually rebuild the viewports until `getViewports` is called
```
1chandu: > // Try to not actually rebuild the viewports until `getViewports` is called

Is this comment valid, we are calling `_rebuildViewportsFromViews()` here.

src/core/lib/layer-manager.js
```diff
@@ -318,19 +350,87 @@ export default class LayerManager {
...
+    // We've just rebuilt the viewports to match the descriptors, so clear the flag
+    this.viewports = newViewports;
+    this.viewDescriptorsChanged = false;
   }
 
-  //
-  // PRIVATE METHODS
-  //
+  // Build a `Viewport` from a view descriptor
+  _makeViewportFromViewDescriptor({viewDescriptor, width, height}) {
```
1chandu: `width` and `height` never used in this method?

src/core/lib/layer-manager.js
```diff
@@ -114,8 +134,14 @@ export default class LayerManager {
       this.layers;
   }
 
-  getViewports() {
-    return this.context.viewports;
+  // Get a set of viewports for a given width and height
+  getViewports({width, height} = {}) {
+    if (width !== this.width || height !== this.height || this.viewDescriptorsChanged) {
```
1chandu: Add an assert on `width` and `height`.

src/core/lib/layer-manager.js
```diff
@@ -237,7 +261,7 @@ export default class LayerManager {
     // render this viewport
     drawLayers(gl, {
       layers: this.layers,
-      viewports: this.context.viewports,
+      viewports: this.getViewports(),
```
1chandu: `getViewports()`, don't we have to pass width/height ? Same for other places we are calling `getViewports()`.

ibgreen:
Will on a new PR on 4.2-dev

PR #1082, opened 2017-10-26
ViewportTransitions: remove recursive children update, fix props update

1chandu:
This includes following two changes :
1. `ViewportController` intercepts and updates all children props to avoid a jump when viewport is updated with transition enabled, this can be achieved by using `shouldComponentUpdate()`, which returns false when a new transition is triggered and hence skipping the render for itself and all its children. All following viewport updates are handled by 'onViewportChange()' called by `TransitionManger`. This should make code a lot less complicated and also makes it easy to integrate Viewport Transitions with Multi-viewports feature.
2. In `TransitionManger`, this.props should be set before calling '_triggerTransition()' which calls `_updateViewport` that uses this.props.

Verified with all 3 viewport-animation examples and layer-manager. #2 fixes an issue in `viewport-animation-3d-heatmap` example, where app uses wrong width/height.

### DIFF
--- a/examples/wip/viewport-animation-geojson/app.js
+++ b/examples/wip/viewport-animation-geojson/app.js
@@ -3,7 +3,7 @@ import React, {Component} from 'react';
 import {render} from 'react-dom';
 import {StaticMap} from 'react-map-gl';
 import DeckGLOverlay from './deckgl-overlay.js';
-import {ViewportController, MapState, experimental} from 'deck.gl';
+import {ViewportController, MapState, TRANSITION_EVENTS} from 'deck.gl';
 import {json as requestJson} from 'd3-request';
 
 // Set your mapbox token here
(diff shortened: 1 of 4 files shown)


src/react/experimental/transition-manager.js
```diff
@@ -45,36 +45,42 @@ export default class TransitionManager {
...
     // NOTE: Be cautious re-ordering statements in this function.
     if (this._shouldIgnoreViewportChange(nextProps)) {
       this.props = nextProps;
-      return;
+      return transitionTriggered;
     }
 
     const isTransitionInProgress = this._isTransitionInProgress();
+    const currentProps = this.props;
+    this.props = nextProps;
```
ibgreen: Why are we setting this here as opposed to at the end? If there is a reason (e.g. other functions using it) that seems subtle and worth a comment (or changing those functions to take a props param).

PR #1083, opened 2017-10-26
Quick fix of fitBounds API in orbit-viewport and examples

jianhuang01:
Current fitBounds API recalculates user's 'lookat' position and distance based on a bounds to make sure canvas size can fit bound size. But changing of 'lookat' position will trigger new behaviors user doesn't expect, like model rotates on a new axis.
The PR here is to keep 'lookat' position untouched unless user changes it, and the bounds is always centered at 'lookat' position.


jianhuang01:
Tested with the two point-cloud examples and plot example

PR #1086, opened 2017-10-27
Add `layerFilter` prop

ibgreen:
A new optional prop that takes (layer, viewport) and allows filtering out layers in specific viewports or from rendering or picking.

Also a bit more functional breakdown in pick-layers.js. No logic changes.

### DIFF
--- a/docs/api-reference/react/deckgl.md
+++ b/docs/api-reference/react/deckgl.md
@@ -62,7 +62,7 @@ The array of deck.gl layers to be rendered. This array is expected to be an arra
 
 ##### `layerFilter`
 
-Optionally takes a function `(layer, viewport) => Boolean` that is called before a layer is rendered. Gives the application an opportunity to filter out layers from the layer list during either rendering or picking. Filtering can be done per viewport or per layer or both. This enables techniques like adding helper layers that work as masks during picking but do not show up during rendering.
+Optionally takes a function `({layer, viewport, isPicking}) => Boolean` that is called before a layer is rendered. Gives the application an opportunity to filter out layers from the layer list during either rendering or picking. Filtering can be done per viewport or per layer or both. This enables techniques like adding helper layers that work as masks during picking but do not show up during rendering.
 
 ##### `viewports`
 
--- a/docs/api-reference/react/deckgl.md
+++ b/docs/api-reference/react/deckgl.md
@@ -56,6 +56,14 @@ Canvas ID to allow style customization in CSS.
 
 The DeckGL component own `canvas` element is rendered last, intentionally on top of all the base maps.
 
+##### `layers` (Array, required)
+
+The array of deck.gl layers to be rendered. This array is expected to be an array of newly allocated instances of your deck.gl layers, created with updated properties derived from the current application state.
+
+##### `layerFilter`
+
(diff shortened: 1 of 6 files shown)


src/core/lib/draw-layers.js
```diff
@@ -165,72 +170,88 @@ function drawLayersInViewport(gl, {
...
-    if (layer.isComposite) {
-      renderStats.compositeCount++;
+
+    // Check if we should draw layer
+    let shouldDrawLayer = layer.props.visible;
+    if (drawPickingColors) {
+      shouldDrawLayer = shouldDrawLayer && layer.props.pickable;
+    }
+    if (shouldDrawLayer && layerFilter) {
+      shouldDrawLayer = layerFilter(layer, viewport);
```
gnavvy: if layerFilter is a function passed down to <DeckGL/>, can we use that function to set the _visible_ and _pickable_ props?

ibgreen: > if layerFilter is a function passed down to , can we use that function to set the visible and pickable props?

Hmm, yes I can see the utility, but how would we do that? Have `layerFilter` return an object with prop overrides?

* Most props cannot be overridden without invoking the layer lifecycle. So those props would be part of a small list of 'special' props that could be overridden in this way?

* We could also add another parameter to `layerFilter` so that it knows if it is picking or rendering?  

If no quick answers, I can make a note and we could take further discussions in the API audit.

docs/api-reference/react/deckgl.md
```diff
@@ -56,6 +56,14 @@ Canvas ID to allow style customization in CSS.
 
 The DeckGL component own `canvas` element is rendered last, intentionally on top of all the base maps.
 
+##### `layers` (Array, required)
+
+The array of deck.gl layers to be rendered. This array is expected to be an array of newly allocated instances of your deck.gl layers, created with updated properties derived from the current application state.
+
+##### `layerFilter`
+
+Optionally takes a function `(layer, viewport) => Boolean` that is called before a layer is rendered. Gives the application an opportunity to filter out layers from the layer list during either rendering or picking. Filtering can be done per viewport or per layer or both. This enables techniques like adding helper layers that work as masks during picking but do not show up during rendering.
```
1chandu: If this can be used to filter a layer differently for rendering and picking, shouldn't it also needs to take parameter that tells if it render flow or picking flow?

src/core/lib/draw-layers.js
```diff
@@ -106,6 +108,7 @@ export function drawPickingBuffer(gl, {
   useDevicePixelRatio,
   pickingFBO,
   deviceRect: {x, y, width, height},
+  layerFilter = (layer, viewport) => true,
```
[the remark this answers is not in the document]
ibgreen: Yes. Changed to null everywhere.

docs/api-reference/react/deckgl.md
ibgreen: Yes makes sense. Added isPicking flag

1chandu:
This PR doesn't say whats been tested, please make sure to test picking/visible props in layer-browser before merging, thanks!

ibgreen:
Did additional testing in layer-browser for pickable/visible props in instanced and composite layers.

PR #1089, opened 2017-10-27
Fix multi-icon-layer default props

ibgreen:


src/experimental-layers/src/text-layer/multi-icon-layer/multi-icon-layer.js
```diff
@@ -25,7 +25,6 @@ import vs from './multi-icon-layer-vertex.glsl';
 import vs64 from './multi-icon-layer-vertex-64.glsl';
 
 const defaultProps = {
-  ...IconLayer.defaultProps,
```
ibgreen: deck.gl uses pure ES6, ... for objects is not supported (but works for arrays).

deck.gl automatically merges defaultProps for derived layers by following the prototype chain.

rivulet-zhang: Thanks for fixing this. I see a few places where we use the spread operator, but they are all for arrays, which makes sense.

PR #1090, opened 2017-10-27
Further simplifications to changeFlags handling

ibgreen:
Also fixes a case where `shouldUpdateState` could get undefined `changeFlags` in composite layers.


1chandu:
@Ib you had chance to test with layer-browser, highlighting is broken for GeoJason and Hexagon layers. It works fine with master before original PR (More rigorous changeFlags handling. (#1051))

ibgreen:
The life cycle refactoring work is being moved to 4.2-dev.

PR #1091, opened 2017-10-27
Use global version from webpack

Apercu:


src/core/lib/init.js
```diff
@@ -21,14 +21,13 @@
 import {global} from '../utils/globals';
 import log from '../utils/log';
 
-// Version detection
-// TODO - this imports a rather large JSON file, we only need one field
-import {version} from '../../../package.json';
-
 const STARTUP_MESSAGE = 'set deck.log.priority=1 (or higher) to trace attribute updates';
+const version = global.DECK_VERSION || 'NODE';
```
ibgreen: Nit: Maybe a comment about why the fallback is needed?

PR #1098, opened 2017-10-27
path+mesh layer: Specify arrow colors

georgios-uber:


src/experimental-layers/src/path-marker-layer/path-marker-layer.js
```diff
@@ -27,8 +27,9 @@ const defaultProps = Object.assign({}, PathOutlineLayer.defaultProps, {
   hightlightIndex: -1,
   highlightPoint: null,
 
-  // getPath: x => x.path,
-  // getColor: x => x.color,
+  getPath: x => x.path,
+  getColor: x => x.color,
+  getDirectionColor: x => [0, 0, 0, 255],
```
ibgreen: If we want to keep this as a general path marker layer, maybe `getMarkerColor`?

PR #1099, opened 2017-10-27
MultiViewport Transitions: Add example, and fix `isMapSynced`.

1chandu:
- Create multi-viewport transition example using existing first-person-map example.
- Add bias to 'isMapSynched' to avoid flashes during transition.

Verified with layer-browser.

### DIFF
--- a/examples/wip/viewport-animation-first-person-map/app.js
+++ b/examples/wip/viewport-animation-first-person-map/app.js
@@ -1,0 +1,346 @@
+/* global window, document, fetch */
+import React, {Component} from 'react';
+import {render} from 'react-dom';
+
+import {
+  COORDINATE_SYSTEM,
+  // Unified controller, together with state that determines interaction model
+  FirstPersonState,
+  // Viewport classes provides various views on the state
+  FirstPersonViewport,
+  ThirdPersonViewport,
+  WebMercatorViewport,
+  PolygonLayer,
+  PointCloudLayer
+} from 'deck.gl';
+
+import TripsLayer from '../../trips/trips-layer';
+
+// deck.gl React components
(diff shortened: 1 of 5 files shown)


examples/wip/viewport-animation-first-person-map/app.js
```diff
@@ -0,0 +1,346 @@
+/* global window, document, fetch */
+import React, {Component} from 'react';
```
ibgreen: Nit: Do we need two examples? Can't we just have one?

src/core/viewports/web-mercator-viewport.js
```diff
@@ -267,14 +267,16 @@ export default class WebMercatorViewport extends Viewport {
...
+    const EPSILON = 0.000001;
     const MAPBOX_LIMITS = {
       pitch: 60,
       zoom: 40
     };
 
     const {pitch, zoom} = this;
 
-    return pitch <= MAPBOX_LIMITS.pitch && zoom <= MAPBOX_LIMITS.zoom;
+    return pitch <= (MAPBOX_LIMITS.pitch + EPSILON) &&
```
ibgreen: Normally one does something like `Math.abs(a - b) < EPSILON` to make sure signs etc don't affect the results. As one option. math.gl exports an `equals` function that does this.

examples/wip/viewport-animation-first-person-map/app.js
1chandu: I will merge them into one in future PR.

src/core/viewports/web-mercator-viewport.js
1chandu: We probably should add lessOrEqual, greatOrEqual version to math as well.

PR #1119, opened 2017-11-07
Fix issues in Orbit Viewport and update examples

jianhuang01:
1. Can specify both 'Y' and 'Z' axis as orbit axis instead of 'Y' only
2. Change translate to world space instead of screen space
3. Change scale to world space in stead of screen space, which makes frustum clipping fails

Tested with point-cloud-laz, point-cloud-ply and plot examples

### DIFF
--- a/examples/plot/app.js
+++ b/examples/plot/app.js
@@ -15,8 +15,9 @@ class Root extends Component {
       viewport: {
         lookAt: [0, 0, 0],
         distance: 3,
-        rotationX: -30,
-        rotationY: 30,
+        pitchAngle: -30,
+        orbitAngle: 30,
+        orbitAxis: 'Y',
         fov: 50,
         minDistance: 1,
         maxDistance: 20,
--- a/src/core/viewports/orbit-viewport.js
+++ b/src/core/viewports/orbit-viewport.js
@@ -31,7 +31,7 @@ export default class OrbitViewport extends Viewport {
     distance, // From eye position to lookAt
     pitchAngle = 0, // Rotating angle around X axis
     orbitAngle = 0, // Rotating angle around orbit axis
-    orbitAxis = 'Z', // Orbit axis with 360 degrees rotating freedom
+    orbitAxis = 'Z', // Orbit axis with 360 degrees rotating freedom, can only be 'Y' or 'Z'
     lookAt = [0, 0, 0], // Which point is camera looking at, default origin
     up = [0, 1, 0], // Defines up direction, default positive y axis
     // projection matrix arguments
--- a/src/core/viewports/orbit-viewport.js
+++ b/src/core/viewports/orbit-viewport.js
@@ -8,7 +8,6 @@ import mat4_translate from 'gl-mat4/translate';
 import mat4_rotateX from 'gl-mat4/rotateX';
 import mat4_rotateY from 'gl-mat4/rotateY';
 import mat4_rotateZ from 'gl-mat4/rotateZ';
-import vec3_add from 'gl-vec3/add';
 
 import {transformVector} from '../math/utils';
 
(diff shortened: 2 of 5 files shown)


examples/plot/app.js
```diff
@@ -15,8 +15,9 @@ class Root extends Component {
       viewport: {
         lookAt: [0, 0, 0],
         distance: 3,
-        rotationX: -30,
-        rotationY: 30,
+        pitchAngle: -30,
```
1chandu: I prefer we follow only one convention either X, Y and Z or Pitch, Roll and Yaw.  `pitchAngle` and `orbitAxis: 'Y'` seem to combine both.

src/core/viewports/orbit-viewport.js
```diff
@@ -43,30 +45,35 @@ export default class OrbitViewport extends Viewport {
...
+    if (orbitAxis === 'Z') {
+      mat4_rotateZ(rotationMatrix, rotationMatrix, -orbitAngle / 180 * Math.PI);
+    } else {
+      mat4_rotateY(rotationMatrix, rotationMatrix, -orbitAngle / 180 * Math.PI);
+    }
+
+    const transformMatrix = mat4_translate([], createMat4(),
+      [translationX / width * 2, translationY / height * 2, 0]);
+    mat4_scale(transformMatrix, transformMatrix, [zoom, zoom, zoom]);
+    mat4_multiply(rotationMatrix, rotationMatrix, transformMatrix);
```
1chandu: Nit: store final result `transformMatrix` as it can contain translations and rotations.

src/core/viewports/orbit-viewport.js
```diff
@@ -43,30 +45,35 @@ export default class OrbitViewport extends Viewport {
     translationY = 0, // in pixels
     zoom = 1
   }) {
-    const eye = vec3_add([], lookAt, [0, 0, distance]);
-    vec3_rotateX(eye, eye, lookAt, rotationX / 180 * Math.PI);
-    vec3_rotateY(eye, eye, lookAt, rotationY / 180 * Math.PI);
+    const rotationMatrix = mat4_rotateX([], createMat4(), -pitchAngle / 180 * Math.PI);
```
1chandu: Earlier, we were rotating 'eye' about 'lookAt', but now we move these rotations into world (as part of transformMatrix), but now we are rotating about Origin, if `lookAt` is not origin wouldn't this be completely different rotations?

examples/plot/app.js
jianhuang01: actually there is only one convention, Pitch is the rotating angle around X, here I use pitchAngle to name the rotation angle around X, but use Y to specify the axis for orbit. So there is no conflict in this area.

src/core/viewports/orbit-viewport.js
jianhuang01: good catch, will fix it. Thanks!

examples/plot/app.js
shaojingli: Is orbit axis Y the perpendicular axis? so what if the axis for orbit is not Y but X or Z. Will the pitch lead to a different rotation direction?

src/core/viewports/orbit-viewport.js
Pessimistress: This is incorrect. `translation` and `scale` were applied to screen space because when panning or scrolling to zoom, the "grabbed" pixel must remain under the cursor. With this change, depending on the orbit angle, the model may travel in the opposite direction of the drag.

I understand the problem of transforms in screen space, and i agree that we need to change it. The old implementation was to avoid tackling the difficult math problem of translating 2d interaction back to 3d camera movement. Now that we're at it, might as well do it right.

jianhuang01: yeah, you are right, the code is still not good enough. I will fix it in later PR instead of a big one with everything inside. This problem is more difficult, so need move forward gradually, but at least we have a better direction to go.

src/core/viewports/orbit-viewport.js
jianhuang01: Done.

examples/plot/app.js
jianhuang01: By default, 'Z' is the perpendicular axis, we can set 'Y' as the orbit axis, too. But can't set 'X' as the orbit axis cause it is not meaningful in right handed 3D coordinate system. For both 'Z' and 'Y' cases, the pitch is consistent. Will add more comments in the code.

PR #1145, opened 2017-11-13
Orbit Controller refactor to improve pan, rotation and zoom

jianhuang01:
This is the second round to improve Orbit Controller, includes these items.
1. Continue to move panning from clip space to world space, which can avoid rendering distortion and inaccurate depth clipping
2. Change the old Orbit State modeling from (lookAt, translateX, translateY, pitchAngle, rotationAngle) to (lookAt, pitchAngle, rotationAngle), cause translate and rotation is dependent with each other which makes the logic much more complicated.
3. Clean up code for pan, rotation and zoom to adapt new modeling

Tested with plot, point-cloud-laz and point-cloud

### DIFF
--- a/src/core/controllers/orbit-state.js
+++ b/src/core/controllers/orbit-state.js
@@ -3,8 +3,8 @@ import assert from 'assert';
 
 const defaultState = {
   lookAt: [0, 0, 0],
-  pitchAngle: 0,
-  orbitAngle: 0,
+  rotationX: 0,
+  rotationOrbit: 0,
   fov: 50,
   near: 1,
   far: 100,
@@ -36,8 +36,8 @@ export default class OrbitState {
     width, // Width of viewport
     height, // Height of viewport
     distance, // From eye to target
-    pitchAngle, // Rotation around x axis
-    orbitAngle, // Rotation around orbit axis
+    rotationX, // Rotation around x axis
+    rotationOrbit, // Rotation around orbit axis
     orbitAxis, // Orbit axis with 360 degrees rotating freedom, can only be 'Y' or 'Z'
     // Bounding box of the model, in the shape of {minX, maxX, minY, maxY, minZ, maxZ}
     bounds,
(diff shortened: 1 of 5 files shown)


src/core/controllers/orbit-state.js
```diff
@@ -50,9 +36,9 @@ export default class OrbitState {
     width, // Width of viewport
     height, // Height of viewport
     distance, // From eye to target
-    pitchAngle, // Rotation around x axis
-    orbitAngle, // Rotation around orbit axis
-
+    rotationX, // Rotation around x axis
```
gnavvy: personally, I agree with the original naming that _pitch_ is relatively easier to understand than _X_.
https://social.sbrick.com/custom/wiki/designer/latest/camera_pitch_yaw_roll.png

but as long as we have documented this clearly it should be fine.

src/core/controllers/orbit-state.js
```diff
@@ -147,19 +136,25 @@ export default class OrbitState {
    * @param {[Number, Number]} pos - position on screen where the pointer is
    */
   pan({pos, startPos}) {
+    if (this._interactiveState.isRotating) {
```
1chandu: Don't have to address in this PR.
Quesiton: Can we make the decision of panning/rotating before pan()/rotate() gets called so we don't need these if conditions?

src/core/controllers/orbit-state.js
```diff
@@ -195,40 +189,23 @@ export default class OrbitState {
...
-    orbitAngle = ensureFinite(orbitAngle, this._viewportProps.orbitAngle);
-    translationX = ensureFinite(translationX, this._viewportProps.translationX);
-    translationY = ensureFinite(translationY, this._viewportProps.translationY);
-
-    const newPitchAngle = clamp(pitchAngle - deltaScaleY * 180, -89.999, 89.999);
-    const newOrbitAngle = (orbitAngle - deltaScaleX * 180) % 360;
-
-    let newTranslationX = translationX;
-    let newTranslationY = translationY;
+    if (this._interactiveState.isPanning) {
```
1chandu: Same comment as above.

src/core/controllers/orbit-state.js
1chandu: On this same topic, we should also document that, OrbitAxis can't be x-axis.

src/core/controllers/orbit-state.js
jianhuang01: I like that, too. The logic is inside ViewControls which will impact all the viewports, and ViewContols is under refactor, need discuss before the change. That's why I didn't do that.

PR #1163, opened 2017-11-16
Fix the bug of distance calculation in orbit controller

jianhuang01:
Make point-cloud-ply, point-cloud-plz and plot examples to have the right size in viewport.

Tested with these examples.


jianhuang01:
@1chandu  based on the fov definition and the math calculation, the size should be half.

Issue #975, opened 2017-10-03
Show message explaining that wind map example doesn't work on safari


jckr:
(http://philogb.github.io/page/wind/)

https://twitter.com/OBuchel/status/915017337211899905

I get:  
[Error] Unhandled Promise Rejection: Error: createGLContext called on canvas 'texture-canvas' before page was loaded
	(anonymous function) (context.js:53)
	promiseReactionJob

ibgreen:
@jckr - This is unfortunately expected as the wind map is a WebGL2 example, currently only supported by Chrome and Firefox. But the error message needs to be much better.

@1chandu, @shaojingli - We need to make this clear and have a nice message on non-supporting browsers.

PR #1166, opened 2017-11-16
Wind demo warning

1chandu:
Windo demo is WebGL2 only application, display a message when run on non-webgl2 browsers.

Fixes : #975 

Testing: Runs fine on Chrome, Firefox, displays this message on Safari.

screen shot 2017-11-16 at 3 45 32 pm


1chandu:
This is on top of my other PR which isn't landed yet, will rebase this on master once it is landed, it should only include one last commit.

1chandu:
@jianhuang01 , yep, this will only include example change, as mentioned above, doc changes are from other approved PR which hasn't landed yet.

examples/wind/src/control-panel.js
```diff
@@ -47,7 +47,7 @@ export default class ControlPanel extends Component {
         { this._renderToggle('showParticles', 'particles') }
         { this._renderToggle('showWind', 'field') }
         { this._renderToggle('showElevation', 'elevation') }
-        { this._renderToggle('useDevicePixels', 'Retina/HD') }
+        { this._renderToggle('useDevicePixels', 'Retina/ HD') }
```
jianhuang01: why do we need this space?

1chandu: Just a non functional change, it looked better this way.

PR #1171, opened 2017-11-18
Fix default alpha blend mode

Pessimistress:
ScatterplotLayer with `getColor: d => [0, 0, 0, 50]`

Before:
screen shot 2017-11-17 at 6 45 20 pm

After:
screen shot 2017-11-17 at 6 45 11 pm

### DIFF
--- a/src/core/pure-js/deck-js.js
+++ b/src/core/pure-js/deck-js.js
@@ -207,7 +207,7 @@ export default class DeckGLJS {
   _onRendererInitialized({gl, canvas}) {
     setParameters(gl, {
       blend: true,
-      blendFunc: [GL.SRC_ALPHA, GL.ONE_MINUS_SRC_ALPHA],
+      blendFuncSeparate: [GL.SRC_ALPHA, GL.ONE_MINUS_SRC_ALPHA, GL.ONE, GL.ONE_MINUS_SRC_ALPHA],
       polygonOffsetFill: true
     });
 


1chandu:
@Pessimistress , I will be very cautious to change this global blend setting, this is going to change the rendering of all deck.gl layers and applications. (consider the case of multiple layers being blended together) It might look good for a specific color case (0, 0, 0, 50) but might not work for other cases.  It is better to change the input alpha value in your example or `opacity` prop to adjust the rendering.

Pessimistress:
@1chandu In the core layers, using color `[0, 0, 0, 50]` and using color `[0, 0, 0]` with opacity `0.2` has the same effect.
The blending problem I am trying to address here is when you overlap multiple instances/layers with alpha. The old default blending mode's behavior is:
1. Drawing `rgba(0, 0, 0, 0.2)` on top of transparency: `rgba(0, 0, 0, 0.2)` (alpha = `0 * 0 + 1 * 0.2`)
2. Drawing `rgba(0, 0, 0, 0.2)` on top of `rgba(0, 0, 0, 0.2)`: `rgba(0, 0, 0, 0.2)` (alpha = `0.2 * 0.2 + 0.8 * 0.2`)
3. Drawing `rgba(0, 0, 0, 0.2)` on top of `rgba(0, 0, 0, 0.4)`: `rgba(0, 0, 0, 0.28)` (alpha = `0.4 * 0.4 + 0.6 * 0.2`)
This is unexpected because drawing a translucent black on top of an existing pixel should make it a little darker, not the same (2) or even lighter (3).

The proposed default blending mode does not change the blending of RGB (hence `blendFuncSeparate`), but treat alpha more intuitively:
1. Drawing `rgba(0, 0, 0, 0.2)` on top of transparency: `rgba(0, 0, 0, 0.2)` (alpha = `0 * 0 + 1 * 0.2`)
2. Drawing `rgba(0, 0, 0, 0.2)` on top of `rgba(0, 0, 0, 0.2)`: `rgba(0, 0, 0, 0.36)` (alpha = `1 * 0.2 + 0.8 * 0.2`)
3. Drawing `rgba(0, 0, 0, 0.2)` on top of `rgba(0, 0, 0, 0.4)`: `rgba(0, 0, 0, 0.52)` (alpha = `1 * 0.4 + 0.6 * 0.2`)

Pessimistress:
CodePen demo of the two blending modes: https://codepen.io/Pessimistress/pen/POQQVX

jianhuang01:
The fix is right, but better to verify our examples to make sure everything is in our expectation because they are built based on the old setting.

1chandu:
As discussed offline , given the breaking nature of the PR (the rendered data looks different) I propose to put this in 5.0 instead of 4.2.

PR #1174, opened 2017-11-20
Fix viewport projection topLeft option

### LINKED ISSUE #1167: Fix Viewport `topleft` option
Need to port it from https://github.com/uber-common/viewport-mercator-project/pull/44

1chandu:
- Fix `project`/`unproject`  `options.topLeft` to behave as per documentation.
- Our viewports are setup with top-left as origin, so perform `y-flip` only when `topLeft=false`.
Fixes #1167 

Tested with: `test-browser`, `layer-browser`, `plot`, `point-cloud-ply` and `point-cloud-laz`.

### DIFF
--- a/src/core/controllers/map-state.js
+++ b/src/core/controllers/map-state.js
@@ -322,7 +322,7 @@ export default class MapState extends ViewState {
 
   _unproject(pos) {
     const viewport = new PerspectiveMercatorViewport(this._viewportProps);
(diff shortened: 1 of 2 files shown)


jianhuang01:
@1chandu can you test point-cloud-ply and point-cloud-laz ?

1chandu:
@jianhuang01 , verified, all good.

PR #1228, opened 2017-12-11
add "babel-loader" dependency for wind example

jianhuang01:
#1213

### DIFF
--- a/examples/wind/package.json
+++ b/examples/wind/package.json
@@ -20,10 +20,11 @@
   },
   "devDependencies": {
     "babel-core": "^6.21.0",
-    "buble-loader": "^0.4.0",
+    "babel-loader": "^6.2.10",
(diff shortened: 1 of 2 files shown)


examples/wind/package.json
```diff
@@ -20,10 +20,11 @@
   },
   "devDependencies": {
     "babel-core": "^6.21.0",
-    "buble-loader": "^0.4.0",
+    "babel-loader": "^6.2.10",
     "babel-preset-es2015": "^6.18.0",
     "babel-preset-react": "^6.16.0",
     "babel-preset-stage-2": "^6.18.0",
+    "buble-loader": "^0.4.0",
```
1chandu: I believe this can be remove altogether.

jianhuang01: no, when we use"npm run start-local", still need the bible-loader.

PR #1230, opened 2017-12-12
Fix Picking Module object highlighting

### LINKED ISSUE #1229: Picking: highlighting with selectedObjectIndex is broken

1chandu:
- Object highlighting uniforms should be set during drawing mode.
- Rename `getPickingModuleParameters` -> `getObjectHighlightParameters`

Fixes #1229 

Verified with layer-browser (`picking`, `selectedObjectIndex` and `autoHighlight`)

### DIFF
--- a/src/core/lib/draw-layers.js
+++ b/src/core/lib/draw-layers.js
@@ -237,8 +237,6 @@ function drawLayerInViewport({gl, layer, layerIndex, drawPickingColors, glViewpo
   });
 
   if (drawPickingColors) {
-    // TODO - Disable during picking
-
     Object.assign(layerParameters, {
       blendColor: [0, 0, 0, (layerIndex + 1) / 255]
     });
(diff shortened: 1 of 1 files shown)


src/core/lib/draw-layers.js
```diff
@@ -238,11 +238,12 @@ function drawLayerInViewport({gl, layer, layerIndex, drawPickingColors, glViewpo
 
   if (drawPickingColors) {
     // TODO - Disable during picking
```
georgios-uber: maybe remove todo?

src/core/lib/draw-layers.js
```diff
@@ -238,11 +238,12 @@ function drawLayerInViewport({gl, layer, layerIndex, drawPickingColors, glViewpo
...
-    Object.assign(moduleParameters, getPickingModuleParameters(layer));
 
     Object.assign(layerParameters, {
       blendColor: [0, 0, 0, (layerIndex + 1) / 255]
     });
+  } else {
+    Object.assign(moduleParameters, getObjectHighlightParameters(layer));
   }
 
   withParameters(gl, layerParameters, () => {
```
Pessimistress: `layer.drawLayer` already calls `withParameters(this.context.gl, parameters, ...)`. Is it necessary to nest another level here?

1chandu: Good find, fixed.

PR #1231, opened 2017-12-12
Remove deprecated Choropleth layers

1chandu:
- Remove Chorpleth layers, docs and tests.
- Add details to `upgrade-guide.md`.

Verified with `run test-browser`.


1chandu:
Dong see `ChoroplethLayer` in examples/sample-layers, it only has `EnhancedChoroplethLayer`.

PR #1260, opened 2017-12-20
Fix pickingSelectedColor, fix mesh-layer regression

1chandu:
- Fix highlighting when `highlightedObjectIndex` is used, set its default value to `null` which disables the feature. When set to a non null value, corresponding instance will be highlighted. Any invalid index value doesn't highlight any object.
- Fix a regression in `mesh-layer`.

Verified with `layer-browser`.


georgios-uber:
The API change is necessary for the fix to work.

PR #1263, opened 2017-12-20
Examples start script

1chandu:
- Fix `plot` example to work when using `npm run start`, `state.models` is iterated using `for .. of` iterator so it should be an array , not a object. 
- Fix build script in few examples.

Verified running `plot` and `point-cloud-laz` example

### DIFF
--- a/examples/plot/plot-layer/axes-layer.js
+++ b/examples/plot/plot-layer/axes-layer.js
@@ -104,11 +104,13 @@ export default class AxesLayer extends Layer {
       instanceIsTitle: {size: 1, update: this.calculateInstanceIsTitle, noAlloc: true}
     });
 
-    this.setState({
-      models: this._getModels(gl),
-      numInstances: 0,
-      labels: null
-    });
+    this.setState(
+      Object.assign({
+          numInstances: 0,
+          labels: null
+        },
+        this._getModels(gl)
+      ));
   }
 
   updateState({oldProps, props, changeFlags}) {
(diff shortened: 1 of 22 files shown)


1chandu:
And for some reason this change is only needed for `npm run start` to work, `npm run start-local` works fine without this change.

examples/svg-interoperability/package.json
```diff
@@ -4,7 +4,11 @@
   "license": "MIT",
   "scripts": {
     "start": "webpack-dev-server --progress --hot --port 3000 --open",
```
gnavvy: shall we remove all the _--port 3000_ to avoid conflicting ports when debugging multiple examples?
webpack-dev-server will assign one if not specified.

examples/plot/plot-layer/axes-layer.js
```diff
@@ -154,18 +158,18 @@ export default class AxesLayer extends Layer {
 
   updateAttributes(props) {
     super.updateAttributes(props);
-    const {attributeManager, models, numInstances} = this.state;
+    const {attributeManager, modelsByName, numInstances} = this.state;
```
gnavvy: @1chandu could you comment on the models => modelsByName change? I might have missed something but I could not find refs in the code why we are destructing modelsByName instead of models from the state.

1chandu: Base Layer class uses ` for .. of` to iterate over `state.models`, it should be an array. So change `state.models` to array and created a new one `modelByName` to use with in `axes-layer`. We are doing same in `SolidPolygonLayer`.

PR #1266, opened 2017-12-21
Ortho zooming example

gnavvy:
Added an example to demo how to zoom with an orthographic viewport.

### DIFF
--- a/docs/api-reference/orthographic-viewport.md
+++ b/docs/api-reference/orthographic-viewport.md
@@ -26,9 +26,13 @@ Parameters:
   * `left` (Number) - Left bound of the frustum
   * `top` (Number) - Top bound of the frustum
 
-  automatically calculated:
+  automatically calculated by default:
   * `right` (Number, optional) - Right bound of the frustum.
   * `bottom` (Number, optional) - Bottom bound of the frustum.
+
+  when zooming with orthographic viewport, the size (width and height) of the viewport and window are no longer the same. In such case, specify `right` and `bottom` together with `left` and `top` explicitly to define the viewport size.
+
+  refer to `examples/experimental/orthographic-zooming` for example.
 
 ```js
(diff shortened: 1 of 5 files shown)


examples/experimental/orthographic-zooming/app.js
```diff
@@ -0,0 +1,132 @@
...
+    const viewport = new OrthographicViewport({
+      width,
+      height,
+      // the size of canvas and viewport are no longer the same as we scale w/ an ortho-viewport
+      // all four are needed - left top right bottom - to decide the viewport size
+      left: -width / 2 * scale,
+      top: -height / 2 * scale,
+      right: width / 2 * scale,
+      bottom: height / 2 * scale
+    });
```
gnavvy: here, we need to explicitly specify all four corners based on the current API design.

docs/api-reference/orthographic-viewport.md
```diff
@@ -26,10 +26,14 @@ Parameters:
   * `left` (Number) - Left bound of the frustum
   * `top` (Number) - Top bound of the frustum
 
-  automatically calculated:
+  automatically calculated by default:
```
1chandu: Given this is a major release, we can make this api cleaner and document it in `upgrade-guide`.
1. Make `right` and `bottom` non optional.  Specifying just `top` and `left` and calculating `right` and `bottom` looks asymmetric any way.
2. Replace `top`, `left`, `bottom` and `right`, with `scale`, inside `OrthographicViewport` we can calculate all 4 parameters using `width`, `height` and `scale`.

gnavvy: I prefer to keep these parameters, it allows the users to specify the range, [-1, 1] or [0, 1] per use case. Using scale instead assumes [-1, 1] which is not always the case.

PR #1295, opened 2018-01-05
add more layer rendering test cases

jianhuang01:
for #999


jianhuang01:
@ibgreen don't know the expected effects, right now it is the same as which is rendered in layer-browser example. also curious why it is not round as the meter version.

Pessimistress:
The point cloud data for lnglat_offset is generated to fit a 0.01deg x 0.01deg x 1000m box. Distance along unit latitude is shorter at this map location.

jianhuang01:
@Pessimistress Thanks for the explanation, understand the details now.

PR #1318, opened 2018-01-12
Use latest luma.gl transform feedback features

Pessimistress:
Depends on https://github.com/uber/luma.gl/pull/383

### DIFF
--- a/src/core/lib/attribute-transition-manager.js
+++ b/src/core/lib/attribute-transition-manager.js
@@ -173,20 +173,9 @@ export default class AttributeTransitionManager {
   _runTransformFeedback({uniforms, buffers}) {
     const {model, transformFeedback} = this;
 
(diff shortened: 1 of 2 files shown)


src/core/lib/attribute-transition-manager.js
```diff
@@ -173,20 +173,9 @@ export default class AttributeTransitionManager {
...
-    transformFeedback.begin(GL.POINTS);
-
-    const parameters = {
-      [GL.RASTERIZER_DISCARD]: true
-    };
-
-    model.draw({uniforms, parameters});
-
-    transformFeedback.end();
+    model.draw({uniforms, transformFeedback});
```
1chandu: As I commented in luma PR, to avoid confusion, parameters with `[GL.RASTERIZER_DISCARD]: true` should still need to sent. Or lets wait until that PR is finalized.

PR #1330, opened 2018-01-20
Merge react-map-gl's map interaction

Pessimistress:


src/core/controllers/map-controls.js
```diff
@@ -38,36 +34,4 @@ export default class MapControls extends ViewportControls {
   _onPan(event) {
     return this.isFunctionKeyPressed(event) ? this._onPanRotate(event) : this._onPanMove(event);
   }
-
```
jianhuang01: without default _onPanRotate(event) function, do we have one in the parent class?

src/core/controllers/map-state.js
```diff
@@ -75,12 +77,11 @@ export default class MapState extends ViewState {
...
-      altitude: ensureFinite(altitude, defaultState.altitude),
-      maxZoom: ensureFinite(maxZoom, MAPBOX_LIMITS.maxZoom),
-      minZoom: ensureFinite(minZoom, MAPBOX_LIMITS.minZoom),
-      maxPitch: ensureFinite(maxPitch, MAPBOX_LIMITS.maxPitch),
-      minPitch: ensureFinite(minPitch, MAPBOX_LIMITS.minPitch)
+      altitude,
+      maxZoom,
+      minZoom,
+      maxPitch,
+      minPitch
```
jianhuang01: why do we remove the range check?

src/core/controllers/viewport-controls.js
```diff
@@ -333,50 +361,66 @@ export default class ViewportControls {
...
-    const key = event.key;
-    const handler = KEY_BINDINGS[key];
-    if (this.viewportState[handler]) {
-      const newViewportState = this.viewportState[handler]();
-      this.updateViewport(newViewportState);
+    const funcKey = this.isFunctionKeyPressed(event);
+    const {viewportStateProps} = this;
+    let newViewportState;
+
+    switch (event.srcEvent.keyCode) {
```
jianhuang01: The old implementation of this key handling function is more clearer, why do we change it?

src/core/controllers/map-controls.js
Pessimistress: Yes

src/core/controllers/map-state.js
Pessimistress: The old fallback mechanism is replaced by default parameters.

PR #1350, opened 2018-01-25
Fix lint error because of unused assert

jianhuang01:


Pessimistress:
Duplicate of https://github.com/uber/deck.gl/pull/1349

PR #1351, opened 2018-01-26
Added Point Cloud VR example

sidx1024:
This is a VR example of point-cloud-laz which uses multi-viewport and WebVR API.
The multi-view mode works perfectly on an Android Device (OnePlus 5, Android 8.0), not tested for high-end hardware.

The "vr/emulated-vr-display.js" is used from the ["vr" branch](https://github.com/uber/deck.gl/blob/vr/examples/vr/src/emulated-vr-display.js) which provides an emulated display when an actual VR Hardware is not present. There are some problems displaying the example without a VR Hardware because the calculated View Matrix & Projection Matrix are not as expected _(working on it)_.


ibgreen:
Thanks for putting this up. @gnavvy worked on the initial example so he should look at this, 

Tried to run after pulling your branch like this
```
git checkout -b sidx1024-6.0-dev 6.0-dev
git pull https://github.com/sidx1024/deck.gl.git 6.0-dev
cd examples/experimental/point-cloud-laz-vr/
npm run start
```
But all I see is a couple of dots in the middle of the screen, no interaction:

Is that because you are still working on the VR controller?

screen shot 2018-01-26 at 8 49 53 am

ibgreen:
I suspect this PR could be opened toward master instead of 6.0-dev. If it doesn't break backwards compatibility for anyone, it can be part of a minor release (e.g. 5.1 or 5.2)

Also @gnavvy do we need two versions for the laz example or should we consolidate?

sidx1024:
I've made few changes just a while ago, now it should look like this.
The position and orientation can be changed in **this.state** to move the camera.

preview
The above preview is at these VRPose parameters:
```
position: [0, 0, 0]
orientation: [0, 0, 0, 1]
```
I think binding mouse events to **emulatedPose** will surely do the trick.

examples/experimental/point-cloud-laz-vr/vr/emulated-vr-display.js
```diff
@@ -0,0 +1,148 @@
...
+    this._leftEyeMatrix = this._getEyeMatrix(this.leftEyeParameters);
+    this._rightEyeMatrix = this._getEyeMatrix(this.rightEyeParameters);
+  }
+
+  getEyeParameters(whichEye) {
+    const viewport = this._getViewportSize();
+
+    if (whichEye === 'right') {
+      return {...this.rightEyeParameters, ...viewport};
+    } else {
```
gnavvy: nit: no need to use else.

examples/experimental/point-cloud-laz-vr/vr/emulated-vr-display.js
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    const timestamp = Date.now();
+
+    const leftProjectionMatrix = new Matrix4();
+    const rightProjectionMatrix = new Matrix4();
+    const leftViewMatrix = new Matrix4();
+    const rightViewMatrix = new Matrix4();
+
+    perspectiveFromFieldOfView(leftProjectionMatrix, this.leftEyeParameters.fieldOfView, this.depthNear, this.depthFar);
```
gnavvy: @sidx1024 could you try eslint this file, several lines are too long and will fail the eslint test before when checking in.

examples/experimental/point-cloud-laz-vr/app.js
```diff
@@ -0,0 +1,245 @@
+/* global document, window,*/
```
gnavvy: _navigator_ is only available in the browser context, need to add it here to make linter happy.
/* global document, window, navigator*/

examples/experimental/point-cloud-laz-vr/app.js
```diff
@@ -0,0 +1,245 @@
...
+      getColor: d => [255, 255, 255, 128],
+      radiusPixels: 1
+    });
+  }
+
+  _initVRDisplay() {
+    // initialize WebVR Polyfill
+    const polyfill = new WebVRPolyfill({
+      PROVIDE_MOBILE_VRDISPLAY: true
+    });
```
gnavvy: polyfill is defined but not used, I noticed the official example of webvr-polyfill does this too...

either use
```
    /* eslint-disable no-unused-vars */
    const polyfill = new WebVRPolyfill({
      PROVIDE_MOBILE_VRDISPLAY: true
    });
    /* eslint-enable no-unused-vars */
```
or
```
    const polyfill = new WebVRPolyfill({  // eslint-disable-line no-unused-vars
      PROVIDE_MOBILE_VRDISPLAY: true
    });
```
to disable the linter rule here.

examples/experimental/point-cloud-laz-vr/app.js
```diff
@@ -0,0 +1,245 @@
...
+    this.setState({
+      viewport: {...this.state.viewport, ...viewport}
+    });
+  }
+
+  _onUpdate() {
+    const {viewport} = this.state;
+    this.setState({
+      viewport: {
+        ...viewport,
```
gnavvy: in my original example, I rotate the viewport for every animation frame.
we can remove the whole _onUpdate function if we don't need auto rotate and just the viewport generated from vrDisplay instead.

examples/experimental/point-cloud-laz-vr/app.js
```diff
@@ -0,0 +1,245 @@
...
+        }),
+        new Viewport({
+          x: width / 2,
+          width: width / 2,
+          height,
+          viewMatrix: frameData.rightViewMatrix,
+          projectionMatrix: frameData.rightProjectionMatrix
+        })
+      ];
+    }
```
gnavvy: we need a default viewport here in case gotFrameData is false.

sidx1024:
@gnavvy 
I have a little confusion over here.
In case of no VR device present, what exactly should be used?
```
1. viewports = [Viewport, Viewport]
    (Stereoscopic View, viewMatrix & projectionMatrix from EmulatedVRDisplay)

2. viewports = [OrbitViewport, OrbitViewport]
    (Stereoscopic View, viewMatrix & projectionMatrix from EmulatedVRDisplay)

3. viewports = [OrbitViewport]
    (Single Viewport with Interaction)
```
Because I've checked and in case 2, the OrbitViewport from OrbitController.getViewport() doesn't seem to use the provided viewMatrix & projectionMatrix.

PR #1366, opened 2018-01-30
Experiment: bezier curve layer

javidhsueh:
- Added a quadratic bezier curve layer.
 - Able to render a multi-edges graph.
multi_graph


examples/bezier/app.js
```diff
@@ -0,0 +1,115 @@
+/* global window,document */
+import React, {Component} from 'react';
+import {render} from 'react-dom';
+import DeckGLOverlay from './deckgl-overlay.js';
+
+import SAMPLE_GRAPH from './sample-graph';
+
+function computeControlPoint(source, target, direction, offset) {
```
gnavvy: shall we add jsdoc or a link to some reference to explain how to use the direction and offset here?

examples/bezier/app.js
```diff
@@ -0,0 +1,115 @@
...
+  const normal = [dy, -dx];
+  const length = Math.sqrt(Math.pow(normal[0], 2.0) + Math.pow(normal[1], 2.0));
+  const normalized = [normal[0] / length, normal[1] / length];
+  return [
+    midPoint[0] + normalized[0] * offset * direction,
+    midPoint[1] + normalized[1] * offset * direction
+  ];
+}
+
+function layoutGraph(graph) {
```
gnavvy: a line of description on the graph data structure would be helpful.
{nodes: [], edges: []}

examples/bezier/bezier-curve-layer/bezier-curve-layer-vertex.glsl.js
```diff
@@ -0,0 +1,102 @@
...
+  vec2 offset_clipspace = project_pixel_to_clipspace(offset_screenspace).xy;
+
+  return offset_clipspace;
+}
+
+float getSegmentRatio(float index) {
+  return smoothstep(0.0, 1.0, index / numSegments);
+}
+
+vec4 computeBezierCurve(vec4 source, vec4 target, vec4 controlPoint, float segmentRatio) {
```
gnavvy: Not in the PR but shall we change this function to handle the more general cubic curve?

If I recall cubic bezier curve with two identical (same xyz) control points generates shape like the quadratic curve. if not we can simply compare if the control points are the same and if yes, fallback to this quadratic calculation.

PR #1405, opened 2018-02-07
Merge experimental SolidPolygonLayer into core-layers

Pessimistress:
Move the performant version of `SolidPolygonLayer` from experimental-layers to core-layers.


jianhuang01:
Feel confused, we have SolidPolygonLayer on both core and experimental layers, what's the difference?

jianhuang01:
Refer to #1420

PR #1407, opened 2018-02-07
Rendering test cases to cover 64bits shaders

jianhuang01:


jianhuang01:
@ibgreen Actually compared with 32bit images,  overall have 2% differences, primarily happens on the AA (alpha blending) pixels, if excluding the AA pixels,  the percentage can be less than 0.1%. Anyway it is impossible to be detected by human eyes.

PR #1409, opened 2018-02-08
Fix Bezier-Curve layer rendering issues

1chandu:
There were two different problems :
1. Triangle ordering was incorrect for the first instance.
2. Depending on number of segments, some of the vertices have wrong z, values.

I have included the fix for both, but lets discuss and finalize the changes.

**Before:**
screen shot 2018-02-07 at 9 04 34 pm
screen shot 2018-02-07 at 9 04 20 pm

**After.**
screen shot 2018-02-07 at 9 05 28 pm
screen shot 2018-02-07 at 9 05 47 pm

### DIFF
--- a/src/experimental-layers/src/bezier-curve-layer/bezier-curve-layer-vertex.glsl.js
+++ b/src/experimental-layers/src/bezier-curve-layer/bezier-curve-layer-vertex.glsl.js
@@ -61,6 +61,7 @@ vec4 computeBezierCurve(vec4 source, vec4 target, vec4 controlPoint, float segme
   float a = mt2;
   float b = mt * segmentRatio * 2.0;
   float c = t2;
+  // TODO: if depth is not needed remove z computaitons.
   vec4 ret = vec4(
     a * source.x + b * controlPoint.x + c * target.x,
     a * source.y + b * controlPoint.y + c * target.y,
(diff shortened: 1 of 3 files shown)


src/experimental-layers/src/bezier-curve-layer/bezier-curve-layer.js
```diff
@@ -25,7 +25,7 @@ import {GL, Model, Geometry} from 'luma.gl';
 import vs from './bezier-curve-layer-vertex.glsl';
 import fs from './bezier-curve-layer-fragment.glsl';
 
-const NUM_SEGMENTS = 40;
+const NUM_SEGMENTS = 80;
```
javidhsueh: I was thinking whether it should be part of the layer api that users can control the number of segments on this curve.

src/experimental-layers/src/bezier-curve-layer/bezier-curve-layer-vertex.glsl.js
```diff
@@ -90,7 +90,9 @@ void main(void) {
   vec4 nextP = computeBezierCurve(source, target, controlPoint, nextSegmentRatio);
 
   // extrude
-  vec2 offset = getExtrusionOffset(nextP.xy - p.xy, positions.y);
+  float direction = float(positions.y);
+  direction = mix(-1.0, 1.0, step(segmentIndex, 0.0)) *  direction;
```
javidhsueh: This should be correct now. It's the same as the implementation of arc layer.

PR #1461, opened 2018-02-22
Fix layer-browser hot reload

Pessimistress:
Retains app state after reload

### DIFF
--- a/examples/layer-browser/src/index.js
+++ b/examples/layer-browser/src/index.js
@@ -1,7 +1,6 @@
 /* global document */
 import React, {Component} from 'react';
 import ReactDOM from 'react-dom';
-import App from './app';
 import FPSStats from 'react-stats-zavatta';
 
 class Root extends Component {
@@ -30,12 +29,17 @@ class Root extends Component {
 
 const container = document.createElement('div');
 document.body.appendChild(container);
-ReactDOM.render(<Root AppComponent={App} />, container);
+
+const render = () => {
+  const App = require('./app').default;
+  ReactDOM.render(<Root AppComponent={App} />, container);
+};
+
+render();
 
 if (module.hot) {
   module.hot.accept('./app', () => {
     console.log('Hot reloading App component'); // eslint-disable-line
-    const NextApp = require('./app').default;
-    ReactDOM.render(<Root AppComponent={NextApp} />, container);
+    render();
   });
 }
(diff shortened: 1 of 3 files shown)


examples/layer-browser/src/index.js
```diff
@@ -0,0 +1,41 @@
...
+      </div>
+    );
+  }
+}
+
+// ---- Main ---- //
+
+const container = document.createElement('div');
+document.body.appendChild(container);
+ReactDOM.render(<Root AppComponent={App} />, container);
```
balthazar: could you create an util function like

```js
const render = (AppComponent) =>
  ReactDOM.render(<Root AppComponent={AppComponent} />, container);

render(App);

if (module.hot) {
  module.hot.accept('./app', () => {
    console.log('Hot reloading App component'); // eslint-disable-line
    render(require('./app').default);
  });
}

File: examples/layer-browser/webpack.config.js [hunk 2]
```diff
@@ -8,7 +8,7 @@ const webpack = require('webpack');
 const CONFIG = {
   // bundle app.js and everything it imports, recursively.
   entry: {
-    app: resolve('./src/app.js')
+    app: resolve('./src/index.js')
```
balthazar: You can `resolve('./src')`

examples/layer-browser/src/index.js
```diff
@@ -0,0 +1,41 @@
+/* global document */
+import React, {Component} from 'react';
+import ReactDOM from 'react-dom';
+import App from './app';
+import FPSStats from 'react-stats-zavatta';
+
+class Root extends Component {
+  constructor(props) {
+    super(props);
+    this._appState = null;
```
balthazar: I think you can initialize `_appState` as a static property (if the babel config there is permits), not sure how much we want to use these kind of features though

examples/layer-browser/src/index.js
```diff
@@ -0,0 +1,41 @@
+/* global document */
+import React, {Component} from 'react';
+import ReactDOM from 'react-dom';
+import App from './app';
+import FPSStats from 'react-stats-zavatta';
```
balthazar: I generally put the "vendor" dependencies before the app ones, maybe we could discuss on what we want to do for ordering?

examples/layer-browser/src/index.js
Pessimistress: It does not

PR #1464, opened 2018-02-22
Update Attribute transitions to use new luma.gl `Transform` API

1chandu:
For luma.gl issue: https://github.com/uber/luma.gl/issues/401

- Make `fromState` a `Buffer` object
- Re-cycle old WebGL `Buffer` handles before resetting them.
- Update transition code flow to use new luma.gl `Transform` API.
- Depends on luma.gl PR: https://github.com/uber/luma.gl/pull/418

Verified with attribute transitions example.


src/core/lib/attribute-transition-model.js
```diff
@@ -1,13 +1,11 @@
...
-
 const ATTRIBUTE_MAPPING = {
   1: 'float',
   2: 'vec2',
   3: 'vec3',
   4: 'vec4'
 };
 
-function getShaders(transitions) {
+export function getShaders(transitions) {
```
1chandu: I will rename this file to `attribute-transition-utils.js` after initial review, kept it as is for easier review.

Issue #1468, opened 2018-02-23
Relative path in md doc doesn't work in website


jianhuang01:
#### Actual Result
When using relative path like '/docs/api-reference/layer.md' in md docs, after building deck.gl website, the URL becomes undefined
#### Expected Result
Link should point to the md file specified
#### Reproduce Steps
- Go to website folder,
- Use "npm run start" to launch deck.gl website
- Go to documentation->what's New
- click 'transitions' link under 'Layer Transitions'
- it didn't go to layer page

#### To Do List
  - [x] Add label and assign to milestone
  - [x] Coding
  - [x] Testing

PR #1469, opened 2018-02-23
Fix link rewrite in website

Pessimistress:
Problem introduced in https://github.com/uber/deck.gl/pull/1364


jianhuang01:
For #1468

jianhuang01:
Tested, it is working well.

PR #1482, opened 2018-02-28
[Website] Render nested categories in side bar

Pessimistress:
For https://github.com/uber/deck.gl/issues/1470

#### Change List
- Support sub headers in table of contents
- Support badges
- Render wider side bar in large window

screen shot 2018-02-27 at 4 38 23 pm

### DIFF
--- a/website/src/components/table-of-contents.js
+++ b/website/src/components/table-of-contents.js
@@ -2,20 +2,37 @@ import React, {Component} from 'react';
 import PropTypes from 'prop-types';
 import {Link} from 'react-router';
 
+const ITEM_HEIGHT = 40;
+const INDENT_BASE = 16;
+const INDENT_INC = 12;
+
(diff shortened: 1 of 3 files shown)


website/src/components/table-of-contents.js
```diff
@@ -2,20 +2,37 @@ import React, {Component} from 'react';
 import PropTypes from 'prop-types';
 import {Link} from 'react-router';
 
+const ITEM_HEIGHT = 40;
+const INDENT_BASE = 16;
+const INDENT_INC = 12;
+
+function getContentHeight(page) {
```
balthazar: neat:

```js
const getContentHeight = page =>
  page.children
    ? page.children.reduce((acc, child) => acc + getContentHeight(child), ITEM_HEIGHT)
    : ITEM_HEIGHT
```

PR #1510, opened 2018-03-06
Fix for experimental-layer arrows with offset coordinates (5.2.0-alpha.2)

georgios-uber:


georgios-uber:
Fails because of https://github.com/uber/deck.gl/pull/1509

PR #1518, opened 2018-03-08
disable path-marker rendering test

jianhuang01:
#### Background
Chrome 65 can't rendering out path-marker layer correctly, disable it for now.

#### Change List
- Disable path-marker rendering test
- use test-render-react cause the test-render has bug which allowed failed rendering test pass the full test set.


shaojingli:
what's the issue? care to put up a screenshot?

jianhuang01:
Chrome 64
image

Chrome 65
image

jianhuang01:
@1chandu Not necessary for now, PathMarkerLayer is still under experimental-layers folder.

PR #1565, opened 2018-03-22
Standalone JS RFC

Pessimistress:
Proposed submodule `@deck.gl/js`

### DIFF
--- a/dev-docs/RFCs/v5.0/purejs-standalone-rfc.md
+++ b/dev-docs/RFCs/v5.0/purejs-standalone-rfc.md
@@ -1,0 +1,148 @@
+# RFC: Standalone version
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


dev-docs/RFCs/v5.0/purejs-standalone-rfc.md
```diff
@@ -0,0 +1,148 @@
...
+```
+
+Hello world example:
+```
+const data = [
+  { pickup: [-122.42, 37.8], dropoff: [-122.48, 37.76] },
+  { pickup: [-122.43, 37.8], dropoff: [-122.42, 37.75] }
+];
+
+const deckgl = new DeckGL({
```
jianhuang01: I think the DeckGL is changed to be Deck on master

Pessimistress: As the title says, this is a pure JS version of deck.gl, published as a new npm module.

PR #1587, opened 2018-03-27
Experimental Floating Solid Polygon Layer

chrisgervang:
For #1525


chrisgervang:
I'm not sure why, but locally 52 tests fail. The tests don't seem to be related to my changes.. am I missing something?

jianhuang01:
@chrisgervang we are fixing the problem on master, there is a bug in recent luma.gl alpha releases, hardcode luma.gl version to "5.2.0-alpha.3" can resolve the test issue.

jianhuang01:
@chrisgervang The test should be OK now, pull the latest code, please. Thanks for reporting the issue.

Pessimistress:
@chrisgervang This PR is getting stale and probably too hard to rebase. There were two features discussed in the original issue:

- setting the base `z` per vertex, which is now supported by the 6.x releases;
- setting extrusion per vertex, which as far as I can tell is not addressed in this PR. We will continue to track the request.

PR #1606, opened 2018-03-30
allow to increase the stroke width of the lines in trips-layer

thuanmb:
Change to use TRIANGLE_TRIP to render the lines of paths in trips-layer.
For #1624

**Note**
Please note that the animation not that smooth. Because:
1. In the previous version, we use GL.LINES, which will draw a line by given set of points (A -> B ->C).
2. But to draw a line with GL.TRIANGLE_STRIP. We need a specific source and target for the lines. For example, to draw a path from A -> B -> C, I will need to split it into 2 lines: A -> B and B -> C. I think it’s the reason make the animation is worse than before.


jianhuang01:
@thuanmb Can you sign the agreement? Thanks!

Pessimistress:
Closing in favor of https://github.com/uber/deck.gl/pull/2708

PR #1647, opened 2018-04-05
fix bench browser test

jianhuang01:
#### Background
bench browser test is broken on master

#### Change List
- fix the tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,13 +43,14 @@
     "test-dist": "node test/start.js test-dist",
     "test-render": "node test/start.js render",
     "test-render-react": "node test/start.js render",
-    "test-render-browser": "webpack-dev-server --env.render --progress --hot --open",
+    "test-render-browser": "webpack-dev-server --config test/webpack.config.js --env.render --progress --hot --open",
     "test-browser": "node test/start.js test-browser",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -43,13 +43,14 @@
...
     "test-render-react": "node test/start.js render",
-    "test-render-browser": "webpack-dev-server --env.render --progress --hot --open",
+    "test-render-browser": "webpack-dev-server --config test/webpack.config.js --env.render --progress --hot --open",
     "test-browser": "node test/start.js test-browser",
     "test-browser-local": "webpack-dev-server --config test/webpack.config.js --env.test_browser --progress --hot --open",
     "test-examples": "node test/node-examples.js",
     "test-size-es6": "npm run build-es6 && NODE_ENV=production webpack --config test/webpack.config.js --env.import-nothing --env.es6",
     "bench": "node test/start.js bench",
-    "bench-browser": "webpack-dev-server --env.bench --progress --hot --open",
+    "bench-browser": "node test/start.js bench-browser",
```
Pessimistress: Do we need to run bench with test?

jianhuang01: don't need for now, could be in future, the tests launched from test/start.js eventually will replace those launched by webpack-dev-server.

PR #1650, opened 2018-04-05
fix broken test-dist

jianhuang01:
#### Background
npm run test-dist is down

#### Change List
- run test against deck.gl es6 build


jianhuang01:
drop this PR cause team has one big refactor ongoing and it will resolve this issue.

PR #1669, opened 2018-04-06
regenerate font atlas when fontFamily changed

gnavvy:
Follow up PR for #976 

#### Change List
- regenerate font atlas when fontFamily changed

### DIFF
--- a/modules/core/src/core-layers/text-layer/text-layer.js
+++ b/modules/core/src/core-layers/text-layer/text-layer.js
@@ -43,7 +43,7 @@ const defaultProps = {
   fontFamily: DEFAULT_FONT_FAMILY,
 
   getText: x => x.text,
-  getPosition: x => x.position || x.coordinates,
+  getPosition: x => x.position,
   getColor: x => x.color || DEFAULT_COLOR,
   getSize: x => x.size || 32,
   getAngle: x => x.angle || 0,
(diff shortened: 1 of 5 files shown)


modules/core/src/core-layers/text-layer/text-layer.js
```diff
@@ -54,23 +54,32 @@ const defaultProps = {
...
   initializeState() {
-    const {gl} = this.context;
-    const {fontFamily} = this.props;
-    const {mapping, texture} = makeFontAtlas(gl, fontFamily);
-    this.state = {
-      iconAtlas: texture,
-      iconMapping: mapping
-    };
+    this.state = {};
+    this.updateFontAtlas(this.props.fontFamily);
```
Pessimistress: This should be removed. `oldProps.fontFamily !== props.fontFamily` is triggered on initialization.

modules/core/src/core-layers/text-layer/text-layer.js
```diff
@@ -54,23 +54,32 @@ const defaultProps = {
...
   }
 
   updateState({props, oldProps, changeFlags}) {
     if (
       changeFlags.dataChanged ||
+      oldProps.fontFamily !== props.fontFamily ||
```
Pessimistress: `transformStringToLetters()` only changes when text is changed, so `fontFamily` has no effect.

Instead, `MultiIconLayer` must invalidate `calculateInstanceOffsets` when `iconMapping` changes.

Pessimistress:
I don't think MultiIconLayer properly invalidates instanceOffsets when font changes.
Can you add a font selector in layer-browser to test?

gnavvy:
Added a control for fontFamily testing in layer-browser.
It seems only monospaced font works for now, the offset calculation for non-monospaced fonts like Helvetica is messed even before fontFamily changes, should not be the instanceOffsets invalidatoin issue.

PR #1671, opened 2018-04-08
Publish standalone bundle with the main package

Pessimistress:
#### Background
- Publish the standalone bundle with the main deck.gl package so that it's accessible via unpkg.com
- "lite" submodule only builds to bundle
- Add tests for standalone bundle

#### Change List
- Include the standalone bundle with the main deck.gl package so that it's accessible via unpkg.com
- Add tests for standalone bundle
- `modules/lite` no longer targets to publish itself to npm. Remove build hacks regarding react-map-gl.


test/src/index.js
```diff
@@ -29,4 +29,5 @@ import './core-layers';
 // TODO - Tests currently only work in browser
 if (typeof document !== 'undefined') {
   require('./experimental-layers');
+  require('./lite');
```
Pessimistress: This is broken under node when importing luma.gl (tries to access `document`). Need to fix luma.

modules/lite/src/index.js
```diff
@@ -1,13 +1,12 @@
...
 import * as lumaGL from 'luma.gl';
 import * as deckGLCore from '@deck.gl/core';
 
 import DeckGL from './deckgl';
 
-if (typeof window !== 'undefined') {
-  window.deck = Object.assign({}, window.deck, deckGLCore, {DeckGL});
-  window.luma = Object.assign({}, window.luma, lumaGL);
-}
+const _global = typeof window === 'undefined' ? global : window;
```
Pessimistress: Support non-browser import; aligned with `core/lib/init.js`

PR #1680, opened 2018-04-09
Support non-monospace fonts in TextLayer

Pessimistress:
For https://github.com/uber/deck.gl/issues/1679

Before & After:
text-layer-font

#### Change List
- Support non-monospace fonts in TextLayer
- Fix minor updateTrigger issues

### DIFF
--- a/modules/core/src/core-layers/text-layer/multi-icon-layer/multi-icon-layer.js
+++ b/modules/core/src/core-layers/text-layer/multi-icon-layer/multi-icon-layer.js
@@ -23,8 +23,8 @@ import IconLayer from '../../icon-layer/icon-layer';
 import vs from './multi-icon-layer-vertex.glsl';
 
 const defaultProps = {
-  getIndexOfIcon: x => x.index || 0,
-  getNumOfIcon: x => x.len || 1,
+  getShiftInQueue: x => x.shift || 0,
+  getLengthOfQueue: x => x.len || 1,
   // 1: left, 0: middle, -1: right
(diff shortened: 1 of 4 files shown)


modules/core/src/core-layers/text-layer/text-layer.js
```diff
@@ -107,6 +110,14 @@ export default class TextLayer extends CompositeLayer {
     this.setState({data: transformedData});
   }
 
+  getIconOffset(datum) {
```
gnavvy: getIconOffset => getLetterOffset? in the context of the text-layer

modules/core/src/core-layers/text-layer/multi-icon-layer/multi-icon-layer.js
```diff
@@ -23,8 +23,8 @@ import IconLayer from '../../icon-layer/icon-layer';
 import vs from './multi-icon-layer-vertex.glsl';
 
 const defaultProps = {
-  getIndexOfIcon: x => x.index || 0,
-  getNumOfIcon: x => x.len || 1,
+  getShiftInQueue: x => x.shift || 0,
```
gnavvy: use offset _instead_ of _shift_ for consistency with the _getPixelOffset_ API?
(unless other multi-icon-layer all uses _.shift_)

Pessimistress: The naming is really confusing. Basically `shift` is applied before scaling to font size, and `pixelOffset` is applied as-is.

PR #1681, opened 2018-04-10
replace deck.gl-layers with @deck.gl/experimental-layers

jianhuang01:
#### Background
deck.gl-layers are replaced by @deck.gl/experimental-layers

#### Change List
- Replace deck.gl-layers with new name
- Fix a dependency issue

### DIFF
--- a/examples/layer-browser/src/examples/experimental-layers.js
+++ b/examples/layer-browser/src/examples/experimental-layers.js
@@ -4,9 +4,7 @@ import {
   PathMarkerLayer,
   AdvancedTextLayer
 } from '@deck.gl/experimental-layers';
-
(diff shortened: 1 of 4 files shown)


examples/layer-browser/src/examples/experimental-layers.js
```diff
@@ -1,6 +1,10 @@
-import {MeshLayer, PathOutlineLayer, PathMarkerLayer, AdvancedTextLayer} from 'deck.gl-layers';
+import {
+  MeshLayer,
+  PathOutlineLayer,
+  PathMarkerLayer,
+  AdvancedTextLayer
+} from '@deck.gl/experimental-layers';
```
Pessimistress: Merge with the next line?

PR #1685, opened 2018-04-11
Fix plot example in website

Pessimistress:
For https://github.com/uber/deck.gl/issues/1684

#### Change List
- Remove reference to removed component `OrbitController`
- Support both standalone and website use case

### DIFF
--- a/website/src/components/demos/plot-demo.js
+++ b/website/src/components/demos/plot-demo.js
@@ -38,28 +38,9 @@ export default class PlotDemo extends Component {
     super(props);
 
     this.state = {
-      viewport: {
-        lookAt: [0, 0, 0],
-        distance: 3,
-        rotationX: -30,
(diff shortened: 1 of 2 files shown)


website/src/components/demos/plot-demo.js
```diff
@@ -90,32 +71,18 @@ export default class PlotDemo extends Component {
...
-    const canvasProps = {
-      ...viewport,
-      width,
-      height
-    };
-    const orbitViewport = OrbitController.getViewport(canvasProps);
+    const {equation, hoverInfo} = this.state;
 
     return (
-      <OrbitController
```
jianhuang01: can't support user interaction in plot demo anymore?

Pessimistress: This  is moved into the overlay component.

PR #1693, opened 2018-04-13
Support characterSet prop in TextLayer

Pessimistress:
For https://github.com/uber/deck.gl/issues/1691

#### Change List
- Fix exception if character is missing. Warn in console instead.
- Add prop `characterSet` for TextLayer
- Update docs

### DIFF
--- a/modules/core/src/core-layers/text-layer/text-layer.js
+++ b/modules/core/src/core-layers/text-layer/text-layer.js
@@ -18,9 +18,10 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-import {CompositeLayer} from '../../core';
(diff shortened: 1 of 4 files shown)


modules/core/src/core-layers/text-layer/text-layer.js
```diff
@@ -37,10 +38,13 @@ const ALIGNMENT_BASELINE = {
 const DEFAULT_FONT_FAMILY = 'Monaco, monospace';
 const DEFAULT_COLOR = [0, 0, 0, 255];
 
+const MISSING_CHAR_WIDTH = 32;
```
gnavvy: change to DEFAULT_CHAR_WIDTH and set as the fallback for getSize?

Pessimistress: I do think we need to redesign how size is handled. Maybe in a follow up PR.

PR #1699, opened 2018-04-16
Fix pixel to clipspace projection

Pessimistress:
For https://github.com/uber/deck.gl/issues/1695

The fix turns out to affect many more layers. This will affect render results of existing applications. We need a compatibility strategy to move forward.

#### Change List
- Fix bug in `project_pixel_to_clipspace` shader util
- Fix render test by cutting pixel values in half
- Note that this fix exposes another `TextLayer` bug in text anchor calculation (every character is shifted 1/4 width to the left in the new golden image) that we need to address in a follow up PR.


gnavvy:
Looking at the changelog we added the _project_pixel_to_clipspace_ support from v5, but I guess the shader logic was there before v5, so likely v4+ users using the following layers will be affected:

Core layers
- arc-layer
- icon-layer
- line-layer
- point-cloud-layer
- text-layer

Experimental layers and layers only used in examples
- arc-brushing-layer
- plot-layer
- advanced-text-layer
- bezier-curve-layer

jianhuang01:
Looks like the primarily impact is the width and radius, I think it is fine to release it, just highlight in what's new, we didn't break anything.

PR #1718, opened 2018-04-19
Update test-utils website pages

jianhuang01:
#### Background
Test-utils website pages are outdated

#### Change List
- Update test-utils docs

### DIFF
--- a/website/contents/pages.js
+++ b/website/contents/pages.js
@@ -491,7 +491,7 @@ export const docPages = generatePath([
       },
       {
         name: 'TestRenderer',
-        content: getDocUrl('api-reference/test-utils/test-renderer.md')
+        content: getDocUrl('api-reference/test-utils/renderer-test.md')
       }
     ]
   }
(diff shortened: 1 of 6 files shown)


website/contents/pages.js
```diff
@@ -481,7 +481,7 @@ export const docPages = generatePath([
       },
       {
         name: 'TestRenderer',
-        content: getDocUrl('api-reference/test-utils/test-renderer.md')
+        content: getDocUrl('api-reference/test-utils/renderer-test.md')
```
Pessimistress: If the class is named TestRenderer let's name the file `test-renderer.md`

Pessimistress:
Can you do a search for `deck.gl-test-utils` in the entire docs directory? I believe there are more mentions.

website/contents/pages.js
jianhuang01: The class is named as RenderTest

jianhuang01:
@Pessimistress Thanks for reminding, it is done.

website/contents/pages.js
Pessimistress: Then you should update the name of this page...

Issue #1749, opened 2018-04-27
Add link to source code of examples in API documentation


pbeshai:
#### Background 
Would be nice if when visiting the API docs there was an easy way to see the source code of the example on the page. For example, look at the [ScatterplotLayer](http://uber.github.io/deck.gl/#/documentation/deckgl-api-reference/layers/scatterplot-layer) documentation. The vis at the top of the screen is pretty nice and interactive, but it's not obvious how to see the source code behind it as an example. 

I was curious to see how the hover behavior and autohighlight worked in this case and was unable to find the source to the example to see.

#### To Do List 
  - [ ] Add label and assign to milestone 
  - [ ] Coding
  - [ ] Doc update
  - [ ] Whats new update 
  - [ ] Test

ianshan0915:
Hey @jianhuang01, I'm trying to work on this. but not sure about the link to source code. is [this link](https://github.com/uber/deck.gl/blob/master/website/src/components/demos/layer-demos.js) the right one? thanks

jianhuang01:
@ianshan0915 Yes, that's the source code for layer demos.

PR #1772, opened 2018-05-04
Add link to source code of examples in API documentation

ianshan0915:
For #1749


Pessimistress:
A little history here: a long time ago, all our examples lived in the `website` directory. When users tried to inspect how they were implemented, the complexity of the one-page app really threw them off. We have since moved all examples to their own minimal, self-contained projects in the `examples` directory. We regularly polish the examples and make sure the code is clean and up to date. Meanwhile, the website code base remains a little messy.

That's why I'm reluctant to add any public link back to the website code. I do agree that the layer docs do not provide any meaningful example code in themselves. I suggest we copy the actual props used from the embedded demo to the markdown itself, i.e. this code snippet in [ArcLayer API](https://uber.github.io/deck.gl/#/documentation/deckgl-api-reference/layers/arc-layer):

```js
  const layer = new ArcLayer({
    id: 'arc-layer',
    data,
    strokeWidth: 2
  });
```

Becomes

```js
  const layer = new ArcLayer({
    id: 'arc-layer',
    data,
    pickable: true,
    strokeWidth: 12,
    getSourcePosition: d => d.from.coordinates,
    getTargetPosition: d => d.to.coordinates,
    getSourceColor: d => [Math.sqrt(d.inbound), 140, 0],
    getTargetColor: d => [Math.sqrt(d.outbound), 140, 0],
    onHover: ({object}) => setTooltip(`${object.from.name} to ${object.to.name}`)
  });
```

PR #1774, opened 2018-05-05
copy the actual props from the embedded demos to their markdowns

ianshan0915:
For #1749 
Sorry that I could not make it work in the old PR, so I create a new one. 
@Pessimistress I did what you suggested.


Pessimistress:
@ianshan0915 Thank you! Can you also update the "example data" in the comments to match the new props?

PR #1806, opened 2018-05-14
Add Transition class

Pessimistress:
#### Background
Consolidate shared code in TransitionManager and AttributeTransitionManager into Transition class.

#### Change List
- Add Transition class
- Add tests

Manually tested: attribute-transition, viewport-transition, viewport-transition-flyto test apps


modules/core/src/controllers/transition-manager.js
```diff
@@ -128,62 +113,53 @@ export default class TransitionManager {
...
+      onUpdate: this._onTransitionUpdate,
+      onInterrupt: this._onTransitionEnd(endProps.onTransitionInterrupt),
+      onEnd: this._onTransitionEnd(endProps.onTransitionEnd)
+    });
+
+    this._onAnimationFrame();
   }
 
-  _onTransitionFrame() {
+  _onAnimationFrame() {
```
1chandu: Super nit: stick with 'Transition' , instead of rename?

modules/core/src/transitions/transition.js
```diff
@@ -0,0 +1,96 @@
...
+  get state() {
+    return this._state;
+  }
+
+  get inProgress() {
+    return this._state === TRANSITION_STATE.PENDING ||
+      this._state === TRANSITION_STATE.IN_PROGRESS;
+  }
+
+  start(props) {
```
1chandu: Add one line comment for all public methods?

Issue #1828, opened 2018-05-18
GPUScreenGridLayer test fails with latest Attribute changes


1chandu:
#### Actual Result
Render test should pass
#### Expected Result
Render test should pass
#### Reproduce Steps
Run render tests on latest master.
#### To Do List 
  - [X] Add label and assign to milestone
  - [ ] Coding
  - [ ] Test

PR #1837, opened 2018-05-21
AttributeManger: Fix external buffer support.

1chandu:
For #1828 

#### Background
- Fix buffer setup code in Attribute class.
- Update GPUScreenGridLayer to setup external buffer.
- Enable render tests.

#### Change List
-  Fix external buffer support.

### DIFF
--- a/modules/core/src/lib/attribute.js
+++ b/modules/core/src/lib/attribute.js
@@ -2,6 +2,13 @@
 import assert from '../utils/assert';
 import {GL, Buffer, experimental} from 'luma.gl';
 const {Attribute} = experimental;
+const DEFAULT_STATE = {
+  isExternalBuffer: false,
+  needsAlloc: false,
+  needsUpdate: false,
+  needsRedraw: false,
+  allocedInstances: -1
+};
 
 export default class LayerAttribute extends Attribute {
   constructor(gl, opts = {}) {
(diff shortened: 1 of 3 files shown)


modules/core/src/lib/attribute.js
```diff
@@ -191,8 +190,6 @@ export default class LayerAttribute extends Attribute {
           state.needsRedraw = true;
         }
       }
-    } else {
-      state.isExternalBuffer = false;
```
Pessimistress: What happens when an attribute is switched from using external buffer to using own update function? This flag is never cleared.

modules/experimental-layers/src/gpu-screen-grid-layer/gpu-screen-grid-layer.js
```diff
@@ -59,12 +59,12 @@ export default class GPUScreenGridLayer extends Layer {
...
     attributeManager.addInstanced({
       instancePositions: {size: 3, update: this.calculateInstancePositions},
-      // TODO: Needed, since attributeManger.update() only updates existing attributes.
       instanceCounts: {
         size: 4,
         transition: true,
         accessor: ['getPosition', 'getWeight'],
-        update: this.calculateInstanceCounts
+        update: this.calculateInstanceCounts,
+        isExternalBuffer: true
```
Pessimistress: `isExternalBuffer` was designed to be an internal state to `LayerAttribute`. The intention was to allow the user to override the attribute values via layer props. This feels like a hack.

1chandu:
Closing , better fix: https://github.com/uber/deck.gl/pull/1846

PR #1846, opened 2018-05-22
Attribute: Add custom Buffer support.

1chandu:
For #1828 

#### Background
Existing `External Buffer` support for Attributes is only through `Layer` props. This state is non persistent, and when buffer is not supported for an attribute through props, it will switch back to normal attribute where `AttributeManger` creates the Buffer object.
For GPGPU layers, we need to support another way of providing `Buffer` object, which will persist until the layer resets it. Added a new flag `isCustomBuffer` to `Attribute` class.

#### Change List
- Attribute: Add custom buffer support.
- Enable GPUScreenGridLayer rendering tests.

### DIFF
--- a/modules/core/src/lib/attribute.js
+++ b/modules/core/src/lib/attribute.js
@@ -2,6 +2,13 @@
 import assert from '../utils/assert';
 import {GL, Buffer, experimental} from 'luma.gl';
 const {Attribute} = experimental;
+const DEFAULT_STATE = {
+  isExternalBuffer: false,
+  needsAlloc: false,
+  needsUpdate: false,
+  needsRedraw: false,
+  allocedInstances: -1
+};
 
 export default class LayerAttribute extends Attribute {
   constructor(gl, opts = {}) {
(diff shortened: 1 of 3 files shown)


modules/experimental-layers/src/gpu-screen-grid-layer/gpu-screen-grid-layer.js
```diff
@@ -136,8 +136,8 @@ export default class GPUScreenGridLayer extends Layer {
   }
 
   calculateInstanceCounts(attribute, {numInstances}) {
-    // Empty method: Attribute manager requires an update method for each added attribute
-    // TODO: remove this when proper external buffer support is added to Attribute Manager.
+    const {aggregatedData} = this.state;
+    attribute.setExternalBuffer(aggregatedData.countsBuffer, numInstances);
```
Pessimistress: `attribute.update({buffer: aggregatedData.countsBuffer})` should work. `setExternalBuffer` is for attributeManager and I don't feel comfortable accessing it in layers.

modules/core/src/lib/attribute.js
```diff
@@ -79,14 +79,16 @@ export default class LayerAttribute extends Attribute {
...
 
-    if (!state.isExternalBuffer) {
-      // Do we need to reallocate the attribute's typed array?
-      const instanceCount = this.getInstanceCount();
-      const needsAlloc = instanceCount === 0 || instanceCount < numInstances;
-      if (needsAlloc && (state.update || state.accessor)) {
-        state.needsAlloc = true;
-        this.setNeedsUpdate(this.id);
-      }
+    if (state.isExternalBuffer || state.isCustomBuffer) {
```
Pessimistress: Is this the only place this new state is used? How about just use the `noAlloc` flag?

1chandu: yeah, `noAlloc` should work fine.

modules/experimental-layers/src/gpu-screen-grid-layer/gpu-screen-grid-layer.js
1chandu: Updated.

PR #1922, opened 2018-06-15
Update luma.gl dependencies in modules

1chandu:
For #

#### Background
When 6.0.0-alpha.1 published, I bumped luma version to 6.0.0-alpha.1 in core module, but other modules still referring to older luma version, manually change versions.

#### Change List
- Update luma.gl dependencies in modules


1chandu:
Actually they have dependency on core module, and I guess no conflict since it gets luma.gl from the core module, let me know if you still think we need to specify explicitly?

PR #1925, opened 2018-06-17
[PART-2] GPUGridAggregation : Add experimental GPUGridLayer.

1chandu:
For #1556 

- Copy existing GridLayer and GridCellLayers to experimental.
- Update layers to use GPUGridAggregator for world space aggregation.
- Add additional utility methods required to enable GridCellLayer to consume aggregated data.
- Add  unit tests.
- Add examples to layer-browser.
- 2 to 3X boost in FPS when using GPU Aggregation.
- Picking and Color/Elevation customizations done on CPU and are not supported in this new experimental layer, those feature need to added by moving more calculation to GPU or by using TF.

#### Change List
- Add new experimental GPUGridLyaer, uses GPU Aggregation

### DIFF
--- a/modules/experimental-layers/src/gpu-grid-layer/gpu-grid-layer.js
+++ b/modules/experimental-layers/src/gpu-grid-layer/gpu-grid-layer.js
@@ -30,7 +30,6 @@ const CPU_MAXCOLOR = [255, 0, 0, 255];
 const GPU_MAXCOLOR = [0, 255, 0, 255];
 
 const defaultProps = {
-
   // elevation
   elevationScale: 1,
 
@@ -57,7 +56,7 @@ export default class GPUGridLayer extends CompositeLayer {
       shaderCache: this.context.shaderCache
     };
     this.state = {
-      gpuGridAggregator: new GPUGridAggregator(gl, options),
+      gpuGridAggregator: new GPUGridAggregator(gl, options)
     };
   }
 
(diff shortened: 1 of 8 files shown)


1chandu:
@ibgreen 
> Before reviewing in detail: Was the decision in today's meeting that we would update the official layers instead, assuming that the GPU aggregation can easily be turned off with a single prop?

Yes thats the plan for ScreenGridLayer. For GridLayer, GPU version is not on parity with CPU version (picking, color/elevation scale need to re-implemented for GPU), so GPUGridLayer will remain experimental until those features ported into shaders.

For ScreenGridLayer, there is some work needed (colorDomain/range) , aiming to finish that this week, so we can replace existing ScreenGridLayer with GPUScreenGridLayer (in experimental), please note there are changes to CPU path as well, due to attribute data format has to match between CPU and GPU.

modules/experimental-layers/src/gpu-grid-layer/gpu-grid-layer.js
```diff
@@ -0,0 +1,138 @@
...
+    const reprojectNeeded = this.needsReProjectPoints(oldProps, props);
+
+    if (changeFlags.dataChanged || reprojectNeeded) {
+      // project data into hexagons, and get sortedBins
+      this.getLayerData();
+    }
+  }
+
+  needsReProjectPoints(oldProps, props) {
+    return oldProps.cellSize !== props.cellSize || oldProps.gpuAggregation !== props.gpuAggregation;
```
heshan0131: needs reproject also need to check whether `getPosition` updateTrigger

1chandu: Addressed, looks like this needed for GridLayer too #1953  ?

Issue #1415, opened 2018-02-09
GPGPU: New layer to render contours.


1chandu:
- This will require : #1414 

#### To Do List 
  - [X] Add label and assign to milestone 
  - [x] Coding
  - [x] Test
  - [x] Docs

1chandu:
Planning: Moving to later milestone.

chrisgervang:
This would be very handy for rendering ground textures (e.g. terrain mapbox api) :)

1chandu:
Need docs.

PR #1938, opened 2018-06-20
RFC: Contour Layer

1chandu:
For #1415 

#### Background
Adding RFC for contour layer.

#### Change List
- RFC: Contour Layer


dev-docs/RFCs/v6.1/contour-layer-rfc.md
```diff
@@ -0,0 +1,75 @@
...
+Kepler and Manifold applications.
+
+
+## Proposed
+
+A new layer with following key props.
+
+* contours (Array): An array with each element containing a threshold and color values. Each element results in at most one contour with specified color.
+
+* cellSize (Number): Cell size in meters, used for aggregating data.
```
gnavvy: in meters? can we specify screen pixel values?

dev-docs/RFCs/v6.1/contour-layer-rfc.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Next Steps
+
+### GPU Acceleration
+
+Contour vertex data generation can be optimized by performing it on GPU, this is currently being explored.
+
+### Screen space contours
+
+This RFC propose world space (lng/lat) layer, but can easily be extended into screen space. `GPUAggregator` already supports screen space aggregation, whats needed is screen space `LineLayer`.
```
gnavvy: We already have the "screen space" LineLayer with coordinateSystem set to IDENTIFY.
Is there any other constraints that prevent from using this layer as it is for non-geospatial applications?

dev-docs/RFCs/v6.1/contour-layer-rfc.md
```diff
@@ -0,0 +1,75 @@
...
+
+Contour layer with 3 sets of contours.
+
+```
+props: {
+  id: 'contourLayer',
+  contours: [
+    {threshold: 1, color: [255, 0, 0]},
+    {threshold: 5, color: [0, 255, 0]},
+    {threshold: 10, color: [0, 0, 255]},
```
gnavvy: prefer _interval_ or just _value_, but not a big issue.

dev-docs/RFCs/v6.1/contour-layer-rfc.md
1chandu: yes do able for screen space layer. To be precise, this in the same space where aggregation happens, if aggregating in lng/lat it will be in meters (converted to lng/lat offsets), if aggregating in screen space, it will be in pixels. Will update once screen space version is ready.

dev-docs/RFCs/v6.1/contour-layer-rfc.md
1chandu: No other known constrains, just hooking it up.

Issue #1972, opened 2018-06-27
GPUScreenGridLayer: not working when view type is OrthographicView.


1chandu:
When GPUScreenGridLayer used in graph.gl , aggregation results don't match with ScreenGridLayer.

#### To Do List 
  - [X] Add label and assign to milestone
  - [x] Coding
  - [x] Test

PR #1973, opened 2018-06-27
Fix aggregation in OrthoGraphicView

1chandu:
For #1972 

#### Background
Perform map projections only for 'WebMercatorViewport'.

#### Change List
- Fix aggregation in OrthoGraphicView
- Fix aggregation when viewport size changes.
- Add render test


1chandu:
> Why would we not want to project points on other viewport types

It is the map projections (same as what `project_position` does).

>Is it a question of the project functions not being properly implemented for non-mercator viewports, then maybe we should open a bug on that.

Not really, this flag just controls whether to do map projections or not in aggregator shader.

I initially wanted to name it something like `mapProjection` but named it similar to `project_position` shader module method. By the way this is still experimental API and we can change it.

PR #1990, opened 2018-07-02
Buffer data removal

Pessimistress:
For https://github.com/uber/deck.gl/issues/1974

#### Change List
- Remove reference to `buffer.data`

### DIFF
--- a/modules/core/src/lib/attribute-transition-utils.js
+++ b/modules/core/src/lib/attribute-transition-utils.js
@@ -108,10 +108,3 @@ export function padBuffer({
 
   fromState.setData({data});
 }
-
-export function getBufferLength(buffer) {
-  if (buffer.data) {
-    return buffer.bytes / buffer.data.BYTES_PER_ELEMENT;
-  }
-  return 0;
-}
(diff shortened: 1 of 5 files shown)


modules/core/src/lib/attribute-transition-utils.js
```diff
@@ -107,3 +108,10 @@ export function padBuffer({
 
   fromState.setData({data});
 }
+
+export function getBufferLength(buffer) {
+  if (buffer.data) {
```
1chandu: Is this temporary? wouldn't this always be false once luma PR lands?

Pessimistress: Doesn't it keep some data for debug purpose? If not, can we store the element length in the Buffer object?

1chandu: Yes we will be adding getLength support to Buffer eventually, it is going to be part of this PR : https://github.com/uber/luma.gl/pull/423

1chandu: if all you need is 'BYTES_PER_ELEMENT', can you use `getTypedArrayFromGLType` + buffer.accessor.type ? We can clean this up once we have full support in luma.gl

1chandu: Added support in luma.gl https://github.com/uber/luma.gl/pull/592

PR #1994, opened 2018-07-03
rerender sublayers when state change

Pessimistress:
For https://github.com/uber/deck.gl/issues/1993

#### Change List
- add `stateChanged` to change flags
- composite layer marks as needs update when `setState` is called
- default layer behavior is not affected

### DIFF
--- a/modules/core/src/lib/composite-layer.js
+++ b/modules/core/src/lib/composite-layer.js
@@ -38,6 +38,12 @@ export default class CompositeLayer extends Layer {
   // Provide empty definition to disable check for missing definition
   initializeState() {}
 
+  // Updates selected state members and marks the composite layer to need rerender
+  setState(updateObject) {
+    super.setState(updateObject);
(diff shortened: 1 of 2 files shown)


modules/core/src/lib/composite-layer.js
```diff
@@ -38,6 +38,12 @@ export default class CompositeLayer extends Layer {
   // Provide empty definition to disable check for missing definition
   initializeState() {}
 
+  // Updates selected state members and marks the composite layer to need rerender
+  setState(updateObject) {
+    super.setState(updateObject);
+    this.setLayerNeedsUpdate();
```
1chandu: this is will make `layermanager` to re-render all existing layers right? Instead can we just iterate over the layers of the composite-layer mark them to re-draw?

1chandu: I see, I also found the note in `updateLayers` method. Thanks for the details.

PR #2005, opened 2018-07-06
fix point cloud examples

jianhuang01:
For #1873

#### Change List
- Fix the two point cloud examples

### DIFF
--- a/examples/website/point-cloud-laz/app.js
+++ b/examples/website/point-cloud-laz/app.js
@@ -101,18 +101,19 @@ class Example extends PureComponent {
   _onResize() {
     this.setState({width: window.innerWidth, height: window.innerHeight});
 
-    const newViewState = Object.assign({}, this.viewState, {
-      distance: OrbitView.getDistance({
-        boundingBox: [1, 1, 1],
-        fov: this.state.viewState.fov
-      })
-    });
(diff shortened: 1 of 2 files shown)


examples/website/point-cloud-ply/app.js
```diff
@@ -11,7 +11,7 @@ const FILE_PATH = 'examples/point-cloud-ply/lucy100k.ply';
 
 const INITIAL_VIEW_STATE = {
   lookAt: [0, 0, 0],
-  distance: 100,
+  distance: 2,
```
jianhuang01: there is a global camera projection change which I believe is a regression somewhere, for now let's change the distance to make this example work, need investigate where the real regression is.

examples/website/point-cloud-laz/app.js
```diff
@@ -157,7 +158,7 @@ class Example extends PureComponent {
       <DeckGL
         width={width}
         height={height}
-        views={view}
+        views={[view]}
```
Pessimistress: `views` accepts both View object and array.

jianhuang01: if that's the case, there is a bug in our core module which doesn't accept a view object.

PR #2007, opened 2018-07-07
Fix website viewport transition

Pessimistress:
For https://github.com/uber/deck.gl/issues/1873

#### Change List
- Fix viewport transition between demos
- Fix home demo size


website/src/components/demo-launcher.js
```diff
@@ -1,34 +1,14 @@
 import React, {Component} from 'react';
 import {connect} from 'react-redux';
 import autobind from 'autobind-decorator';
+import {FlyToInterpolator} from 'react-map-gl';
```
1chandu: Why not use deck.gl's version?

Pessimistress: The website examples interaction is using the react-map-gl controller.

Issue #2012, opened 2018-07-09
Picking documentation needs an update


NikhilM98:
Changes required:

1.  Instead of `instancePickingColors`, the attribute key is `pickingColors` as declared inside the attributeManager. In the Vertex Shader code snippet, the `instancePickingColors` attribute needs to be changed to `pickingColors`.
2. `instancePickingColors` attribute is used inside Fragment Shader unnecessarily. Also, attributes cannot be used inside Fragment Shader.

PR #2013, opened 2018-07-09
Documentation correction picking.md

NikhilM98:
For #2012 

#### Change List
- `instancePickingColors` replaced with `pickingColors`
- Use of `instancePickingColors` attribute in Fragment Shader removed


1chandu:
@NikhilM98 , thanks for the PR, but this change is not needed, the attribute name is `instancePickingColros` , ref: https://github.com/uber/deck.gl/blob/b8f6aeb793c6e609ef3f3005aaa2d67e6947418b/modules/core/src/lib/layer.js#L714

PR #2017, opened 2018-07-09
Various trips demo issue fixes

Pessimistress:
For https://github.com/uber/deck.gl/issues/1873

#### Change List
- Consolidate timer
- Fix trips layer in Website not rendered bug: in demos with multiple data sources, prevent prop fallback during loading


website/src/utils/animation-utils.js
```diff
@@ -1,20 +0,0 @@
-/* global window */
```
1chandu: This file is used here too: https://github.com/uber/deck.gl/blob/c534253cdad3cc26abf358d3b5017c4e28a3de9b/website/src/components/home.js#L8

Pessimistress: Removed in this PR

1chandu: oh yes, all good, thanks!

PR #2020, opened 2018-07-10
Fix TransitionManager crash on start

Pessimistress:
#### Change List
- Pass `viewState` when constructing `TransitionManager`
- Do not merge view states
- Update viewport-transitions test app to v6 API


modules/core/src/controllers/transition-manager.js
```diff
@@ -45,7 +47,7 @@ export default class TransitionManager {
     let transitionTriggered = false;
     const currentProps = this.props;
     // Set this.props here as '_triggerTransition' calls '_updateViewport' that uses this.props.
-    nextProps = Object.assign({}, this.props, nextProps);
+    nextProps = Object.assign({}, DEFAULT_PROPS, nextProps);
```
1chandu: Wouldn't this trigger invalid transition in following cases :
```
DEFAULT_PROPS  = {a: 1, b: 1}
this.props = {a: 10, b: 10}
passed in nextProps = {a: 20}
```
triggers transition between {a:10 b: 10} to {a:20, b: 1} , but actually we should trigger transition {a: 10, b:10} to {a: 20, b: 10}

test/apps/viewport-transitions/app.js
```diff
@@ -1,9 +1,8 @@
-/* global window,document */
+/* global document */
```
1chandu: can you update `viewport-transitions-flyTo`, should be same set of changes.

PR #2021, opened 2018-07-10
Fix plot example

jianhuang01:
For #1873

#### Change List
- cast view object to array

### DIFF
--- a/modules/core/src/lib/deck.js
+++ b/modules/core/src/lib/deck.js
@@ -362,9 +362,8 @@ export default class Deck {
   // Get the view descriptor list
   _getViews(props) {
     // Default to a full screen map view port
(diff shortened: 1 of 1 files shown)


modules/core/src/lib/deck.js
```diff
@@ -362,7 +362,8 @@ export default class Deck {
   // Get the view descriptor list
   _getViews(props) {
     // Default to a full screen map view port
-    const views = props.views || [new MapView({id: 'default-view'})];
+    let views = props.views || [new MapView({id: 'default-view'})];
+    views = views instanceof Array ? views : [views];
```
Pessimistress: Use `Array.isArray`

Issue #2019, opened 2018-07-10
Orthographic view issues on 6.0


1chandu:
1. Zoom is broken
2. ScreenGridLayer rendering is bad on first instance, re-rendering fixes it, could be related to above point.

#### To Do List 
  - [x] Add label and assign to milestone
  - [x] Coding
  - [x] Test

1chandu:
CC: @javidhsueh

1chandu:
Need to re-test with #2024 for item#2

javidhsueh:
I've confirmed that zoom works again

PR #2024, opened 2018-07-11
Fix view state comparison

Pessimistress:
#### Change List
- Fix bug where multi-view does not rerender
- Add tests


1chandu:
Also fixes #2019

PR #2026, opened 2018-07-11
Delete controller when view is removed

Pessimistress:
#### Change List
- Fix: first `controller.processViewportChange` throws error for missing width/height
- Delete controller when view is removed
- Move default view state to each view
- Properly finalize controller and transition manager

### DIFF
--- a/modules/core/src/views/view-manager.js
+++ b/modules/core/src/views/view-manager.js
@@ -25,15 +25,13 @@ import Viewport from '../viewports/viewport';
 import log from '../utils/log';
 import {flatten} from '../utils/flatten';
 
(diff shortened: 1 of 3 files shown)


modules/core/src/views/view-manager.js
```diff
@@ -232,26 +229,42 @@ export default class ViewManager {
...
         return viewport;
       });
 
+      // Remove unused controllers
+      for (const id in oldControllers) {
+        if (!this.controllers[id]) {
+          oldControllers[id].finalize();
+        }
+      }
+
```
1chandu: nit: add `oldControllers = {}` to remove holding any object references.

Pessimistress: `oldControllers` should be GC'ed at the end of this function.

Issue #2001, opened 2018-07-05
.shortestPathFrom is not a function


eric-kimbrel:
#### summary 
I am using an OrbitView / OrbitController  I get an error when double clicking on the canvas.

#### Actual Result:
transition-manager.js:119 Uncaught TypeError: (intermediate value).shortestPathFrom is not a function
    at TransitionManager._triggerTransition (transition-manager.js:119)
    at TransitionManager.processViewStateChange (transition-manager.js:75)
    at OrbitController.setProps (controller.js:189)
    at Deck.setProps (deck.js:202)
    at DeckGL._updateFromProps (deckgl.js:203)
    at DeckGL.componentWillReceiveProps (deckgl.js:89)
    at callComponentWillReceiveProps (react-dom.development.js:11527)
    at updateClassInstance (react-dom.development.js:11719)
    at updateClassComponent (react-dom.development.js:13153)
    at beginWork (react-dom.development.js:13824)
    at performUnitOfWork (react-dom.development.js:15863)
    at workLoop (react-dom.development.js:15902)
    at HTMLUnknownElement.callCallback (react-dom.development.js:100)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:138)
    at invokeGuardedCallback (react-dom.development.js:187)
    at replayUnitOfWork (react-dom.development.js:15310)
    at renderRoot (react-dom.development.js:15962)
    at performWorkOnRoot (react-dom.development.js:16560)
    at performWork (react-dom.development.js:16482)
    at performSyncWork (react-dom.development.js:16454)
    at requestWork (react-dom.development.js:16354)
    at scheduleWork$1 (react-dom.development.js:16218)
    at Object.enqueueSetState (react-dom.development.js:11299)
    at DeckglGraphView../node_modules/react/cjs/react.development.js.Component.setState (react.development.js:270)
    at DeckglGraphView../src/DeckGlGraphView/index.tsx.DeckglGraphView._onViewStateChange (index.tsx:116)
    at Deck._onViewStateChange (deck.js:441)
    at OrbitController.updateViewport (controller.js:270)
    at OrbitController._onDoubleTap (controller.js:501)
    at OrbitController.handleEvent (controller.js:120)
    at Array.<anonymous> (event-manager.js:342)
    at Manager.emit (hammer.js:2519)
    at TapRecognizer.emit (hammer.js:2128)
    at TapRecognizer.tryEmit (hammer.js:1555)
[...]

PR #2033, opened 2018-07-11
Make doubleTap and keyboard transitions exlusive to MapController

1chandu:
For #2001 

**NOTE**: I will port the changes to 6.0 and master.

#### Background
Make doubleTap and keyboard transitions exlusive to MapController

#### Change List
- Make doubleTap and keyboard transitions exlusive to MapController

### DIFF
--- a/modules/core/src/controllers/controller.js
+++ b/modules/core/src/controllers/controller.js
@@ -18,19 +18,11 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-import LinearInterpolator from '../transitions/linear-interpolator';
-import TransitionManager, {TRANSITION_EVENTS} from './transition-manager';
+import TransitionManager from './transition-manager';
 import assert from '../utils/assert';
 
 const NO_TRANSITION_PROPS = {
   transitionDuration: 0
-};
-
-const LINEAR_TRANSITION_PROPS = {
-  transitionDuration: 300,
-  transitionEasing: t => t,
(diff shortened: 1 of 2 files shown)


modules/core/src/controllers/controller.js
```diff
@@ -423,4 +422,9 @@ export default class Controller {
     return this.updateViewport(newControllerState, LINEAR_TRANSITION_PROPS);
   }
   /* eslint-enable complexity */
+
+  _getTransitionControls(props) {
```
Pessimistress: This seems strangely specific.
Instead of `LINEAR_TRANSITION_PROPS`, can just call a method `_getLinearTransitionProps`?

1chandu: We have two flags that control keyboard/double tap transitions, this method can be more customizable (to enable, disable one or both transitions) by child controllers.

Pessimistress: The crash is caused by transition, not the interaction itself. Force disabling the interactions is excessive.

modules/core/src/controllers/controller.js
```diff
@@ -383,7 +376,7 @@ export default class Controller {
     const isZoomOut = this.isFunctionKeyPressed(event);
 
     const newControllerState = this.controllerState.zoom({pos, scale: isZoomOut ? 0.5 : 2});
-    return this.updateViewport(newControllerState, LINEAR_TRANSITION_PROPS);
+    return this.updateViewport(newControllerState, this._transitionProps);
```
Pessimistress: If you call `_getTransitionProps()` instead of caching it, the method can return different results based on the interaction state.

PR #2034, opened 2018-07-11
Fix getCursor bugs

Pessimistress:
#### Background
In React, `props.getCursor` is not called when app state updates

#### Change List
- Update cursor in `renderFrame`
- fix cursor CSS for non-Chrome browsers

### DIFF
--- a/modules/core/src/utils/css-vendor-prefix.js
+++ b/modules/core/src/utils/css-vendor-prefix.js
@@ -1,0 +1,36 @@
+// Copyright (c) 2015 - 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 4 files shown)


modules/core/src/utils/css-vendor-prefix.js
```diff
@@ -0,0 +1,36 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+/* global document */
+let venderPrefix = '';
+
+// Get CSS vendor prefix
+if (typeof document !== 'undefined') {
```
1chandu: a unit test?

PR #2043, opened 2018-07-12
Multi-view picking perf: filter viewports by pointer position

Pessimistress:
#### Background
`pickLayers` should only render the views that contain the picked position

#### Change List
- Add `viewport.containsPixel` method
- `viewManager.getViewports` accepts an optional pixel/rect as filter
- Tests

### DIFF
--- a/modules/core/src/views/view-manager.js
+++ b/modules/core/src/views/view-manager.js
@@ -81,7 +81,11 @@ export default class ViewManager {
 
   // Get a set of viewports for a given width and height
   // TODO - Intention is for deck.gl to autodeduce width and height and drop the need for props
-  getViewports() {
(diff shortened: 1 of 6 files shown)


modules/core/src/views/view-manager.js
```diff
@@ -81,7 +81,11 @@ export default class ViewManager {
 
   // Get a set of viewports for a given width and height
   // TODO - Intention is for deck.gl to autodeduce width and height and drop the need for props
-  getViewports() {
+  // @param rect ([object]) - x, y, [width], [height] of the pixel/area that the viewports should cover
```
1chandu: Nit: Update the comment, to reflect `rect` is optional, and returns all viewports when not provided.

Issue #2046, opened 2018-07-12
Make ViewState Transitions API official


1chandu:
#### Background 
Now that Controller refactor is done in 6.0, we can make Viewport Transitions API official.

#### To Do List 
  - [x] Add label and assign to milestone 
  - [x] Add support for OrbitController, Update an example use viewport transition.
  - [x] Rename exports : _Name to Name
  - [x] Doc update
       - [x]  Update view-state to specify all transitions props
       - [x]  Add docs for linear/fly to interpolators

  - [x] Whats new update 
  - [x] Port to 6.0 branch.

PR #2047, opened 2018-07-12
OrbitController: Add support for viewport transitions

1chandu:
For #2046

#### Background
OrbitController: Add support for viewport transitions
Update `point-cloud-ply` to use linear transitions for rotation

#### Change List
- OrbitController: Add support for viewport transitions

### DIFF
--- a/modules/core/src/controllers/orbit-controller.js
+++ b/modules/core/src/controllers/orbit-controller.js
@@ -2,6 +2,15 @@ import {clamp} from 'math.gl';
 import Controller from './controller';
 import OrbitViewport from '../deprecated/orbit-viewport';
 import assert from '../utils/assert';
+import LinearInterpolator from '../transitions/linear-interpolator';
+import {TRANSITION_EVENTS} from './transition-manager';
+
+const LINEAR_TRANSITION_PROPS = {
+  transitionDuration: 300,
+  transitionEasing: t => t,
+  transitionInterpolator: new LinearInterpolator(),
+  transitionInterruption: TRANSITION_EVENTS.BREAK
+};
 
 const defaultState = {
   lookAt: [0, 0, 0],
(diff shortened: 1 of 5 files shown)


modules/core/src/controllers/orbit-controller.js
```diff
@@ -2,6 +2,15 @@ import {clamp} from 'math.gl';
 import Controller from './controller';
 import OrbitViewport from '../deprecated/orbit-viewport';
 import assert from '../utils/assert';
+import LinearInterpolator from '../transitions/linear-interpolator';
+import {TRANSITION_EVENTS} from './transition-manager';
+
+const LINEAR_TRANSITION_PROPS = {
+  transitionDuration: 300,
+  transitionEasing: t => t,
+  transitionInterpolator: new LinearInterpolator(),
```
Pessimistress: I think the default props include longitude/latitude?

1chandu: Right, all we need is to transition 'zoom' on double-tap, we can update this to `zoom` only and also update `_getTransitionProps` to return these props for double-tap only?

Pessimistress: What about keyboard navigation?

modules/core/src/controllers/orbit-controller.js
```diff
@@ -293,4 +312,10 @@ export default class OrbitController extends Controller {
   constructor(props) {
     super(OrbitState, props);
   }
+
+  _getTransitionProps(event) {
+    // Enables Transitions on double-tap.
+    const props = event.type === 'doubletap' ? LINEAR_TRANSITION_PROPS : NO_TRANSITION_PROPS;
```
Pessimistress: Is this necessary? The transition would not be triggered unless zoom is changed.

PR #2053, opened 2018-07-13
Make ViewState Transitions API official

1chandu:
For #2046 

#### Background
- Move ViewState transition API exports from experimental to official
- Add docs and update `whats-new`.

#### Change List
- ViewportTranstions: Add docs, make exports official


Pessimistress:
Shall we rename `ViewportFlyToInterpolator` to `FlyToInterpolator`? We're moving away from the term viewport in favor of viewState, and it will align with react-map-gl.

PR #2076, opened 2018-07-17
Add RFC for Layer Intersections.

1chandu:
For #

#### Background
Propose an RFC for layer intersection operations.

#### Change List
- Add RFC for Layer Intersections.


Pessimistress:
"Intersection" implies a commutative operation, but this feature is not. I'd call it a "mask" layer after [this](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/mask).

PR #2081, opened 2018-07-17
Fix website in Firefox

Pessimistress:
For https://github.com/uber/deck.gl/issues/2070

Polyfill breaks async props (`fetch(...) instanceof Promise` returns `false`)

#### Change List
- Remove polyfill


1chandu:
Does this fix IconLayer demo?

Pessimistress:
@1chandu yes

Issue #2121, opened 2018-07-25
GPUAggregation: Returns invalid results on Intel GPU due to FP64 division not working.


1chandu:
#### Actual Result
GPU aggregation results do not match with CPU Aggregation.
#### Expected Result
GPU aggregation results should match with CPU Aggregation.
#### Reproduce Steps
Run ScreenGrid demo on Intel GPU (Intel Iris Pro 1536 or 6100).
#### To Do List 
  - [X] Add label and assign to milestone
  - [x] Coding
  - [x] Test

PR #2122, opened 2018-07-25
GPUAggregator: Set ONE uniform for fp64 module

1chandu:
For #2121

#### Background
On Intel GPUs `fp64` shader module uses `LUMA_FP64_CODE_ELIMINATION_WORKAROUND`, which uses a const uniform `ONE` always. For all layers this is provided as part of base `Layer` class. Set this uniform explicitly for GPU Aggregation path. 

#### Change List
- GPUAggregator: Set ONE uniform for fp64 module

### DIFF
--- a/modules/core/src/experimental/utils/gpu-grid-aggregator.js
+++ b/modules/core/src/experimental/utils/gpu-grid-aggregator.js
@@ -372,7 +372,8 @@ export default class GPUGridAggregator {
         gridSize,
         uProjectionMatrix: gridTransformMatrix,
         uProjectionMatrixFP64,
-        projectPoints
+        projectPoints,
+        ONE: 1.0
       }
     });
     gridAggregationFramebuffer.unbind();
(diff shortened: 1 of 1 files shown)


modules/core/src/experimental/utils/gpu-grid-aggregator.js
```diff
@@ -372,7 +372,8 @@ export default class GPUGridAggregator {
         gridSize,
         uProjectionMatrix: gridTransformMatrix,
         uProjectionMatrixFP64,
-        projectPoints
+        projectPoints,
+        ONE: 1.0
```
1chandu: This uniform should actually be set by `fp64` module. I will fix this separately on `master` (this is for 6.0-release branch). I will add `getUniform` method for `fp64` module in luma.gl and remove this uniform setting from `Layer` class in deck.gl

1chandu:
Fixing this properly in `master` branch: This uniform should actually be set by `fp64` module. I will fix this separately on `master` (this is for 6.0-release branch). I will add `getUniform` method for `fp64` module in luma.gl and remove this uniform setting from `Layer` class in deck.gl.

Done, luma.gl: https://github.com/uber/luma.gl/pull/622 deck.gl: https://github.com/uber/deck.gl/pull/2124

PR #2143, opened 2018-08-01
Scripting example fixes

Pessimistress:
#### Change List
- Add missing accessors (due to v6 accessor default value change)
- Fix an issue in standalone bundle where calling `setProps` without `viewState` resets the viewport

### DIFF
--- a/examples/gallery/src/hexagon-layer.html
+++ b/examples/gallery/src/hexagon-layer.html
@@ -100,7 +100,7 @@
       OPTIONS.forEach(key => {
         const value = document.getElementById(key).value;
         document.getElementById(key + '-value').innerHTML = value;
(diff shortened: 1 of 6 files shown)


examples/gallery/src/hexagon-layer.html
```diff
@@ -100,7 +100,7 @@
       OPTIONS.forEach(key => {
         const value = document.getElementById(key).value;
         document.getElementById(key + '-value').innerHTML = value;
-        options[key] = value;
+        options[key] = Number(value);
```
Pessimistress: @1chandu `HTMLInputElement.value` is a string. luma.gl no longer recasts the type of the uniform.

PR #2147, opened 2018-08-01
Wrap longitudes over the 180th meridian

Pessimistress:
For https://github.com/uber/deck.gl/issues/2051

Before:
wrapping-1

After:
wrapping-0

#### Change List
- By default, wrap longitude around the viewport center.
- Add a `wrapCoordinates` prop to layers. If set to `false`, fallback to the old projection behavior.

### DIFF
--- a/docs/whats-new.md
+++ b/docs/whats-new.md
@@ -28,6 +28,10 @@ Release date: TBD, Target late Aug, 2018
 
 The projection algorithm used for geospatial coordinates (layers with `coordinateSystem: COORDINATE_SYSTEM.LNGLAT`) has been replaced with a "hybrid" projection/offset based implementation that rivals 64-bit precision at 32-bit speeds. This makes the use of the `fp64` mode unnecessary for most applications, and should increase application performance and avoid issues on untested graphics drivers.
 
+#### Dynamic Meridian
+
+LNGLAT projection modes automatically wrap coordinates over the 180th meridian for the best placement in the current viewport. Set the `wrapLongitude` prop in a layer to `true` to enable this behavior.
+
 
 ### JSON API (Experimental)
 
@@ -37,7 +41,6 @@ A new experimental module `@deck.gl/json` provides a set of classes that allows 
 ### Enhanced Multiview Support
 
 deck.gl's multiview support has been significantly enhanced. New `View` properties give applications more control over rendering, making it possible to implement e.g. overlapping views, partially synchronized views (share some but not all view state props), views with different background colors etc.
-
 
 
 ## deck.gl v6.0
--- a/modules/core/src/shaderlib/project/project.glsl.js
+++ b/modules/core/src/shaderlib/project/project.glsl.js
@@ -28,8 +28,8 @@ const float COORDINATE_SYSTEM_LNGLAT_EXPERIMENTAL = 4.;
 
 uniform float project_uCoordinateSystem;
 uniform float project_uScale;
-uniform bool project_uWrapCoordinates;
-uniform vec2 project_uLngLatCenter;
+uniform bool project_uWrapLongitude;
+uniform float project_uAntimeridian;
 uniform vec3 project_uPixelsPerMeter;
 uniform vec3 project_uPixelsPerDegree;
 uniform vec3 project_uPixelsPerUnit;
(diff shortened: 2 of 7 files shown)


docs/whats-new.md
```diff
@@ -2,10 +2,18 @@
...
 This page contains highlights of each deck.gl release. Also check our [vis.gl blog](https://medium.com/vis-gl) for news about new releases and features in deck.gl.
 
+## deck.gl v6.1
+
+Release date: TBD
+
+### Projection Modes
+
+- `LNGLAT_EXPERIMENTAL` mode
+- LNGLAT projection modes automatically wrap coordinates over the 180th meridian for the best placement in the current viewport. Set the `wrapCoordinates` prop in a layer to `false` to disable this behavior.
```
ibgreen: I think the feature should have a name "Dynamic Antemeridian"?

ibgreen:
Also, would it make any sense to allow the user choose the antemeridian (in addition to having it dynamic of course)?

Pessimistress:
"Wrapping" was probably not the best choice to describe this behavior. To expand on the issue itself:
- The map controller ensures that the longitude of the viewport center always falls within `[-180, 180]`.
- Standard geolocations also have their longitude within the range `[-180, 180]`.
- Web mercator projection maps longitudes to pixel X on the map using a linear transform: `(radians(lng) + PI) * WORLD_SCALE * project_uScale`.

In an extreme example, if the viewport center is at `[179.99, 0]` and the data point being projected is `[-179.99, 0]`, our mercator projection shader function places it one full map width away (`512 * 2 ^ zoom`), instead of close to the center. See this Codepen [drawing an arc from New Zealand to American Samoa](https://codepen.io/Pessimistress/pen/rrKYmj):

image

The solution in this PR is to remap all longitudes from `[-180, 180]` to `[mapCenterLng - 180, mapCenterLng + 180]` before applying mercator projection:

image

modules/core/src/shaderlib/project/project.glsl.js
```diff
@@ -89,8 +91,13 @@ vec4 project_offset_(vec4 offset) {
 // Projecting positions - non-linear projection: lnglats => unit tile [0-1, 0-1]
 //
 vec2 project_mercator_(vec2 lnglat) {
+  float x = lnglat.x;
+  if (project_uWrapCoordinates) {
```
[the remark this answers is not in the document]
Pessimistress: This PR is not about handling out-of-range coordinates in the *source data*, but an inherited problem in web mercator projection. See my comment below.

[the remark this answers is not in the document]
ibgreen: > This PR is not about handling out-of-range coordinates in the source data, but an inherited problem in web mercator projection. See my comment below.

Yes that is why I suggested not using modular arithmetic, since that changes the behavior of out of range coordinates.

ibgreen:
> The solution in this PR is to remap all longitudes from [-180, 180] to [mapCenterLng - 180, mapCenterLng + 180] before applying mercator projectio

Yes, but we are not "removing" the antemeridian, we are just moving it, so the same artifact can pop up somewhere else, if there is data across the globe?

Pessimistress:
> so the same artifact can pop up somewhere else, if there is data across the globe?

You are correct. I think this is safest applied for point features and/or regional datasets at high zoom levels. That's an argument for disabling it by default.

modules/core/src/shaderlib/project/viewport-uniforms.js
```diff
@@ -189,6 +197,10 @@ function calculateViewportUniforms({
     // Projection mode values
     project_uCoordinateSystem: shaderCoordinateSystem,
     project_uCenter: projectionCenter,
+    project_uWrapCoordinates: wrapCoordinates,
```
ibgreen: nit: `project_uLngLatCenter` 

We only use longitude, not both values, and they are specific to one coordinate system.

I wonder if the uniform could either be:

* minimized to just `project_uMeridian` or similar.
* generalized to not be lng/lat (`project_uUnprojectedCenter`)

modules/core/src/shaderlib/project/project.glsl.js
nrabinowitz: Do we have the information at this point to determine whether the antimeridian is in the viewport? If not, we could skip this step, right?

jasonsoden:
This shouldn't be marked as fixed yet and is a blocker for anyone who wants to create a GeoJsonLayer for countries at the global zoom level.

wrapLongitude : false

image

wrapLongitude: true

image

PR #2159, opened 2018-08-03
Make ArcLayer and PathLayer work with LNGLAT_EXPERIMENTAL mode

Pessimistress:
For https://github.com/uber/deck.gl/issues/1937

#### Change List
- ArcLayer
- PathLayer


georgios-uber:
Also, can we patch to 6.0.X ?

Issue #2194, opened 2018-08-15
Website build is broken


1chandu:
#### Actual Result
Crashes on processing `JSON` string.
#### Expected Result
Should run website.
#### Reproduce Steps
cd website, yarn and npm run start.
#### To Do List 
  - [X] Add label and assign to milestone
  - [x] Coding

PR #2195, opened 2018-08-15
Fix website build

1chandu:
For #2194 

#### Background
Simple fix for for string match.

#### Change List
-  Fix website build

### DIFF
--- a/website/contents/pages.js
+++ b/website/contents/pages.js
@@ -16,7 +16,7 @@ function generatePath(tree, parentPath = '', depth = 0) {
 
   tree.depth = depth;
   if (tree.name) {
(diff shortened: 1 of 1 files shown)


website/contents/pages.js
```diff
@@ -16,8 +16,11 @@ function generatePath(tree, parentPath = '', depth = 0) {
 
   tree.depth = depth;
   if (tree.name) {
-    tree.path = tree.name.match(/(GeoJson|3D|API|DeckGL|[A-Z]?[a-z'0-9\.]+|\d+)/g)
-      .join('-').toLowerCase().replace(/[^\w-]/g, '');
+    tree.path = tree.name
```
1chandu: Only change is adding `JSON` to the list, remaining is lint.

Issue #2111, opened 2018-07-23
Hexagons rendered at (0, 0)


SymbolixAU:
#### Issue

I'm trying to recreate the [hexagon example](https://beta.observablehq.com/@pessimistress/deck-gl-hexagonlayer-example), but the hexagons aren't being rendered at the correct locations.  

The data is the road accidents in the UK as per
```
data = {
  return d3.csv('https://raw.githubusercontent.com/uber-common/deck.gl-data/master/examples/3d-heatmap/heatmap-data.csv', row => {
    return {lng: +row.lng, lat: +row.lat};
  })
```

#### Actual Result

In this screenshot I'm showing two layers - a **scatterplot** and a **hexagon**

![screen shot 2018-07-23 at 12 40 30 pm](https://user-images.githubusercontent.com/18344164/43054585-9f138f7a-8e75-11e8-9863-40e0debc8d14.png)

The scatterplot has rendered as expected - the blue dots over the UK

The hexagons have rendered at coordinates (0, 0) (you can see a small blue dot at location (0, 0) off the central-east coast of Africa at the bottom of the image), and I don't know why.

If it helps, this is the log of the hexagon and scatterplot layers

![screen shot 2018-07-23 at 12 50 11 pm](https://user-images.githubusercontent.com/18344164/43054873-fc73a7bc-8e76-11e8-82fd-d42a1841173c.png)

#### Code Example 

The two functions I'm using to create the hexagon and scatterplots are:

```
const hexagonLayer = new deck.HexagonLayer({
  id: 'hexagon'
  data,
  pickable: true,
  extruded: true,
  elevationRange: [0, 1000],
  elevationScale: 250,
  opacity: 1,
  getPosition: d => [d.lng, d.lat]
});
```

```
const scatterLayer = new deck.ScatterplotLayer({
  id: 'scatterplot',
  data,
  radiusScale: 1,
  radiusMinPixels: 1,
  getRadius: d => d.radius,
  getPosition: d => [d.lng, d.lat],
  getColor: d => hexToRgb( d.fill_colour )
});
```

Both functions are using the same `getPosition: d => [d.lng, d.lat]` function, and receive the same data object.  

#### Background / extra detail

I'm an R programmer, not a javascript programmer, so I may have missed something obvious. 

I'm making an `R` package [(github repo here)](https://github.com/SymbolixAU/mapdeck) which wraps `deck.gl`. Therefore my js functions are slightly different to the examples, as I have to receive the data from R.

[...]

PR #2196, opened 2018-08-15
HexagonLayer: Aggregate data using valid viewport.

1chandu:
For #2111 

#### Background
Re-aggregate and update hexbins when viewport is changed.

#### Change List
- HexagonLayer: Re-aggregate data when viewport is changed.

### DIFF
--- a/modules/core/src/lib/layer-manager.js
+++ b/modules/core/src/lib/layer-manager.js
@@ -54,7 +54,7 @@ const INITIAL_CONTEXT = Object.seal({
   stats: null, // for tracking lifecycle performance
 
   // Make sure context.viewport is not empty on the first layer initialization
-  viewport: new Viewport(), // Current viewport, exposed to layers for project* function
+  viewport: new Viewport({id: 'DEFAULT-INITIAL-VIEWPORT'}), // Current viewport, exposed to layers for project* function
 
   // GL Resources
   shaderCache: null,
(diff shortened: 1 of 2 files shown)


1chandu:
>Are the results of the aggregation actually viewport dependent (lng/lat/zoom) or do we just need a mercator projection? If the latter, can we just create a "working viewport" during aggregration (instead of depending on the passed in viewport)?

It uses viewport for 1. calls `projectFlat` and `unprojectFlat` 2. uses viewport's `distanceScales`.

So if any change in above factors, we should re-aggregate.

1chandu:
Digged more into this

1. During the layer initialization, the viewport available in `layer.context` is the default viewport (non geospatial)
2. Actual webmercator viewport is created after layer is initialized and after GLContext is created, at this point it uses canvas `width` and `height`.
3. During `render` time, each viewport is activated using `_activateViewport`, where layer's `context.viewport` is updated, `viewportChanged` flag is set.
4. Layer gets chance to update its state (provided `shouldUpdateState` is overwritten)

For hexagon layers, we do need to re-project points at #4 , so we have to check for `viewportChanged` flag. But this might trigger aggregation un-necessarily for each pan/zoom events. So we need a new flag that is only set when mercator projection needs to be updated.

1chandu:
Update the PR with possible solution (couple of tests failing, will fix before landing them)

We need to project points the first time viewport changed to a geospatial (WebMercatorViewport) viewport, since initial viewport is not a WebMercatorViewport. Added changes to aggregate only when layer has geospatial viewport and `shouldUpdate` returns only when `viewportChanged` and data is not aggregated yet.

1chandu:
@ibgreen created https://github.com/uber/deck.gl/issues/2215

PR #2225, opened 2018-08-23
Change Back Projection mode to the 6.0 default

georgios-uber:


georgios-uber:
Yes, I know. I will not push this until I have tried to fix all the issues.

PR #2280, opened 2018-09-15
Handle GeoJSON GeometryCollection

supersonicclay:
Handle geometry with `type='GeometryCollection'`.

`GeometryCollection` is described in [section 3.1.8 of the GeoJSON spec](https://tools.ietf.org/html/rfc7946#section-3.1.8).

TODO
[X] Tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ dist-demo/
 node_modules/
 coverage/
 .nyc_output/
+.vscode/
 
 dist-bundle.js
 examples/winds/data/*.html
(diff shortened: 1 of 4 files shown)


Pessimistress:
Looks good. Can you add a simple test case?

supersonicclay:
Tweaked the existing tests and implementation so that `GeometryCollection` is handled just like any other geometry type.

supersonicclay:
I'm good with merging this when you are @Pessimistress

PR #2285, opened 2018-09-19
Match projectPosition on CPU and GPU

Pessimistress:

#### Background

`projectPosition` is the CPU utility function that projects a point from any coordinate system to the world position in another coordinate system. It is currently used by the lighting module.

In a follow up PR, I will update the `project` method on the base Layer class to use this util, so that layers can work with different coordinate systems on the CPU.

#### Change List
- Remove LNGLAT_AUTO_OFFSET hack from `lighting.getUniforms`
- Handle LNGLAT_AUTO_OFFSET mode in `projectPosition`
- Add tests that compare the returned values from `projectPosition` and compiled vertex shader function `project_position`.

### DIFF
--- a/modules/core/src/shaderlib/project/project-functions.js
+++ b/modules/core/src/shaderlib/project/project-functions.js
@@ -3,13 +3,55 @@
  * TODO: move to Viewport class?
  */
 import {COORDINATE_SYSTEM} from '../../lib/constants';
+import {LNGLAT_AUTO_OFFSET_ZOOM_THRESHOLD} from './viewport-uniforms';
+
 import vec4_transformMat4 from 'gl-vec4/transformMat4';
 import vec3_sub from 'gl-vec3/subtract';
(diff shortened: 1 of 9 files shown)


modules/core/src/shaderlib/project/project-functions.js
```diff
@@ -37,25 +81,27 @@ export function projectPosition(
...
     [x, y, z] = vec4_transformMat4([], [x, y, z, 1.0], modelMatrix);
   }
-  if (fromCoordinateSystem === undefined) {
-    fromCoordinateSystem = coordinateSystem;
-  }
-  if (fromCoordinateOrigin === undefined) {
-    fromCoordinateOrigin = coordinateOrigin;
+
+  let offsetMode = false;
+  switch (coordinateSystem) {
```
1chandu: Nit: Add all enums here and assert on default, it will make things easier when new enums area added.

PR #2286, opened 2018-09-19
update layer.project to use actual coordinate system

Pessimistress:

#### Background

`Layer.project` does not respect the coordinate system of the position

#### Change List
- Use the current coordinate system for projection
- **Proposed** deprecate `Layer.projectFlat` and replace with `Layer.projectPosition`
- **Proposed** deprecate `Layer.unprojectFlat`. There is no easy way to unproject back to the original coordinate system.

### DIFF
--- a/modules/core/src/lib/layer.js
+++ b/modules/core/src/lib/layer.js
@@ -30,9 +30,12 @@ import log from '../utils/log';
 import GL from 'luma.gl/constants';
 import {withParameters} from 'luma.gl';
 import assert from '../utils/assert';
+import {projectPosition} from '../shaderlib/project/project-functions';
 
 import Component from '../lifecycle/component';
 import LayerState from './layer-state';
+
+import {worldToPixels} from 'viewport-mercator-project';
 
 const LOG_PRIORITY_UPDATE = 1;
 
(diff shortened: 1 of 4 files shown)


Pessimistress:
> Should we add some test cases with actual coordinate systems?

Tests are added in https://github.com/uber/deck.gl/pull/2285

modules/core/src/lib/layer.js
```diff
@@ -155,27 +158,44 @@ export default class Layer extends Component {
...
   projectFlat(lngLat) {
+    log.deprecated('layer.projectFlat', 'layer.projectPosition');
     const {viewport} = this.context;
     assert(Array.isArray(lngLat));
     return viewport.projectFlat(lngLat);
   }
 
+  // DEPRECATE: This is not meaningful in offset modes
   unprojectFlat(xy) {
+    log.deprecated('layer.unprojectFlat');
```
1chandu: Nit: should this be `log.deprecated('layer.unprojectFlat', 'layer.unproject`);` ?

Pessimistress: There is no replacement for `layer.unprojectFlat`.

Issue #2272, opened 2018-09-13
Projection module and Viewport issues


Pessimistress:
##### Shader projection bugs related to y-flipping

`project` module does not behave consistently when handling y-offset and Mercator projection. This issue is explained and partially patched in https://github.com/uber/deck.gl/pull/2230. As of today, `project_position` handles Y correctly, while `project_scale` does not, leading to the issue that https://github.com/uber/deck.gl/pull/2167 tried to address.

##### Inconsistent modelMatrix support

- Modes that respect `modelMatrix`: LNGLAT, METER_OFFSETS, IDENTITY
- Modes that do not respect `modelMatrix`: LNGLAT_OFFSETS, LNGLAT_AUTO_OFFSET

##### CPU and GPU projection disparity

The shader functions in `project` module and the methods in the `Viewport` class have diverged:

- No CPU equivalent to project coordinates in METER_OFFSETS, LNGLAT_OFFSETS modes
- `Viewport.project` does not produce the same result as `project_position_to_clipspace` (bug)

### To Do List
  - [x] Add label and assign to milestone
  - [x] Add tests for project module
  - [x] Bug fixes
  - [x] Whats new update

PR #2287, opened 2018-09-19
Add Transform based GPU tests for project module.

1chandu:
For #2272 

#### Background
Add `Transform` based unit tests for `project` module. Verified on AMD and INTEL GPUs.

#### Change List
- Add Transform based GPU tests for project module.

### DIFF
--- a/test/modules/core/shaderlib/project/project-glsl.spec.js
+++ b/test/modules/core/shaderlib/project/project-glsl.spec.js
@@ -23,6 +23,8 @@ import test from 'tape-catch';
 import {COORDINATE_SYSTEM, Viewport, WebMercatorViewport} from 'deck.gl';
 import {project} from '@deck.gl/core/shaderlib';
 import {Matrix4, equals, config} from 'math.gl';
+import {gl} from '@deck.gl/test-utils';
+import {isWebGL2, Transform, Buffer} from 'luma.gl';
 
 import {compileVertexShader} from '../shaderlib-test-utils';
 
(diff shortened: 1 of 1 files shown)


test/modules/core/shaderlib/project/project-glsl.spec.js
```diff
@@ -62,6 +64,67 @@ const TEST_VIEWPORT_ORTHO = new Viewport({
...
+  project_position: pos => `\
+varying vec4 outValue;
+
+void main()
+{
+  outValue = project_position(vec4(${pos[0].toFixed(MAX_FRACTION_DIGITS)}, ${pos[1].toFixed(
+    MAX_FRACTION_DIGITS
+  )}, ${pos[2].toFixed(MAX_FRACTION_DIGITS)}, ${pos[3].toFixed(
+    MAX_FRACTION_DIGITS
+  )}), vec2(0., 0.));
```
Pessimistress: Take advantage of the xy64Low argument to improve precision

test/modules/core/shaderlib/project/project-glsl.spec.js
```diff
@@ -258,14 +362,17 @@ test('project#vs', t => {
     const module = projectVS(uniforms);
 
     testCase.tests.forEach(c => {
-      const actual = c.func(module);
+      let actual = c.func(module);
```
Pessimistress: Only run this if webgl is not available?

test/modules/core/shaderlib/project/project-glsl.spec.js
1chandu: Added code to use low64 bits, but LNGLAT mode test still fails, as these not used in base project module. We can remove the tolerance for project-64 module tests.

test/modules/core/shaderlib/project/project-glsl.spec.js
1chandu: Done, run tests using Transform when supported, otherwise fallback to transpiler.

PR #2288, opened 2018-09-19
Fix 3d projection in Mapbox integration

Pessimistress:

mapbox-integration

### DIFF
--- a/modules/mapbox-layers/src/deck-layer.js
+++ b/modules/mapbox-layers/src/deck-layer.js
@@ -10,13 +10,16 @@ export default class DeckLayer {
   }
 
   _getViewState() {
-    const {lng, lat} = this.map.getCenter();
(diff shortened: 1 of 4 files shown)


modules/mapbox-layers/src/deck-layer.js
```diff
@@ -45,6 +47,7 @@ export default class DeckLayer {
 
     this.deck.setProps({viewState});
     this.deck._drawLayers();
-    this.map.triggerRepaint();
+
+    // this.map.triggerRepaint();
```
Pessimistress: @ibgreen Why was this needed?

Pessimistress:
> It would be good with an example with more deck.gl 3D structure

I will consolidate the 2d and 3d examples into one in a follow up PR.

PR #2308, opened 2018-09-25
Fix Browser Test

Pessimistress:
#### Background
Fix test-browser on master


Pessimistress:
> Should we also remove usage of _customRender in deck.js ?

It's needed by mapbox integration. The prop was switched from boolean to function in a recent PR.

PR #2315, opened 2018-09-27
Add advanced event handling RFC

Pessimistress:
For https://github.com/uber/deck.gl/issues/2099

#### Background
- See RFC

#### Change list
- RFC catalog
- Advanced event handling RFC


supersonicclay:
@Pessimistress I'm late to the party, but the other thing that'd be great to include is the ability for a layer to interrupt or prevent certain interactions (e.g. dynamically prevent map panning).

PR #2322, opened 2018-10-01
Added nebula.gl to docs intro and website header

ralphstodomingo:
...so newcomers to the docs may notice it.

For #

#### Background
Feature editing capabilities are one of the things some people might look for or ask when they come across deck.gl.

#### Change List
- Added nebula.gl to the links rendered in the doc website header
- Added nebula.gl mention in the doc intro


Pessimistress:
Should add kepler.gl too.

Issue #2325, opened 2018-10-02
Examples are broken master branch


jianhuang01:
#### Actual Result
Examples are broken because of bad dependencies
#### Expected Result
Should run well
#### Reproduce Steps
npm run start-local under example folder
#### To Do List 
  - [x] Add label and assign to milestone
  - [x] Coding
  - [x] Test

jordmccordgcv:
#2324

PR #2326, opened 2018-10-02
Example fix

XiaokaiUber:
For #2325 

#### Change List
- fixed the bad dependencies
- fixed the style issue in docs


Pessimistress:
The versions of deck.gl should not be locked. Otherwise we will have to modify them again after each patch.

Issue #2274, opened 2018-09-13
project / unproject support for orthographicView 


gnavvy:
#### Background 
Adding project / unproject support at framework level for orthographic views.
project: convert world/data space coordinate to screen space.
unproject: convert screen space coordinate back data/world space.

For InfoVis apps using an orthographic view, users often need to convert the cursor position (relative to the top left corner of the deck.gl canvas) to the data coordinate space for picking and range selection.

The conversion needs to consider both the offset (can be derived from left, right, top, bottom of the orthographic view) and the scale (the left, right, top, bottom, together with the width and height of the view) of the view.

See @Pessimistress's comment in this [PR](https://github.com/uber/deck.gl/pull/2128) for example, in which we need to convert the cursor position to data space, and apply the changes to the orthographic view state, to achieve the zooming by cursor effect.

Also, this is a step towards the project / unproject support for custom coordinate systems, as described in this [RFC](https://github.com/uber/deck.gl/blob/master/dev-docs/RFCs/v6.x/project-unproject-rfc.md).

#### To Do List 
  - [x] Add label and assign to milestone 
  - [ ] Coding
  - [ ] Doc update
  - [ ] Whats new update 
  - [ ] Test

gnavvy:
[This](https://github.com/uber/deck.gl/blob/master/modules/core/src/controllers/map-controller.js#L221-L255) part of the logic need to be implemented in the [zoom](https://github.com/uber/deck.gl/blob/master/modules/core/src/controllers/orthographic-controller.js#L110-L116) function in orthographic-controller.js as well. 
The latter only applies zoom without considering the offset and cursor position now. 

^ can be reproduced and tested in the [bezier](https://github.com/uber/deck.gl/tree/master/examples/experimental/bezier) example.

PR #2336, opened 2018-10-05
improve orthographic zoom

jianhuang01:
For #2274 

#### Change List
- Implemented a new zoom logic


modules/core/src/controllers/orthographic-controller.js
```diff
@@ -107,11 +107,16 @@ class OrthographicState extends ViewState {
    * @param {Number} scale - a number between [0, 1] specifying the accumulated
    *   relative scale.
    */
-  zoom({scale}) {
-    const {zoom} = this._viewportProps;
+  zoom({pos, scale}) {
```
1chandu: Nit: update the description, zoom with the current mouse pointer as center? assuming that is what this method is doing.

Issue #2350, opened 2018-10-12
ScreenGridLayer: no rendering at high zoom


1chandu:
#### Actual Result
Doesn't render aggregated cells.
#### Expected Result
Should render aggregated cells.
#### Reproduce Steps
Run screen-grid demo and increase the zoom (above 12)
#### To Do List 
  - [x] Add label and assign to milestone
  - [x] Coding

PR #2351, opened 2018-10-12
GPUGridAggregator: Project to clip space to handle offset project modes.

1chandu:
For #2350 

#### Background
ScreenGridLayer uses GPUAggregator to to perform aggregation in screen (viewport) space. It first uses `project_position` to perform mercator projection then trnasforms the result into viewport (screen) sapce to perform aggregation.
With the introduction of new LNGLAT_OFFSET (now LNGLAT) projection mode, when zoom level is more than a certain level (12), `project_position` only returns projected offsets instead of actual points. And required offset is added back during `project_to_clipspace`.

The fix is, replaced `project_position` with `project_position_to_clipspace` and instead of using `pixelProjectionMatrix` use `viewportMatrix` to convert projected points from NDC to viewport (screen) space.

#### Change List
- GPUGridAggregator: Project to clip space to handle offset project modes

### DIFF
--- a/modules/core/src/viewports/viewport.js
+++ b/modules/core/src/viewports/viewport.js
@@ -445,11 +445,13 @@ export default class Viewport {
      */
 
     // matrix for conversion from world location to screen (pixel) coordinates
-    const m = createMat4();
-    mat4_scale(m, m, [this.width / 2, -this.height / 2, 1]);
-    mat4_translate(m, m, [1, -1, 0]);
-    mat4_multiply(m, m, this.viewProjectionMatrix);
-    this.pixelProjectionMatrix = m;
+    const viewportMatrix = createMat4(); // matrix from NDC to viewport.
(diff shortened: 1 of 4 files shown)


modules/core/src/viewports/viewport.js
```diff
@@ -445,11 +445,13 @@ export default class Viewport {
      */
 
     // matrix for conversion from world location to screen (pixel) coordinates
-    const m = createMat4();
-    mat4_scale(m, m, [this.width / 2, -this.height / 2, 1]);
-    mat4_translate(m, m, [1, -1, 0]);
-    mat4_multiply(m, m, this.viewProjectionMatrix);
-    this.pixelProjectionMatrix = m;
+    const viewportMatrix = createMat4(); // matrix from NDC to viewport.
```
1chandu: No function changes here except, adding a new matrix `this.viewportMatrix`, that can convert from NDC to viewport (screen) space.

PR #2352, opened 2018-10-12
Fix upgrade guide

1chandu:
For #2240 

#### Change List
- Fix upgrade guide


1chandu:
@Pessimistress , merging this, will address if you have an comments, don't want this to be slipped out of website publish.

PR #2357, opened 2018-10-16
Added TileLayer

starry97:
#### Change List
- Added `TileLayer` component 
- Changed `MapboxTiles` to use `TileLayer`


modules/experimental-layers/src/tile-layer/tile-layer.js
```diff
@@ -0,0 +1,50 @@
...
+    return info;
+  }
+
+  renderLayers() {
+    // eslint-disable-next-line no-unused-vars
+    const {fetchData, renderSubLayers, ...geoProps} = this.props;
+    return this.state.tiles.map(tile => {
+      return renderSubLayers({
+        ...geoProps,
+        id: `${tile.x}-${tile.y}-${tile.z}`,
```
Pessimistress: Use `${this.id}-...` to avoid potential collision

modules/experimental-layers/src/tile-layer/tile-layer.js
```diff
@@ -0,0 +1,50 @@
...
+      tileCache: new TileCache({fetchData: this.props.fetchData})
+    };
+  }
+
+  shouldUpdateState({changeFlags}) {
+    return changeFlags.somethingChanged;
+  }
+
+  updateState({props, oldProps, context, changeFlags}) {
+    if (props.fetchData !== oldProps.fetchData) {
```
Pessimistress: We usually avoid directly comparing two props incase inline functions are used, e.g.:

```js
new ScatterplotLayer({
  ...
  getRadius: d => d.radius
})
```

Then `props.getRadius !== oldProps.getRadius` would always be true.

Instead, deck.gl requires explicit [updateTriggers](http://deck.gl/#/documentation/deckgl-api-reference/layers/layer?section=data-properties) to trigger data regeneration. You can access the comparison results via:

```js
changeFlags.updateTriggersChanged &&
        (changeFlags.updateTriggersChanged.all || changeFlags.updateTriggersChanged.fetchData))
```

modules/experimental-layers/src/tile-layer/tile-layer.js
```diff
@@ -0,0 +1,50 @@
+import {GeoJsonLayer, CompositeLayer} from 'deck.gl';
+import TileCache from './utils/tile-cache';
+
+const defaultProps = {
+  renderSubLayers: props => new GeoJsonLayer(props)
```
Pessimistress: Add all layer-specific props here

modules/experimental-layers/src/tile-layer/tile-layer.js
```diff
@@ -0,0 +1,50 @@
...
+};
+
+export default class TileLayer extends CompositeLayer {
+  initializeState() {
+    this.state = {
+      tiles: [],
+      tileCache: new TileCache({fetchData: this.props.fetchData})
```
Pessimistress: Call it `getTileData` to align with other layers' naming conventions?

modules/experimental-layers/src/tile-layer/utils/tile-cache.js
```diff
@@ -0,0 +1,56 @@
+import Tile from './tile';
+import {getTileIndices} from './viewport-util';
+
+/**
+ * Manages loading and purging of tiles data
+ */
+
+export default class TileCache {
+  constructor({fetchData, size = 10, onUpdate}) {
```
Pessimistress: Remove `onUpdate` if not used

modules/experimental-layers/src/tile-layer/utils/tile-cache.js
```diff
@@ -0,0 +1,56 @@
+import Tile from './tile';
+import {getTileIndices} from './viewport-util';
+
+/**
+ * Manages loading and purging of tiles data
+ */
+
+export default class TileCache {
```
Pessimistress: This internal class can benefit from some JSDoc

modules/experimental-layers/src/tile-layer/utils/tile-cache.js
```diff
@@ -0,0 +1,56 @@
...
+          y,
+          z
+        });
+        this._push(tile);
+      }
+      return tile;
+    });
+
+    while (cache.length > size) {
+      cache.shift();
```
Pessimistress: This is not the correct logic (the fist tile may still be visible).
Add a TODO here if this is not going to be addressed in the current PR.

modules/experimental-layers/src/tile-layer/utils/tile.js
```diff
@@ -0,0 +1,50 @@
...
+    this._loader = this._loadData();
+  }
+
+  _loadData() {
+    const {x, y, z} = this;
+    return this.fetchData(x, y, z).then(buffers => {
+      return new Promise((resolve, reject) => {
+        try {
+          this._data = buffers;
+          resolve(buffers);
```
Pessimistress: - No need for the `new Promise` wrapper, you can just return `buffers`.
- No need for try...catch as nothing can break here

modules/experimental-layers/src/tile-layer/utils/tile.js
```diff
@@ -0,0 +1,50 @@
...
+    return this._data !== null;
+  }
+
+  overlaps(tile) {
+    const {x, y, z} = this;
+    const m = Math.pow(2, tile.z - z);
+    return Math.floor(tile.x / m) === x && Math.floor(tile.y / m) === y;
+  }
+
+  _init() {
```
Pessimistress: Just drop this method if it's only one line

modules/experimental-layers/src/tile-layer/utils/viewport-util.js
```diff
@@ -44,3 +44,16 @@ export function getTileIndices(viewport) {
 
   return indices;
 }
+
+export function long2tile(lon, zoom) {
```
Pessimistress: Is this used anywhere?

modules/experimental-layers/src/tile-layer/utils/tile.js
```diff
@@ -0,0 +1,50 @@
...
+    this._loader = this._loadData();
+  }
+
+  _loadData() {
+    const {x, y, z} = this;
+    return this.fetchData(x, y, z).then(buffers => {
```
Pessimistress: Prefer named parameters (`{x, y, z}`) in case we need to send more arguments in the future.

test/apps/mapbox-tile/app.js
```diff
@@ -1,10 +1,11 @@
-/* global document */
+/* global document fetch window */
 /* eslint-disable no-console */
 import React, {Component} from 'react';
 import {render} from 'react-dom';
 import DeckGL from 'deck.gl';
+import TileLayer from '../../../modules/experimental-layers/src/tile-layer/tile-layer';
```
Pessimistress: Use `@deck.gl/experimental-layers/...`
So that this app will work standalone.

test/apps/mapbox-tile/utils/decode.js
```diff
@@ -0,0 +1,41 @@
+import Protobuf from 'pbf';
+import {VectorTile} from '@mapbox/vector-tile';
+import {worldToLngLat} from 'viewport-mercator-project';
+import {vectorTileFeatureToGeoJSON} from './feature';
+
+const TILE_SIZE = 512;
+
+export function decodeTiles(x, y, z, arrayBuff) {
```
Pessimistress: Use full name for arguments

test/apps/mapbox-tile/utils/decode.js
```diff
@@ -0,0 +1,41 @@
...
+  const tile = new VectorTile(new Protobuf(arrayBuff));
+
+  const result = [];
+  const projectFunc = project.bind(null, x, y, z);
```
Pessimistress: `Math.pow(2, z)`, `x * TILE_SIZE` and `y * TILE_SIZE` are repeatedly calculated inside `project`. We can do it once before binding to improve perf.

Pessimistress:
Please add a README.md in the tile-layer directory following [this format](https://github.com/uber/deck.gl/blob/master/docs/layers/scatterplot-layer.md). It will be moved to official documentation when this layer graduates.

modules/experimental-layers/src/tile-layer/tile-layer.js
starry97: Makes sense. 

Is `changeFlags.updateTriggersChanged.all` checking whether all accessor props are changed?

test/apps/mapbox-tile/utils/decode.js
starry97: What do you mean? Use `arrayBuffer` instead of `arrayBuff`?

modules/experimental-layers/src/tile-layer/tile-layer.js
Pessimistress: The user can have an updateTrigger called `all` which would invalidate all accessors.

test/apps/mapbox-tile/utils/decode.js
Pessimistress: Yes

modules/experimental-layers/src/tile-layer/utils/tile.js
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  _loadData() {
+    const {x, y, z} = this;
+    return (
+      this.getTileData &&
+      this.getTileData({x, y, z}).then(buffers => {
+        this._data = buffers;
+        return buffers;
+      })
```
Pessimistress: Need to handle loading errors

modules/experimental-layers/src/tile-layer/tile-layer.js
```diff
@@ -0,0 +1,89 @@
+import {GeoJsonLayer, CompositeLayer} from 'deck.gl';
+import TileCache from './utils/tile-cache';
+
+const defaultLineColor = [0, 0, 0, 255];
+const defaultFillColor = [0, 0, 0, 255];
+
+const defaultProps = {
+  stroked: true,
```
Pessimistress: Since the sub layers are not necessarily GeoJsonLayers, we should not presume the props here. The only two default props needed are `getTileData` and `renderSubLayers`

modules/experimental-layers/src/tile-layer/tile-layer.md
```diff
@@ -0,0 +1,236 @@
...
+        return decodeTiles(x, y, z, buffer);
+      });
+  }
+
+  render() {
+    return (
+      <DeckGL
+        layers={[
+          new TileLayer({
+            ...MAP_LAYER_STYLES,
```
Pessimistress: This constant is undefined in this code sample -- just expand it in place with a few examples.

modules/experimental-layers/src/tile-layer/tile-layer.md
```diff
@@ -0,0 +1,236 @@
...
+This TileLayer takes in a function `getTileData` that fetches tiles, and renders it in a GeoJsonLayer or with the layer returned in `renderSubLayers`.
+
+```js
+import DeckGL from 'deck.gl';
+import TileLayer from '@deck.gl/experimental-layers/tile-layer/tile-layer';
+
+getTileData({x, y, z}) {
+    const mapSource = `https://a.tiles.mapbox.com/v4/mapbox.mapbox-streets-v7/${z}/${x}/${y}.vector.pbf?access_token=${MAPBOX_TOKEN}`;
+    return fetch(mapSource)
+      .then(response => {
```
Pessimistress: `.then(response => response.arrayBuffer())`

test/apps/mapbox-tile/utils/decode.js
```diff
@@ -0,0 +1,40 @@
+import Protobuf from 'pbf';
+import {VectorTile} from '@mapbox/vector-tile';
+import {worldToLngLat} from 'viewport-mercator-project';
+import {vectorTileFeatureToGeoJSON} from './feature';
+
+const TILE_SIZE = 512;
+
+export function decodeTiles(x, y, z, arrayBuffer) {
```
Pessimistress: `decodeTile`

modules/experimental-layers/src/tile-layer/utils/tile.js
starry97: Is logging the stack trace enough? Or do we have other mechanism to handle errors.

modules/experimental-layers/src/tile-layer/tile-layer.md
starry97: In this code sample, is it necessary to process the fetched tile data? (`decodeTile` is not defined in this snippet either)

Pessimistress: Let's add a minimal version of the decoder:

```js
import {VectorTile} from '@mapbox/vector-tile';
import Protobuf from 'pbf';

function vectorTileToGeoJSON(buffer, x, y, z) {
  const tile = new VectorTile(new Protobuf(buffer));
  const features = [];

  for (const layerName in tile.layers) {
    const vectorTileLayer = tile.layers[layerName];

    for (let i = 0; i < vectorTileLayer.length; i++) {
      const vectorTileFeature = vectorTileLayer.feature(i);
      const feature = vectorTileFeature.toGeoJSON(x, y, z);
      features.push(feature);
    }
  }

  return features;
}
```

modules/experimental-layers/src/tile-layer/tile-layer.md
```diff
@@ -0,0 +1,243 @@
...
+```
+
+## Properties
+
+Inherits from all [Base Layer](/docs/api-reference/layer.md) properties, along with `renderSubLayer`
+and `getTileData`
+
+### Render Options
+
+##### `filled` (Boolean, optional)
```
Pessimistress: Remove GeoJsonLayer props from here as well

modules/experimental-layers/src/tile-layer/utils/tile.js
Pessimistress: Consider the case where you don't have tiles above MAX_ZOOM but you can still allow users to zoom in. When the tiles at the current zoom level fail to load (either throw errors or resolve to null), you need to mark the tiles as invalid and fallback to using data from an available zoom level.

Let's land this layer as is and address this in a follow up PR.

starry97: Yep. Added a TODO comment.

PR #2362, opened 2018-10-19
Handle TileLayer Caching Better

starry97:
**Tile Cache**
- Set default caching size based on the size of the viewport
- When cache is full, only remove tiles that are not visible
- Use cached parents tiles when we are waiting for a new tile to be loaded

**Tile Layer**
- Only Render tiles that are between minZoom and maxZoom level

### DIFF
--- a/modules/experimental-layers/src/tile-layer/utils/tile-cache.js
+++ b/modules/experimental-layers/src/tile-layer/utils/tile-cache.js
@@ -37,10 +37,20 @@ export default class TileCache {
    */
   update(viewport, onUpdate) {
     const {_cache, _getTileData, _maxSize} = this;
-    _cache.forEach(cachedTile => {
-      cachedTile.isVisible = false;
-    });
+    this._markOldTiles();
     const tileIndices = getTileIndices(viewport);
+    if (!tileIndices || tileIndices.length === 0) {
+      onUpdate(tileIndices);
+      return;
+    } else if (
+      this._minZoom &&
+      this._minZoom === parseInt(this._minZoom, 10) &&
+      tileIndices[0].z < this._minZoom
+    ) {
+      // Do not show anything if zoom level is too small (avoid loading too many tiles)
+      onUpdate([]);
+      return;
+    }
     const viewportTiles = new Set();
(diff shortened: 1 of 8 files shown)


modules/experimental-layers/src/tile-layer/utils/tile-cache.js
```diff
@@ -30,48 +36,93 @@ export default class TileCache {
...
+      this._minZoom === parseInt(this._minZoom, 10) &&
+      tileIndices[0].z < this._minZoom
+    ) {
+      // Do not show anything if zoom level is too small (avoid loading too many tiles)
+      onUpdate([]);
+      return;
+    }
+    const viewportTiles = new Set();
+    _cache.forEach(cachedTile => {
+      if (tileIndices.some(tile => cachedTile.isOverlapped(tile))) {
```
Pessimistress: Do not get the fallback tiles for loaded tiles?

modules/experimental-layers/src/tile-layer/utils/tile-cache.js
```diff
@@ -30,48 +36,93 @@ export default class TileCache {
...
-      let tile = this._find(x, y, z);
+    const {_cache, _getTileData, _maxSize} = this;
+    this._markOldTiles();
+    const tileIndices = getTileIndices(viewport);
+    if (!tileIndices || tileIndices.length === 0) {
+      onUpdate(tileIndices);
+      return;
+    } else if (
+      this._minZoom &&
+      this._minZoom === parseInt(this._minZoom, 10) &&
```
Pessimistress: What does this do?

starry97: This checks whether `this._minZoom` is an integer

Pessimistress: You can normalize minZoom and maxZoom in the constructor once and for all.

starry97: Yep, moved this to the constructor already.

PR #2364, opened 2018-10-21
Remove excessive redraws in layer update

Pessimistress:
#### Background
CompositeLayers do not have models but are marked needs redraw in `_updateBaseUniforms()`

#### Change List
- only update attributes and uniforms for non-composite layers
- use model's needsRedraw flag

### DIFF
--- a/modules/core/src/lib/layer.js
+++ b/modules/core/src/lib/layer.js
@@ -513,20 +513,20 @@ export default class Layer extends Component {
 
     // Call subclass lifecycle methods
     this.updateState(updateParams);
-
-    // Render or update previously rendered sublayers
-    if (this.isComposite) {
-      this._renderLayers(updateParams);
-    }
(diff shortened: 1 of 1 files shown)


modules/core/src/lib/layer.js
```diff
@@ -513,20 +513,20 @@ export default class Layer extends Component {
...
       this._renderLayers(updateParams);
-    }
-    // End subclass lifecycle methods
-
-    // Add any subclass attributes
-    this.updateAttributes(this.props);
-    this._updateBaseUniforms();
+    } else {
+      // Add any subclass attributes
+      this.updateAttributes(this.props);
```
1chandu: I wonder if these two lines 
```
this.updateAttributes(this.props);
this._updateBaseUniforms();
```
can be moved inside of `if (this.state.model)` block.

PR #2372, opened 2018-10-22
Implement prop types RFC

Pessimistress:
For https://github.com/uber/deck.gl/issues/1371

#### Change List
- Implement prop types
- Update prop diffing to use prop type equal functions
- Use prop types in base Layer default props

### DIFF
--- a/modules/core/src/lifecycle/prop-types.js
+++ b/modules/core/src/lifecycle/prop-types.js
@@ -1,19 +1,13 @@
-import assert from '../utils/assert';
-
 const TYPE_DEFINITIONS = {
   boolean: {
     validate(value, propType) {
       return true;
+    },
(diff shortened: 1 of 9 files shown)


modules/core/src/lifecycle/prop-types.js
```diff
@@ -1,29 +1,78 @@
...
     validate(value, propType) {
       return (
         Number.isFinite(value) &&
         (!('max' in propType) || value <= propType.max) &&
         (!('min' in propType) || value >= propType.min)
       );
     }
+  },
+  color: {
+    validate(value, propType) {
```
1chandu: Is `compare` field assumed to be true for this case, also for other cases where it is not used? I also didn't see any doc changes, please add these details to docs in this or next PR.

Issue #2215, opened 2018-08-21
All Layers rendered with default viewport which is not always correct.


1chandu:
#### Actual Result
All layers are first rendered with default basic `Viewport` instance, instead of `Viewport` created from the provided view state.
#### Expected Result
All layers should only be rendered with `Viewport` created from the provided view state.
#### Reproduce Steps
Usually only happens if you provide static data , instead of URL.  When providing the URL this issue is hidden as data load triggers another render, by which time viewport has been updated.
See #2196

#### To Do List 
  - [X] Add label and assign to milestone
  - [x] Coding
  - [x] Test

1chandu:
Temporary workaround added in #2196 can be removed once this is resolved.

PR #2376, opened 2018-10-23
LayerManager: fix initial viewport

1chandu:
For #2111 and #2215 

#### Background
Before version 6, it was guaranteed that layers are always initialized with a viewport built with user provided parameters, during multi-viewport support this code was re-structured and now layers are initialized with a default viewport (not built with user provided `viewState` or `initialViewState`) causing issues. For example, layers that use `context.viewport` to perform projection on `data`.  Actual viewport is only set during draw time.

To fix this, build `viewport` from user provided `view`/`viewState` or default when not available, and use it to initialize the `LayerManager`, this will match the old behavior.

This change also reverts previously added WA for HexagonLayer : #2239

#### Change List
- Revert "Hexagon aggregation with valid viewport (#2239)"
- LayerManager: Initialize with correct viewport.

### DIFF
--- a/modules/core/src/lib/deck.js
+++ b/modules/core/src/lib/deck.js
@@ -477,8 +477,7 @@ export default class Deck {
     // viewManager must be initialized before layerManager
     // layerManager depends on viewport created by viewManager.
     assert(this.viewManager);
-    const viewports = this.viewManager.getViewports();
-    const viewport = viewports.length > 0 ? viewports[0] : null;
+    const viewport = this.viewManager.getViewports()[0];
     // Note: avoid React setState due GL animation loop / setState timing issue
     this.layerManager = new LayerManager(gl, {
       stats: this.stats,
--- a/modules/core/src/lib/deck.js
+++ b/modules/core/src/lib/deck.js
@@ -396,6 +396,20 @@ export default class Deck {
     return views;
   }
 
+  // Returns default viewport based on initialViewState/viewState.
+  _getDefaultViewport(props) {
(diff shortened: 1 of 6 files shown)


modules/core/src/lib/deck.js
```diff
@@ -396,6 +396,20 @@ export default class Deck {
     return views;
   }
 
+  // Returns default viewport based on initialViewState/viewState.
+  _getDefaultViewport(props) {
```
Pessimistress: This is duplicating what ViewManager does.
Maybe we can send `views` and `viewState` to ViewManager constructor then use `deck.getViewports()` instead?

modules/core/src/lib/layer-manager.js
```diff
@@ -96,7 +93,9 @@ export default class LayerManager {
         // For callback tracking and autohighlight
         index: -1,
         layerId: null
-      }
+      },
+      // Make sure context.viewport is not empty on the first layer initialization
+      viewport: viewport || new Viewport() // Current viewport, exposed to layers for project* function
```
Pessimistress: Keep the `DEFAULT-INITIAL-VIEWPORT` id in default fallback?

1chandu: Oh yeah, will add it back.

modules/core/src/lib/deck.js
Pessimistress: Add a comment where the two instances are initialized on the importance of the order?

PR #2378, opened 2018-10-23
Proptypes validation

Pessimistress:
For #1371 	

#### Change List
- Validate layer props if deck has prop `debug: true`
- Tests

### DIFF
--- a/examples/layer-browser/src/app.js
+++ b/examples/layer-browser/src/app.js
@@ -20,6 +20,7 @@ import LayerSelector from './components/layer-selector';
 import LayerControls from './components/layer-controls';
 
 import LAYER_CATEGORIES from './examples';
+import 'luma.gl/debug';
 
 /* eslint-disable no-process-env */
 const MapboxAccessToken =
@@ -292,7 +293,7 @@ export default class App extends PureComponent {
           onLayerHover={this._onHover}
           onLayerClick={this._onClick}
           useDevicePixels={useDevicePixels}
-          debug={false}
+          debug={true}
           drawPickingColors={drawPickingColors}
         >
           <View id="basemap">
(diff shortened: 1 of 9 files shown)


modules/experimental-layers/src/gpu-grid-layer/gpu-grid-cell-layer.js
```diff
@@ -31,7 +31,7 @@ const DEFAULT_MAXCOLOR = [0, 255, 0, 255];
 const AGGREGATION_DATA_UBO_INDEX = 0;
 
 const defaultProps = {
-  cellSize: {type: 'number', min: 0, max: 20, value: 20},
+  cellSize: {type: 'number', min: 0, max: 1000, value: 1000},
```
Pessimistress: @1chandu is there a reason to limit `cellSize` to 20?

1chandu: Nope, there shouldn't be any, this change is fine.

examples/layer-browser/src/app.js
```diff
@@ -292,7 +293,7 @@ export default class App extends PureComponent {
           onLayerHover={this._onHover}
           onLayerClick={this._onClick}
           useDevicePixels={useDevicePixels}
-          debug={false}
+          debug={true}
```
1chandu: This flag also used to create debug `WebGL` context, wondering if overloading is fine or we create a new flag `validateProps` ?

Pessimistress: I really don't want to use another prop. Ideally I wish this could do what React does (switch between dev and prod build by looking for `NODE_ENV=production`)

Pessimistress: Opened issue https://github.com/uber/deck.gl/issues/2386

PR #2389, opened 2018-10-26
Add a width scale to arc layer so it maintains projected size

akre54:
#### Background
example
As you zoom out, the ArcLayer's line widths appear to grow instead of staying consistent with the projection of the viewport. 

#### Change List
-This adds a `widthScale` uniform float prop meant to be used to determine the scaling factor based on the zoom level:

```
new ArcLayer({
  // ...
  widthScale: Math.pow(2, viewport.zoom) / scaleFactor,
})
```

Alternatively, the layer could unproject the viewport's zoom by default, and call worldToPixels on the scaleFactor. Open to suggestions.

### DIFF
--- a/docs/layers/arc-layer.md
+++ b/docs/layers/arc-layer.md
@@ -36,6 +36,7 @@ const App = ({data, viewport}) => {
     data,
     pickable: true,
     getStrokeWidth: 12,
+    widthScale: Math.pow(viewport.zoom, 2) / 4096,
     getSourcePosition: d => d.from.coordinates,
     getTargetPosition: d => d.to.coordinates,
     getSourceColor: d => [Math.sqrt(d.inbound), 140, 0],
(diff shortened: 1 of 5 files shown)


modules/layers/src/arc-layer/arc-layer-vertex.glsl.js
```diff
@@ -52,7 +53,7 @@ vec2 getExtrusionOffset(vec2 line_clipspace, float offset_direction) {
   // rotate by 90 degrees
   dir_screenspace = vec2(-dir_screenspace.y, dir_screenspace.x);
 
-  vec2 offset_screenspace = dir_screenspace * offset_direction * instanceWidths / 2.0;
+  vec2 offset_screenspace = dir_screenspace * offset_direction * instanceWidths * widthScale / 2.0;
```
Pessimistress: Same needs to be added to the fp64 shader.

docs/layers/arc-layer.md
```diff
@@ -59,6 +60,12 @@ Inherits from all [Base Layer](/docs/api-reference/layer.md) properties.
 
 Whether the layer should be rendered in high-precision 64-bit mode. Note that since deck.gl v6.1, the default 32-bit projection uses a hybrid mode that matches 64-bit precision with significantly better performance.
 
+##### `widthScale` (number, optional)
+
+* Default: 1
+
+The scaling factor for the width of each arc. Usually `Math.pow(viewport.zoom, 2) / Math.pow(2, 12)`, corresponding to the current zoom level and the width of 1 pixel at zoom level 12. You can limit the minimum size of the arc with this property.
```
Pessimistress: s/Usually/If set to

Pessimistress: Also `Math.pow(2, viewport.zoom - 12)`

docs/layers/arc-layer.md
```diff
@@ -36,6 +36,7 @@ const App = ({data, viewport}) => {
     data,
     pickable: true,
     getStrokeWidth: 12,
+    widthScale: Math.pow(viewport.zoom, 2) / 4096,
```
Pessimistress: Remove this - this code sample is supposed to match the embedded demo

PR #2390, opened 2018-10-26
Implement Event handling RFC

Pessimistress:
#### Background

[RFC](https://github.com/uber/deck.gl/blob/master/dev-docs/RFCs/v6.x/event-handling-rfc.md)

#### Change List
- Support `onDragStart`, `onDrag`, `onDragEnd` callbacks in both layer and deck props.

### DIFF
--- a/docs/api-reference/layer.md
+++ b/docs/api-reference/layer.md
@@ -118,7 +118,40 @@ Requires `pickable` to be true.
 
 This callback will be called when the mouse clicks over an object of this deck.gl layer with a single parameter [`info`](/docs/get-started/interactivity.md#the-picking-info-object).
 
-If this callback returns a truthy value, the `hover` event is marked as handled and will not bubble up to the [`onLayerClick`](/docs/api-reference/react/deckgl.md#-onlayerclick-function-optional-) callback of the `DeckGL` canvas.
+If this callback returns a truthy value, the `click` event is marked as handled and will not bubble up to the [`onLayerClick`](/docs/api-reference/react/deckgl.md#-onlayerclick-function-optional-) callback of the `DeckGL` canvas.
+
+Requires `pickable` to be true.
+
+##### `onDragStart` (Function, optional)
+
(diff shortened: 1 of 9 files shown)


docs/api-reference/layer.md
```diff
@@ -118,7 +118,40 @@ Requires `pickable` to be true.
...
+Requires `pickable` to be true.
+
+##### `onDragStart` (Function, optional)
+
+This callback will be called when the mouse starts dragging an object of this deck.gl layer with the following parameters:
+
+* [`info`](/docs/get-started/interactivity.md#the-picking-info-object)
+* `event` - the source event
+
+If this callback returns a truthy value, the `dragstart` event is marked as handled and will not bubble up to the [`onDragStart`](/docs/api-reference/react/deckgl.md#-ondragstart-function-optional-) callback of the `DeckGL` canvas.
```
supersonicclay: Does this mean that this is how a layer could cancel/disable the map panning action?

docs/api-reference/layer.md
```diff
@@ -118,7 +118,40 @@ Requires `pickable` to be true.
 
 This callback will be called when the mouse clicks over an object of this deck.gl layer with a single parameter [`info`](/docs/get-started/interactivity.md#the-picking-info-object).
 
-If this callback returns a truthy value, the `hover` event is marked as handled and will not bubble up to the [`onLayerClick`](/docs/api-reference/react/deckgl.md#-onlayerclick-function-optional-) callback of the `DeckGL` canvas.
+If this callback returns a truthy value, the `click` event is marked as handled and will not bubble up to the [`onLayerClick`](/docs/api-reference/react/deckgl.md#-onlayerclick-function-optional-) callback of the `DeckGL` canvas.
+
+Requires `pickable` to be true.
+
+##### `onDragStart` (Function, optional)
```
supersonicclay: Can these new events be handled by a `CompositeLayer` now or still only usable from whatever is instantiating those layers?

Pessimistress: The composite layer will have to overwrite the `on*` props in its constructor.

PR #2393, opened 2018-10-26
Fix FirstPersonView offset

Pessimistress:
#### Background

After https://github.com/uber/deck.gl/pull/2284, `pixelPerMeters` already adjusts for Mercator plane Y-flip.

#### Change List
- Correct offset projection in FirstPersonView


Pessimistress:
> Would this affect infoviz (orthographic) use cases too

This only applies to geospatial viewports. I will add tests in a follow up PR.

PR #2404, opened 2018-11-02
Improve mapbox module

Pessimistress:
#### Change List
- Fix a bug for React integration: `props.layerFilter` is reset every render due to default props
- Accept a new `renderingMode` prop to `MapboxLayer`. Value can be `'2d'` or `'3d'` (default)
- When using React, also draw layers that are not added to Mapbox

### DIFF
--- a/modules/mapbox/src/deck-utils.js
+++ b/modules/mapbox/src/deck-utils.js
@@ -11,13 +11,8 @@ export function getDeckInstance({map, gl, deck}) {
     width: '100%',
     height: '100%',
     useDevicePixels: true,
-    // layerFilter needs to be changed inside a rendering/picking cycle
-    // But calling setProps({layerFilter}) will trigger another rerender which sets off an infinite loop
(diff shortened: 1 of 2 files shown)


modules/mapbox/src/deck-utils.js
```diff
@@ -56,21 +52,35 @@ export function updateLayer(deck, layer) {
...
-  const {layerFilter} = deck.props.userData;
-
-  if (typeof layerFilter !== 'string') {
-    return layerFilter;
+function afterRender(deck) {
+  const {mapboxLayers, isExternal} = deck.props.userData;
+
+  if (isExternal) {
+    // Draw non-Mapbox layers
+    const mapboxLayerIds = Array.from(mapboxLayers).map(layer => layer.id);
```
akre54: Minor, but `Array.from` can take a second mapper argument, no need for the `map`

```suggestion
    const mapboxLayerIds = Array.from(mapboxLayers, layer => layer.id);
```

PR #2417, opened 2018-11-07
GPUGridAggregator: Add support for MEAN operations

1chandu:
For #2291

#### Background
Adding Mean operation support.
CPU: After initial aggregation, calculate mean weights (RGB) based on count (Alpha).
GPU: Use `Transform`'s texture API to calculate Mean.

#### Change List
- CPU: Add Mean operation support, and unit tests.
- GPU: Add Mean operation support

### DIFF
--- a/modules/core/src/experimental/utils/gpu-grid-aggregation/transform-mean-vs.glsl.js
+++ b/modules/core/src/experimental/utils/gpu-grid-aggregation/transform-mean-vs.glsl.js
@@ -1,0 +1,34 @@
+// Copyright (c) 2015 - 2018 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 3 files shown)


modules/core/src/experimental/utils/gpu-grid-aggregation/transform-mean-vs.glsl.js
```diff
@@ -0,0 +1,34 @@
...
+#define SHADER_NAME gpu-aggregation-transform-mean-vs
+attribute vec4 aggregationValues;
+varying vec4 meanValues;
+
+void main()
+{
+  // TODO: Use 64-bit division ?? not needed given this is aggregation ??
+  bool isCellValid = bool(aggregationValues.w);
+  // aggregationValues:  XYZ contain aggregated values, W contains count
+  meanValues.xyz = isCellValid ? aggregationValues.xyz/aggregationValues.w : vec3(0, 0, 0);
```
Pessimistress: Do you see a precision issue doing the division here?

Issue #2414, opened 2018-11-06
userData is not present in context on initial render


goszczynskip:
When supplying `userData` attribute to deck.gl component isn't present in `context` passed to `initializeState` method.
Version: 6.2.3

#### Code Example
```
class CustomLayer extends CompositeLayer {
  initializeState(context) {
   console.log(context.userData.foo)
  }
}

// in react component
<DeckGL 
  userData={{ foo: "bar" }} 
  layers={[new CustomLayer({ data:[] })]} 
  {...otherProps}>
  {children}
</DeckGL>
```
#### Actual Result
`console.log` prints `undefined`

#### Expected Result
`console.log` should print `"bar"`

#### Potential culprit
`setLayers` are called before setting userData. 
https://github.com/uber/deck.gl/blob/7cde1e1e01acbecef5e6634fc88510be5576ce87/modules/core/src/lib/layer-manager.js#L165
https://github.com/uber/deck.gl/blob/7cde1e1e01acbecef5e6634fc88510be5576ce87/modules/core/src/lib/layer-manager.js#L184
Changing their order should fix this bug.

Pessimistress:
Published 6.2.4.

PR #2418, opened 2018-11-07
fix userdata is not present on initial render

goszczynskip:
For #2414

#### Change List
- moves setting userData before layer initialization

### DIFF
--- a/modules/core/src/lib/layer-manager.js
+++ b/modules/core/src/lib/layer-manager.js
@@ -166,6 +166,11 @@ export default class LayerManager {
       this._debug = props.debug;
     }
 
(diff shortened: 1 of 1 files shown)


modules/core/src/lib/layer-manager.js
```diff
@@ -185,11 +190,6 @@ export default class LayerManager {
       }
     }
 
-    // A way for apps to add data to context that can be accessed in layers
-    if ('userData' in props) {
-      this.context.userData = props.userData;
-    }
-
     if ('useDevicePixels' in props) {
```
Pessimistress: Move this too since it's also populating the context?

PR #2421, opened 2018-11-10
Fallback to gl.canvas.height when gl.canvas.clientHeight are not avai…

sakitam-fdd:
For #2296
#### Change List
Fallback to gl.canvas.height when gl.canvas.clientHeight are not available
-

### DIFF
--- a/modules/core/src/lib/draw-layers.js
+++ b/modules/core/src/lib/draw-layers.js
@@ -37,7 +37,8 @@ export const getPixelRatio = ({useDevicePixels}) => {
 // Convert viewport top-left CSS coordinates to bottom up WebGL coordinates
 const getGLViewport = (gl, {viewport, pixelRatio}) => {
   // TODO - dummy default for node
-  const height = gl.canvas ? gl.canvas.clientHeight : 100;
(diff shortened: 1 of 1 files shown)


sakitam-fdd:
@1chandu Yes, I am very happy to see your changes. But I don't think this change is fully functional. In my last test, I found that in addition to the [deck.js](https://github.com/uber/deck.gl/blob/master/modules/core/src/lib/deck.js) needs to be rolled back, the [draw-layer](https://github.com/uber/deck.gl/blob/master/modules/core/src/lib/draw-layers.js) also needs to be rolled back. So I made this change.

fuzhenn:
Hi @1chandu ,

#2405 is not enough to fix the off-screen canvas bug described in #2296. `getGLViewport` will still blank the canvas as it only reads `canvas.clientHeight` as viewport height which is 0 when canvas is off-screen. It also needs to fallback canvas.height when canvas.clientHeight is 0.

As I have tested, this PR will have `getGLViewport` get the correct viewport height and perfectly fix the off-screen bug, please consider to reopen this PR and have it merged.

Thanks very much!

1chandu:
@fuzhenn , this looks good, #2405 missed `getGLViewport`, I just merged into master.

PR #2427, opened 2018-11-13
Allow layers to implement event handling via class methods

Pessimistress:
#### Change List
- RFC
- Update event handling logic

### DIFF
--- a/modules/core/src/lib/layer.js
+++ b/modules/core/src/lib/layer.js
@@ -40,7 +40,6 @@ import {worldToPixels} from 'viewport-mercator-project';
 const LOG_PRIORITY_UPDATE = 1;
 
 const EMPTY_ARRAY = Object.freeze([]);
-const noop = () => {};
 
 const defaultProps = {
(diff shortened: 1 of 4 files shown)


modules/core/src/lib/layer.js
```diff
@@ -244,6 +243,19 @@ export default class Layer extends Component {
...
+      this.props.onHover(info);
+    }
+  }
+
+  onClick(info) {
+    if (this.props.onClick) {
+      this.props.onClick(info);
+    }
+  }
+
```
javidhsueh: how about onDrag?

javidhsueh: onDragStart, onDrag, onDragEnd

Pessimistress: See [deck.js:617](https://github.com/uber/deck.gl/pull/2427/files#diff-2c3ba220c52980e05b7f0f83e67bb7e5R617) - The new callbacks are added dynamically so they don't have default layer implementations.

Issue #2180, opened 2018-08-10
Contour Layer - support Isobands


1chandu:
#### Background 
Current Contour layer supports rendering of contour lines (iso-lines), enhance this layer to support iso-bands.

#### To Do List 
  - [X] Add label and assign to milestone 
  - [ ] Coding
  - [ ] Doc update
  - [ ] Whats new update 
  - [ ] Test

PR #2429, opened 2018-11-13
 IsoBands Part-1: Add Marching Squares utility methods

1chandu:
For #2180 

#### Background
- Add support for `saddle` cases for Iso Lines.
- Add `Marching Squares` methods to generate Iso Bands.
- Add Unit tests.
- Some cleanup.

#### Change List
- Rename thresholdValue to threshold
- IsoBands Part-1: Add marching squares utility methods


modules/layers/src/contour-layer/marching-squares-codes.js
```diff
@@ -0,0 +1,444 @@
...
+  // where T: Top, TR: Top-right, R: Right and C: current, each will be either 0, 1 or 2
+  // final code is binary represenation of above code , where takes 2 digits
+  // for example:  code 2-2-2-1 => 10-10-10-01 => 10101001 => 169
+
+  // no contours
+
+  // 0000
+  0: [],
+  // 2222
+  170: [],
```
Pessimistress: Does it make sense to write it like this:

```js
[parseInt('2222', 4)]: []
```

To eliminate conversion errors?

modules/layers/src/contour-layer/marching-squares.js
```diff
@@ -1,43 +1,51 @@
...
+const Z_OFFSET = 0.002;
+
+const DEFAULT_THRESHOLD_DATA = {
+  zIndex: 0,
+  zOffsetScale: 1
 };
 
+// Utility methods
+
+function getVertexCode({weight, threshold}) {
```
Pessimistress: How frequently is this called? For private utility functions that are performance-sensitive, you can skip named parameters to avoid the creation and garbage collection of small objects.

modules/layers/src/contour-layer/marching-squares.js
```diff
@@ -1,43 +1,51 @@
 // All utility mehtods needed to implement Marching Squres algorithm
 // Ref: https://en.wikipedia.org/wiki/Marching_squares
 import assert from 'assert';
```
Pessimistress: Just noticed this - `assert` is not a dependency and only works because of webpack polyfill. This has been replaced in the core module by https://github.com/uber/deck.gl/blob/master/modules/core/src/utils/assert.js and the layers module by https://github.com/uber/deck.gl/blob/master/modules/layers/src/geojson-layer/geojson.js#L24.

Generally speaking, we only use assert checks on things that involve user data or user implementation. If it is something strictly generated by the layer itself, you should implement a safe guard instead of throwing an error.

This needs to be fixed in a separate PR.

1chandu: Done, addressed in 2438

modules/layers/src/contour-layer/marching-squares.js
1chandu: yeah this is performed for each cell, i.e. called very frequently, I will pass individual parameters, to avoid object creation.

test/modules/experimental-layers/contour-layer/marching-squares.spec.js
```diff
@@ -1,10 +1,24 @@
 import test from 'tape-catch';
-import {getCode, getVertices} from '@deck.gl/layers/contour-layer/marching-squares';
+import {getCode, getVertices, CONTOUR_TYPE} from '@deck.gl/layers/contour-layer/marching-squares';
```
Pessimistress: This file should be moved to `test/modules/layers/` now?

PR #2437, opened 2018-11-15
Fix async prop comparison bug

Pessimistress:
These new test cases capture a bug: if an async prop (e.g. `data`) is not specified by the user, instead of resolving to default (`[]`), it is assumed to be "loading". As a result, when later a value is supplied to this prop via layer update, `changeFlags.dataChanged` is set to `false`.

This issue does not block any applications (can be worked around by setting `data: null`), but it causes unexpected behavior in some of our core layer tests.

#### Change List
- Resolve undefined async props correctly
- Remove a duplicate setAsyncProps call

### DIFF
--- a/modules/core/src/lib/layer.js
+++ b/modules/core/src/lib/layer.js
@@ -835,9 +835,6 @@ ${flags.viewportChanged ? 'viewport' : ''}\
       model.userData.layer = this;
     }
 
-    // Ensure any async props are updated
-    this.internalState.setAsyncProps(this.props);
-
     this.diffProps(this.props, this.internalState.getOldProps());
   }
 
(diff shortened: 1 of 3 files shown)


modules/core/src/lib/layer.js
```diff
@@ -835,9 +835,6 @@ ${flags.viewportChanged ? 'viewport' : ''}\
       model.userData.layer = this;
     }
 
-    // Ensure any async props are updated
-    this.internalState.setAsyncProps(this.props);
```
Pessimistress: This is a duplicate of line:831

Pessimistress:
> Is this specific to data prop only

No. Though the test cases only verify it for `data`.

PR #2438, opened 2018-11-16
Remove assert dependency from modules

1chandu:
For #

#### Background
As commented in #2429
- Use internal `assert` in core module.
- Remove assert usage in marching-squares/contour private utility methods, most of these were added for internal validation.
- Add `assert` definition for mesh layer, not sure if we need this, but added to be safe.

#### Change List
- Remove assert dependency from modules


modules/layers/src/screen-grid-layer/screen-grid-layer.js
```diff
@@ -362,7 +359,6 @@ export default class ScreenGridLayer extends Layer {
 
     if (oldProps.colorRange !== props.colorRange) {
       const colorRangeUniform = [];
-      assert(props.colorRange && props.colorRange.length === COLOR_RANGE_LENGTH);
```
1chandu: This `props` is already documented to be an array of 6.

Issue #2399, opened 2018-11-01
Fix invalid accessor comparisons in all layers


1chandu:
#### Background 
All layers should make use of `changeFlags.updateTriggersChanged` instead of comparing `function` type accessors, since function instances are always different for every time react render call. (layers are re-initialized, giving new function instances)

Also props comparison is shallow, and it doesn't compare `accessor` type props, so `updateState` is not called by default.

Correct usage (PathLayer) : 
https://github.com/uber/deck.gl/blob/7cde1e1e01acbecef5e6634fc88510be5576ce87/modules/layers/src/path-layer/path-layer.js#L129

In correct usage (GridLayer) :
https://github.com/uber/deck.gl/blob/7cde1e1e01acbecef5e6634fc88510be5576ce87/modules/layers/src/grid-layer/grid-layer.js#L86

#### To Do List 
  - [x] Add label and assign to milestone 
  - [x] Search all layers with invalid usage and fix them.
  - [x] Test

1chandu:
@XiaokaiUber , I believe your #2442 resolved this, please re-open for any pending issues.

PR #2442, opened 2018-11-16
fixed invalid accessor comparisons in contour layer and grid layer

XiaokaiUber:
For #2399 

#### Background
All layers should make use of changeFlags.updateTriggersChanged instead of comparing function type accessors.

#### Change List
- modified the incorrect comparison in contour layer and grid layer


Pessimistress:
The issue also exists in HexagonLayer. Either fix it or change your PR title.

XiaokaiUber:
> The issue also exists in HexagonLayer. Either fix it or change your PR title.

@Pessimistress it's updated.

PR #2448, opened 2018-11-17
Allow user to change min/max zoom of the orthographic controller

Pessimistress:
#### Background

minZoom and maxZoom should not be hard-coded.

#### Change List
- OrthographicController
- docs

### DIFF
--- a/modules/core/src/controllers/orthographic-controller.js
+++ b/modules/core/src/controllers/orthographic-controller.js
@@ -3,11 +3,10 @@ import ViewState from './view-state';
 
 import {Vector2, clamp} from 'math.gl';
 
-// TODO: not sure whether we should have these constants
-// as part of viewport props
 const MOVEMENT_SPEED = 10; // per keyboard click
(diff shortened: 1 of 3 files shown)


modules/core/src/controllers/orthographic-controller.js
```diff
@@ -3,11 +3,10 @@ import ViewState from './view-state';
...
 import {Vector2, clamp} from 'math.gl';
 
-// TODO: not sure whether we should have these constants
-// as part of viewport props
 const MOVEMENT_SPEED = 10; // per keyboard click
-const MIN_ZOOM = 0.1;
-const MAX_ZOOM = 10;
+// TODO - make default unlimited in the next major version
+const DEFAULT_MIN_ZOOM = 0.1;
+const DEFAULT_MAX_ZOOM = 10;
```
gnavvy: I feel [0.1, 10] is good.
Users can at least see something if they messed up the zoom and scale at the beginning.

Issue #2447, opened 2018-11-17
OrthographicView zoom is reversed


Pessimistress:
#### Actual Result
When zooming in, `zoom` value decreases

#### Expected Result
When zooming in, `zoom` value increases

#### Reproduce Steps

https://codepen.io/Pessimistress/pen/RqZJEg

zoom in/out and watch the zoom value.

#### To Do List
  - [x] Add label and assign to milestone
  - [x] Coding
  - [ ] Test

Pessimistress:
@javidhsueh @gnavvy Is this something by design?

javidhsueh:
> The issue is here: https://github.com/uber/deck.gl/blob/master/modules/core/src/views/orthographic-view.js#L33
> 
> When zoom level is higher, the camera should be looking at a smaller bounding box, not larger.

Does it mean we should do:
```
      left = (-width / 2 + offset[0]) / zoom, // Left bound of the frustum
      top = (-height / 2 + offset[1]) / zoom, // Top bound of the frustum
      right = (width / 2 + offset[0]) / zoom, // Right bound of the frustum
      bottom = (height / 2 + offset[1]) / zoom, // Bottom bound of the frustum
```

Pessimistress:
Essentially yes.
A separate alignment issue is that if we follow the naming convention of the MapController, `zoom` is logarithmic and `scale = Math.pow(2, zoom)`.

PR #2466, opened 2018-11-28
Reverse zoom in Orthographic controller

javidhsueh:
For #2447

### DIFF
--- a/modules/core/src/controllers/orthographic-controller.js
+++ b/modules/core/src/controllers/orthographic-controller.js
@@ -19,10 +19,12 @@ class OrthographicState extends ViewState {
     maxZoom = DEFAULT_MAX_ZOOM,
 
     /** Interaction states */
+    /* The point on the view being grabbed when the operation first started */
     startPanPosition,
+    /* The offset on the view being grabbed when the operation first started */
     startPanOffset,
-    startRotatePosition,
-    startRotateOffset
+    /** The zoom level when the first zoom operation started */
+    startZoom
   }) {
     super({
       width,
@@ -35,8 +37,7 @@ class OrthographicState extends ViewState {
     this._interactiveState = {
       startPanPosition,
       startPanOffset,
-      startRotatePosition,
-      startRotateOffset
+      startZoom
     };
   }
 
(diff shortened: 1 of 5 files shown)


modules/core/src/controllers/orthographic-controller.js
```diff
@@ -105,15 +106,43 @@ class OrthographicState extends ViewState {
...
+      // pinch started).
+      // If startZoom state is defined, then use the startZoom state;
+      // otherwise assume discrete zooming
+      startZoom = this._viewportProps.zoom;
+    }
+
+    const newZoom = this._calculateNewZoom({scale, startZoom});
     const centerX = width / 2 - offset[0];
     const centerY = height / 2 - offset[1];
     const dX = (pos[0] - centerX) * (zoom / newZoom - 1);
```
Pessimistress: Zoom center is broken. I believe this needs to be reversed as well.

Pessimistress: When you zoom in/out the plot is supposed to scale around your pointer location. On your branch:

ortho-controller

javidhsueh: zoom

javidhsueh:
zoom

modules/core/src/controllers/orthographic-controller.js
```diff
@@ -105,25 +108,82 @@ class OrthographicState extends ViewState {
...
+      startZoom = this._viewportProps.zoom;
+      startZoomPosition = startPos || pos;
+    }
+
+    const newZoom = this._calculateNewZoom({scale, startZoom});
     const centerX = width / 2 - offset[0];
     const centerY = height / 2 - offset[1];
-    const dX = (pos[0] - centerX) * (zoom / newZoom - 1);
-    const dY = (pos[1] - centerY) * (zoom / newZoom - 1);
+    const dX = (startZoomPosition[0] - centerX) * (zoom / newZoom - 1);
```
Pessimistress: - Why use `startZoomPosition` instead of `pos`?
- This is still incorrect. If `zoom` is logarithmic you need to use `scale` and `newScale` here.

modules/core/src/views/orthographic-view.js
```diff
@@ -30,10 +30,10 @@ export default class OrthographicView extends View {
...
     const {
       // projection matrix arguments
-      left = (-width / 2 + offset[0]) * zoom, // Left bound of the frustum
-      top = (-height / 2 + offset[1]) * zoom, // Top bound of the frustum
-      right = (width / 2 + offset[0]) * zoom, // Right bound of the frustum
-      bottom = (height / 2 + offset[1]) * zoom, // Bottom bound of the frustum
+      left = (-width / 2 + offset[0]) / zoom, // Left bound of the frustum
+      top = (-height / 2 + offset[1]) / zoom, // Top bound of the frustum
+      right = (width / 2 + offset[0]) / zoom, // Right bound of the frustum
+      bottom = (height / 2 + offset[1]) / zoom, // Bottom bound of the frustum
```
Pessimistress: Should use scale which is 2^zoom

modules/core/src/controllers/orthographic-controller.js
Pessimistress: orthographic

PR #2472, opened 2018-12-01
Update layer docs regarding tooltip display

Pessimistress:
For https://github.com/uber/deck.gl/issues/1861

#### Change List
- Add tooltip examples to "adding interactivity" page
- Update each layer example with link


Pessimistress:
Yeah it is probably a pretty small effort on our end. This has proved to be a pain point especially for users unfamiliar with React.

PR #2475, opened 2018-12-03
fix updateTriggers not being called issue when transitioning between 0 and null

XiaokaiUber:
For #2068 

#### Background
updateTriggers doesn't work when transitioning between 0 and null

#### Change List
- modified diffUpdateTrigger in props.js. It should return 0 instead of {} when the trigger is 0.

Example demo(in branch `update-trigger-bug`)  
update_trigger_bug

### DIFF
--- a/modules/core/src/lifecycle/props.js
+++ b/modules/core/src/lifecycle/props.js
@@ -147,7 +147,7 @@ function diffUpdateTriggers(props, oldProps) {
 
   // If the 'all' updateTrigger fires, ignore testing others
   if ('all' in props.updateTriggers) {
-    const diffReason = diffUpdateTrigger(oldProps, props, 'all');
+    const diffReason = diffUpdateTrigger(props, oldProps, 'all');
     if (diffReason) {
       return {all: true};
     }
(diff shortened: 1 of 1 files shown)


modules/core/src/lifecycle/props.js
```diff
@@ -170,8 +170,10 @@ function diffUpdateTriggers(props, oldProps) {
 }
 
 function diffUpdateTrigger(props, oldProps, triggerName) {
-  const newTriggers = props.updateTriggers[triggerName] || {};
-  const oldTriggers = oldProps.updateTriggers[triggerName] || {};
+  let newTriggers = props.updateTriggers[triggerName];
```
jianhuang01: why do we have a legitimate "newTriggers === 0"  case?

XiaokaiUber: @jianhuang01 
In the demo example, the updateTriggers is defined as `{getColor: this.state.colorIndex}`. When calling [diffUpdateTrigger](https://github.com/uber/deck.gl/blob/master/modules/core/src/lifecycle/props.js#L173-L174)
`updateTriggers[triggerName]` may return 0.

Pessimistress:
Please add tests for this use case.

test/modules/core/lib/layer.spec.js
```diff
@@ -209,6 +210,38 @@ test('Layer#diffProps', t => {
   t.ok(spy.called, 'updateTriggers fired');
   spy.restore();
 
+  layer = new SubLayer(Object.assign({}, LAYER_PROPS, {updateTriggers: {time: 0}}));
+  testInitializeLayer({layer});
+  layer.diffProps(
+    new SubLayer(Object.assign({}, LAYER_PROPS, {updateTriggers: {time: 0}})).props,
+    layer.props
+  );
+  t.false(layer.getChangeFlags().updateTriggersChanged, 'updateTriggers not fired');
```
Pessimistress: `updateTriggersChanged`

PR #2476, opened 2018-12-04
Add bootstrap to publish script

Pessimistress:
#### Background

The dependencies on master and release branches have diverged. We had one incident in 6.2.3 when deck.min.js was bundled with the wrong version of mjolnir.js, which could not be caught by our tests. This change will ensure that the correct dependencies are installed whenever we publish.

#### Change List
- Always run bootstrap before build


package.json
```diff
@@ -24,8 +24,8 @@
     "cover": "NODE_ENV=test nyc --reporter html --reporter cobertura --reporter=lcov npm run test cover",
     "lint": "scripts/lint.sh",
     "lint-docs": "scripts/lint-markdown.sh",
-    "publish-prod": "npm run build && npm run test && npm run test dist && scripts/publish.sh prod",
-    "publish-beta": "npm run build && npm run test && npm run test dist && scripts/publish.sh beta",
+    "publish-prod": "npm run bootstrap && npm run build && npm run test && npm run test dist && scripts/publish.sh prod",
```
1chandu: We should do the same for luma.gl too.

Issue #2468, opened 2018-11-29
highlightColor doesn't work on HexagonLayer


contra:
![kapture 2018-11-29 at 12 41 28](https://user-images.githubusercontent.com/425716/49240816-2722ab80-f3d4-11e8-8776-0845d3489a77.gif)

`autoHighlight` works great on the hexagon layer, but the color seems to be ignored and the default is used. This would also be a good opportunity to document these options for the hexagon layer. If this seems uncontroversial I'd be happy to submit a PR - let me know.

#### Actual Result

Highlighting is the default `cyan` highlightColor.

#### Expected Result

Highlighting would be black, given the code sample below.

#### Code Example

```js
  new HexagonLayer({
    id: 'test',
	data,
    extruded: true,
    radius: 25,
    autoHighlight: true,
    highlightColor: [ 0, 0, 0 ]
  })
```

#### To Do List 
  - [ ] Add label and assign to milestone
  - [ ] Coding
  - [ ] Test

#### To Do List 
  - [ ] Add label and assign to milestone 
  - [ ] Coding
  - [ ] Doc update
  - [ ] Whats new update 
  - [ ] Test

contra:
Ah - I seem to have found the issue. If opacity is missing from the color it gets ignored silently.

1chandu:
Thanks for reporting, I will take a look.

1chandu:
@contra , you should see following warnings from luma and webgl : 
```
luma: Uniform size should be multiples of 4 (3) [0, 0, 0]
uniforms.js:16 WebGL: INVALID_VALUE: uniform4fv: invalid size
```
I  will update the code to append alpha value similar to how we handle any other color values.

1chandu:
Resolved by #2479

PR #2479, opened 2018-12-04
Use default Alpha for picking highlight color

1chandu:
For #2468

When Alpha not provided for picking highlight color, use default value.

#### Change List
- Use default Alpha for picking highlight color

### DIFF
--- a/docs/api-reference/layer.md
+++ b/docs/api-reference/layer.md
@@ -167,7 +167,7 @@ Requires `pickable` to be true.
 
 * Default: `[0, 0, 128, 128]`
 
-RGBA color to be used to render highlighted object.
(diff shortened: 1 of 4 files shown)


docs/api-reference/layer.md
```diff
@@ -167,7 +167,7 @@ Requires `pickable` to be true.
 
 * Default: `[0, 0, 128, 128]`
 
-RGBA color to be used to render highlighted object.
+RGBA color to be used to render highlighted object. When 3 component (RGB) array is specified, a default value of 128 is used for alpha.
```
Pessimistress: We should use `255` for consistency with the rest of the layer APIs.
This won't break any existing use case as 3-component value doesn't work right now.

PR #2487, opened 2018-12-07
Add callback function onDataLoaded in TileLayer

starry97:
Add callback function -- `onDataLoaded` in TileLayer, which is called when all tiles in the viewport are loaded.

### DIFF
--- a/modules/experimental-layers/src/tile-layer/tile-layer.js
+++ b/modules/experimental-layers/src/tile-layer/tile-layer.js
@@ -27,7 +27,7 @@ export default class TileLayer extends CompositeLayer {
   }
 
   updateState({props, oldProps, context, changeFlags}) {
-    const {onDataLoaded} = props;
+    const {onDataLoaded, onGetTileDataError} = props;
     if (
       changeFlags.updateTriggersChanged &&
       (changeFlags.updateTriggersChanged.all || changeFlags.updateTriggersChanged.getTileData)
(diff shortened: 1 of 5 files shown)


modules/experimental-layers/src/tile-layer/tile-layer.js
```diff
@@ -4,6 +4,7 @@ import TileCache from './utils/tile-cache';
 const defaultProps = {
   renderSubLayers: props => new GeoJsonLayer(props),
   getTileData: ({x, y, z}) => Promise.resolve(null),
+  onDataLoaded: data => null,
```
Pessimistress: Nit: Use the standard noOp `() => {}` as default

modules/experimental-layers/src/tile-layer/tile-layer.js
```diff
@@ -43,9 +45,12 @@ export default class TileLayer extends CompositeLayer {
           const allCurrTilesLoaded = currTiles.every(tile => tile.isLoaded);
```
Pessimistress: Currently in Tile class, if loading errors out, then `isLoaded` will never be set. This causes unnecessary overhead every time the viewport changes. I wonder if it makes more sense to make `isLoaded` true and keep `_data` empty in this case.

modules/experimental-layers/src/tile-layer/utils/tile.js
```diff
@@ -28,10 +29,15 @@ export default class Tile {
...
-      return buffers;
-    });
+    getTileDataPromise
+      .then(buffers => {
+        this._data = buffers;
+        return buffers;
+      })
+      .catch(err => {
+        // eslint-disable-next-line
+        console.warn(err);
```
Pessimistress: It's always better to supply a callback instead of logging to console.

PR #2503, opened 2018-12-14
UnitTests: for path layer shaders

1chandu:
For #
**NOTE** requires : https://github.com/uber/luma.gl/pull/812

As discussed in https://github.com/uber/deck.gl/pull/2495

#### Background
Reference implementation to add unit tests for a layer shader. Shader should be divded into multiple glsl functions and these functions can be tested individually.
More robust unit tests can be added using this as reference, for which we need the correct values for intermediate shader variables.

#### Change List
- UnitTests: for path layer shaders

### DIFF
--- a/test/modules/layers/path-layer/path-layer-vertex.spec.js
+++ b/test/modules/layers/path-layer/path-layer-vertex.spec.js
@@ -1,0 +1,41 @@
+import test from 'tape-catch';
+
+// import {COORDINATE_SYSTEM, Viewport, WebMercatorViewport} from 'deck.gl';
+import {gl} from '@deck.gl/test-utils';
+import {Transform, Buffer} from 'luma.gl';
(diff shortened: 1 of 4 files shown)


test/modules/layers/path-layer/path-layer-vertex.spec.js
```diff
@@ -0,0 +1,41 @@
+import test from 'tape-catch';
+
+// import {COORDINATE_SYSTEM, Viewport, WebMercatorViewport} from 'deck.gl';
+import {gl} from '@deck.gl/test-utils';
+import {Transform, Buffer} from 'luma.gl';
+import VS from '../../../../modules/layers/src/path-layer/path-layer-vertex.glsl';
```
Pessimistress: Should import from `@deck.gl/layers/path-layer/...` and rely on module-alias to resolve the path.

Using relative path may break dist test.

PR #2515, opened 2018-12-18
PathLayer shader: 2-dimensional vPathPosition

Pessimistress:
#### Background
The old varying `float vPathPosition` is set to the position of a fragment along the length of the path, normalized by width. This PR makes the varying a `vec2`, adding the position along the width of the path as the `x` component.

As an attempt to make the PathLayer more customizable, this change addresses an internal use case where we need to subclass the PathLayer to add custom fill logic in the fragment shader, including double line, path offset, and custom marking patterns.

Example to draw double line:

```js
class MyPathLayer extends PathLayer {
  getShaders() {
    return {
      ...super.getShaders(),
      inject: {
        'fs:#main-start': `if (abs(vPathPosition.x) < 0.1) discard;`
      }
    }
  }
}
```

#### Change List
- Change `vPathPosition` varying from `float` to `vec2`

### DIFF
--- a/modules/layers/src/path-layer/path-layer-fragment.glsl.js
+++ b/modules/layers/src/path-layer/path-layer-fragment.glsl.js
@@ -31,6 +31,11 @@ varying vec4 vColor;
 varying vec2 vCornerOffset;
 varying float vMiterLength;
 varying vec2 vDashArray;
+/*
+ * vPathPosition represents the relative coordinates of the current fragment on the path segment.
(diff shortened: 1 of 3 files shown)


Pessimistress:
> Maybe layer shaders could export a public interface.

I like the sound of this. How will it work though? Do we offer an JavaScript API for more specific shader injection than `fs:#main-start`?

Pessimistress:
@1chandu Setting the alpha to zero has the advantage that the fragment is still pickable. It's actually possible to discard a fragment outside the main() function, so the user does not have to choose between one or another.

PR #2520, opened 2018-12-19
specify types for defaultProps in mesh layer

XiaokaiUber:
For #2477 

#### Background
In mesh layer, the defaultProps don't have types specified. We need to specify type for those props.

#### Change List
- Add type to the defaultProps in the mesh layer

note: tested with render test and layer browser

### DIFF
--- a/modules/experimental-layers/src/mesh-layer/mesh-layer.js
+++ b/modules/experimental-layers/src/mesh-layer/mesh-layer.js
@@ -106,13 +106,13 @@ const defaultProps = {
   lightSettings: {},
 
   getPosition: {type: 'accessor', value: x => x.position},
-  getColor: {type: 'accessor', DEFAULT_COLOR},
+  getColor: {type: 'accessor', value: DEFAULT_COLOR},
 
   // yaw, pitch and roll are in degrees
   // https://en.wikipedia.org/wiki/Euler_angles
-  getYaw: {type: 'accessor', value: 0},
-  getPitch: {type: 'accessor', value: 0},
-  getRoll: {type: 'accessor', value: 0}
+  getYaw: {type: 'accessor', value: x => x.yaw || x.angle || 0},
+  getPitch: {type: 'accessor', value: x => x.pitch || 0},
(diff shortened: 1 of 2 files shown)


modules/experimental-layers/src/mesh-layer/mesh-layer.js
```diff
@@ -105,14 +105,14 @@ const defaultProps = {
   // Optional settings for 'lighting' shader module
   lightSettings: {},
 
-  getPosition: x => x.position,
-  getColor: x => x.color || DEFAULT_COLOR,
+  getPosition: {type: 'accessor', value: x => x.position},
+  getColor: {type: 'accessor', DEFAULT_COLOR},
```
1chandu: don't we need `value: x => x.color`, same for the ones below.

modules/experimental-layers/src/mesh-layer/mesh-layer.js
```diff
@@ -137,9 +136,13 @@ export default class MeshLayer extends Layer {
       instanceRotations: {
         size: 3,
         accessor: ['getYaw', 'getPitch', 'getRoll'],
-        update: this.calculateInstanceRotations
+        defaultValue: [0, 0, 0]
```
1chandu: should we match the order of array elements in `accessor` with how they used in `calculateInstanceRotations` ?

Pessimistress:
> Add a render test for mesh-layer

There is already a render test for path marker layer, which uses mesh layer. Render test does not pass on this branch, please fix it.

There is also a mesh layer test in layer browser, where you can test a combination of prop values.

modules/experimental-layers/src/mesh-layer/mesh-layer.md
```diff
@@ -99,7 +99,7 @@ See [Euler angles](https://en.wikipedia.org/wiki/Euler_angles).
 
 ##### `getColor` (Function, optional)
 
-- Default: `object => object.color || [0, 0, 0, 255]`
+- Default: `object => [0, 0, 0, 255]`
```
Pessimistress: Default is not a function.

PR #2521, opened 2018-12-19
add type for defaultProps in path outline layer

XiaokaiUber:
For #2477 

#### Background
In path outline layer, the defaultProps don't have types specified. We need to specify type for those props.

#### Change List
- Add type to the defaultProps in path outline layer

note: tested with render test and layer-browser. compared the rendering result with that in 6.3-release

### DIFF
--- a/modules/experimental-layers/src/path-outline-layer/path-outline-layer.js
+++ b/modules/experimental-layers/src/path-outline-layer/path-outline-layer.js
@@ -2,7 +2,6 @@ import {PathLayer} from '@deck.gl/layers';
 import GL from '@luma.gl/constants';
 import {Framebuffer, Texture2D} from 'luma.gl';
 import outline from '../shaderlib/outline/outline';
-import assert from '../../../core/src/utils/assert';
 
 // TODO - this should be built into assembleShaders
 function injectShaderCode({source, declarations = '', code = ''}) {
@@ -137,11 +136,7 @@ export default class PathOutlineLayer extends PathLayer {
     attribute.value = pathTesselator._updateAttribute({
       target: attribute.value,
       size: 1,
-      getValue: (object, index) => {
-        const zLevel = getZLevel(object, index);
-        assert(!isNaN(zLevel), '`getZLevel` does not return a number!');
-        return [zLevel];
-      }
+      getValue: (object, index) => [getZLevel(object, index) || 0]
     });
   }
 }
--- a/modules/experimental-layers/src/path-outline-layer/path-outline-layer.js
+++ b/modules/experimental-layers/src/path-outline-layer/path-outline-layer.js
@@ -62,6 +62,7 @@ export default class PathOutlineLayer extends PathLayer {
       instanceZLevel: {
         size: 1,
         type: GL.UNSIGNED_BYTE,
+        update: this.calculateZLevels,
         accessor: 'getZLevel',
         defaultValue: 0
       }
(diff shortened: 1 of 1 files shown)


modules/experimental-layers/src/path-outline-layer/path-outline-layer.js
```diff
@@ -62,8 +62,8 @@ export default class PathOutlineLayer extends PathLayer {
       instanceZLevel: {
         size: 1,
         type: GL.UNSIGNED_BYTE,
-        update: this.calculateZLevels,
```
Pessimistress: This cannot be removed.

modules/experimental-layers/src/path-outline-layer/path-outline-layer.js
```diff
@@ -63,7 +63,8 @@ export default class PathOutlineLayer extends PathLayer {
         size: 1,
         type: GL.UNSIGNED_BYTE,
         update: this.calculateZLevels,
-        accessor: 'getZLevel'
+        accessor: 'getZLevel',
+        defaultValue: 0
```
Pessimistress: This does nothing if auto update is not used.

Pessimistress:
Test your change in the layer browser.

Pessimistress:
Did you test this layer with `getZLevel` set to `0`?

XiaokaiUber:
> Did you test this layer with `getZLevel` set to `0`?

Yes. If `0` or any number is passed in, the image shows up but there is no outline. This is the same as that in `6.2-release`. But in `6.3-release`, it will crash.

modules/experimental-layers/src/path-outline-layer/path-outline-layer.js
```diff
@@ -136,7 +136,10 @@ export default class PathOutlineLayer extends PathLayer {
     attribute.value = pathTesselator._updateAttribute({
       target: attribute.value,
       size: 1,
-      getValue: (object, index) => getZLevel(object, index) || 0
+      getValue: (object, index) => {
+        const zLevel = getZLevel(object, index);
+        return isNaN(zLevel) ? 0 : [zLevel];
```
Pessimistress: Why is this function sometimes returning a number sometimes returning an array?

Pessimistress: 1. Your implementation should not rely on some niche behavior of another utility. The other function could be changed from checking `value.length` to Array.isArray(value)` and your layer will be broken.
2. If an accessor ever returns NaN, the layer should throw an error instead of silently not updating the attribute. It is certainly not the user's intention and they should know it.

modules/experimental-layers/src/path-outline-layer/path-outline-layer.js
```diff
@@ -2,6 +2,7 @@ import {PathLayer} from '@deck.gl/layers';
 import GL from '@luma.gl/constants';
 import {Framebuffer, Texture2D} from 'luma.gl';
 import outline from '../shaderlib/outline/outline';
+import assert from '../../../core/src/utils/assert';
```
Pessimistress: You cannot import something from another module. This is not going to be the directory structure when the module is published.

modules/experimental-layers/src/path-outline-layer/path-outline-layer.js
```diff
@@ -136,7 +137,11 @@ export default class PathOutlineLayer extends PathLayer {
     attribute.value = pathTesselator._updateAttribute({
       target: attribute.value,
       size: 1,
-      getValue: (object, index) => getZLevel(object, index) || 0
+      getValue: (object, index) => {
+        const zLevel = getZLevel(object, index);
+        assert(!isNaN(zLevel), '`getZLevel` does not return a number!');
```
Pessimistress: Just keep the old behavior before your change.

PR #2549, opened 2019-01-06
Apply TinySDF to TextLayer

xintongxia:
#### Background
[issue](https://github.com/uber/deck.gl/issues/2546)

This PR focuses on the first and second items

- Use SDF in font atlas generation
- fontSmoothing: Number

#### Change List
- Applied `TinySDF` library to encode the distance (distance between a data point and nearest character edge) into texture alpha channel

Rendering comparison:

1. looking:
```
sizeScale: 48
getSize: 100
```

Left is the original code, right is with `sdf` applied. 

tiny-sdf

2. performance:

```
fontSize: 64 // used in generating `fontAtlas`
```
Time used for generating fontAtlas:
```
original: <5 milliseconds 
sdf:      ~50 - 100 milliseconds 
```

### DIFF
--- a/docs/layers/text-layer.md
+++ b/docs/layers/text-layer.md
@@ -49,8 +49,10 @@ const App = ({data, viewport}) => {
 
 ## Example: rendering with sdf (Signed Distance Fields)
 
-This approach is slower but can provide a better this font is slower to generate but can provide a sharper look
-when used with very large or small font sizes.
+You may consider using [`sdf` (Signed Distance Fields)](http://cs.brown.edu/people/pfelzens/papers/dt-final.pdf)
+option when rendering with very large or small font sizes, which will provide a sharper look. `TextLayer` integrates 
+with [`TinySDF`](https://github.com/mapbox/tiny-sdf) which implements the `sdf` algorithm. You can enable `sdf` by 
+setting `sdf` to true or pass customized options, which will be used in `TinySDF`.
 
 ```js
 import DeckGL from 'deck.gl';
@@ -69,14 +71,14 @@ const App = ({data, viewport}) => {
     id: 'text-layer',
     data,
    
-   // sdf will be used to construct `TinySDF` instance
+   // sdf object will be used to construct `TinySDF` instance
    // https://github.com/mapbox/tiny-sdf
     sdf: {
       fontSize: 64,
-      buffer: 0,
+      buffer: 1,
       radius: 3,
       cutoff: 0.25,
-      fontWeight: 'normal',
+      fontWeight: 'normal'
     },
 
     pickable: true,
--- a/docs/layers/text-layer.md
+++ b/docs/layers/text-layer.md
@@ -8,6 +8,10 @@
 
 The text layer renders text labels on the map using texture mapping. This Layer is extended based on [Icon Layer](/docs/layers/icon-layer.md) and wrapped using [Composite Layer](/docs/api-reference/composite-layer.md).
 
+Auto pack required `characterSet` into a shared texture `fontAtlas`.
+
+## Example
+
 ```js
 import DeckGL from 'deck.gl';
 import TextLayer from './text-layer';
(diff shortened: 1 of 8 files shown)


docs/layers/text-layer.md
```diff
@@ -110,6 +110,23 @@ Whether the layer should be rendered in high-precision 64-bit mode. Note that si
...
+If you would like to tune the text rendering for deep zoom level, you can manipulate `cutoff` and `fontSmoothing`. 
+The implementation is based on [Drawing Text with Signed Distance Fields in Mapbox GL](https://blog.mapbox.com/drawing-text-with-signed-distance-fields-in-mapbox-gl-b0933af6f817) 
+
+##### `cutoff` (Number, optional)
+
+* Default: 192 / 255
+
+`cutoff` decides each pixel is inside or outside the glyph. `0` to `cutoff` indicates inside, `cutoff` to `1` indicates outside.
+
+##### `fontSmoothing` (Number, optional)
```
Pessimistress: Just `smoothing`?

modules/layers/src/text-layer/font-atlas.js
```diff
@@ -35,6 +44,15 @@ export function makeFontAtlas(
   const canvas = document.createElement('canvas');
   const ctx = canvas.getContext('2d');
 
+  const tinySDF = new TinySDF(
+    fontSize,
+    0, // buffer
+    3, // radius
+    0.25, // cutoff
+    fontFamily,
+    'normal' // fontWeight
```
Pessimistress: Let user set this with a layer prop?

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer-fragment.glsl.js
```diff
@@ -0,0 +1,61 @@
...
+const float buffer = 196.0 / 255.0;
+
+void main(void) {
+  vec4 texColor = texture2D(iconsTexture, vTextureCoords);
+  float distance = texture2D(iconsTexture, vTextureCoords).a;
+  float alpha = smoothstep(cutoff - smoothing, cutoff + smoothing, distance);
+
+  // if colorMode == 0, use pixel color from the texture
+  // if colorMode == 1 or rendering picking buffer, use texture as transparency mask
+  vec3 color = mix(texColor.rgb, vColor.rgb, vColorMode);
```
Pessimistress: We should not need this since `mask` is always true for this layer.

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer.js
```diff
@@ -21,8 +21,12 @@
 import IconLayer from '../../icon-layer/icon-layer';
 
 import vs from './multi-icon-layer-vertex.glsl';
+import fs from './multi-icon-layer-fragment.glsl';
 
 const defaultProps = {
+  fontSmoothing: 0,
+  cutoff: 192.0 / 255.0,
```
Pessimistress: - Define [prop type](http://deck.gl/#/documentation/developer-guide/writing-custom-layers/layer-prop-types) and range
- Add a comment on where this number comes from

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer.js
```diff
@@ -64,6 +69,18 @@ export default class MultiIconLayer extends IconLayer {
     }
   }
 
+  draw({uniforms}) {
+    const {cutoff, fontSmoothing, sizeScale} = this.props;
+
+    super.draw({
+      uniforms: Object.assign({}, uniforms, {
+        cutoff,
+        // TODO: handle for individual sizes using this.props.getSize()
```
Pessimistress: This needs to be handled in the vertex shader and calculated for each object.
Setting `getSize: 1, sizeScale: 10` and `getSize: 10, sizeScale: 1` should not result in any difference.

Pessimistress:
1. Can you add before & after screenshots to the summary of this PR?
2. I've noticed that it is much slower to generate the font atlas with this method. A few options we might want to consider:
  - Add the old method back as an user option.
  - Reuse the generated texture between multiple TextLayers and/or when the layer is removed then added back.

modules/layers/src/text-layer/font-atlas.js
```diff
@@ -35,6 +44,15 @@ export function makeFontAtlas(
   const canvas = document.createElement('canvas');
   const ctx = canvas.getContext('2d');
 
+  const tinySDF = new TinySDF(
+    fontSize,
+    0, // buffer
+    3, // radius
+    0.25, // cutoff
```
Pessimistress: How does this relate to `props.cutoff`?

docs/layers/text-layer.md
```diff
@@ -8,6 +8,53 @@
 
 The text layer renders text labels on the map using texture mapping. This Layer is extended based on [Icon Layer](/docs/layers/icon-layer.md) and wrapped using [Composite Layer](/docs/api-reference/composite-layer.md).
 
+##### Auto packing fontAtlas
```
Pessimistress: They are both auto packed.

docs/layers/text-layer.md
```diff
@@ -24,6 +71,16 @@ const App = ({data, viewport}) => {
   const layer = new TextLayer({
     id: 'text-layer',
     data,
+    
+    sdf: false,
```
Pessimistress: `true`

docs/layers/text-layer.md
```diff
@@ -8,6 +8,53 @@
...
+    }
+  });
+
+  return <DeckGL {...viewport} layers={[layer]} />;
+};
+```
+
+### Rendering With sdf (Signed Distance Fields)
+
+This is slower than `Auto packing fontAtlas` but can provide a better antialiased look.
```
Pessimistress: - Move the link to SDF paper here
- It's not really about antialiasing. try `this font is slower to generate but can provide a sharper look when used with very large or small font sizes.`

modules/layers/src/text-layer/font-atlas.js
```diff
@@ -16,6 +16,15 @@ for (let i = 32; i < 128; i++) {
   DEFAULT_CHAR_SET.push(String.fromCharCode(i));
 }
 
+function makeRGBAImageData(ctx, alphaChannel, size) {
+  const imageData = ctx.createImageData(size, size);
```
Pessimistress: What is the cost of creating may ImageData objects? Can we reuse?

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer-vertex.glsl.js
```diff
@@ -38,10 +38,12 @@ attribute vec2 instancePixelOffset;
 
 uniform float sizeScale;
 uniform vec2 iconsTextureDim;
+uniform float smoothing;
 
 varying float vColorMode;
```
Pessimistress: Remove

modules/layers/src/text-layer/text-layer.js
```diff
@@ -36,14 +36,33 @@ const ALIGNMENT_BASELINE = {
...
+
 const defaultProps = {
   fp64: false,
   sizeScale: 1,
-  fontFamily: DEFAULT_FONT_FAMILY,
+  // used in `makeFontAtlas`
+  sdf: false,
   characterSet: DEFAULT_CHAR_SET,
+  fontFamily: DEFAULT_FONT_FAMILY,
+  fontSize: DEFAULT_FONT_SIZE,
```
Pessimistress: - `fontSize` will be very confusing as we also have `getSize` and `sizeScale`.
- We could allow `sdf` be either a boolean or an object. The user can specify SDF generation options by supplying `sdf: {fontSize, buffer, ...}`.

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer-fragment.glsl.js
```diff
@@ -0,0 +1,62 @@
...
+
+export default `\
+#define SHADER_NAME multi-icon-layer-fragment-shader
+
+precision highp float;
+
+uniform float opacity;
+uniform sampler2D iconsTexture;
+uniform float buffer;
+uniform float sdf;
```
Pessimistress: You can just use `bool` as a uniform

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer-fragment.glsl.js
```diff
@@ -0,0 +1,62 @@
...
+  
+  float alpha = texColor.a;
+  // if enable sdf (signed distance fields)
+  if (sdf > 0.5) {
+    float distance = texture2D(iconsTexture, vTextureCoords).a;
+    alpha = smoothstep(buffer - vGamma, buffer + vGamma, distance);
+  }
+
+  // Take the global opacity and the alpha from vColor into account for the alpha component
+  float a = alpha * opacity * vColor.a;
```
Pessimistress: Do the `opacity * instanceColor.a` multiplication in the vertex shader.

modules/layers/src/text-layer/text-layer.js
```diff
@@ -39,11 +39,21 @@ const DEFAULT_COLOR = [0, 0, 0, 255];
 
 const MISSING_CHAR_WIDTH = 32;
 
+// used in TinySDF
+const SDFPreset = {
```
Pessimistress: `SDF_PRESET` for constant

modules/layers/src/text-layer/text-layer.js
```diff
@@ -74,9 +86,20 @@ export default class TextLayer extends CompositeLayer {
...
-  updateFontAtlas(fontFamily, characterSet) {
+  updateFontAtlas() {
+    const startTime = Date.now();
+
     const {gl} = this.context;
-    const {scale, mapping, texture} = makeFontAtlas(gl, {fontFamily, characterSet});
+    const {sdf, fontFamily, characterSet} = this.props;
+
+    const {scale, mapping, texture} = makeFontAtlas(gl, {
+      sdf: sdf ? SDFPreset : null,
```
Pessimistress: - User settings should be merged with the preset
- Override `fontFamily` with the value from props

Issue #2556, opened 2019-01-08
ScreenGridLayer: picking under WebGL1 not working (Safari)


1chandu:
#### Actual Result
Picking not working
#### Expected Result
Picking should work
#### Reproduce Steps
Run ScreenGridLayer in Safari
#### To Do List 
  - [x] Add label and assign to milestone
  - [x] Coding
  - [ ] Test

tgorkin:
Released a fix for this in deck.gl patch release 6.3.4.

PR #2557, opened 2019-01-08
ScreenGridLayer: fix picking under WebGL1

1chandu:
For #2556 

#### Background
When running on Safari, aggregation happens on CPU and results are in JS Arrays, for picking, check JS Array's first before reading from `Buffer` objects. (For WebGL2, data is read from Buffer objects).

#### Change List
- ScreenGridLayer: fix picking under WebGL1

### DIFF
--- a/modules/layers/src/screen-grid-layer/screen-grid-layer.js
+++ b/modules/layers/src/screen-grid-layer/screen-grid-layer.js
@@ -187,15 +187,13 @@ export default class ScreenGridLayer extends Layer {
   getPickingInfo({info, mode}) {
     const {index} = info;
     if (index >= 0) {
-      let {aggregationResults} = this.state;
-      if (!aggregationResults) {
-        aggregationResults = {
-          aggregationData: this.state.aggregationBuffer.getData(),
-          maxData: this.state.maxBuffer.getData()
-        };
-        // Cache aggregationResults to avoid multiple buffer reads.
-        this.setState({aggregationResults});
-      }
+      const {aggregationResults} = this.state;
+
+      // Cache aggregationResults to avoid multiple buffer reads.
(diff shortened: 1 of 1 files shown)


modules/layers/src/screen-grid-layer/screen-grid-layer.js
```diff
@@ -187,15 +187,13 @@ export default class ScreenGridLayer extends Layer {
...
-          maxData: this.state.maxBuffer.getData()
-        };
-        // Cache aggregationResults to avoid multiple buffer reads.
-        this.setState({aggregationResults});
-      }
+      const {aggregationResults} = this.state;
+
+      // Cache aggregationResults to avoid multiple buffer reads.
+      aggregationResults.aggregationData =
+        aggregationResults.aggregationData || this.state.aggregationBuffer.getData();
```
Pessimistress: Shouldn't the array vs. buffer difference be handled inside `GPUAggregator`? Do all users of the class need to be aware of this caveat?

1chandu: Good point.
Couple of things going on here:

1. GPUAggregator always returns `Buffer` objects, whether it is WebGL1 or WebGL2, but if the user wants to read the results into CPU memory, it needs to know whether it is under WebGL1 or WebGL2 because, it under WebGL1, buffer.getData() is not supported, and also under WebGL1, data is aggregated by CPU and results are available in JS Arrays. So here I am checking if the results already contains JS Arrays (WebGL1) , if not I am reading it from `Buffer` object (WebGL2).
2. Once aggregation is completed, and if the application is using WebGL2, we only need to read data from buffer once and cache it (invalidated when data is re-aggregated).

I will address this in a separate PR (#2561).

Issue #2541, opened 2019-01-03
Calculating mean in ScreenGridLayer


mkrkelj1:
Are we able to calculate mean/min/max w/ the public API via passing in props or some other method? This features appears to have been added and tested: [Issue: 2291](https://github.com/uber/deck.gl/issues/2291) , but I am having trouble locating documentation.

tgorkin:
It looks like the GPUAggregator class has been improved in v6.3 to support mean/min/max, but the system is still experimental and the new functionality has not yet been exposed all the way up to the ScreenGridLayer API.  Hopefully we'll get to those changes soon.  If you wanted to start tinkering with the new functionality directly at the GPUAggregator level, or even helping to flush out the implementation at the ScreenGridLayer level, there are some docs here: https://github.com/uber/deck.gl/blob/master/modules/core/src/experimental/utils/gpu-grid-aggregation/gpu-grid-aggregator.md.

1chandu:
Thanks @tgorkin 

Yes, we will be exposing these with Grid and ScreenGrid aggregation layers soon, but happy to look into any PRs as well.

1chandu:
This will be supported as **Experimental** feature in 7.0, #2560

tgorkin:
Fixed in master and release-6.4 branch.  Will be published to npm with 6.4 release.

PR #2560, opened 2019-01-08
ScreenGridLayer: add support Min/Max/Mean aggregation

1chandu:
For #2541 

#### Background
- Add a new experimental prop `aggregation`, to use Sum, Mean, Min or Max as aggregation operation. Default behavior is un-changed, we use `Sum` as aggregation operation.
- When performing `Min`, cell's that do not have any data points will have max float value, so update shaders to discard all cells, that do not have any data points.

#### Change List
- ScreenGridLayer: add support Min/Max/Mean aggregation


1chandu:
> As discussed, please keep the prop experimental for now (underscore and mark in doc), so that we have the liberty to change it after API audit.

Yep, already marked as `Experimental` in the doc.

PR #2586, opened 2019-01-17
Add deprecated flag to prop types

Pessimistress:
For https://github.com/uber/deck.gl/issues/2584

#### Change List

- Support `getValue` callback in prop type definition
- Fix `LineLayer` constructor support for multiple arguments
- Tests


Pessimistress:
> props should be independent

This was added so solve a very specific use case (backward-compatibility by support a deprecated prop), but I can see it being abused.

Alternatively, I could make it work this way:

```js
MyLayer.defaultProps = {
  ...
  getStrokeWidth: {type: 'accessor', value: 1},
  strokeWidth: {type: 'number', deprecated: true, newPropName: 'getStrokeWidth'}
}
```

PR #2587, opened 2019-01-18
Pass the missing startZoomPosition into Orthographic Controller interactiveState

javidhsueh:
#### Background
The startZoomPosition should be passed down to the interactiveState in Orthographic Controller.

#### Change List
-


Pessimistress:
Align with the `mapController` options? http://deck.gl/#/documentation/deckgl-api-reference/controllers/controller

Pessimistress:
In fact, shouldn't the base controller options work already?

javidhsueh:
@Pessimistress You're right. I didn't know the API of the controller. Revert most changes but pass down the missing interactive state.

Issue #2584, opened 2019-01-17
All core layer constructors should support multiple arguments


Pessimistress:
#### Description

Base layer constructor supports any number of arguments per [documentation](http://deck.gl/#/documentation/deckgl-api-reference/layers/layer?section=constructor).
Some core layers override the constructor and only consider the first argument, for example [LineLayer](https://github.com/uber/deck.gl/blob/master/modules/layers/src/line-layer/line-layer.js#L41).
All core layer constructors should support multiple arguments.

#### TODO

- [x] Add labels and milestone
- [x] Coding
- [x] Test

PR #2590, opened 2019-01-18
Fix HexagonLayer constructors

Pessimistress:
For #2584 

#### Change List
- Use the prop type system for validation and fallback


modules/layers/src/hexagon-cell-layer/hexagon-cell-layer.js
```diff
@@ -30,8 +30,8 @@ const DEFAULT_COLOR = [255, 0, 255, 255];
 
 const defaultProps = {
   hexagonVertices: null,
-  radius: null,
-  angle: null,
+  radius: {type: 'number', min: 1, value: 1000},
```
heshan0131: radius can also be 0

PR #2594, opened 2019-01-19
Update import path for TileLayer in example

mccalluc:
#### Background

Copying and pasting the original did not work for me, but this does.

Also, the default viewport just gave me a blue square... Not sure if that's the right behavior? This looks better for me, but I'm not sure how that should be integrated into this document.
```
  <App viewport={new WebMercatorViewport({
    width:600, height:600, zoom: 10, longitude: -71.1, latitude:42.4
  })}/>
```
#### Change List
- `@deck.gl/experimental-layers/tile-layer/tile-layer` -> `@deck.gl/experimental-layers`
- `TileLayer` -> `{TileLayer}`

### DIFF
--- a/modules/experimental-layers/src/tile-layer/tile-layer.md
+++ b/modules/experimental-layers/src/tile-layer/tile-layer.md
@@ -4,7 +4,7 @@ This TileLayer takes in a function `getTileData` that fetches tiles, and renders
 
 ```js
 import DeckGL from 'deck.gl';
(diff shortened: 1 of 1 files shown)


Pessimistress:
> the default viewport just gave me a blue square

Can you elaborate?

We have an test app for this feature here: https://github.com/uber/deck.gl/tree/master/test/apps/mapbox-tile

You can run it with `yarn start-local`

Pessimistress:
Merging this now, for any remaining problems you may open an issue.

PR #2599, opened 2019-01-20
Document the sublayers of all composite layers

Pessimistress:
RFC: https://github.com/uber/deck.gl/blob/master/dev-docs/RFCs/vNext/composite-layer-customization-rfc.md

#### Change List
- Document the sublayer types and ids of all composite layers
- Some sublayers are renamed in GeoJsonLayer and TextLayer for clarity


docs/api-reference/composite-layer.md
```diff
@@ -113,9 +113,16 @@ The default implementation returns `pickParams.info` without any change.
...
+Parameters:
+
+* `subLayerProps` (Object)
+  + `id` (String) - the sublayer id. The parent layer `id` will be prepended to the sublayer id.
+  + `updateTriggers` - the sublayer update triggers. Will be merged with the parent layer update triggers.
+
+Returns:
+
+An object with the properties of the sublayer built using the passed props and
 key base `Layer` props that `CompositeLayer`s typically forward to their sublayers.
```
xintongxia: what is `key` base `Layer` props?

docs/api-reference/composite-layer.md
```diff
@@ -113,9 +113,16 @@ The default implementation returns `pickParams.info` without any change.
 
 ##### `getSubLayerProps`
 
-Returns an object with the properties of the sublayer built using the passed props and
+Parameters:
+
+* `subLayerProps` (Object)
```
xintongxia: why only pass these two props? how about the other props of a certain `subLayer`?

PR #2600, opened 2019-01-20
Update composite layer customization RFC

Pessimistress:


dev-docs/RFCs/vNext/composite-layer-customization-rfc.md
```diff
@@ -29,9 +29,9 @@ There are two types of naming possible
 The `forwardBaseLayerProps()` method could check for `transferAllProps`
 
 
-### Proposal: CompositeLayers take new a `subLayerProps` prop
+### Proposal: CompositeLayers take a new `subLayerProps` prop
 
-The sublayerProps would be a recursive object. A special prop `layer` can be used to override layer type.
+The `subLayerProps` prop would be a recursive object. The keys are sublayer ids and the values are object of props to override. A special prop `type` can be used to override layer type.
```
Pessimistress: We use `type` to specify class constructors in both the `controller` API and the mapbox custom layers integration.

dev-docs/RFCs/vNext/composite-layer-customization-rfc.md
```diff
@@ -40,62 +40,61 @@ const customizedGeoJsonLayer = new GeoJsonLayer({
...
-  	...,
-  	subLayerProps: {
-  	  PointLayer: ...
-  	}
-  })
-```
 
+#### Open Questions
 
+**Should keys be layer class name or sublayer id?**
```
xintongxia: In the first proposal, it also mentioned  `There are two types of naming possible: ...`. Is it talking about the same issue as here?

https://github.com/uber/deck.gl/blob/4284dfa4c3c8b79b6f82f2a58a922b0de0eb8cda/dev-docs/RFCs/vNext/composite-layer-customization-rfc.md#proposal-compositelayers-document-which-sub-layers-they-render

dev-docs/RFCs/vNext/composite-layer-customization-rfc.md
xintongxia: Maybe describe what values does `type` could be?

dev-docs/RFCs/vNext/composite-layer-customization-rfc.md
```diff
@@ -40,62 +40,61 @@ const customizedGeoJsonLayer = new GeoJsonLayer({
...
+
+Composite layers may render multiple sublayers of the same type. For example, the GeoJsonLayer renders two `PathLayer` instances, one for polygon outlines and one for line strings. Differentiating them with unique ids would give the user finer grain control.
+
+**How do user props merge with props created by the parent layer?**
+
+The composite layer themselves should decide which props are "overridable" by the user. This can be done via supplying multiple arguments to the sublayer constructor:
```
xintongxia: Is there a way for the user to know which `props` are "overridable" and which are not?

Pessimistress: These will need to be documented on a case-by-case basis

dev-docs/RFCs/vNext/composite-layer-customization-rfc.md
Pessimistress: Yes

dev-docs/RFCs/vNext/composite-layer-customization-rfc.md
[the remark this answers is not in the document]
Pessimistress: Correct. The sublayer ids are added to the documentation of all composite layers: https://github.com/uber/deck.gl/pull/2599

dev-docs/RFCs/vNext/composite-layer-customization-rfc.md
```diff
@@ -40,62 +40,61 @@ const customizedGeoJsonLayer = new GeoJsonLayer({
...
+    {
+      // "overridable" props
+      stroked: this.props.stroked,
+      filled: this.props.filled
+    },
+    // where subLayerProps are merged in
+    this.getSubLayerProps({id: 'child-layer'}),
+    {
+      // "non-overridable" props
+      data
```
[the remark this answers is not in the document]
Pessimistress: The only "non-overridable" props are the ones that are generated internally to the parent layer (e.g. the aggregated data in a HexagonLayer) and the accessors that unwrap that specific format. Overriding these props will very likely break the layer. Of course, arguably, it'll be the user's fault if they do...

See how this affect individual layers in https://github.com/uber/deck.gl/pull/2600

dev-docs/RFCs/vNext/composite-layer-customization-rfc.md
Pessimistress: Updated the section to clarify: 1) default behavior (external always overrides internal) 2) generic recommendation (all props are overridable) 3) mechanism to block override and when to use it

Issue #2597, opened 2019-01-19
Most gallery examples do not work with Firefox 64.0


mccalluc:
#### Description
Of the 9 examples on https://deck.gl/showcases/gallery/, only ScreenGrid and Scatterplot work on Firefox.

#### Repro Steps
Visit https://deck.gl/showcases/gallery/boiler-plate (for example): Nothing is rendered

#### Environment (please complete the following information):
 - Framework Version: 6.0.0 on your site, but I've downloaded a couple, upped the version (`https://unpkg.com/deck.gl@6.3.4/deckgl.min.js`), and the behavior does not change.
 - Browser Version: Firefox 64.0
 - OS: MacOS 10.11.6

#### Logs
```
Error: WebGL warning: texImage2D: Alpha-premult and y-flip are deprecated for non-DOM-Element uploads.
texture.js:70:12
Error in parsing value for ‘cursor’.  Declaration dropped. data:%3Chtml%3E%0A%20%20%3Chead%3E%0A%20%20%20%20%3Ctitle%3EGet%20Started%20with%20deck.gl%3C/title%
Error in parsing value for ‘cursor’.  Declaration dropped. data:%3Chtml%3E%0A%20%20%3Chead%3E%0A%20%20%20%20%3Ctitle%3EGet%20Started%20with%20deck.gl%3C/title%
Unknown descriptor ‘"normal"’ in @font-face rule.  Skipped to next declaration. refresh.css:1:42967
Unknown descriptor ‘"normal"’ in @font-face rule.  Skipped to next declaration. refresh.css:1:86438
Unknown descriptor ‘"normal"’ in @font-face rule.  Skipped to next declaration. refresh.css:1:129781
Unknown descriptor ‘"normal"’ in @font-face rule.  Skipped to next declaration. refresh.css:1:172668
Unknown descriptor ‘"normal"’ in @font-face rule.  Skipped to next declaration. refresh.css:1:214746
Stylesheets without CSSOM changes reparsed to check for errors. Refresh the page to also see errors from stylesheets changed from CSSOM and from style attributes.
Error in parsing value for ‘cursor’.  Declaration dropped.
data:%3Chtml%3E%0A%20%20%3Chead%3E%0A%20%20%20%20%3Ctitle%3EGet%20Started%20with%20deck.gl%3C/title%
```
(It doesn't show up in the copy, but the last `Error in parsing value` is being hit in an infinite loop, but only the first time after browser start.)

[...]

PR #2608, opened 2019-01-23
Fix gallery in Firefox

Pessimistress:
For #2597 

#### Change List
- Generate dynamic URL for iframe src
- Fix boiler plate example size
- Fix geojson example lighting

### DIFF
--- a/examples/gallery/build-tools/page.html
+++ b/examples/gallery/build-tools/page.html
@@ -45,11 +45,16 @@
         <a href="./">↩ Back to Gallery</a>
       </div>
       <h2>Demo</h2>
-      <iframe width="100%" height="450" frameborder="0" src="{{iframeSrc}}" ></iframe>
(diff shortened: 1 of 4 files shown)


examples/gallery/build-tools/page.html
```diff
@@ -45,11 +45,16 @@ <h1>{{title}}</h1>
...
-      <iframe width="100%" height="450" frameborder="0" src="{{iframeSrc}}" ></iframe>
+      <iframe width="100%" height="450" frameborder="0" ></iframe>
       <h2>Source</h2>
       <div id="source">
         <code class="prettyprint">{{content}}</code>
       </div>
     </div>
+    {{{iframeSrc}}}
+    <script type="text/javascript">
+      var blob = new Blob([src], {type: 'text/html'})
```
xintongxia: semicolon  `var blob = new Blob([src], {type: 'text/html'});`

PR #2609, opened 2019-01-23
Increase buffer when generating fontAtlas for TextLayer

xintongxia:
For #1550 

#### Change List
- Increase `buffer` when generating fontAtlas for TextLayer
- Bug fix for layout characters in each row

Before:
screen shot 2019-01-23 at 10 16 41 am

After:
screen shot 2019-01-23 at 10 14 57 am


Pessimistress:
Since this is an edge case, I would make this a user-configurable option instead of increasing the buffer for everyone. The resulting font atlas could potentially be very large if using CJK fonts.

Pessimistress:
Closing for https://github.com/uber/deck.gl/pull/2628

PR #2627, opened 2019-01-25
add text layer api change RFC

xintongxia:
For #1550 

#### Change List
- add text layer api change RFC

### DIFF
--- a/dev-docs/RFCs/vNext/text-layer-font-setting-api.md
+++ b/dev-docs/RFCs/vNext/text-layer-font-setting-api.md
@@ -1,0 +1,70 @@
+# RFC: Text Layer API Change
+
+* **Authors**: Xintong Xia 
+* **Date**: December 2018
+* **Status**: Draft 
+
+## Motivation 
+
+When `TextLayer` generates shared `fontAtlas` for required `characterSet`, it applied default settings, like `fontSize`, 
+`buffer`(Whitespace surround each character), etc.. However one setting can not be suitable for all the different fonts. 
+For example, [Cinzel](https://fonts.google.com/specimen/Cinzel) has a special shape of `Q` character, which requires
+a big buffer [See detail](https://github.com/uber/deck.gl/pull/2609). Big buffer around each character can be very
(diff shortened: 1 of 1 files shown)


dev-docs/RFCs/vNext/text-layer-font-setting-api.md
```diff
@@ -0,0 +1,63 @@
...
+const textLayer = new TextLayer({
+  ...,
+  characterSet: 'abcdefg',
+  fontFamily: 'Monaco, monospace',
+
+  fontSettings: {
+    // shared options between non-sdf and sdf
+    // this fontSize if only applied for generating fontAtlas
+    // it does not impact the size of the text labels 
+    fontSize: 64,
```
Pessimistress: Need a better name to avoid confusion with the actual display size (i.e. `getSize` and `sizeScale` props)

dev-docs/RFCs/vNext/text-layer-font-setting-api.md
```diff
@@ -0,0 +1,63 @@
...
+    // Whitespace buffer around each side of the character
+    buffer: 2,
+    
+    // `sdf` only options
+    // https://github.com/mapbox/tiny-sdf
+    sdf: true,
+    // if `sdf` is false, the following parameters are not appliable
+    radius: 3,
+    cutoff: 0.25,
+    fontWeight: 'normal'
```
Pessimistress: I would move this option to a top-level prop as it is a fairly common config that most user would want to have access to.

More generally, we need to define what this settings object is. I see two aspects:
- These are parameters used **only** in font atlas generation, that an **advanced** user can fine-tune for better perf/appearance. For most people, they should not have to understand any of its concepts.
- Like accessors, this prop may change shallowly if it is defined inline. Arguably, there is no need to regenerate the font atlas if nothing else has changed. It is reasonable to assume that these are not things that users want to constantly adjust at runtime. Following this line of thinking, maybe `sdf` should be top-level too.

Pessimistress:
Once implemented, we will need good documentation of each font setting that explain how changing its value affect appearance/performance and when someone might want to use them.

PR #2629, opened 2019-01-25
Remove prefixes for grab and grabbing cursor values.

### LINKED ISSUE #2618: Loop of CSS Warnings about `cursor` in Firefox
#### Description
Firefox Devtools console is spammed with `Error in parsing value for ‘cursor’. Declaration dropped.` messages.

#### Repro Steps
1. Visit https://deck.gl/showcases/gallery/boiler-plate on Firefox
1. Open the console, and enable the "CSS" filter if it wasn't
1. Close devtools
1. Reopen the console

`Error in parsing value for ‘cursor’. Declaration dropped.` is displayed, and the repeat counter keeps updating.

 - Framework Version: unknown
 - Browser Version: Firefox 66
 - OS: OSX

---

In the page code, I'm seeing:
```js
this.canvas && (this.canvas.style.cursor = this.props.getCursor(this.interactiveState))
```

A quick evaluation in the console indicates that `this.props.getCursor(this.interactiveState)` returns `"-webkit-grab"`, which is not supported by Firefox, hence the warning message.
Then my guess is that this code is executed in a requestAnimationFrame callback or another loop, triggering the warning message on each iteration.

A quick test indicates that the following: `data:text/html,<meta charset=utf8><h1 style="cursor:grab">grab</h1><h1 style="cursor:grabbing">grabbing</h1>` does show grab/grabbing cursors in Chrome, so maybe we don't need the `PREFIX` part in https://github.com/uber/deck.gl/blob/049b6836f394507988853713f02ed21327476753/modules/core/src/lib/deck.js#L43-L44 ?

I'd be happy to create a PR if you can confirm that I can get rid of them (I only tested Chrome Canary, and I don't know what browser version you want to support).

nchevobbe:
Those prefixes are not needed in the latest version of Chrome
and Safari, so we can safely remove them.

Fixes #2618.


nchevobbe:
Everything looks cool. Could someone review these changes?

modules/core/src/lib/deck.js
```diff
@@ -31,17 +31,14 @@ import {Stats} from 'probe.gl';
 import {EventManager} from 'mjolnir.js';
 
 import assert from '../utils/assert';
-import VENDOR_PREFIX from '../utils/css-vendor-prefix';
```
Pessimistress: I believe this file can be safely removed.

modules/react/src/utils/cursors.js
```diff
@@ -18,15 +18,8 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-// import browser from 'bowser';
-
-const PREFIX = '-webkit-';
```
Pessimistress: I don't see this file being imported anywhere. Just remove?

nchevobbe:
Thanks for the reviews. I deleted all unused file, inlined `grab` and `grabbing` in the only place it was used, and removed the `CURSOR` object which has no use (`POINTER` was not used anywhere).

PR #2632, opened 2019-01-25
fix projection for point lights

jianhuang01:
For #2096 

#### Change List
- Provide the right coordinate system for point lights projection
- Improve comment for projectPosition function


modules/core/src/lighting/point-light.js
```diff
@@ -1,11 +1,7 @@
 import {PointLight as BasePointLight} from 'luma.gl';
-import {COORDINATE_SYSTEM} from '../lib';
 
 export default class PointLight extends BasePointLight {
```
Pessimistress: Is there other features planned for this class? Can just re-export luma's PointLight class?

Issue #2553, opened 2019-01-08
TextLayer's auto highlighting is per-character instead of per-object


Pessimistress:
#### Actual Result
On hover over, TextLayer auto-highlights individual characters.

#### Expected Result
TextLayer's auto-highlight should be applied per-object.

#### Reproduce Steps
Turn on `autoHighlight` of any TextLayer

#### To Do List 
  - [x] Add label and assign to milestone
  - [x] Coding
  - [x] Test

PR #2633, opened 2019-01-26
 fix text-layer per object highlighting

xintongxia:
For #2553

#### Change List
- assign the same pickingColorIndex to all the characters in one text label


modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer.js
```diff
@@ -105,6 +108,20 @@ export default class MultiIconLayer extends IconLayer {
...
     }
   }
+
+  calculateInstancePickingColors(attribute) {
+    const {data, getPickingIndex} = this.props;
+    const {value} = attribute;
+    let i = 0;
+    for (const point of data) {
+      const index = getPickingIndex(point);
+      const pickingColor = this.encodePickingColor(index);
```
Pessimistress: This will break the current picking logic. You need to override the default `getPickingInfo` as well.

modules/layers/src/text-layer/text-layer.js
```diff
@@ -119,17 +119,21 @@ export default class TextLayer extends CompositeLayer {
   }
 
   getPickingInfo({info}) {
+    // because `TextLayer` assign the same pickingInfoIndex for one text label,
+    // here info.index refers the index of text label in props.data
     return Object.assign(info, {
       // override object with original data
-      object: info.object && info.object.object
+      object: info.index > -1 && this.props.data[info.index]
```
Pessimistress: This syntax is fine when used with an `if` statement. However, the pick info object is a user-facing API. Doing a shorthand will set `object` to `false` when nothing is picked.

```js
info.index >= 0 ? this.props.data[info.index] : null
```

Issue #2623, opened 2019-01-24
v6.4 release tracker


Pessimistress:
- [x] Cut release branch
- [x] Update documentation links
- [x] What's New / Upgrade Guide
- Website testing:

  - [ ] Windows/IE11
         - No website or documentation pages load in IE11.  Same issue occurs on currently live version of website.  Console errors: `SCRIPT438: Object doesn't support property or method 'assign'`
  - [x] Windows/Edge
        - http://pessimistress.github.io/deck.gl/#/documentation/deckgl-api-reference/layers/geojson-layer
`Cannot find linked doc:  docs/layers/solid-polygon-layer.md`
        - Occasional spam of WebGL errors when clicking around between examples.  Occurs on prod website as well. 
`100 WEBGL11163: drawArrays: An enabled vertex array has no buffer.  bundle.js (35,649859)`
        - New deprecation warnings in Arc and Arc Brushing examples:
```
  deck: `ScatterplotBrushingLayer: getColor` is deprecated and will be removed in a later version. Use `getFillColor/getLineColor` instead
bundle.js (46,91854)
 deck: `ScatterplotBrushingLayer: strokeWidth` is deprecated and will be removed in a later version. Use `getLineWidth` instead
bundle.js (46,91854)
 deck: `ScatterplotBrushingLayer: outline` is deprecated and will be removed in a later version. Use `stroked` instead
bundle.js (46,91854)
 deck: `ArcBrushingLayer: strokeWidth` is deprecated and will be removed in a later version. Use `getStrokeWidth` instead
 deck: `ArcLayer: `strokeWidth`` is deprecated and will be removed in a later version. Use ``getStrokeWidth`` instead
```
  - [x] Chrome
        - Documentation/"JSON API reference"/JSONDeck page is not found
        - Lots of deprecated API warnings
        - Lots of "Too many active WebGL contexts" and "Invalid operation" warnings while visiting core layer examples, and occasionally the base map can't show up
        - "Cannot find linked doc:  docs/layers/solid-polygon-layer.md" while visiting Documentation/"Deck.gl API reference"/layers/ PolygonLayer
  - [x] Firefox
		- http://pessimistress.github.io/deck.gl/#/documentation/deckgl-api-reference/layers/scatterplot-layer: the outline is messed up when zoom in/out 
		- http://pessimistress.github.io/deck.gl/#/documentation/json-api-reference-experimental/json-deck-experimental: link not found
  - [x] Safari
[...]

PR #2636, opened 2019-01-29
Website issue fixes

Pessimistress:
For #2623 

#### Change List

- Fix target rings in brushing example
- Add polyfill to website for IE

### DIFF
--- a/website/src/main.js
+++ b/website/src/main.js
@@ -1,4 +1,6 @@
 /* global document */
+import '@babel/polyfill';
+
 import React from 'react';
 import ReactDOM from 'react-dom';
 import {Provider} from 'react-redux';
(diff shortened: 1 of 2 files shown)


website/src/main.js
```diff
@@ -1,4 +1,6 @@
 /* global document */
+import '@babel/polyfill';
```
tgorkin: Does this dependency need to be added to website/package.json?  Or does it work fine with it added at the top level ./package.json?

tgorkin: Also curious about the documentation saying it should be added to depenendencies instead of devDependencies: [https://babeljs.io/docs/en/babel-polyfill/](https://babeljs.io/docs/en/babel-polyfill/).  I guess if it worked in our testing..

Pessimistress: If this is a standalone application it should be added to dependencies. It doesn't matter in the website case since we only distribute the compiled bundle.

PR #2639, opened 2019-01-29
add font atlas manager

xintongxia:
#### Background
Current `TextLayer` is not efficient in generating shared `fontAtlas` and the corresponding mapping.

1. Recalculate mapping, regenerate `fontAtlas` image data and create a new texture when `characterSet` changed.

It is not necessary if `characterSet` only partially update. So in this PR, it will only calculate mapping of the new chars.

2. If there are more than one `TextLayer`, which shared the same `characterSet`, it is not necessary to generate a mapping, and a `fontAtlas` for each layer. In this PR, it allows different text layers share the same `fontAtlas` and mapping if settings are the same. 

#### Change List
- refactor `font-atlas`, move to `font-atlas-manager`
- cache generated `fontAtlas` canvas
- update cached `fontAtlas` with new characters when `characterSet` changed

Problems:
In this PR, when update cached `fontAtlas` with new characters, it needs first redraw the canvas with the old canvas, and then append the new characters. This needs to be improved.


modules/layers/src/text-layer/font-atlas-manager.js
```diff
@@ -0,0 +1,263 @@
...
+        this[prop] = props[prop];
+      }
+    });
+
+    // update cache key
+    const oldKey = this._key;
+    this._key = this._getKey();
+
+    const charSet = getNewChars(this._key, this.characterSet);
+    const cachedFontAtlas = cache.get(this._key);
```
Pessimistress: Textures cannot be reused cross GL context. Need to use `gl` as a key as well.

modules/layers/src/text-layer/util.js
```diff
@@ -0,0 +1,46 @@
+// TODO merge with icon-layer/icon-manager
```
Pessimistress: Nit: font-atlas-utils.js

modules/layers/src/text-layer/cache.js
```diff
@@ -0,0 +1,50 @@
...
+// update order, first is oldest, last is newest
+let _order = [];
+
+/**
+ * Cache class with limit
+ *
+ * Update timestamp for each get/set operation
+ * Delete oldest when reach given limit
+ */
+export default class Cache {
```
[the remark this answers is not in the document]
xintongxia: Yes. you are right. It shouldn't be global.

At beginning, these code stayed in `font-atlas-maanger`, after I extracted them to a separate class file. I didn't think through.

modules/layers/src/text-layer/font-atlas-manager.js
```diff
@@ -0,0 +1,263 @@
...
+  'cutoff',
+  'radius'
+];
+
+/**
+ * get all the chars not in cache
+ * @param key cache key
+ * @param characterSet (Array|Set)
+ * @returns {Array} chars not in cache
+ */
```
xintongxia: new function

modules/layers/src/text-layer/util.js
```diff
@@ -0,0 +1,46 @@
...
+  getFontWidth,
+  fontHeight,
+  buffer,
+  maxCanvasWidth,
+  mapping = {},
+  xOffset = 0,
+  yOffset = 0
+}) {
+  let row = 0;
+  let x = xOffset;
```
xintongxia:
```
 let x = 0;
```

modules/layers/src/text-layer/font-atlas-utils.js
```diff
@@ -0,0 +1,66 @@
...
+  fontHeight,
+  buffer,
+  maxCanvasWidth,
+  mapping = {},
+  xOffset = 0,
+  yOffset = 0
+}) {
+  let row = 0;
+  // continue from x position of last character in the old mapping
+  let x = xOffset;
```
xintongxia: This `buildMapping` changed to allow appending new characters to existing mapping

Old code
```
  const mapping = {};
  let row = 0;
  let x = 0;
```

modules/layers/src/text-layer/font-atlas-utils.js
```diff
@@ -0,0 +1,66 @@
...
+  Array.from(characterSet).forEach((char, i) => {
+    if (!mapping[char]) {
```
xintongxia: Add checking if `char` already exists in the mapping.

modules/layers/src/text-layer/font-atlas-utils.js
```diff
@@ -0,0 +1,66 @@
...
+  });
+
+  const rowHeight = fontHeight + buffer * 2;
+
+  return {
+    mapping,
+    xOffset: x,
+    yOffset: yOffset + row * rowHeight,
+    canvasHeight: nextPowOfTwo(yOffset + (row + 1) * rowHeight)
+  };
```
xintongxia: return object add 2 more fields:

```
xOffset: x position (right) of last char
yOffset: y position (top) last char
``

File: modules/layers/src/text-layer/font-atlas-manager.js [hunk 9]
```diff
@@ -0,0 +1,264 @@
...
+      ctx.putImageData(imageData, 0, 0);
+    }
+    setTextStyle(ctx, fontFamily, fontSize, fontWeight);
+
+    // 3. layout characters
+    if (sdf) {
+      const tinySDF = new TinySDF(fontSize, buffer, radius, cutoff, fontFamily, fontWeight);
+      // used to store distance values from tinySDF
+      // tinySDF.size equals `fontSize + buffer * 2`
+      const imageData = ctx.getImageData(0, 0, tinySDF.size, tinySDF.size);
```
xintongxia: #235: try to reuse the canvas image data

old code
```
    const imageData = ctx.createImageData(tinySDF.size, tinySDF.size);
```

modules/layers/src/text-layer/font-atlas-manager.js
```diff
@@ -0,0 +1,264 @@
...
+
+    return {
+      xOffset,
+      yOffset,
+      mapping,
+      data: canvas,
+      width: canvas.width,
+      height: canvas.height
+    };
+  }
```
xintongxia: return object structure changed.

old code 

```
  return {
    scale: HEIGHT_SCALE,
    mapping,
    texture: new Texture2D(gl, {
      pixels: canvas,
      // padding is added only between the characters but not for borders
      // enforce CLAMP_TO_EDGE to avoid any artifacts.
      parameters: {
        [GL_TEXTURE_WRAP_S]: GL_CLAMP_TO_EDGE,
        [GL_TEXTURE_WRAP_T]: GL_CLAMP_TO_EDGE
      }
    })
  };
```

modules/layers/src/text-layer/font-atlas-manager.js
```diff
@@ -0,0 +1,264 @@
...
+
+    // update fontAtlas with new settings
+    const fontAtlas = this._generateFontAtlas(this._key, charSet, cachedFontAtlas);
+    this._updateTexture(fontAtlas);
+
+    // update cache
+    cache.set(this._key, fontAtlas);
+  }
+
+  _updateTexture({data: canvas, width, height}) {
```
xintongxia: new func

modules/layers/src/text-layer/font-atlas-manager.js
```diff
@@ -0,0 +1,264 @@
...
+  get mapping() {
+    const data = cache.get(this._key);
+    return data && data.mapping;
+  }
+
+  get scale() {
+    return HEIGHT_SCALE;
+  }
+
+  setProps(props = {}) {
```
xintongxia: new func.

Also the main logic to decide if needs generating a new fontAtlas or if fontAtlas needs update

modules/layers/src/text-layer/lru-cache.js
```diff
@@ -0,0 +1,78 @@
...
+
+    this.clear();
+  }
+
+  clear() {
+    this._cache = {};
+
+    const head = this._getNode({key: `${this}._head`});
+    const tail = this._getNode({key: `${this}._tail`});
+    this._head = Object.assign(head, {next: tail});
```
Pessimistress: The pointers seem a little excessive. A simple array should do?

modules/layers/src/text-layer/font-atlas-manager.js
```diff
@@ -0,0 +1,264 @@
...
+  ctx.font = `${fontWeight} ${fontSize}px ${fontFamily}`;
+  ctx.fillStyle = '#000';
+  ctx.textBaseline = 'baseline';
+  ctx.textAlign = 'left';
+}
+
+export default class FontAtlasManager {
+  constructor(gl) {
+    this.gl = gl;
+    this.fontFamily = DEFAULT_FONT_FAMILY;
```
Pessimistress: Nit: use one object e.g. `this.options` to manage the font settings.

Issue #2665, opened 2019-02-10
Implement binary data RFC


Pessimistress:
[RFC](https://github.com/uber/deck.gl/blob/master/dev-docs/RFCs/v7.x/binary-data-rfc.md)

#### To Do List 
  - [ ] Add label and assign to milestone 
  - [ ] Coding
  - [ ] Doc update
  - [ ] What’s new update 
  - [ ] Test

PR #2666, opened 2019-02-11
Implement binary data RFC (PR 1/3)

Pessimistress:
For #2665 

#### Change List
- Add experimental core utility `forEach`
- Add layer method `iterateData`
- Support non-iterable data in ScatterplotLayer
- Support non-iterable data in Tesselator class
- Add tests


modules/core/src/utils/array-utils.js
```diff
@@ -18,6 +18,34 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
+export function forEach(data, visitor) {
```
jianhuang01: Add a comment to explain the forEach API?

Issue #2681, opened 2019-02-13
OrbitView: allow pan instead of rotate for the 'panmove' event


danmarshall:
#### Target Use case
I would like the ability to have OrbitView's `panmove` event to behave similarly to the MapView, where panning is the default action, and rotating is the alternate action (shift/ctrl/alt key). 

#### Proposed feature
In [controller.js](https://github.com/uber/deck.gl/blob/master/modules/core/src/controllers/controller.js#L57), add an additional flag for `invertPanRotate` just like `invertPan`. To achieve the target use case, one can subclass OrbitViewController and set these flags to manipulate panmove and panrotate directions.

#### To Do List 
  - [ ] Add label and assign to milestone 
  - [ ] Coding
  - [ ] Doc update
  - [ ] What’s new update 
  - [ ] Test

danmarshall:
It seems that this can already be accomplished by seeting `this.invertPan = true` and overriding `_onPanRotate`:
```
_onPanRotate(event) {
    if (!this.dragRotate) {
      return false;
    }
    return this._onPanRotateMap(event);
  }
```
So, closing this issue. Keeping the PR open so the flag is not used for `_onPanRotate`.

PR #2682, opened 2019-02-13
Don't use invertPan flag in _onPanRotate method

danmarshall:
For #2681

### DIFF
--- a/modules/core/src/controllers/controller.js
+++ b/modules/core/src/controllers/controller.js
@@ -55,6 +55,7 @@ export default class Controller {
     this.onViewStateChange = null;
     this.onStateChange = null;
     this.invertPan = false;
+    this.invertPanRotate = false;
 
     this.handleEvent = this.handleEvent.bind(this);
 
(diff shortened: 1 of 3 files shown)


Pessimistress:
Thank you for bringing up this issue. You are correct that we should not be overloading `invertPan` for multiple purposes.

Revisiting this code I think we should make the base class `_onPanRotate` method use only the `_onPanRotateStandard` behavior. Then, `MapController` should override `_onPanRotate` with the `_onPanRotateMap` behavior, since the logic created specifically to match mapbox-gl's interaction.

If you feel up to it you can update this PR, otherwise we're happy to open a new PR to unblock you.

danmarshall:
@Pessimistress ok I see. Let me give it a try, and I will update this PR 👍

danmarshall:
@Pessimistress thanks for the advice. The original issue is closed because the result is achievable with the override you suggested. This PR is now just for removing the overloaded dependency on the flag.

PR #2684, opened 2019-02-13
[For Discussion] Layer extension POC

Pessimistress:
RFC thread: https://github.com/uber/deck.gl/pull/2524

#### Change List
- Add `LayerExtension` class and `extendLayer` utility
- Update `test/apps/data-filter` to demonstrate implementing and using an extension

data-filter-2


test/apps/data-filter/src/app.js
```diff
@@ -2,11 +2,17 @@
...
 import {render} from 'react-dom';
-import DeckGL, {COORDINATE_SYSTEM} from 'deck.gl';
+import DeckGL, {ScatterplotLayer, COORDINATE_SYSTEM} from 'deck.gl';
 
-import ScatterplotLayer from './scatterplot-layer';
+import DataFilterExtension from './data-filter-extension';
 import POINTS from './data-sample';
 
+import {extendLayer} from '@deck.gl/core';
+const FilteredScatterplotLayer = extendLayer(
```
Pessimistress: @heshan0131 The intention is that we will have a collection of extensions (e.g. data-filter, brushing) that can be used like this with any core layer.

test/apps/data-filter/src/data-filter-extension.js
```diff
@@ -0,0 +1,119 @@
+import {LayerExtension} from '@deck.gl/core';
+
+const DATA_TYPE_FROM_SIZE = {
```
heshan0131: what's the reason of having 4 different data types here?

Pessimistress: This is so that the app can specify the number of filters to use. Say your earthquake data contains magnitude and time. You will create a data filter of size 2, which allows you to specify independent ranges for each value.

Pessimistress: We could potentially expand this to support larger sizes but 4 is probably good enough to cover most use cases?

modules/layers/src/scatterplot-layer/scatterplot-layer-vertex.glsl.js
```diff
@@ -47,6 +47,8 @@ varying vec2 unitPosition;
 varying float innerUnitRadius;
 
 void main(void) {
+  DECKGL_MAIN_START();
```
Pessimistress: @tsherif I find it pretty awkward to have to add this to every layer. Can I define module injection into standard locations (`vs:main-start` and such) without registering my own hook?

tsherif: I suppose we could make those strings options for the `shaderHook` argument to `setModuleInjection`. Would that be enough?

PR #2691, opened 2019-02-15
Rendering & Picking refactor clean up

jianhuang01:
For #2662 

#### Change List
- remove old files and integrate new passes
- clean up LayerManager
- fix tests


modules/core/src/index.js
```diff
@@ -38,6 +38,7 @@ export {default as LayerManager} from './lib/layer-manager';
 export {default as AttributeManager} from './lib/attribute-manager';
 export {default as Layer} from './lib/layer';
 export {default as CompositeLayer} from './lib/composite-layer';
+export {default as DeckRenderer} from './lib/deck-renderer';
```
Pessimistress: Let's not export this unless necessary.

modules/core/src/lib/deck-picker.js
```diff
@@ -38,39 +41,41 @@ export default class DeckPicker {
     };
   }
 
+  setProps(props) {
+    if ('useDevicePixels' in props) {
+      this.useDevicePixels = props.useDevicePixels;
```
Pessimistress: call `getPixelRatio` here and store the result instead.

modules/core/src/lib/deck-renderer.js
```diff
@@ -1,44 +1,58 @@
...
+    this.layerFilter = null;
+    this.drawPickingColors = false;
     this.drawLayersPass = new DrawLayersPass(gl);
+    this.pickLayersPass = new PickLayersPass(gl);
     this.renderCount = 0;
   }
 
+  setProps(props) {
+    if ('useDevicePixels' in props) {
+      this.useDevicePixels = props.useDevicePixels;
```
Pessimistress: Ditto

modules/core/src/lib/deck.js
```diff
@@ -216,11 +218,37 @@ export default class Deck {
...
+    }
+
     this.stats.timeEnd('deck.setProps');
   }
 
   // Public API
 
+  updateNeedsRedraw(props) {
+    if (this.props) {
+      let reason = this._needsRedraw;
```
Pessimistress: Use `DeckRenderer`'s needsRedraw flag instead?

modules/core/src/index.js
jianhuang01: we need it in test-utils/src/lifecycle-test.js.

PR #2697, opened 2019-02-18
Improve picking color generation

Pessimistress:
Calculate picking color for 1M instances, before: 45 iterations/s; after: 412 iterations/s

Since `layer.calculateInstancePickingColors` always produces the same sequence, we can save the largest generated array and copy its value into others.

We can potentially make all `instancePickingColors` attributes share the same WebGLBuffer, but that would introduce more complexity around:
- Retrieving the shared buffer by GL context
- Creating a temporary buffer when performing multi-depth picking

#### Change List
- Optimize `layer.calculateInstancePickingColors`
- Benchmark case

### DIFF
--- a/modules/core/src/lib/layer.js
+++ b/modules/core/src/lib/layer.js
@@ -406,19 +406,22 @@ export default class Layer extends Component {
     const {value, size} = attribute;
 
     if (value[0] === 1) {
-      // already populated
+      // This can happen when data has changed, but the attribute value typed array
+      // has sufficient size and does not need to be re-allocated.
+      // This attribute is already populated, we do not have to recalculate it
       return;
     }
 
     const cacheSize = pickingColorCache.length / size;
 
+    // Copy the last calculated picking color sequence into the attribute
     value.set(
       numInstances < cacheSize
         ? pickingColorCache.subarray(0, numInstances * size)
         : pickingColorCache
     );
 
-    // add 1 to index to seperate from no selection
+    // If the attribute is larger than the cache, populate the missing chunk
     for (let i = cacheSize; i < numInstances; i++) {
       const pickingColor = this.encodePickingColor(i);
       value[i * size + 0] = pickingColor[0];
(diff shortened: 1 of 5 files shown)


modules/core/src/lib/layer.js
```diff
@@ -402,13 +404,35 @@ export default class Layer extends Component {
...
+    for (let i = cacheSize; i < numInstances; i++) {
       const pickingColor = this.encodePickingColor(i);
       value[i * size + 0] = pickingColor[0];
       value[i * size + 1] = pickingColor[1];
       value[i * size + 2] = pickingColor[2];
     }
+
+    if (cacheSize < numInstances) {
+      // Save the largest picking color sequence
+      pickingColorCache = value;
```
tsherif: This seems like an error. Do you really want pickingColorCache to reference the array created internally in the attribute?

modules/core/src/lib/layer.js
```diff
@@ -268,7 +270,7 @@ export default class Layer extends Component {
   // Returns the picking color that doesn't match any subfeature
   // Use if some graphics do not belong to any pickable subfeature
   encodePickingColor(i) {
-    assert((((i + 1) >> 24) & 255) === 0, 'index out of picking color range');
+    assert(i < 16777215, 'index out of picking color range');
     return [(i + 1) & 255, ((i + 1) >> 8) & 255, (((i + 1) >> 8) >> 8) & 255];
```
tsherif: Not part of your updates, but since you're optimizing, you might get a bit of a win by pre-allocating the 3-element array, pulling out the increment, and collapsing the shifts. E.g. something like:
```
let idx = i + 1;
pickColor[0] = idx & 255;
pickColor[1] = (idx >> 8) & 255;
pickColor[2] = (idx >> 16) & 255;
return pickColor;
```

Pessimistress: Good point. Will do.

PR #2703, opened 2019-02-20
Use shader attributes for solid polygon layer

tsherif:
#### Background
Use `shaderAttributes` to replace SolidPolygonLayer attribute hijacking in `updateAttributes`. 

#### Change List
- Use new `shaderAttributes` instead of creating new attributes in `updateAttributes`. The `updateAttributes` override is is still necessary to switch to instance drawing for the sides.
- Update vertex shader module for SolidPolygonLayer to inject source code instead of branching. Was necessary to prevent the topModel from trying to access attributes it doesn't use.


tsherif:
@Pessimistress much cleaner! Thanks for the `#define` tip :)

modules/layers/src/solid-polygon-layer/solid-polygon-layer.js
```diff
@@ -228,20 +255,6 @@ export default class SolidPolygonLayer extends Layer {
           });
         }
       }
-      if (newAttributes.positions) {
```
Pessimistress: 🎉

modules/layers/src/solid-polygon-layer/solid-polygon-layer.js
```diff
@@ -87,9 +94,29 @@ export default class SolidPolygonLayer extends Layer {
         transition: ATTRIBUTE_TRANSITION,
         accessor: 'getPolygon',
         update: this.calculatePositions,
-        noAlloc
+        shaderAttributes: {
+          positions: {
+            size: 3
```
Pessimistress: Is the `size` override necessary?

tsherif: It's not, though having a 'positions' key is necessary. I figure this is a bit clearer than having an empty object?

PR #2709, opened 2019-02-21
Model attribute updaters

tsherif:
Possible solution to the overriding up `updateAttributes` in the SolidPolygonLayer. Allow layers to have a `modelAttributeUpdaters` prop: an object keyed by the models name in the `state`, value is a function that will be called in `Layer.updateAttributes` to do any custom attribute updating.

### DIFF
--- a/modules/core/src/lib/layer.js
+++ b/modules/core/src/lib/layer.js
@@ -390,10 +390,23 @@ export default class Layer extends Component {
       ignoreUnknownAttributes: true
     });
 
-    const model = this.getSingleModel();
+    let model;
+
+    model = this.getSingleModel();
     if (model) {
       const changedAttributes = attributeManager.getChangedAttributes({clearChangedFlags: true});
       model.setAttributes(changedAttributes);
+    }
+
+    if (props.modelAttributeUpdaters) {
+      const changedAttributes = attributeManager.getChangedAttributes({clearChangedFlags: true});
+      for (const modelName in props.modelAttributeUpdaters) {
+        model = this.state[modelName];
+        const updater = props.modelAttributeUpdaters[modelName];
+        if (model && updater) {
+          updater(this, model, changedAttributes);
+        }
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


tsherif:
Hmmm... not sure why those extra commits are in there... I merged master into this branch.

jianhuang01:
> Hmmm... not sure why those extra commits are in there... I merged master into this branch.

You need rebase master after pull the latest, because master branch history was modified yesterday.

tsherif:
@jianhuang01 Ah, right! Thanks for the tip. I'll fix that now.

modules/core/src/lib/layer.js
```diff
@@ -390,11 +390,24 @@ export default class Layer extends Component {
...
-    const model = this.getSingleModel();
+    let model;
+
+    model = this.getSingleModel();
     if (model) {
       const changedAttributes = attributeManager.getChangedAttributes({clearChangedFlags: true});
       model.setAttributes(changedAttributes);
     }
+
+    if (props.modelAttributeUpdaters) {
```
Pessimistress: This logic is internal to the layer and should not be part of the props system. Props reflect user-customizable layer settings. Do you consider adding an extra descriptor to each layer attribute?

tsherif: I don't think there's a way to do everything that needs to be done with extra attribute properties. There are 4 things that happened in the custom `updateAttributes`:
- Filter certain attributes to one model or the other
- Switch attributes between instanced and non-instanced depending on the model.
- Update the vertexCount in one model
- Update the instanceCount in the other model

I don't love the idea of using callbacks to solve this problem, but I'm not sure how to do all of that without them. Would moving the callbacks out of `props` (maybe just making them member properties of the layer?) be enough?

Pessimistress: There are two different things we want to do here:
1. Upload new attribute values to the GPU. These can be expensive so we only want to do it when they are recalculated.
2. Set `vertexCount`/`instanceCount`. They are not tied to any particular attribute. They also cost virtually nothing.

I think it's fair to limit this exploration to 1.

tsherif: Actually, I have a new idea for this. I think the shaderAttributes + conditional compilation should be enough to handle at least filtering and switching b/w instanced and non-instanced. I don't think I can filter out the indices for the sideModel, though. I'll take a crack at this when I get into the office.

Issue #2721, opened 2019-02-25
LineLayer API update


XiaokaiUber:
Switch from screen-pixel-width to meter-width.

- getStrokeWidth: deprecate
- getWidth: Accessor | Number - in meters
- widthMinPixels: Number - default `1`
- widthMaxPixels: Number

#### To Do List
  - [ ] Add label and assign to milestone
  - [ ] Coding
  - [ ] Doc update
  - [ ] Whats new update
  - [ ] Test

PR #2723, opened 2019-02-25
Line layer api update

XiaokaiUber:
For #2721

#### Background
update LineLayer APIs for 7.0-release

#### Change List
- added new props: `widthMinPixels`, `widthMaxPixels` and `getWidth`
- deprecated props: `getStrokeWidth`
- updated docs


docs/layers/line-layer.md
```diff
@@ -97,6 +97,35 @@ The stroke width of each object, in pixels.
...
+The stroke width of each object, in meters.
+
+* If a number is provided, it is used as the stroke width for all objects.
+* If a function is provided, it is called on each object to retrieve its stroke width.
+
+##### `widthScale` (number, optional)
+
+* Default: 1
+
+The scaling factor for the width of each line. If you set the property to `Math.pow(2, viewport.zoom - 12)` it will keep the width constant, corresponding to the current zoom level and the width of 1 pixel at zoom level 12. You can also limit the minimum size of the line with this property.
```
Pessimistress: The "If you set the property to..." sentence no longer applies with this update.

modules/layers/src/line-layer/line-layer.js
```diff
@@ -34,10 +34,14 @@ const defaultProps = {
...
   getColor: {type: 'accessor', value: DEFAULT_COLOR},
-  getStrokeWidth: {type: 'accessor', value: 1},
+  getWidth: {type: 'accessor', value: 1},
+  widthScale: {type: 'number', value: 1, min: 0},
+  widthMinPixels: {type: 'number', value: 1, min: 0},
+  widthMaxPixels: {type: 'number', value: Number.MAX_SAFE_INTEGER, min: 0},
 
   // deprecated
-  strokeWidth: {deprecatedFor: 'getStrokeWidth'}
+  strokeWidth: {deprecatedFor: 'getStrokeWidth'},
```
Pessimistress: `deprecatedFor` needs to be an existing prop.

docs/layers/line-layer.md
```diff
@@ -97,6 +97,35 @@ The stroke width of each object, in pixels.
...
+The stroke width of each object, in meters.
+
+* If a number is provided, it is used as the stroke width for all objects.
+* If a function is provided, it is called on each object to retrieve its stroke width.
+
+##### `widthScale` (number, optional)
+
+* Default: 1
+
+The scaling factor for the width of each line. You can also limit the minimum size of the line with this property.
```
Pessimistress: > You can also limit the minimum size of the line with this property.

Remove

Pessimistress:
1. Do not change the LineLayer golden images. Update the test cases to use the new API and verify that they can produce the same outcome.
2. ContourLayer's API is not changed, therefore its appearance should not change. This is the point of the render test. You cannot just replace the image so that the test passes.

docs/layers/line-layer.md
```diff
@@ -97,6 +97,35 @@ The stroke width of each object, in pixels.
 * If a number is provided, it is used as the stroke width for all objects.
 * If a function is provided, it is called on each object to retrieve its stroke width.
 
+It will be overridden by `getWidth` if this new accessor is provided.
+
+##### `getWidth` (Function|Number, optional) ![transition-enabled](https://img.shields.io/badge/transition-enabled-green.svg?style=flat-square")
```
Pessimistress: Why is `getStrokeWidth` still in the documentation if it's deprecated?

Issue #2725, opened 2019-02-26
Add IconAtlas Caching


xintongxia:
#### Target Use case

Current `IconLayer` supports [auto packing `iconAtlas`](http://deck.gl/#/documentation/deckgl-api-reference/layers/icon-layer?section=example-auto-packing-iconatlas). However, it will regenerate the shared `iconAtlas` and rebuild `iconMapping` when one of the following condition satisfies.
 - `changeFlags.dataChanged`
 - `changeFlags.updateTriggersChanged.all`
 - `changeFlags.updateTriggersChanged.getIcon`

There are some performance drawbacks:
1. In certain cases `dataChanged` does not effect `iconAtlas` and `iconMapping`, therefore should not regenerate them. For example, when user moving icons around, the positions of the icons changed, but `iconAtlas` does not change.
2. It is not necessary to regenerate the entire `iconAtlas` if only part of the icons changed.
3. When some existing icons are updated with new image data but kept with the same size (width, height), `IconMapping` does not need to change for those icons.

#### Proposed feature

**API change:**
When using [`auto packing iconAtlas`](http://deck.gl/#/documentation/deckgl-api-reference/layers/icon-layer?section=example-auto-packing-iconatlas), `getIcon` could provide an identifier `id`, if `id` is not specified, fallback to `url`.

```
    getIcon: d => ({
      id: 'marker-1',
      url: d.avatar_url,
      width: 128,
      height: 128,
      anchorY: 128,
      mask: true
    })
```

Internally, `IconLayer` should cache the generated `iconAtlas` and `iconMapping`.

#### To Do List 
  - [ ] Add label and assign to milestone 
  - [x] Coding
  - [x] Doc update
  - [ ] What’s new update 
  - [ ] Test

PR #2728, opened 2019-02-27
Support caching icons

xintongxia:
#### Background
#2725 

#### Change List
- update document
- In `auto packing` case: `getMapping` could return an object with an additional field `id`, used to track which icons are already fetched

### DIFF
--- a/modules/layers/src/icon-layer/icon-manager.js
+++ b/modules/layers/src/icon-layer/icon-manager.js
@@ -1,9 +1,9 @@
 /* global document */
 import GL from '@luma.gl/constants';
-import {Texture2D} from 'luma.gl';
+import {Texture2D, readPixelsToBuffer} from 'luma.gl';
 import {loadImage} from '@loaders.gl/core';
 
-const MAX_CANVAS_WIDTH = 1024;
+const DEFAULT_CANVAS_WIDTH = 1024;
 const DEFAULT_BUFFER = 4;
 
 const DEFAULT_TEXTURE_MIN_FILTER = GL.LINEAR_MIPMAP_LINEAR;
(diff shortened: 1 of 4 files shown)


modules/layers/src/icon-layer/icon-manager.js
```diff
@@ -12,6 +14,19 @@ const DEFAULT_TEXTURE_MAG_FILTER = GL.LINEAR;
...
+ *   [gl]: {
+ *     texture, // Texture2D instance containing the icons
+ *     mapping, // Object, {[id]: {width, height, }}
+ *     xOffset, // right position of last icon
+ *     yOffset, // top position of last icon
+ *     canvasHeight // canvas height
+ *   }
+ * }
+ */
+const cached = new LRUCache(3);
```
Pessimistress: Comparing to the text layer, it is less significant for icon layers to be able to share textures with each other. This is also a potential source of bugs if several icon layers try to use the same icon id but different images.

modules/layers/src/icon-layer/icon-manager.js
```diff
@@ -34,50 +34,78 @@ function resizeImage(ctx, imageData, width, height) {
...
+    width: oldWidth,
+    height: oldHeight,
+    parameters: {
+      [GL.TEXTURE_MIN_FILTER]: DEFAULT_TEXTURE_MIN_FILTER,
+      [GL.TEXTURE_MAG_FILTER]: DEFAULT_TEXTURE_MAG_FILTER
+    }
+  });
+  texture.generateMipmap();
+
+  return texture;
```
Pessimistress: delete the buffer after use

Pessimistress:
Can you add a render test for auto packing with a follow up PR?

xintongxia:
> Can you add a render test for auto packing with a follow up PR?

Issue #2738

Issue #2732, opened 2019-03-01
Deck.gl mutates color variables


manassra:
#### Description

In a project where we use DeckGL, @rolyatmax noticed that `getFillColor` wasn't behaving correctly. When using a color variable, and spreading into the function we see this artifact. 
```
const COLOR = [229, 73, 55];
// later...
getFillColor: () => [...COLOR, 20]
```
![image](https://user-images.githubusercontent.com/1332450/53664657-f5f98780-3c1d-11e9-9aef-122332cc0f08.png)

As it turns out, color accessors mutate the color array and set an alpha value if it is missing. For example:
```
const COLOR = [0, 1, 2]
// later...
getLineColor: () => COLOR
// then...
console.log(COLOR) // [0, 1, 2, 255]
```

#### Repro Steps
* Declare an RGB color array as a variable
* return this variable in `getLineColor`
* spread and set a value for the alpha chanel, and return in `getFillColor`

#### Environment (please complete the following information):
 - Framework Version: [e.g. deck.gl 6.3.0] 
6.3.4
 - Browser Version: [e.g. Chrome 71.0]
72.0.3626.119 
 - OS: [e.g. Mac OS X 10.4]
10.14.3

rolyatmax:
It also looks like it's assuming the color array is size 4 and loading the entire array into the color attribute's buffer, and then reading it out in size 4. A color value that looks like `[R, G, B, A, X]` will result in an attribute buffer that looks like `[R, G, B, A, X, R, G, B, A, X, R, G, B, A, X]`. This gets cut up into vec4s and passed to each vertex. So the colors look like:

```
RGBA
XRGB
AXRG
BAXR
```

That's why we've got the funky rainbow color scheme above.

manassra:
Yeah, I wonder whether we should either:
1) Only slice to the first 4 values if the color array is larger than RGBA
2) Pass it through and warn the user

PR #2733, opened 2019-03-01
Prevent mutation of color variables passed to color setters

manassra:
**Background**
  #2732

**Changes**
Instead of mutating the color variables when missing the alpha value, return a new color array instead.

### DIFF
--- a/modules/layers/src/solid-polygon-layer/polygon-tesselator.js
+++ b/modules/layers/src/solid-polygon-layer/polygon-tesselator.js
@@ -70,9 +70,9 @@ export default class PolygonTesselator extends Tesselator {
           target,
           size: 4,
           getValue: object => {
-            const color = accessor(object);
-            if (isNaN(color[3])) {
-              color[3] = 255;
+            let color = accessor(object);
+            if (color.length < 4) {
+              color = [...color, 255];
             }
             return color;
           }
--- a/modules/layers/src/solid-polygon-layer/polygon-tesselator.js
+++ b/modules/layers/src/solid-polygon-layer/polygon-tesselator.js
@@ -29,6 +29,10 @@ const {Tesselator} = experimental;
 import {fp64 as fp64Module} from 'luma.gl';
 const {fp64LowPart} = fp64Module;
 
+// colorArray is used to copy over color values if the passed color is an RGB
+// array instead of RGBA
+const colorArray = [0, 0, 0, 255];
+
 // This class is set up to allow querying one attribute at a time
 // the way the AttributeManager expects it
 export default class PolygonTesselator extends Tesselator {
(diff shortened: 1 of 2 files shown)


modules/layers/src/solid-polygon-layer/polygon-tesselator.js
```diff
@@ -70,9 +70,9 @@ export default class PolygonTesselator extends Tesselator {
           target,
           size: 4,
           getValue: object => {
-            const color = accessor(object);
-            if (isNaN(color[3])) {
-              color[3] = 255;
+            let color = accessor(object);
+            if (color.length < 4) {
+              color = [...color, 255];
```
Pessimistress: This can be very expensive. Can you create an intermediate target e.g.

```
const colorArray = [0, 0, 0, 255];
return this._updateAttribute({
  ...
  getValue: object => {
    const color = accessor(object);
    if (color.length === 4} {
      return color;
    }
    colorArray[0] = color[0];
    colorArray[1] = color[1];
    colorArray[2] = color[2];
    return colorArray;
  }
```

manassra: @Pessimistress how would this ensure that the color values are correctly copied to the buffer before `colorArray` is mutated by the next call of the accessor?

tsherif: @manassra `_updateAttributes` immediately copies the result of `getValue` into the attribute array: https://github.com/uber/deck.gl/blob/b520d5d91324780f9604ff801b9085af6da32e26/modules/core/src/utils/tesselator.js#L126-L129

Further, I'd recommend that the `colorArray` that @Pessimistress suggested be pulled out of the class completely so it's only ever allocated once.

Pessimistress: > how would this ensure that the color values are correctly copied to the buffer before colorArray is mutated by the next call of the accessor?

The color object is used once to populate a chunk in the attribute typed array. It is no longer needed after the current object is processed.

modules/layers/src/solid-polygon-layer/polygon-tesselator.js
```diff
@@ -70,9 +70,9 @@ export default class PolygonTesselator extends Tesselator {
           target,
           size: 4,
           getValue: object => {
-            const color = accessor(object);
-            if (isNaN(color[3])) {
-              color[3] = 255;
+            let color = accessor(object);
+            if (color.length < 4) {
```
rolyatmax: Is it worth warning when the accessor returns a color that is not a vec3 or vec4? This is one of the issues we ran into with the bug described in the issue linked to this PR.

Pessimistress:
`getColor` is handled similarly in `PathTesselator`.

modules/layers/src/solid-polygon-layer/polygon-tesselator.js
manassra: Yeah that's a good point. @Pessimistress's suggestion ensures that we always return a vec4, which handles incorrect inputs better than my initial fix.

modules/layers/src/solid-polygon-layer/polygon-tesselator.js
manassra: @Pessimistress @tsherif thank you for confirming! I just wanted to make sure the value doesn't get overwritten before it is actually used.

manassra:
> `getColor` is handled similarly in `PathTesselator`.

Fixed this one as well!

PR #2737, opened 2019-03-02
New Layer Submodule Structure

Pessimistress:
For https://github.com/uber/deck.gl/issues/2736

#### Change List
- Reorganized layer code into new submodule directories
- Fixed test and example imports
- Linked documentations for new layers


modules/geo-layers/src/tile-layer/utils/viewport-util.js
```diff
@@ -33,10 +33,11 @@ export function getTileIndices(viewport, maxZoom, minZoom) {
     return [];
   }
 
-  viewport = new viewport.constructor({
-    ...viewport,
-    zoom: z
-  });
+  viewport = new viewport.constructor(
```
gnavvy: Is the `view.constructor` part expected? (instead of new Viewport({...))

modules/aggregation-layers/src/index.js
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) 2015 - 2017 Uber Technologies, Inc.
```
gnavvy: nit: do we need to update the year range to 2019?

modules/geo-layers/src/tile-layer/utils/viewport-util.js
Pessimistress: The user-supplied viewport class is used so that this layer works with any custom (non-web-mercator) projection.

modules/aggregation-layers/src/index.js
Pessimistress: Will do in follow up PR.

PR #2742, opened 2019-03-04
Delete path-marker-layer and path-outline-layer

georgios-uber:
To be moved to nebula.gl


xintongxia:
These files are also referred in index file. Could you also cleanup all the places mentioning these files?

Pessimistress:
Addressed in https://github.com/uber/deck.gl/pull/2737

PR #2780, opened 2019-03-11
Don't fire panmove if drag started somewhere else

### LINKED ISSUE #2779: OrthographicView: Panning outside of viewport causes panmove to fire inside of viewport
#### Description

When there are multiple views each with its own controller, panning outside of a viewport currently would cause panmove inside of the viewport to be fired.

This shouldn't happen. 

"_onPanStart" already checks for this problem:

```
      if (!this.isPointInBounds(pos)) {
        return false;
      }
```

"_onPanMove" needs checks too.

#### Repro Steps

```
<DeckGL 
        views={[
            new OrthographicView({
                id: 'v1',
                width: '50%',
                controller: {
                    dragRotate: false,
                    dragPan: true
                }
            }),
            new OrthographicView({
                id: 'v2',
                width: '50%',
                x: '50%',
                controller: {
                    dragRotate: false,
                    dragPan: true
                }
            })
            })
        ]} 
        layers={[layer]} 
        // initialViewState={initialViewState}
        viewState={viewState}
        onViewStateChange={ e=>setViewState(e.viewState)}
    />)
```

 - Framework Version: [deck.gl 6.4.5]
 - Browser Version: [Chrome 72.0.3626.121 ]
 - OS: [Mac OS X 10.14.3]

zhihua-chen:
Fixing a bug. When there are multiple views each with its own controller, panning outside of the current viewport would cause the controller inside the current viewport to fire "panmove", which would throw an error. Adding checks here so panmove doesn't get fired.

fixes #2779 

#### Change List
- ` modules/core/src/controllers/controller.js`

### DIFF
--- a/modules/core/src/controllers/controller.js
+++ b/modules/core/src/controllers/controller.js
@@ -263,7 +263,7 @@ export default class Controller {
       return false;
     }
     const pos = this.getCenter(event);
-    if (!this.isPointInBounds(pos) || !this.isDragging()) {
+    if (!this.isDragging()) {
       return false;
     }
     const newControllerState = this.controllerState.pan({pos});
--- a/modules/core/src/controllers/controller.js
+++ b/modules/core/src/controllers/controller.js
@@ -263,6 +263,9 @@ export default class Controller {
       return false;
     }
     const pos = this.getCenter(event);
+    if (!this.isPointInBounds(pos) || !this.isDragging()) {
+      return false;
+    }
     const newControllerState = this.controllerState.pan({pos});
     return this.updateViewport(newControllerState, NO_TRANSITION_PROPS, {isDragging: true});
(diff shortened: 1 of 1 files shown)


modules/core/src/controllers/controller.js
```diff
@@ -263,6 +263,9 @@ export default class Controller {
       return false;
     }
     const pos = this.getCenter(event);
+    if (!this.isPointInBounds(pos) || !this.isDragging()) {
```
Pessimistress: The `this.isDragging()` check makes sense. However we do want to allow a dragging action to start inside this viewport and continue outside.

zhihua-chen: Make sense. Corrected.

Issue #2745, opened 2019-03-04
IconLayer (Autopacking) Icons losing definition when Zooming out 


Lowhearth:
First of all  I'd like to say thanks for this amazing library.

Using IconLayer with autopacking, the icon correctly diminishes in size when Zooming out but loses significant definition  (Feels like its not using the MIPMAP filtering for minification).
This behavior is not shown when I previously pack the exact same image on a sprite, and use the prepacked atlas IconLayer configuration.
For diminishing the size of the icon I'm using a linear function that reduces the iconSize prop when zoom changes (working fine with prepacked atlas)

I've tried to follow the SourceCode of the IconLayerManager to find the root cause but had no luck.

Thanks in advance!

xintongxia:
@Lowhearth 

1. What prop you are using to control the size of icon?
In `IconLayer`, there is a prop `sizeScale` and accessor `getSize`, which can manipulate the size of icon. Is this what you are using?

As I tried to correlate the size and zoom level, I couldn't reproduce your issue.

2. What `height` and `width` you are used in `getIcon` callback? The `height` and `width` are the size we used to pack the iconAtlas in `IconManager`.

3. Could you provide a small .gif to demonstrate the issue?

Lowhearth:
@xintongxia 

Sorry for the delayed response, I've got all you need :

![grab-landing-page](https://media.giphy.com/media/1wPWLjUOWY6TkdA7OJ/giphy.gif)

For better resolution: https://i.ibb.co/DfMbB3n/issuedem.gif

``` import React, { Component } from "react";
import { render } from "react-dom";
import { StaticMap } from "react-map-gl";
import DeckGL, { IconLayer } from "deck.gl";

import IconClusterLayer from "./icon-cluster-layer";

const MAPBOX_TOKEN =""
const iconImage =
  "https://cors-anywhere.herokuapp.com/http://www.naturephotographysimplified.com/wp-content/uploads/2014/12/Nature-Photography-Simplified-Elements-of-Design-Diagonal-Line-On-Ground.jpg?x92028";
const DATA = [
  {
    coordinates: [28.96, 13.66033],
    name: "Al Zarnkh",
    class: "LL5",
    mass: "700",
    year: 2001
  }
];

export const INITIAL_VIEW_STATE = {
  longitude: -35,
  latitude: 36.7,
  zoom: 1.8,
  maxZoom: 20,
  pitch: 0,
  bearing: 0
};


export class App extends Component {
  constructor(props) {
[...]

PR #2790, opened 2019-03-13
Fix IconLayer autopacking: generate mipmap when constructing Texture

xintongxia:
For #2745


modules/layers/src/icon-layer/icon-manager.js
```diff
@@ -278,7 +278,11 @@ export default class IconManager {
       if (!this._texture) {
         this._texture = new Texture2D(this.gl, {
           width: this._canvasWidth,
-          height: this._canvasHeight
+          height: this._canvasHeight,
+          parameters: {
```
Pessimistress: Make default texture parameters a constant that is shared between pre-packed and auto-packed

PR #2791, opened 2019-03-13
Generic Prop Animation (PR 1/4)

Pessimistress:
For https://github.com/uber/deck.gl/issues/2363

#### Change List
- Add `LayerPropAnimation` class
- Tests

### DIFF
--- a/modules/core/src/lifecycle/prop-animation.js
+++ b/modules/core/src/lifecycle/prop-animation.js
@@ -28,7 +28,11 @@ export default class LayerPropAnimation extends Animation {
         return lerp;
 
       case 'array':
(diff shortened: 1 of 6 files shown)


modules/core/src/lifecycle/prop-animation.js
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    switch (startType) {
+      case 'number':
+        return lerp;
+
+      case 'array':
+        if (
+          startValue.length === endValue.length &&
+          startValue.length <= 16 &&
```
jianhuang01: better to add comment to explain why we use 16 here

Issue #2363, opened 2018-10-20
Implement Generic Prop Animation


Pessimistress:
#### Background

[RFC](https://github.com/uber/deck.gl/blob/master/dev-docs/RFCs/vNext/generic-layer-prop-animation-rfc.md)

#### To Do List 
  - [x] Add label and assign to milestone 
  - [ ] Coding
  - [ ] Doc update
  - [ ] Whats new update 
  - [ ] Test

charlieforward9:
Looking through old PRs here for entertainment - was this what eventually became the `animate` / `transition` props? Is this still something that would be in the cards to implement or is it largely solved?

PR #2792, opened 2019-03-13
Generic Prop Animation (PR 2/4)

Pessimistress:
For #2363

#### Change List
- Handle `LayerPropAnimation` in prop creation
- Check animation states in `layerManager.needsUpdate`
- Update animation props in `layer._initialize` and `layer._update`
- Use prop animation in heatmap example

### DIFF
--- a/examples/website/3d-heatmap/app.js
+++ b/examples/website/3d-heatmap/app.js
@@ -1,8 +1,7 @@
-/* global window */
 import React, {Component} from 'react';
 import {render} from 'react-dom';
 import {StaticMap} from 'react-map-gl';
-import DeckGL, {HexagonLayer} from 'deck.gl';
+import DeckGL, {HexagonLayer, LayerPropAnimation} from 'deck.gl';
 
 // Set your mapbox token here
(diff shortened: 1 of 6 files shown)


examples/website/3d-heatmap/app.js
```diff
@@ -99,7 +52,7 @@ export class App extends Component {
         coverage,
         data,
         elevationRange: [0, 3000],
-        elevationScale: this.state.elevationScale,
+        elevationScale: data ? new LayerPropAnimation(0).to({value: 50, duration: 2000}) : 0,
```
jianhuang01: do we need this if statement here? won't do anything anyway when data is null

modules/core/src/lib/layer.js
```diff
@@ -558,6 +565,7 @@ export default class Layer extends Component {
     // initializeState callback tends to clear state
     this.setChangeFlags({dataChanged: true, propsChanged: true, viewportChanged: true});
 
+    this._updateAnimationProps();
```
jianhuang01: kind of confusing here, we claim to check if needs updating here, but actually does the update.

PR #2796, opened 2019-03-14
Consolidate cell layers

Pessimistress:
For https://github.com/uber/deck.gl/issues/2736

#### Change List
- `HexagonCellLayer` is removed; add generic `ColumnLayer` (see layer API doc)
- `GridCellLayer` now extends `ColumnLayer`
- The old cell layers were not projecting meter elevations correctly. This is a breaking change for `GridCellLayer`, `GridLayer` and `HexagonLayer` (see updated golden images)
- Documentation
- Update layer browser example and unit tests


heshan0131:
This change will break h3 hexagon layer in kepler.gl and the fix wouldn't be easy. H3 hexagon is not a perfect hexagon, the vertices are slightly distorted based on which part of the world it is in. Previously when we can pass `hexagonVertices` to hexagon cell, we can actually pass in h3 hex vertices to achieve good rendering result. 

https://twitter.com/heshan_cheri/status/1033929329976037376

heshan0131:
That being said, I like the idea of consolidating hexagon-cell and grid-cell layer into a column layer. Instead of rescrit the geometry to be a disk with x number of sides can we also allow pass in `diskVertices`?  And if `diskVertices` is passed, we don't need to calculate radius / angle

Pessimistress:
@heshan0131 you can always calculate radius and angle in your own H3 Hexagon layer before passing them to the sub layer.

With that said, we do want to implement a proper H3Layer as part of the deck.gl layer catalog.

heshan0131:
@Pessimistress Since H3 is not actually a perfect hexagon shape, we need to pass in all 6 vertices to render it. Angle and radius won't do

Pessimistress:
@heshan0131 That is not how I read the current [HexagonCellLayer](https://github.com/uber/deck.gl/blob/6.4-release/modules/layers/src/hexagon-cell-layer/hexagon-cell-layer.js). `hexagonVertices` is only used to calculate `angle` and `radius`. I think kepler is using a custom version?

Pessimistress:
@heshan0131 we can add a prop to support custom vertices, but they need to be in *meter offsets* instead of lng/lat like the old API did. Core layers must support all coordinate systems and views, and the old treatment was specialized to only geospatial data in LNGLAT mode.

heshan0131:
Sounds good, we can work with that

PR #2799, opened 2019-03-15
TileLayer API audit

xintongxia:
#### Background
deck.gl v7.0 Layers API Audit

#### Change List
- onDataLoaded: Rename to onTileLoaded
- onGetTileDataError: Rename to onTileError
- add test


docs/layers/tile-layer.md
```diff
@@ -4,31 +4,34 @@ This TileLayer takes in a function `getTileData` that fetches tiles, and renders
...
-    const tile = new VectorTile(new Protobuf(buffer));
-    const features = [];
-    for (const layerName in tile.layers) {
-      const vectorTileLayer = tile.layers[layerName];
-      for (let i = 0; i < vectorTileLayer.length; i++) {
-        const vectorTileFeature = vectorTileLayer.feature(i);
-        const feature = vectorTileFeature.toGeoJSON(x, y, z);
-        features.push(feature);
-      }
+const MAP_SOURCE = `https://a.tiles.mapbox.com/v4/mapbox.mapbox-streets-v7/${z}/${x}/${y}.vector.pbf?access_token=${MAPBOX_TOKEN}`;
```
Pessimistress: x y z are undefined here

docs/layers/tile-layer.md
```diff
@@ -41,48 +44,53 @@ export const App = ({viewport}) => {
 
 ## Properties
 
-Inherits from all [Base Layer](/docs/api-reference/layer.md) properties, along with `renderSubLayer`, `getTileData`, `onGetTileDataError`, `onDataLoaded`, `maxZoom`, `minZoom` and `maxCacheSize`.
+Inherits from all [Base Layer](/docs/api-reference/layer.md) properties, along with `renderSubLayer`, `getTileData`, `onTileError`, `onTileLoaded`, `maxZoom`, `minZoom` and `maxCacheSize`.
```
Pessimistress: Just say "along with the following props"

examples/experimental/bezier/src/sample-graph.json
```diff
@@ -0,0 +1,82 @@
+{
```
Pessimistress: This file should not be in this PR

modules/geo-layers/src/tile-layer/tile-layer.js
```diff
@@ -3,22 +3,22 @@ import {GeoJsonLayer} from '@deck.gl/layers';
...
+  getTileData: {type: 'function', value: ({x, y, z}) => Promise.resolve(null)},
+  onTileLoaded: {type: 'function', value: () => {}},
   // eslint-disable-next-line
-  onGetTileDataError: err => console.error(err),
-  maxZoom: null,
-  minZoom: null,
-  maxCacheSize: null
+  onTileError: {type: 'function', value: err => console.error(err)},
+  maxZoom: {type: 'number', min: 0, value: null},
+  minZoom: {type: 'number', min: 0, value: null},
```
Pessimistress: - `null` will throw an error when checking against type `number`. It's ok to not specify the type of a prop.
- Default `minZoom` should be 0.

modules/geo-layers/src/tile-layer/tile-layer.js
```diff
@@ -56,10 +56,10 @@ export default class TileLayer extends CompositeLayer {
           if (!allCurrTilesLoaded) {
             Promise.all(currTiles.map(tile => tile.data)).then(() => {
               this.setState({isLoaded: true});
-              onDataLoaded(currTiles.filter(tile => tile._data).map(tile => tile._data));
+              onTileLoaded(currTiles.filter(tile => tile._data).map(tile => tile._data));
```
Pessimistress: Reading this again - `onTileLoaded` is not really called when new tile data is loaded, but whenever the viewport is changed. The name is very misleading. Maybe call it `onViewportLoaded`?

xintongxia: `onViewportLoaded` is also confusing.

test/modules/geo-layers/tile-layer/tile-cache.spec.js
```diff
@@ -3,6 +3,10 @@ import TileCache from '@deck.gl/geo-layers/tile-layer/utils/tile-cache';
 import Tile from '@deck.gl/geo-layers/tile-layer/utils/tile';
 import {WebMercatorViewport} from '@deck.gl/core';
 
+/* eslint-disable no-console, no-undef */
+const onError = console.error;
```
Pessimistress: There should not be any console error log from tests. If errors are expected let's test for that. Otherwise they need to throw.

modules/geo-layers/src/tile-layer/tile-layer.js
```diff
@@ -3,22 +3,22 @@ import {GeoJsonLayer} from '@deck.gl/layers';
...
+  getTileData: {type: 'function', value: ({x, y, z}) => Promise.resolve(null)},
+  onViewportLoaded: {type: 'function', value: () => {}},
   // eslint-disable-next-line
-  onGetTileDataError: err => console.error(err),
-  maxZoom: null,
-  minZoom: null,
-  maxCacheSize: null
+  onTileError: {type: 'function', value: err => console.error(err)},
+  maxZoom: {min: 0, value: null},
+  minZoom: {min: 0, value: 0},
```
Pessimistress: - Type must be specified if default is an object.
- There should not be a min value.

Just use `maxZoom: null, minZoom: 0`

PR #2801, opened 2019-03-15
Fix onAfterUpdate callback in tests

xintongxia:
1. onAfterUpdate should be called in every test
2. onAfterUpdate should work when data is empty

### DIFF
--- a/modules/test-utils/src/generate-layer-tests.js
+++ b/modules/test-utils/src/generate-layer-tests.js
@@ -89,7 +89,9 @@ export function generateLayerTests({
     onAfterUpdate(params);
     // Default assert
     if (params.layer.isComposite) {
-      assert(params.subLayers.length, 'Layer should have sublayers');
+      if (Object.keys(params.layer.props).length) {
+        assert(params.subLayers.length, 'Layer should have sublayers');
+      }
     } else {
       assert(params.layer.getModels().length, 'Layer should have models');
     }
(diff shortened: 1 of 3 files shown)


modules/test-utils/src/generate-layer-tests.js
```diff
@@ -98,9 +98,10 @@ export function generateLayerTests({
   testCases.forEach(testCase => {
     testCase.title = wrapTestCaseTitle(testCase.title);
     testCase.onBeforeUpdate = onBeforeUpdate;
-    testCases.onAfterUpdate = _onAfterUpdate;
   });
 
+  testCases[2].onAfterUpdate = _onAfterUpdate;
```
xintongxia: I am not quite sure about this, I am wondering is it necessary to call onAfterUpdate when there is no data?

Pessimistress: There can be many test cases, you cannot only add the callback to one of them.
You can check if data is empty inside `onAfterUpdate`.

modules/test-utils/src/generate-layer-tests.js
```diff
@@ -89,7 +89,9 @@ export function generateLayerTests({
     onAfterUpdate(params);
     // Default assert
     if (params.layer.isComposite) {
-      assert(params.subLayers.length, 'Layer should have sublayers');
+      if (Object.keys(params.layer.props).length) {
```
Pessimistress: Why not just check if `props.data` is empty? This cannot be reliable.

xintongxia: Because it is uncertain that no sublayer will be generated when there is no data.

Pessimistress: What is the use case you're fixing for? Can you add description to the PR summary?

Pessimistress: If you check `GeoJsonLayer`, it does not render any sub layers if `data` is empty. This test utility must work with the most common denominator. If a layer does not require `data` to render, you should add the additional check to its own tests.

PR #2808, opened 2019-03-16
Add H3 layers

Pessimistress:
For https://github.com/uber/deck.gl/issues/2443

H3HexagonLayer (instanced):

h3-hexagon

H3ClusterLayer (renders geojson polygons):

h3-cluster

#### Change List
- Add H3HexagonLayer and H3ClusterLayer
- API documentation
- Tests


modules/geo-layers/src/h3-layers/h3-hexagon-layer.js
```diff
@@ -0,0 +1,100 @@
...
+      return worldPosition;
+    });
+
+    this.setState({centerHex: hex, vertices});
+  }
+
+  renderLayers() {
+    const {data, getHexagon, updateTriggers} = this.props;
+
+    const SubLayerClass = this.getSubLayerClass('hexagon-cell', ColumnLayer);
```
heshan0131: Does `ColumnLayer` handle calculation of normals based on vertices in the primitive geometry?

test/render/test-cases.js
```diff
@@ -1121,5 +1124,44 @@ export const TEST_CASES = [
...
+      zoom: 11,
+      pitch: 30,
+      bearing: 0
+    },
+    layers: [
+      new H3HexagonLayer({
+        data: h3.kRing('882830829bfffff', 4),
+        getHexagon: d => d,
+        getColor: (d, {index}) => [255, index * 5, 0],
+        getElevation: (d, {index}) => index * 100
```
heshan0131: is it possible to add another `getCoverage` accessor?

modules/geo-layers/src/h3-layers/h3-hexagon-layer.js
Pessimistress: Not yet, it currently assume that normals are close enough to the regular polygon. We do not render normals correctly anyways (normals are interpolated instead of flat for each side).

test/render/test-cases.js
Pessimistress: Yes? What is the use case, do we need different coverage for each hexagon?

website/contents/pages.js
```diff
@@ -407,6 +407,14 @@ export const docPages = generatePath([
             name: 'GridCellLayer',
             content: getDocUrl('layers/grid-cell-layer.md')
           },
+          {
+            name: 'H3HexagonLayer',
+            content: getDocUrl('layers/h3-hexagon-layer.md')
```
ibgreen: Nit: I think we should have a separate index rather than duplicate layers in two places just to get an alphabetic list. An index could be easily auto-generated by ocular once we port the deck website after v7 release. The docs should just contain folders corresponding to the modules. If keeping this.

modules/geo-layers/src/h3-layers/h3-hexagon-layer.js
```diff
@@ -0,0 +1,100 @@
...
+
+    return new SubLayerClass(
+      this.props,
+      this.getSubLayerProps({
+        id: 'hexagon-cell',
+        updateTriggers
+      }),
+      {
+        data,
+        diskResolution: 6,
```
ibgreen: What is `diskResolution`? Worth a comment?

test/render/test-cases.js
heshan0131: As shown here https://twitter.com/carnby/status/1059530716269154306
It's useful for odt analysis

kepler.gl currently edited shader to enable `instanceCoverage`
https://github.com/uber/kepler.gl/blob/master/src/deckgl-layers/h3-hexagon-cell-layer/h3-hexagon-cell-layer.js#L26-L41

modules/geo-layers/src/h3-layers/h3-hexagon-layer.js
Pessimistress: The `ColumnLayer` is currently using `CylinderGeometry` from luma. I feel this option needs to be added at the source.

docs/layers/h3-hexagon-layer.md
```diff
@@ -0,0 +1,131 @@
...
+## Sub Layers
+
+The H3HexagonLayer renders the following sublayers:
+
+* `hexagon-cell` - a [ColumnLayer](/docs/layers/column-layer.md) rendering all hexagons.
+
+
+## Remarks
+
+* Each hexagon in the H3 indexing system is [slightly different in shape](https://uber.github.io/h3/#/documentation/core-library/coordinate-systems). To draw a large number of hexagons efficiently, the H3HexagonLayer assumes that all hexagons within the current viewport have the same shape as the one at the center of the current viewport. This strategy is usually sufficient. However, the discrepancy may become visually significant at rare geolocations. In that case, the [H3ClusterLayer] can be used as an alternative by trading performance for accuracy.
```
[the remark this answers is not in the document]
Pessimistress: It is mentioned in the description of `getHexagon`.

test/render/test-cases.js
Pessimistress: Will do in a follow up PR.

docs/layers/h3-cluster-layer.md
```diff
@@ -0,0 +1,82 @@
+<!-- INJECT:"PolygonLayerDemo" -->
+
+<p class="badges">
+  <img src="https://img.shields.io/badge/64--bit-support-blue.svg?style=flat-square" alt="64-bit" />
+  <img src="https://img.shields.io/badge/extruded-yes-blue.svg?style=flat-square" alt="extruded" />
+</p>
+
+# H3ClusterLayer
+
+The `H3ClusterLayer` renders regions represented by hexagon sets from the [h3](https://uber.github.io/h3/) geospatial indexing system.
```
isaacbrodsky: nit: capitalize `H3` here

docs/layers/h3-hexagon-layer.md
```diff
@@ -0,0 +1,131 @@
+<!-- INJECT:"PolygonLayerDemo" -->
+
+<p class="badges">
+  <img src="https://img.shields.io/badge/64--bit-support-blue.svg?style=flat-square" alt="64-bit" />
+  <img src="https://img.shields.io/badge/extruded-yes-blue.svg?style=flat-square" alt="extruded" />
+</p>
+
+# H3HexagonLayer
+
+The `H3HexagonLayer` renders hexagons from the [h3](https://uber.github.io/h3/) geospatial indexing system.
```
isaacbrodsky: nit: capitalize `H3` again

docs/layers/h3-hexagon-layer.md
```diff
@@ -0,0 +1,131 @@
...
+
+Inherits from all [Base Layer](/docs/api-reference/layer.md) and [CompositeLayer](/docs/api-reference/composite-layer.md) properties.
+
+### Render Options
+
+##### `coverage` (Number, optional)
+
+* Default: `1`
+
+Radius multiplier, between 0 - 1. The radius of the disk is calculated by
```
isaacbrodsky: `the disk` seems a little unclear to me, I assume this is scaling the radius of the hexagon? Could the text be updated to clarify what `the disk` is?

docs/layers/h3-hexagon-layer.md
```diff
@@ -0,0 +1,131 @@
...
+
+Column elevation multiplier. The elevation of column is calculated by
+`elevationScale * getElevation(d)`. `elevationScale` is a handy property
+to scale all hexagon elevations without updating the data.
+
+##### `extruded` (Boolean, optional)
+
+* Default: `true`
+
+Whether to extrude hexagon. If se to false, all hexagons will be set to flat.
```
isaacbrodsky: If se -> If set

docs/layers/h3-hexagon-layer.md
```diff
@@ -0,0 +1,131 @@
+<!-- INJECT:"PolygonLayerDemo" -->
+
+<p class="badges">
+  <img src="https://img.shields.io/badge/64--bit-support-blue.svg?style=flat-square" alt="64-bit" />
+  <img src="https://img.shields.io/badge/extruded-yes-blue.svg?style=flat-square" alt="extruded" />
```
isaacbrodsky: Do we want to add an H3 version badge, similar to what we have on the binding readme: http://github.com/uber/h3-js?

modules/geo-layers/src/h3-layers/h3-hexagon-layer.js
```diff
@@ -0,0 +1,100 @@
...
+      this._updateVertices(this.context.viewport);
+    }
+  }
+
+  _updateVertices(viewport) {
+    const {resolution, centerHex} = this.state;
+    if (resolution < 0) {
+      return;
+    }
+    const hex = geoToH3(viewport.latitude, viewport.longitude, resolution);
```
isaacbrodsky: What happens if this is a pentagon?

docs/layers/h3-hexagon-layer.md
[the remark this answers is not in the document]
dfellis: If I can interject, every aperture has a mix of hexagons and pentagons. The pentagons have simply been placed in water so common use-cases do not run into them.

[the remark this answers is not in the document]
Pessimistress: This will require some changes in the `ColumnLayer`, will do in a follow up PR. Do you have a hex id I can test with?

docs/layers/h3-hexagon-layer.md
Pessimistress: I don't expect the JavaScript API to be important here, but the version of the indexing system definitely matters. How stable is the `h3-js` API? Could we relax it to something like ">=2" so users may pick their version?

docs/layers/h3-hexagon-layer.md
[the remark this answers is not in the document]
dfellis: Sure. One example pentagon at resolution 9 is `891c0000003ffff`.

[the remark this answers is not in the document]
dfellis: (To be fair, when I said "a mix of hexagons and pentagons" there are always 12 pentagons at each resolution, so the higher the resolution it becomes far, far less likely to run into one.)

You may also want to run a `kRing` of 10 or 20 around that pentagon, since pentagons are at the epicenter of the hexagon distortions and you're going to run into essentially 5 different hexagon shapes in that area (one for each of the icosahedron faces the hexagons are on, and technically some really weird ones that are split across the icosahedron boundaries).

modules/geo-layers/src/h3-layers/h3-hexagon-layer.js
ibgreen: That could make sense. Opened luma issue https://github.com/uber/luma.gl/issues/994, so that deck can reference this in a follow-up tracker ticket for the H3 layers.

docs/layers/h3-hexagon-layer.md
isaacbrodsky: As long as it's the same major version it should be fine. cc @nrabinowitz

nrabinowitz: Honestly even the major version of the library shouldn't matter, as long as it's the same index version.

modules/geo-layers/src/h3-layers/h3-hexagon-layer.js
nrabinowitz: Discussed offline with @Pessimistress - the responsibility is on the user to use this fast version only if they're willing to deal with occasional issues arising from pentagons and crossing icosahedron edges.

PR #2815, opened 2019-03-20
Add contentBase config for local debugging

xintongxia:
#### Background
When debugging locally, I need manually change the docs url each time, I would like to automatically switch to correct url when env changed. 

#### Change List
- webpack config in website/
- docs url prefix


website/webpack.config.js
```diff
@@ -13,20 +13,17 @@ const ALIASES = require('ocular-dev-tools/config/ocular.config')({
...
 const BABEL_CONFIG = {
-  presets: [
-    'es2015',
-    'stage-2',
-    'react'
-  ].map(name => require.resolve(`babel-preset-${name}`)),
-  plugins: [
-    'transform-decorators-legacy'
-  ].map(name => require.resolve(`babel-plugin-${name}`))
+  presets: ['es2015', 'stage-2', 'react'].map(name => require.resolve(`babel-preset-${name}`)),
```
Pessimistress: The `map` call is no longer necessary.

website/webpack.config.js
```diff
@@ -94,14 +93,20 @@ const addDevConfig = config => {
...
     ])
-
   });
-
 };
 
 const addProdConfig = config => {
+  config.plugins = config.plugins.concat(
+    new webpack.DefinePlugin({
+      DOCS_DIR: JSON.stringify('https://raw.githubusercontent.com/uber/deck.gl/master')
```
Pessimistress: This file needs to be added to scripts/update-release-branch.sh

website/webpack.config.js
xintongxia: It is needed website still in babel 6. I can upgrade to 7.

PR #2818, opened 2019-03-21
[Discussion] User external h3

Pessimistress:
This change avoids installing and bundling h3-js if the layers are not used, and also allows users to install an alternative implementation of h3 (e.g. an earlier version from the internal module)

#### Change List
- Add required `h3` props to `H3HexagonLayer` and `H3ClusterLayer`.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -3,16 +3,12 @@ const getBabelConfig = require('ocular-dev-tools/config/babel.config');
 module.exports = api => {
   const config = getBabelConfig(api);
 
-  config.plugins.push(
-    'version-inline',
-    'inline-webgl-constants',
-    [
-      'remove-glsl-comments',
-      {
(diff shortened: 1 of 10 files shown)


Pessimistress:
> Or maybe the bulk of the 150KB is in the core projection code?

libh3.js (the EMSCRIPTEN transpiled code) is 147K and minified.

Pessimistress:
Regarding the S2Layer:

- As far as I can tell, there is no versioning in the S2 system, while `h3-js` is only v3.
- `s2-geometry` as a dependency is only 13.5KB minified and 3.7KB gzipped.
- `s2-geometry` can be imported with a `<script>` tag, which creates a global `S2` object. We exclude this module from the standalone bundle and look for the global object instead. There is no similar way to use H3.

Issue #2802, opened 2019-03-15
Add example for BitmapLayer


xintongxia:
#### Proposed feature
Related issue #2507

#### To Do List 
  - [ ] Add label and assign to milestone 
  - [ ] Coding
  - [ ] Doc update
  - [ ] What’s new update 
  - [ ] Test

PR #2821, opened 2019-03-21
Add OpenStreetMap example using TileLayer and BitmapLayer

xintongxia:
For #2802 

#### Change List
- example
openstreet

### DIFF
--- a/examples/website/openstreet-map/app.js
+++ b/examples/website/openstreet-map/app.js
@@ -1,0 +1,103 @@
+import React, {PureComponent} from 'react';
+import {render} from 'react-dom';
+
+import DeckGL, {TileLayer, BitmapLayer} from 'deck.gl';
+
+export const INITIAL_VIEW_STATE = {
+  latitude: 47.65,
+  longitude: 7,
+  zoom: 4.5,
+  // https://wiki.openstreetmap.org/wiki/Zoom_levels
+  maxZoom: 20,
+  pitch: 50,
+  bearing: 0
+};
+
+// https://wiki.openstreetmap.org/wiki/Slippy_map_tilenames#Tile_servers
+// `s` represents sub-domain of tile servers
+const s = 'c';
+
+export class App extends PureComponent {
+  constructor(props) {
+    super(props);
+    this.state = {};
+    this._onHover = this._onHover.bind(this);
+    this._renderTooltip = this._renderTooltip.bind(this);
+  }
+
+  _onHover({x, y, sourceLayer, tile}) {
+    this.setState({x, y, hoveredObject: {sourceLayer, tile}});
+  }
+
+  _renderTooltip() {
+    const {x, y, hoveredObject} = this.state;
+    const {sourceLayer, tile} = hoveredObject || {};
+    return (
+      sourceLayer &&
+      tile && (
+        <div className="tooltip" style={{left: x, top: y}}>
+          <div className="tooltip-item">
+            <div>tile:</div>
+            <div>
+              x: {tile.x}, y: {tile.y}, z: {tile.z}
+            </div>
+          </div>
+          <div className="tooltip-item">
+            <div>boundary:</div>
+            <div>
+              north: {sourceLayer.props.bounds[3].toFixed(4)}, west:{' '}
+              {sourceLayer.props.bounds[0].toFixed(4)}
+            </div>
+          </div>
+          <div className="tooltip-item">
+            <div>coordinates:</div>
+            <div>
+              x: {x.toFixed(4)}, y: {y.toFixed(4)}
+            </div>
+          </div>
+        </div>
+      )
+    );
+  }
+
+  _renderLayers() {
+    return [
+      new TileLayer({
+        pickable: true,
+        onHover: this._onHover,
+        renderSubLayers: props => {
+          const {x, y, z, bbox} = props.tile;
+          const {west, south, east, north} = bbox;
+
+          return new BitmapLayer({
+            id: `bitmap-${props.id}`,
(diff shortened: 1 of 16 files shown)


Pessimistress:
Screenshot?

examples/website/openstreet-map/app.js
```diff
@@ -0,0 +1,101 @@
...
+            this.setState({
+              [layerId]: {
+                image: data.image,
+                bounds: data.bounds
+              }
+            });
+          });
+
+          return new BitmapLayer({id: layerId, ...this.state[layerId]});
+        }
```
xintongxia: From my understanding, `renderSubLayers` callback makes sense when the SubLayer expects `data` field, deck.gl Layer auto resolves the `data` promise and feed resolved data to that SubLayer. 

But `BitmapLayer` does not rely on data, and it uses `image` and `bounds` to place the image. So I have to manipulate the react state to trigger the layer re-render when data resolved.

examples/website/openstreet-map/app.js
```diff
@@ -0,0 +1,70 @@
+import React, {Component} from 'react';
+import {render} from 'react-dom';
+
+import DeckGL, {TileLayer, BitmapLayer} from 'deck.gl';
+
+export const INITIAL_VIEW_STATE = {
+  latitude: 47.65,
+  longitude: 7,
+  zoom: 4.5,
+  maxZoom: 16,
```
Pessimistress: Any reason to set a max zoom? Comment?

examples/website/openstreet-map/app.js
```diff
@@ -0,0 +1,70 @@
...
+  _renderLayers() {
+    return [
+      new TileLayer({
+        renderSubLayers: props => {
+          const {x, y, z} = props.tile;
+
+          return new BitmapLayer({
+            id: `bitmap-${props.id}`,
+            image: `https://${s}.tile.openstreetmap.org/${z}/${x}/${y}.png`,
+            bounds: tileIndexToBounds({x, y, z})
```
Pessimistress: Just a thought: the lnglat bounds of a tile could be something we provide as part of the `Tile` class. I can see a lot of use cases needing these numbers.

examples/website/openstreet-map/index.html
```diff
@@ -0,0 +1,37 @@
...
+  <style>
+    body {
+      margin: 0;
+      font-family: sans-serif;
+      width: 100vw;
+      height: 100vh;
+      overflow: hidden;
+    }
+
+    .tooltip {
```
Pessimistress: Idea: auto highlight the tile under the cursor and show a tooltip of its x,y,z coordinates.

examples/website/openstreet-map/package.json
```diff
@@ -0,0 +1,22 @@
...
+  "license": "MIT",
+  "scripts": {
+    "start-local": "webpack-dev-server --env.local --progress --hot --open",
+    "start": "webpack-dev-server --progress --hot --open"
+  },
+  "dependencies": {
+    "deck.gl": "^7.0.0-alpha",
+    "react": "^16.3.0",
+    "react-dom": "^16.3.0",
+    "react-map-gl": "^3.3.0"
```
Pessimistress: This can be removed now?

modules/layers/src/bitmap-layer/bitmap-layer.js
```diff
@@ -192,7 +198,12 @@ export default class BitmapLayer extends Layer {
 
     if (typeof image === 'string') {
       loadImage(image).then(data => {
-        this.setState({bitmapTexture: new Texture2D(gl, {data})});
+        this.setState({
+          bitmapTexture: new Texture2D(gl, {
+            data,
+            parameters: DEFAULT_TEXTURE_PARAMETERS
```
Pessimistress: Does this help the image quality?

examples/website/openstreet-map/app.js
```diff
@@ -0,0 +1,70 @@
...
+export const INITIAL_VIEW_STATE = {
+  latitude: 47.65,
+  longitude: 7,
+  zoom: 4.5,
+  maxZoom: 16,
+  pitch: 50,
+  bearing: 0
+};
+
+const s = 'c';
```
Pessimistress: What does this mean?

examples/website/openstreet-map/app.js
```diff
@@ -0,0 +1,70 @@
...
+
+export class App extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {};
+  }
+
+  _renderLayers() {
+    return [
+      new TileLayer({
```
Pessimistress: What are the min and max zoom of OSM tiles? Let's set them explicitly.

modules/layers/src/bitmap-layer/bitmap-layer.js
xintongxia: Not really. It looks the same with or without these parameters.

TrantorM:
I tried to run your openstreet-map example and got an error shown in the attached image. To use the example standalone I deleted the bottom line `module.exports = env...` in the file `webpack.config.js` according the description within that file.

Bildschirmfoto 2019-03-28 um 14 59 08

Any Idea?

ibgreen:
@TrantorM FWIW:

If you are testing things on master, you probably want to use `npm run start-local` to build the example against the local source on master rather than against installed modules.

The reason is that since deck.gl v7 is still in development the published modules may not contain what you need.

TrantorM:
Not able to build it on master with `npm run start-local` either. Can't resolve '@loaders.gl/core'.

Bildschirmfoto 2019-03-29 um 12 12 13

A `git checkout tags/v7.0.0-beta.1`  results in the same problem.

PR #2850, opened 2019-03-28
Set picking color only when hovering

xintongxia:


modules/core/src/lib/picking/pick-info.js
```diff
@@ -108,9 +110,6 @@ export function processPickInfo({
       infos.set(info.layer.id, info);
     }
 
-    const pickingSelectedColor =
-      layer.props.autoHighlight && pickedLayer === layer ? pickedColor : null;
-
     layer.setModuleParameters({
```
Pessimistress: You're still clearing picked color in all modes

PR #2852, opened 2019-03-28
Update to attribute and geometry APIs

tsherif:
Made updates based on new Geometry and attribute APIs. A bit cleaner than the last attempt. Wasn't able to run the tests as they were failing on me.


jianhuang01:
@tsherif I changed the default lighting material prop in luma.gl, if it failed because the lighting is slightly different, you can just update the golden images.

tsherif:
@jianhuang01 Oh that might explain something I was seeing before. But the latest failure was actually preventing the tests from running.

PR #2864, opened 2019-03-29
Integrate lighting in mesh layer

jianhuang01:
For #2833 

#### Change List
- Add module dependency
- Change shaders with lighting APIs

### DIFF
--- a/modules/mesh-layers/src/simple-mesh-layer/simple-mesh-layer-vertex.glsl.js
+++ b/modules/mesh-layers/src/simple-mesh-layer/simple-mesh-layer-vertex.glsl.js
@@ -19,20 +19,21 @@ attribute vec3 instanceTranslation;
 
 // Outputs to fragment shader
 varying vec2 vTexCoord;
+varying vec3 cameraPosition;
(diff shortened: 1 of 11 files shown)


modules/mesh-layers/src/simple-mesh-layer/simple-mesh-layer-vertex.glsl.js
```diff
@@ -32,7 +34,9 @@ void main(void) {
 
   vTexCoord = texCoords;
   vColor = instanceColors;
-  // vLightWeight = lighting_getLightWeight(position_commonspace.xyz, project_normal(normals));
+  vec3 normals_commonspace = project_normal(instanceModelMatrix * normals);
+  lightColor = lighting_getLightColor(instanceColors.rgb, project_uCameraPosition, position_commonspace.xyz, normals_commonspace);
+  specularColor = lighting_getSpecularLightColor(project_uCameraPosition, position_commonspace.xyz, normals_commonspace);
```
tsherif: You'll get discontinuities if you calculate specular highlights in the vertex shader.

PR #2866, opened 2019-03-29
Add highPrecision prop to H3HexagonLayer

Pessimistress:
Follow up of #2808

h3-hexagon-high-precision

#### Change List
- Add `highPrecision` prop to `H3HexagonLayer` (see docs for details)
- Add tests

@dfellis


dfellis:
Looks great for those running into this corner case :)

modules/geo-layers/src/h3-layers/h3-hexagon-layer.js
```diff
@@ -37,21 +47,34 @@ export default class H3HexagonLayer extends CompositeLayer {
...
       const {iterable, objectInfo} = createIterable(props.data);
       for (const object of iterable) {
         objectInfo.index++;
-        const sampleHex = props.getHexagon(object, objectInfo);
-        resolution = h3GetResolution(sampleHex);
-        break;
+        const hexId = props.getHexagon(object, objectInfo);
+        // Take the resolution of the first hex
+        resolution = resolution < 0 ? h3GetResolution(hexId) : resolution;
+        hasPentagon = hasPentagon || h3IsPentagon(hexId);
```
dfellis: You could technically put a break conditional on `hasPentagon === true` in this loop. It won't do anything for the regular cases, but would reduce the iterations when someone has force-enabled high precision mode. Not sure of the perf tradeoff evaling that conditional on each run of the loop, though.

PR #2880, opened 2019-04-02
Draft: google maps module RFC

Pessimistress:
@donmccurdy

### DIFF
--- a/dev-docs/RFCs/v7.0/google-maps-integration.md
+++ b/dev-docs/RFCs/v7.0/google-maps-integration.md
@@ -17,7 +17,7 @@ This RFC proposes a new module `@deck.gl/google-maps` that handles the synchroni
 The proposed `@deck.gl/google-maps` module will have a single export `DeckOverlay`. This class implements the [OverlayView](https://developers.google.com/maps/documentation/javascript/reference/#OverlayView) interface and can be used as any other Google Maps overlay:
 
 ```js
-  import DeckOverlay from '@deck.gl/google-maps';
+  import {DeckOverlay} from '@deck.gl/google-maps';
   import {GeoJsonLayer} from '@deck.gl/layers';
 
   // Create map
@@ -44,27 +44,24 @@ The proposed `@deck.gl/google-maps` module will have a single export `DeckOverla
 
 `const deckOverlay = new DeckOverlay(props)`
 
-Parameters:
+`props` are forwarded to a `Deck` instance. The following [Deck](/docs/api-reference/deck.md) props are supported:
 
-* `props.api` - the client from which we pull the Maps API classes. Default `google.maps`.
-* The following [Deck](/docs/api-reference/deck.md) props are supported:
-  - `layers`
-  - `layerFilter`
-  - `effects`
-  - `parameters`
-  - `pickingRadius`
-  - `useDevicePixels`
-  - `onWebGLInitialized`
-  - `onBeforeRender`
-  - `onAfterRender`
-  - `onLoad`
+- `layers`
+- `effects`
+- `parameters`
+- `pickingRadius`
+- `useDevicePixels`
+- `onWebGLInitialized`
+- `onBeforeRender`
+- `onAfterRender`
+- `onLoad`
 
 
 ##### `setMap`
 
 `deckOverlay.setMap(map)`
 
-Add/remove the overlay from a map.
+Add/remove the overlay from a map. An overlay can be temporarily hidden from a map by calling `setMap(null)`. Removing an overlay does not destroy the WebGL context; use `finalize()` if the overlay should be permanently removed.
 
 ##### `setProps`
 
@@ -103,15 +100,24 @@ Supported features:
 
 - Layers
 - Effects
-- Auto-highliting
+- Auto-highlighting
 - Transitions / Animations
-- Interaction callbacks
+- `onHover` and `onClick` callbacks
 
 Not supported features:
 
 - Multi view
 - Controller
 - React integration (there is no official React wrapper of Google Maps)
+- Gesture event callbacks (e.g. `onDrag*`)
+
+### Support tilting?
+
+According to official docs:
+
+> Controls the automatic switching behavior for the angle of incidence of the map. The only allowed values are 0 and 45. The value 0 causes the map to always use a 0° overhead view regardless of the zoom level and viewport. The value 45 causes the tilt angle to automatically switch to 45 whenever 45° imagery is available for the current zoom level and viewport, and switch back to 0 whenever 45° imagery is not available (this is the default behavior). 45° imagery is only available for satellite and hybrid map types, within some locations, and at some zoom levels.
+
+There is no obvious way to get the fractional tilting of the base map. We will hide the deck.gl layers when tilt is larger than 0.
 
 ### How do we test these features?
 
--- a/dev-docs/RFCs/v7.0/google-maps-integration.md
+++ b/dev-docs/RFCs/v7.0/google-maps-integration.md
@@ -1,0 +1,74 @@
+# RFC: Google Maps Integration
+
+* **Authors**: Xiaoji Chen
+* **Date**: April 2019
+* **Status**: Draft
+
+## Overview
+
+deck.gl was initially designed to work with mapbox-gl. Its MapView is implemented to precisely match the camera behavior of Mapbox. So far, Mapbox has been the only base map solution for deck.gl. There has been multiple attempts from external developers to use deck.gl with Google Maps API. It is no trivial task, because Google Maps is much more opaque comparing to mapbox-gl.
+
+This RFC proposes a new module `@deck.gl/google-maps` that handles the synchronization between Deck and Google Maps. This offering will make it a lot easier for Google Maps devlopers to leverage deck.gl for data visualization.
+
+## Proposal
+
+### DeckOverlay
+
+The proposed `@deck.gl/google-maps` module will have a single export `DeckOverlay`. This class implements the [OverlayView](https://developers.google.com/maps/documentation/javascript/reference/#OverlayView) interface and can be used as any other Google Maps overlay:
+
+```js
+  import DeckOverlay from '@deck.gl/google-maps';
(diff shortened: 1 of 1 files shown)


dev-docs/RFCs/v7.0/google-maps-integration.md
```diff
@@ -0,0 +1,120 @@
...
+This RFC proposes a new module `@deck.gl/google-maps` that handles the synchronization between Deck and Google Maps. This offering will make it a lot easier for Google Maps devlopers to leverage deck.gl for data visualization.
+
+## Proposal
+
+### DeckOverlay
+
+The proposed `@deck.gl/google-maps` module will have a single export `DeckOverlay`. This class implements the [OverlayView](https://developers.google.com/maps/documentation/javascript/reference/#OverlayView) interface and can be used as any other Google Maps overlay:
+
+```js
+  import DeckOverlay from '@deck.gl/google-maps';
```
donmccurdy: I somewhat prefer named exports for forward-compatibility, and that would be more consistent with `@deck.gl/mapbox` too.

dev-docs/RFCs/v7.0/google-maps-integration.md
```diff
@@ -0,0 +1,120 @@
...
+  });
+
+  // Create overlay instance
+  const overlay = new DeckOverlay({
+    layers: [
+      new GeoJsonLayer({
+        ...
+      })
+    ]
+  });
```
donmccurdy: For consistency with other Google Maps API layers, I think developers may expect to be able to pass a `map` option to this constructor. I realize that's less consistent with the `Deck` class, which is a tradeoff. Which do you prefer?

dev-docs/RFCs/v7.0/google-maps-integration.md
```diff
@@ -0,0 +1,120 @@
...
+  - `onBeforeRender`
+  - `onAfterRender`
+  - `onLoad`
+
+
+##### `setMap`
+
+`deckOverlay.setMap(map)`
+
+Add/remove the overlay from a map.
```
donmccurdy: Should I expect `setMap` to destroy the WebGL context? If not (preferred), what happens if I later add the layer to another map? If so (less preferred), we'll need another mechanism to toggle the layer on/off. With existing Maps API layers, `setMap()` is the preferred way to show/hide a layer without destroying it.

dev-docs/RFCs/v7.0/google-maps-integration.md
```diff
@@ -0,0 +1,120 @@
...
+  overlay.setMap(map);
+```
+
+##### `Constructor`
+
+`const deckOverlay = new DeckOverlay(props)`
+
+Parameters:
+
+* `props.api` - the client from which we pull the Maps API classes. Default `google.maps`.
```
donmccurdy: Is this for test setup? The official library will always be a global, `google.maps`.

dev-docs/RFCs/v7.0/google-maps-integration.md
```diff
@@ -0,0 +1,120 @@
...
+
+### What Deck features are supported?
+
+Supported features:
+
+- Layers
+- Effects
+- Auto-highliting
+- Transitions / Animations
+- Interaction callbacks
```
donmccurdy: Unless there's another way to keep the DeckGL canvas synced to the Maps API DOM than the one I've used in [this prototype](https://gist.github.com/donmccurdy/d3aaedf98bd05a1b4a6bbd563804611f#file-googlemapslayer-ts-L95-L99), the canvas will not directly receive any DOM events unless we specifically re-emit them based on Maps API events.

Which interaction callbacks do you mean here? Certainly things like `onLayerHover` and `onLayerClick` seem helpful. But I don't think the layer needs to replicate `onDrag[Start|End]`, which would already be provided by Maps API.

Should the interaction callbacks use the Maps API `addListener` pattern, or something more `EventTarget`-like?

dev-docs/RFCs/v7.0/google-maps-integration.md
```diff
@@ -0,0 +1,120 @@
...
+
+This module will be much simpler than `@deck.gl/mapbox`. Because there is no context sharing, there is no need for WebGL state management, custom rendering order, etc. All layers are rendered once the overlay is redrawn.
+
+### What Deck features are supported?
+
+Supported features:
+
+- Layers
+- Effects
+- Auto-highliting
```
donmccurdy: Per offline conversation, it sounds like we'll need a new API for triggering auto-highlighting, if the canvas is not directly receiving DOM events?

dev-docs/RFCs/v7.0/google-maps-integration.md
```diff
@@ -0,0 +1,120 @@
...
+```js
+  import DeckOverlay from '@deck.gl/google-maps';
+  import {GeoJsonLayer} from '@deck.gl/layers';
+
+  // Create map
+  const map = new google.maps.Map(document.getElementById('map'), {
+    mapTypeId: 'hybrid',
+    center: { lat: 40, lng: -100 },
+    zoom: 5,
+    tilt: 0
```
donmccurdy: Note that the `tilt` option has a nontrivial definition: 

> Controls the automatic switching behavior for the angle of incidence of the map. The only allowed values are 0 and 45. The value 0 causes the map to always use a 0° overhead view regardless of the zoom level and viewport. The value 45 causes the tilt angle to automatically switch to 45 whenever 45° imagery is available for the current zoom level and viewport, and switch back to 0 whenever 45° imagery is not available (this is the default behavior). 45° imagery is only available for satellite and hybrid map types, within some locations, and at some zoom levels.

I assume that's why you're setting 0º here, but just to be explicit, we probably can't support the 45º mode with DeckOverlay.

Pessimistress: We can disable drawing when `getTilt()` is not 0.

dev-docs/RFCs/v7.0/google-maps-integration.md
Pessimistress: This is what we are doing in the mapbox module (wrap the event and pass to a private method):

https://github.com/uber/deck.gl/blob/master/modules/mapbox/src/deck-utils.js#L138

I think this is good enough for now?

dev-docs/RFCs/v7.0/google-maps-integration.md
Pessimistress: Good call, drag events are not going to be supported.

dev-docs/RFCs/v7.0/google-maps-integration.md
Pessimistress: When writing this I was hoping that there might be other ways to import the library (e.g. via npm package). If that's not the case let's take it out.

dev-docs/RFCs/v7.0/google-maps-integration.md
Pessimistress: `deckOverlay.finalize()` destroys the WebGL context. We can make it so that the old context is kept unless you add this overlay to a different map.

donmccurdy: I can't think of a case where you'd need to move a layer back and forth between maps without losing the WebGL context, so that sounds fine to me. 👍

dev-docs/RFCs/v7.0/google-maps-integration.md
donmccurdy: Yeah, even with the unofficial npm packages the library still ends up in the global namespace.

dev-docs/RFCs/v7.0/google-maps-integration.md
donmccurdy: Agreed, that should work. 👍

dev-docs/RFCs/v7.0/google-maps-integration.md
Pessimistress: I was looking at some of the built-in overlays, e.g. [GroundOverlay](https://developers.google.com/maps/documentation/javascript/examples/groundoverlay-simple) is constructed without a map.

dev-docs/RFCs/v7.0/google-maps-integration.md
```diff
@@ -103,15 +100,24 @@ Supported features:
...
 - React integration (there is no official React wrapper of Google Maps)
+- Gesture event callbacks (e.g. `onDrag*`)
+
+### Support tilting?
+
+According to official docs:
+
+> Controls the automatic switching behavior for the angle of incidence of the map. The only allowed values are 0 and 45. The value 0 causes the map to always use a 0° overhead view regardless of the zoom level and viewport. The value 45 causes the tilt angle to automatically switch to 45 whenever 45° imagery is available for the current zoom level and viewport, and switch back to 0 whenever 45° imagery is not available (this is the default behavior). 45° imagery is only available for satellite and hybrid map types, within some locations, and at some zoom levels.
+
+There is no obvious way to get the fractional tilting of the base map. We will hide the deck.gl layers when tilt is larger than 0.
```
donmccurdy: The tilt is never fractional – it's either 0º or 45º, and nothing in between. See [this example](https://developers.google.com/maps/documentation/javascript/examples/aerial-rotation) and try zooming out once. It's [possible to get the tilt](https://stackoverflow.com/questions/37313360/google-maps-javascript-api-v3-how-do-i-make-gettilt-method-return-current-ti) by listening for a `tilt_changed` event. However, I'm not confident the actual imagery is _exactly_ at 45º, and would be surprised if it lines up easily and consistently.

dev-docs/RFCs/v7.0/google-maps-integration.md
donmccurdy: GroundOverlay can [still take a map as a constructor argument](https://developers.google.com/maps/documentation/javascript/reference/image-overlay#GroundOverlayOptions), e.g.:

```js
historicalOverlay = new google.maps.GroundOverlay(
  'https://www.lib.utexas.edu/maps/historical/newark_nj_1922.jpg',
  imageBounds,
  {map}
);
```

But in any case this isn't high priority.

PR #2889, opened 2019-04-02
Fix point cloud example for the OribitView change

xintongxia:
Screen Shot 2019-04-02 at 4 23 37 PM


gnavvy:
can we reduce the opacity of the points? 
or is it a blending issue?

xintongxia:
> can we reduce the opacity of the points?
> or is it a blending issue?

The previous uploaded .gif is in low resolution, I uploaded an image with higher resolution.

Issue #2890, opened 2019-04-02
[7.0] SimpleMeshLayer issue tracker


Pessimistress:
- [x] Bug: transform attributes never update. This is due to https://github.com/uber/deck.gl/blob/master/modules/mesh-layers/src/utils/matrix.js#L167 using a stale version of a layer - it always retrieves the props from the layer instance at initialization.
- [x] Feature request: add `wireframe` mode to draw meshes using `GL.LINE` (see avs.auto demo)
- [x] Support using loaders.gl to fetch mesh, similar to https://github.com/uber/deck.gl/blob/master/modules/mesh-layers/src/scenegraph-layer/scenegraph-layer.js#L40
- [x] Attribute validation is overly strict: missing texCoords should be allowed

tsherif:
@Pessimistress for wireframe, just draw the same thing with gl.LINE_STRIP? Or do we we want to re-organize the mesh so the wireframe shows the actual mesh?

Pessimistress:
The streetscape.gl version just used `gl.LINE_STRIP`.

Pessimistress:
When the mesh attributes are invalid, the layer emits an error `deck.gl: undefined`. Can we add a readable message to this, considering this is going to be user-facing?

PR #2901, opened 2019-04-04
Fix stale layer bug when using matrix attributes

tsherif:
For #2890 

Matrix attributes no longer hold on to layer at initialization. Required an update to Luma.gl.


tsherif:
@Pessimistress I haven't been able to reproduce this in the layer browser. Any suggestions? I figured messing around with the controls would trigger it, but they work fine.

PR #2908, opened 2019-04-05
disable lighting in picking pass

jianhuang01:
For #2833 

#### Change List
- add empty light source for picking
- add test

### DIFF
--- a/modules/core/src/passes/pick-layers-pass.js
+++ b/modules/core/src/passes/pick-layers-pass.js
@@ -19,7 +19,8 @@ export default class PickLayersPass extends LayersPass {
     onViewportActive,
     pickingFBO,
     deviceRect: {x, y, width, height},
-    redrawReason = ''
+    redrawReason = '',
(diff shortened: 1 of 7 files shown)


modules/core/src/passes/pick-layers-pass.js
```diff
@@ -19,7 +19,8 @@ export default class PickLayersPass extends LayersPass {
     onViewportActive,
     pickingFBO,
     deviceRect: {x, y, width, height},
-    redrawReason = ''
+    redrawReason = '',
+    effectProps = {lightSources: {}}
```
Pessimistress: Why is the default value needed? The picking pass should not be aware of specific effects.

jianhuang01: didn't we discuss to use the empty lightSources to disable lighting? if we don't fill it here, it can be in deck-picker or base LayersPass, where do you prefer?

PR #2913, opened 2019-04-07
Align all website example setup and READMEs

Pessimistress:
Delete point-cloud-ply


xintongxia:
What is the reason to delete the `point-cloud-ply` example?

Pessimistress:
> What is the reason to delete the point-cloud-ply example?

After switching to loaders.gl this example is essentially the same as the point-cloud-laz example.

PR #2919, opened 2019-04-08
Fix attribute transition

Pessimistress:
#### Background

Attribute transition is throwing `Found instanced attributes on non-instanced model transform-model` error. E.g. `instancePositionsFrom` and `instancePositionsTo` are not supposed to be instanced in the transform feedback model, but they are auto-inferred to be based on their names.

#### Change List
- Stop using attribute descriptors in AttributeTransitionManager.

### DIFF
--- a/test/apps/attribute-transition/app.js
+++ b/test/apps/attribute-transition/app.js
@@ -3,13 +3,7 @@
 import React, {Component} from 'react';
 import {render} from 'react-dom';
 import {StaticMap} from 'react-map-gl';
-import DeckGL, {
-  COORDINATE_SYSTEM,
-  MapView,
-  ScatterplotLayer,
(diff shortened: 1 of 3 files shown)


test/apps/attribute-transition/app.js
```diff
@@ -83,14 +77,12 @@ class Root extends Component {
...
       <div>
         <DeckGL
-          views={new MapView({id: 'map'})}
           controller={MapController}
           viewState={viewState}
           onViewStateChange={evt => this.setState({viewState: evt.viewState})}
           layers={layers}
         >
           <StaticMap
-            viewId="map"
```
georgios-uber: Is this related?

Pessimistress: No.

PR #2924, opened 2019-04-08
Add transitions to ScenegraphLayer and update demo to use them

georgios-uber:


examples/website/scenegraph-layer/app.js
```diff
@@ -51,6 +54,7 @@ export class App extends Component {
       .then(resp => {
         if (resp && resp.states) {
           this.setState({data: resp.states});
+          setTimeout(() => this._loadData(), REFRESH_TIME);
```
Pessimistress: Clear the time out when the app unmounts.

examples/website/scenegraph-layer/app.js
```diff
@@ -45,12 +49,32 @@ export class App extends Component {
...
+  }
+
   _loadData() {
     fetch(DATA_URL)
       .then(resp => resp.json())
       .then(resp => {
         if (resp && resp.states) {
-          this.setState({data: resp.states});
+          const nextTimeoutId = setTimeout(() => this._loadData(), REFRESH_TIME);
+          this.setState({data: this._sort(resp.states, this.state.data), nextTimeoutId});
```
Pessimistress: This will fail if the app has already been unmounted.

georgios-uber: But I cancel the timeout in that case.

Pessimistress: If your promise is resolved after the app is unmounted,

- The timeout is not cancelled because it's not set yet
- Calling `this.setState` will throw

georgios-uber: How do I prevent that?

Pessimistress: Set a flag when the app unmounts.

georgios-uber: ok 🎏

georgios-uber: Done

examples/website/scenegraph-layer/app.js
```diff
@@ -64,43 +97,94 @@ export class App extends Component {
...
+          {baseMap && (
+            <StaticMap
+              reuseMaps
+              mapStyle={MAPBOX_STYLE}
+              preventStyleDiffing={true}
+              mapboxApiAccessToken={MAPBOX_TOKEN}
+            />
+          )}
+        </DeckGL>
+        <div
```
Pessimistress: The information panel is rendered and styled by the website, see https://github.com/uber/deck.gl/blob/master/website/src/components/demos/geojson-demo.js#L27

PR #2925, opened 2019-04-08
Fix onHover behavior

Pessimistress:
#### Background

The new `onHover` callback (replaces `onLayerHover`) should always return a non-empty first argument per documentation.

#### Change List

- `pickObject` returns a valid picking info object (with `index: -1` and current pixel information) even if nothing is picked
- Fix layer browser crash

### DIFF
--- a/modules/core/src/lib/deck-picker.js
+++ b/modules/core/src/lib/deck-picker.js
@@ -215,17 +215,12 @@ export default class DeckPicker {
       }
     }
 
-    if (result.length === 0 && infos) {
-      // result should always contain something
-      result.push(infos.get(null));
-    }
-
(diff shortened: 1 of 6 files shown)


modules/core/src/lib/deck-picker.js
```diff
@@ -218,7 +220,7 @@ export default class DeckPicker {
       layers[layerId].restorePickingColors(affectedLayers[layerId])
     );
 
-    return result;
+    return {result, emptyInfo: infos && infos.get(null)};
```
jianhuang01: what are we expecting from infos.get(null)? Isn't that always 'undefined'?

PR #2932, opened 2019-04-10
Fix examples/gallery local debugging config

xintongxia:
#### Background
v7.0 audit

#### Change List
- fix `LOCAL_BUNDLE` path
- update readme


examples/gallery/README.md
```diff
@@ -34,7 +34,7 @@ To test all examples with the local version, run:
 
 ```bash
 # rebuild bundle
-cd ../../modules/lite && npm run build && cd -
+cd ../../modules/main && npm run build-bundle && cd -
```
Pessimistress: Run prepublish which will give you both dust and dist.min. dist is easier to use when debugging.

examples/gallery/build-tools/constants.js
```diff
@@ -12,7 +12,7 @@ if (!MAPBOX_TOKEN) {
 }
 /* eslint-enable */
 
-const LOCAL_BUNDLE = path.resolve(__dirname, '../../../node_modules/deck.gl/dist.js');
+const LOCAL_BUNDLE = path.resolve(__dirname, '../../../modules/main/dist.js');
```
Pessimistress: These two should be the same after running bootstrap at root.

PR #2934, opened 2019-04-10
Fix external gl context usage

Pessimistress:
External gl context must be supplied to `AnimationLoop` as a prop for polyfill to work. This breaks the mapbox custom layer use case.


Pessimistress:
> Do we understand what has changed (in luma.gl) here?

This is related to https://github.com/uber/luma.gl/pull/954 but it only handles external context provided using `props.gl`, not `props.createContext`

PR #2939, opened 2019-04-10
Improve React synchronization

Pessimistress:
For https://github.com/uber/deck.gl/issues/2831
https://github.com/uber/deck.gl/issues/2458

#### Background

The root cause of the synchronization issue is that there are up to 3 independent render loops: Deck, Mapbox and React. Each component in this mix may notify others to update, but they each update on its own schedule.

There are the following issues in the DeckGL React component implementation:

- Deck may not redraw immediately after component render, making it lag behind the map and other overlays.
- It keeps a copy of the current view state in the component state as a mechanism to trigger rerender.
  + Even if the user supplies the `viewState` via props, the internal state still gets updated and can potentially go out of sync or cause double rerender.
  + In order to mimic the view state handling logic, the component duplicates a considerable amount of code from the Deck class.
- It calls `deck.setProps` inside the render method.
  + This is often unnecessary, when the render is triggered by an internal view state change.
  + In the case of viewport transition, `setProps` triggers another `onViewStateChange` call which calls `setState`. Calling `setState` inside of `render` causes React to emit an error. Sometimes this also causes the map to flash to the destination viewport before resume transitioning.

This PR rewrites the React component logic to something hopefully simpler and more generic.

#### Change List
- Remove DeckGL component states
- Remove DeckGL view state/size change handlers
- Check for view state change in a custom render callback and trigger React render if needed


modules/react/src/deckgl.js
```diff
@@ -89,25 +87,18 @@ export default class DeckGL extends React.PureComponent {
...
 
-    // If initialViewState was set on creation, auto track position
-    if (this.state.viewState) {
-      this.setState({
-        viewState: Object.assign({}, this.state.viewState, {
-          [params.viewId]: viewState
-        })
-      });
+    if (viewports !== this.viewports) {
+      // Viewports have changed, update children props first
```
xintongxia: what is the reason of this comment `Viewports have changed, update children props first`?

Issue #2941, opened 2019-04-10
Controller option for smooth zoom transition


Lowhearth:
#### Target Use case
- Anyone who wants smooth transitions between zoom points using the default controller.

#### Proposed feature
- A controller option/prop that will set a transitionDuration for this event.

#### To Do List 
  - [X] Add label and assign to milestone 
  - [X] Coding
  - [X] Test
  - [ ] Doc update
  - [ ] What’s new update

PR #2942, opened 2019-04-10
Controller prop for smooth zoom in/out 

Lowhearth:
Hello,
thanks again for this amazing library, I'm learning a lot just by browsing the source code and experimenting.

I had to implement the feature of smooth zooming and came up with this solution, a zoomTransition Controller option that will set the transitionDuration for the animation. #2941 

Let me know what you think and how can I improve it  

  - [X] Add label and assign to milestone 
  - [X] Coding
  - [X] Test
  - [ ] Doc update
  - [ ] What’s new update 

#### Change List
- Added a prop to the Controller class

### DIFF
--- a/modules/core/src/controllers/controller.js
+++ b/modules/core/src/controllers/controller.js
@@ -158,7 +158,8 @@ export default class Controller {
       doubleClickZoom = true,
       touchZoom = true,
       touchRotate = false,
-      keyboard = true
+      keyboard = true,
+      zoomTransition = 0
     } = props;
 
     // Register/unregister events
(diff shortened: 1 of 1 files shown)


Pessimistress:
You can achieve this by extending the default `MapController` class:

```js
import DeckGL, {MapController} from 'deck.gl'

class MyController extends MapController {
  _getTransitionProps() {
    return {transitionDuration: 0};
  }
}

<DeckGL ... controller={MyController} />

```

Lowhearth:
Definitely taking this approach if we decide that the feature is not worthy to be included in the controller class.
¿ Wouldn't it be good to have this built in feature ?

Pessimistress:
There are a lot of hard-coded behaviors in the controller class, e.g. transition duration/interpolator, increment for keyboard navigation, etc. I think it's a great suggestion that we should expose them as options. I would avoid making individual additions like this and come up with a comprehensive list of options.

Lowhearth:
I couldn't agree more

Issue #2935, opened 2019-04-10
deck.gl not working with vue.js


DominusVilicus:
Hello,

I'm trying to make deck.gl work inside a Vue application, however, it's not working for me. The canvas is being recognised and attached and the canvas becomes interactive (can grab) **However,** my red dot doesn't appear in the canvas. I'm not sure why this is happening

```html
<template>
    <div style="height: 100%">
          <canvas id="my-canvas"></canvas>
    </div>
</template>
<script>
import { Deck } from "@deck.gl/core"
import { ScatterplotLayer } from "@deck.gl/layers"

export default {
    data(){
        return {
            deck: null
        }
    },
    mounted(){
        const INITIAL_VIEW_STATE = {
            latitude: 37.78,
            longitude: -122.45,
            zoom: 3
        }

        this.deck = new Deck({
            canvas: 'my-canvas',
            initialViewState: INITIAL_VIEW_STATE,
            controller: true,
            layers: [
                new ScatterplotLayer({
                    data: [{
                        position: [-122.45, 37.8],
                        color: [255, 0, 0],
                        radius: 100
                    }]
                })
            ]
        })
    },

}
</script>
```

DominusVilicus:
Here's a codesandbox https://codesandbox.io/s/1mx1v97q3

DominusVilicus:
I'd be glad to help with the development of deck.gl and expand its reach into other frameworks like Vue, but I can't get a simple standalone working.

Pessimistress:
It is working, it's just that your layer does not render anything visible.

First, you need to specify the `getRadius` prop of the ScatterplotLayer. The default is `1`. Second, the radius is in meters, so it's not visible at zoom level 3.

You can try adding the following to the layer:

```
getRadius: d => d.radius,
getColor: d => d.color,
radiusMinPixels: 1
```

DominusVilicus:
@Pessimistress lol, thankyou! I've suggested some doc updates to make the standalone "get-started" guide simpler to get started. I think it'll help with adoption of deck.gl

PR #2945, opened 2019-04-10
Fixed standalone example

DominusVilicus:
**Fixed docs to be more hello-world friendly for standalone**
Makes the red dot visible (at zoom 3 or 1 it was not visible) #2935 

For #2935

### DIFF
--- a/docs/get-started/using-standalone.md
+++ b/docs/get-started/using-standalone.md
@@ -26,7 +26,6 @@ const INITIAL_VIEW_STATE = {
 };
 
 const deckgl = new Deck({
-  canvas: 'my-deck-canvas', // This is the canvas id in your html (<canvas id="my-deck-canvas">)
   initialViewState: INITIAL_VIEW_STATE,
   controller: true,
   layers: [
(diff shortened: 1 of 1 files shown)


docs/get-started/using-standalone.md
```diff
@@ -11,6 +11,8 @@ Our [examples](https://github.com/uber/deck.gl/tree/master/examples/get-started)
 
 The [Deck](/docs/api-reference/deck.md) class takes deck.gl layer instances and viewport parameters, and renders those layers as a transparent overlay.
 
+You must create a canvas inside the html with the id `my-deck-canvas` for it to work.
```
xintongxia: Move this comment close to line `29` to avoid confusion

DominusVilicus: Done

docs/get-started/using-standalone.md
```diff
@@ -20,20 +20,22 @@ import {Deck} from '@deck.gl/core';
...
-  latitude: 37.78,
+  latitude: 37.8,
   longitude: -122.45,
-  zoom: 3
+  zoom: 15
 };
 
 const deckgl = new Deck({
-  canvas: 'my-deck-canvas',
+  canvas: 'my-deck-canvas', // This is the canvas id in your html (<canvas id="my-deck-canvas">)
```
Pessimistress: We can just remove this line. The default parent will be document.body

DominusVilicus: I've removed it, but I feel like most users may be more comfortable with mounting to a dom element, rather than it being automatically created

PR #2949, opened 2019-04-11
Improve react perf with memoized extractors

Pessimistress:
#### Change List
- Export `memoize` from core
- Avoid re-extracting JSX components if props are not changed

### DIFF
--- a/modules/core/src/index.js
+++ b/modules/core/src/index.js
@@ -79,8 +79,9 @@ export {default as log} from './utils/log';
 import {flattenVertices, fillArray} from './utils/flatten'; // Export? move to luma.gl or math.gl?
 
 export {createIterable} from './utils/iterable-utils';
-import {default as Tesselator} from './utils/tesselator'; // Export? move to luma.gl or math.gl?
+import Tesselator from './utils/tesselator'; // Export? move to luma.gl or math.gl?
 import {count} from './utils/count';
+import memoize from './utils/memoize';
 
 // lighting
 export {AmbientLight, PointLight, DirectionalLight} from '@luma.gl/core';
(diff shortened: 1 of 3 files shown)


modules/react/src/utils/position-children-under-views.js
```diff
@@ -0,0 +1,57 @@
+import {createElement} from 'react';
+import {View, log} from '@deck.gl/core';
+import {inheritsFrom} from './inherits-from';
```
xintongxia: not important, but maybe could also make these `imports` consistent `import inheritsFrom from './inherits-from';`

modules/core/src/index.js
```diff
@@ -79,8 +79,9 @@ export {default as log} from './utils/log';
 import {flattenVertices, fillArray} from './utils/flatten'; // Export? move to luma.gl or math.gl?
 
 export {createIterable} from './utils/iterable-utils';
-import {default as Tesselator} from './utils/tesselator'; // Export? move to luma.gl or math.gl?
+import Tesselator from './utils/tesselator'; // Export? move to luma.gl or math.gl?
```
xintongxia: Are we actually revisiting these `TODO`s

xintongxia:
Did you test, how much perf could be improved with this `memorize`?

modules/core/src/index.js
Pessimistress: Out of scope for this PR

Pessimistress:
> how much perf could be improved with this memorize

I can confirm that `extractJSXLayers` is no longer called every render. The perf improvements will depend on how many children `<DeckGL>` contains.

PR #2958, opened 2019-04-12
Fix gallery examples

Pessimistress:
#### Change List
- Rename dev bundle to dist.dev.js
- Update reference in gallery start-local script
- Fix point cloud standalone example


Pessimistress:
We're not using the dev bundle when published, the setting is for local debug.

PR #2959, opened 2019-04-12
clean up in mesh layer

jianhuang01:
#### Background
clean up unused and duplicated code

#### Change List
- clean up

### DIFF
--- a/modules/mesh-layers/src/simple-mesh-layer/simple-mesh-layer.js
+++ b/modules/mesh-layers/src/simple-mesh-layer/simple-mesh-layer.js
@@ -24,9 +24,8 @@
 
 import {Layer, createIterable} from '@deck.gl/core';
 import GL from '@luma.gl/constants';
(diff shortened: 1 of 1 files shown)


modules/mesh-layers/src/simple-mesh-layer/simple-mesh-layer.js
```diff
@@ -128,7 +127,6 @@ const defaultProps = {
   wireframe: false,
   // Optional material for 'lighting' shader module
   material: defaultMaterial,
-  sizeScale: {type: 'number', value: 1, min: 0},
```
tsherif: Was this removed at some point from the shader? I thought we wanted to keep it as a way to change size via a uniform.

jianhuang01: we have another sizeScale in the same 'defaultProps', so they are duplicated.

PR #2965, opened 2019-04-12
Update testing docs

Pessimistress:
#### Change List

- Consolidate testing documentation and align with other submodules
- Replace outdated examples


docs/developer-guide/testing.md
```diff
@@ -0,0 +1,153 @@
...
+```
+
+
+## Integration Tests
+
+While unit tests are good at capturing issues in layer initialization and prop updates, they do not guarantee that the layer will be correctly rendered to screen. Some issues in e.g. the WebGL shaders can only be spotted in an integration test.
+
+The `@deck.gl/test-utils` module offers a [SnapshotTestRunner](/docs/api-reference/test-utils/snapshot-test-runner.md) that works with the [probe.gl](https://uber-web.github.io/probe.gl) library's [`BrowserTestDriver`](https://github.com/uber-web/probe.gl/blob/master/docs/api-reference/test-utils/browser-test-driver.md) class to perform this task. Together, they enable the following scenario:
+
+* start a Chrome browser instance
```
jianhuang01: Chrome or Chromium?

PR #2966, opened 2019-04-12
SimpleMeshLayer bug fixes

tsherif:
- Update luma version: fixes deletion of input mesh properties.
- Fix hasNormals check for meshes without "attributes" property.


modules/mesh-layers/src/simple-mesh-layer/simple-mesh-layer.js
```diff
@@ -195,8 +195,10 @@ export default class SimpleMeshLayer extends Layer {
       }
       if (props.mesh) {
         this.setState({model: this.getModel(props.mesh)});
+
+        const attributes = props.mesh.attributes ? props.mesh.attributes : props.mesh;
```
Pessimistress: Just `props.mesh.attributes || props.mesh` ?

Issue #2892, opened 2019-04-02
7.0 Release Checklist


tgorkin:
- [x] Verify whats-new (@tgorkin )
- [x] Verify upgrade-guide (@tgorkin )
- [x] Verify all 7.0 issues are closed or move out (@tgorkin )
- [ ] Verify all API changes are documented (@Pessimistress )
- [ ] Website (docs and example) testing: [staging](https://jianhuang01.github.io/deck.gl/)
    - [ ] Testing on Chrome (@1chandu )
    - [x] Testing on Safari (@jianhuang01 )
    - [x] Testing on Firefox (@tgorkin )
    - [ ] Testing on Edge + IE11 (@tsherif)
    - [ ] Testing on iOS (@Pessimistress )
    - [ ] Testing on Android (@Pessimistress)

Testing procedure
- access staged doc website
- go through all examples
- verify functionality in control panels
- look for console errors/warnings
- add results of testing as comments on this issue

Pessimistress:
- [ ] `isWebGL2` is returning true in Safari. Broken example: http://pessimistress.github.io/deck.gl/#/examples/core-layers/screen-grid-layer

@tsherif

tsherif:
The problem is that we have several versions of `isWebGL2`. The broken one is https://github.com/uber/luma.gl/blob/master/modules/webgl-state-tracker/src/utils/utils.js#L10-L13

It breaks after the gl constant replacement. I'll remove that function use the one in webgl/webgl-utils everywhere: https://github.com/uber/luma.gl/blob/master/modules/webgl/src/webgl-utils/webgl-checks.js#L12-L19

tgorkin:
Testing results with Firefox 66.0.2 (64-bit) on MacOSX:

**General warnings on most pages:**
Error: WebGL warning: getContext: Disallowing antialiased backbuffers due to blacklisting.
Error: WebGL warning: texImage2D: Alpha-premult and y-flip are deprecated for non-DOM-Element uploads.
Error: WebGL warning: texSubImage2D: Alpha-premult and y-flip are deprecated for non-DOM-Element uploads.
luma: `Attribute.instanced` is deprecated and will be removed in a later version. Use `undefined` instead
luma: accessor size mismatch
luma: `Buffer.type` is deprecated and will be removed in a later version. Use `Buffer.accessor.type` instead
luma: Model.render() is deprecated. Use Model.setUniforms() and Model.draw()

**Line Layer Example**
[...]

PR #2967, opened 2019-04-12
Remove deprecated props of ScreenGridLayer from examples and tests.

kaushikb11:
For #2892 

- Remove deprecated props `minColor` and `maxColor` of ScreenGridLayer from examples.

- Modify test cases for ScreenGridLayer.


kaushikb11:
Modified `screengrid-infoviz.png` image for a test case of ScreenGridLayer.

Screenshot from 2019-04-13 02-55-01

test/render/test-cases.js
```diff
@@ -779,28 +777,6 @@ export const TEST_CASES = [
     ],
     goldenImage: './test/render/golden-images/grid-lnglat.png'
   },
-  {
```
Pessimistress: Why is this one deleted? We need tests for both GPU aggregation and CPU aggregation.

kaushikb11: GPU aggregation is covered in the `screengrid-lnglat-colorRange` test case as default aggregation happens on GPU.

kaushikb11: There was an extra test case as it had `minColor` and `maxColor` props.

PR #2998, opened 2019-04-17
fix website warnings

jianhuang01:
For #2892 

#### Change List
- update doc link
- remove deprecated APIs

### DIFF
--- a/examples/website/plot/plot-layer/surface-layer.js
+++ b/examples/website/plot/plot-layer/surface-layer.js
@@ -91,11 +91,12 @@ export default class SurfaceLayer extends Layer {
   draw({uniforms}) {
     const {lightStrength} = this.props;
 
-    this.state.model.render(
+    this.state.model.setUniforms(
       Object.assign({}, uniforms, {
         lightStrength
       })
     );
+    this.state.model.draw();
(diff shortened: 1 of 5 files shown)


examples/website/plot/plot-layer/surface-layer.js
```diff
@@ -91,11 +91,12 @@ export default class SurfaceLayer extends Layer {
   draw({uniforms}) {
     const {lightStrength} = this.props;
 
-    this.state.model.render(
+    this.state.model.setUniforms(
```
1chandu: Nit: you can just do setUniforms(...).draw()

examples/website/trips/app.js
```diff
@@ -99,7 +99,7 @@ export class App extends Component {
         data: buildings,
         extruded: true,
         wireframe: false,
-        fp64: true,
+        fp64: false,
```
1chandu: any specific reason?

jianhuang01: we don't need specify fp64 now, there is an automatic switch in zoom 12, from regular mode to offset mode. With fp64 flag, there is a warning printed.

Issue #2968, opened 2019-04-12
Add examples for core layers in gallery


xintongxia:
#### Proposed feature
Deck.gl [examples](http://deck.gl/#/examples/overview) page currently does not have all the demos for all the layers.

For starting point, port [ArcLayer](https://codepen.io/vis-gl/pen/dmmQqa) example from CodePen to deck.gl gallery.

Read the [article](https://github.com/uber/deck.gl/tree/master/examples/gallery) about how to work on gallery locally.

 
#### To Do List 
  - [ ] Add label and assign to milestone 
  - [ ] Coding
     - [x] ArcLayer
     - [x] BitmapLayer
     - [x] ColumnLayer
     - [x] ContourLayer
     - [x] GreatCircleLayer
     - [x] IconLayer
     - [x] LineLayer
     - [ ] MeshLayer
     - [ ] S2Layer
     - [x] TextLayer
     - [ ] TileLayer
  - [ ] Doc update
  - [ ] What’s new update 
  - [ ] Test

manishmz:
@xintongxia what data source should I use to implement a column layer demo?

xintongxia:
> column layer

You can try this dataset
https://github.com/uber-common/deck.gl-data/blob/master/website/hexagons.json

Refer this [example](https://github.com/uber/deck.gl/blob/master/website/src/components/demos/layer-demos.js#L37)

xintongxia:
@manishmz 

Are you planning to add all the examples listed here?

manishmz:
> @manishmz
> 
> Are you planning to add all the examples listed here?

Yes, I am thinking about to add all examples.

manishmz:
@xintongxia I have tried creating examples for SimpleMeshLayer but it's not showing any data visualization and also not giving any error.
Also for the implementing S2Layer example getting an error of 'S2Cell' of undefined.
I tried reading the documentation of these layers but not helping me.

Are there any examples for these layers to refer or anything do you think I am missing?

xintongxia:
@manishmz 

> I have tried creating examples for SimpleMeshLayer but it's not showing any data visualization and also not giving any error.

You can post your code and error message here.

> Also for the implementing S2Layer example getting an error of 'S2Cell' of undefined.
I tried reading the documentation of these layers but not helping me.
> Are there any examples for these layers to refer or anything do you think I am missing?

[...]

PR #3003, opened 2019-04-18
Add IconLayer demo to the gallery

kaushikb11:
For #2968 

- Add IconLayer demo to the gallery. Used auto packing to get Icons. In the example, avatars of Github contributors to the `deckgl` repo were used as Icons. 

- The `edgeLength` is scaled to the number of contributions made by the contributor.

Peek 2019-04-19 02-58


xintongxia:
Great work! @kaushikb11 

A few suggestions,
1. Node `deck.gl` could use [website](http://deck.gl/#/) background.
2. Could try d3 `force-layout` to separate the nodes from each other, right now some of them are overlapped. (this could be a follow up PR)
3. I think correlate the icon size with the contributions would be more visible to the users. 
4. Could change the background to white and use a more muted color for edges.

kaushikb11:
**Additional changes**

I have used the logic of quartiles and 90th percentile to make the correlation of icon size to the no of contributions more visible. As there were outliers, just scaling the contributions wasn't enough. I think even after using force layout when the links are concentrated, there was a bit of overlapping of icons because of Icon sizes. So, I made the links evenly distributed throughout the `main node`. There's almost no overlapping now.

Peek 2019-04-23 00-19

examples/gallery/src/icon-layer.html
```diff
@@ -0,0 +1,191 @@
...
+      #tooltip {
+        font-family: Helvetica, Arial, sans-serif;
+        font-size: 11px;
+        position: absolute;
+        padding: 4px;
+        margin: 8px;
+        background: rgba(0, 0, 0, 0.8);
+        color: #fff;
+        max-width: 300px;
+        font-size: 10px;
```
xintongxia: duplicate

examples/gallery/src/icon-layer.html
```diff
@@ -0,0 +1,191 @@
...
+      const {firstQuartile, secondQuartile, thirdQuartile} = calculateQuartiles(listOfContributions);
+      const ninetiethPercentile = [...listOfContributions]
+        .sort((a, b) => a - b)[Math.floor(0.90 * listOfContributions.length)];
+
+      listOfContributions.forEach((number, index) => {
+        const linkDistance = 20 * ((number - Math.min(...listOfContributions))
+         / (Math.max(...listOfContributions) - Math.min(...listOfContributions))) + 30;
+
+        addTargetPositionToContributor(data[index], linkDistance);
+      })
```
xintongxia: In your code, you calculate the min and max for every contribution, which is not good for performance.

You can use a temporary variable to store this `Math.min(...listOfContributions)` and `Math.max(...listOfContributions)`

xintongxia:
This is great. We can have a follow up PR to make this layout more interactive.

examples/gallery/src/icon-layer.html
```diff
@@ -0,0 +1,196 @@
...
+
+      const maxContributions = Math.max(...listOfContributions);
+      const minContributions = Math.min(...listOfContributions);
+
+      listOfContributions.forEach((number, index) => {
+        const linkDistance = 20 * ((number - minContributions) 
+        / (maxContributions - minContributions)) + 30;
+
+        addTargetPositionToContributor(data[index], linkDistance);
+      })
```
xintongxia: Could extract the logic to get contributor nodes' positions to a separate function.

PR #3012, opened 2019-04-22
Minor changes to the IconLayer AutoPacking Docs Example

kaushikb11:
**IconLayer AutoPacking Example**

- Change Octokit request format from `octokit.repos.listContributors` to `octokit.repos.getContributors`

- Change the tooltip parameters according to example dataset.

- For the autopacking example, I don't think Icons need to be treated as a transparency mask. Keep the default value for mask which is false and the user-defined `getColor` accessor is not required.

### DIFF
--- a/docs/layers/icon-layer.md
+++ b/docs/layers/icon-layer.md
@@ -91,7 +91,7 @@ const App = ({data, viewport}) => {
    */
   const layer = new IconLayer({
     id: 'icon-layer',
-    data: octokit.repos.listContributors({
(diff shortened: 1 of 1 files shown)


docs/layers/icon-layer.md
```diff
@@ -102,17 +102,15 @@ const App = ({data, viewport}) => {
       width: 128,
       height: 128,
       anchorY: 128,
```
xintongxia: remove the trailing comma for line 104.

PR #3014, opened 2019-04-22
Add TextLayer demo to the gallery & modify control-panel in the examples

kaushikb11:
For #2968 

- Add TextLayer demo to the gallery.

- Remove `font-size` duplicate from the examples

Peek 2019-04-14 22-08


examples/gallery/src/text-layer.html
```diff
@@ -0,0 +1,123 @@
...
+        padding: 4px;
+        margin: 8px;
+        background: rgba(0, 0, 0, 0.8);
+        color: #fff;
+        max-width: 300px;
+        font-size: 10px;
+        z-index: 9;
+        pointer-events: none;
+      }
+      #control-panel {
```
xintongxia: Could you apply similar styles of [website](http://deck.gl/#/documentation/deckgl-api-reference/layers/text-layer) demo for this `control-panel`?

For example, `box-shadow` could make the `control-panel` look better

examples/gallery/src/text-layer.html
```diff
@@ -0,0 +1,127 @@
...
+        left: 0;
+        margin: 12px;
+        padding: 10px;
+        font-size: 12px;
+        line-height: 1.5;
+        z-index: 1;
+        background-color: #fcfcfc;
+        border: solid 1px #ccc;
+        border-bottom-color: #bbb;
+        border-radius: 3px;
```
xintongxia: remove `line 40 - line 42`, not quite necessary.

examples/gallery/src/text-layer.html
```diff
@@ -0,0 +1,127 @@
...
+        box-shadow: inset 0 -1px 0 #bbb;
```
xintongxia: Let's use the same shadow as [website control](http://deck.gl/#/documentation/deckgl-api-reference/layers/text-layer) does.

kaushikb11: I have used the same one.

PR #3017, opened 2019-04-22
Add Changes to the LineLayer Demo

kaushikb11:
For #2968 

- Fix the `deck.gl` URL and add `onHover` to the demo to make it interactive for the user. 

Peek 2019-04-14 20-26

### DIFF
--- a/examples/gallery/src/line-layer.html
+++ b/examples/gallery/src/line-layer.html
@@ -16,7 +16,6 @@
       }
       #tooltip {
         font-family: Helvetica, Arial, sans-serif;
-        font-size: 11px;
         position: absolute;
         padding: 4px;
         margin: 8px;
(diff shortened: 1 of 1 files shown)


examples/gallery/src/line-layer.html
```diff
@@ -12,20 +11,32 @@
...
+      #tooltip {
+        font-family: Helvetica, Arial, sans-serif;
+        font-size: 11px;
+        position: absolute;
+        padding: 4px;
+        margin: 8px;
+        background: rgba(0, 0, 0, 0.8);
+        color: #fff;
+        max-width: 300px;
+        font-size: 10px;
```
xintongxia: duplicate

kaushikb11: I thought this is the standard tooltip we are using for the examples. Should I change the css? @xintongxia

PR #3019, opened 2019-04-23
Fix GoogleMapsOverlay remove and finalize

Pessimistress:
For #2981 

#### Change List
- Implement `overlay.onRemove`
- Remove listeners correctly
- Remove canvas when finalized

### DIFF
--- a/modules/google-maps/src/utils.js
+++ b/modules/google-maps/src/utils.js
@@ -63,14 +63,17 @@ function createDeckCanvas(overlay) {
  * @param deck (Deck) - a previously created instances
  */
 export function destroyDeckInstance(deck) {
-  const {_googleMap: map, _eventListeners: eventListeners} = deck.props.userData;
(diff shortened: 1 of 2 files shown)


modules/google-maps/src/utils.js
```diff
@@ -63,14 +63,17 @@ function createDeckCanvas(overlay) {
...
   // Unregister event listeners
   for (const eventType in eventListeners) {
-    map.removeListener(eventListeners[eventType]);
+    eventListeners[eventType].remove();
   }
 
   deck.finalize();
+
+  // Remove canvas
+  deck.canvas.parentNode.removeChild(deck.canvas);
```
xintongxia: Isn't `deck.finalize()` supposed to do this?

Pessimistress: Only if the canvas is created by the Deck instance.

Issue #2894, opened 2019-04-03
Walls of extruded polygons are transparent


SymbolixAU:
#### Description

Walls of extruded polygons are transparent when specifying a transition duration greater than zero, but without actually using the transition. 

A live version showing this issue is [hosted here](https://symbolix.shinyapps.io/debuggingdeckgl/) where you can access the console logs. 

Here is a screenshot of what's happening. 

![Screen Shot 2019-04-03 at 1 25 47 pm](https://user-images.githubusercontent.com/18344164/55448604-0da89f00-5614-11e9-947f-af82794adc63.png)

In the console, for each map I am logging

- `Object` : the transition object
- `l` : the `PolygonLayer` object

The map on the left has `transitions: { getElevation: 0 }` 
The map on the right has `transitions: { getElevation: 600 }`

Therefore I am setting the `transition` prop, but not actually using it in either of the maps.

This has only been happening since v7.0.0-beta. Previous versions have worked as intended

The data is the same in both maps - [widget_melbourne.json](https://github.com/SymbolixAU/debugging/widget_melbourne.json) - a geojson file of polygons in Melbourne

#### Environment

 - Framework: deck.gl 7.0.0-beta.1 (https://unpkg.com/deck.gl@7.0.0-beta.1/dist.min.js)
 - Browser:  Chrome version 73.0.3683.86 (Official Build) (64-bit)
 - Mac OS X 10.14.1 (Mojave)

1chandu:
@SymbolixAU , what layers are you using, is this specific to one layer?  Can you reproduce this with any of the standalone [examples](https://github.com/uber/deck.gl/tree/master/examples/website) provided in the repo, if so it will be easy for us to reproduce and debug.

dcooley:
@1chandu (I'm on my other github account) 

Here's a [codepen](https://codepen.io/dcooley/pen/vMJxpe) which demonstrates the issue. 

The code i'm using is

```js

/** Data **/
const data = [
  {
	  // Simple polygon (array of coords)
	  contour: [[-122.4, 37.75], [-122.4, 37.8], [-122.5, 37.8], [-122.5, 37.75], [-122.4, 37.75]],
	  elevation: 3000
  }
];

/** DeckGL **/
new deck.DeckGL({
  container: 'container',
  mapboxApiAccessToken: 'pk.eyJ1IjoidWJlcmRhdGEiLCJhIjoidGllX1gxUSJ9.gElUooDF7u51guCQREmAhg',
[...]

PR #3023, opened 2019-04-23
Fix constant shader attributes

tsherif:
For #2894

Had to introduce `shaderValues` prop to cover constant matrices. Hope to clean up when the shader attribute stuff is moved into the `Attribute` class.

### DIFF
--- a/modules/mesh-layers/src/utils/matrix.js
+++ b/modules/mesh-layers/src/utils/matrix.js
@@ -115,11 +115,11 @@ export const MATRIX_ATTRIBUTES = {
 
       const valueMatrix = new Float32Array(matrix);
       const valueTranslation = new Float32Array(modelTranslation);
-      const shaderAttributes = attribute.userData.shaderAttributes;
-      shaderAttributes.instanceModelMatrix__LOCATION_0.value = valueMatrix.subarray(0, 3);
-      shaderAttributes.instanceModelMatrix__LOCATION_1.value = valueMatrix.subarray(3, 6);
-      shaderAttributes.instanceModelMatrix__LOCATION_2.value = valueMatrix.subarray(6, 9);
-      shaderAttributes.instanceTranslation.value = valueTranslation;
+
+      attribute.userData.shaderValues.instanceModelMatrix__LOCATION_0 = valueMatrix.subarray(0, 3);
+      attribute.userData.shaderValues.instanceModelMatrix__LOCATION_1 = valueMatrix.subarray(3, 6);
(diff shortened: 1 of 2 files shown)


modules/mesh-layers/src/utils/matrix.js
```diff
@@ -115,11 +115,11 @@ export const MATRIX_ATTRIBUTES = {
 
       const valueMatrix = new Float32Array(matrix);
       const valueTranslation = new Float32Array(modelTranslation);
-      const shaderAttributes = attribute.userData.shaderAttributes;
-      shaderAttributes.instanceModelMatrix__LOCATION_0.value = valueMatrix.subarray(0, 3);
-      shaderAttributes.instanceModelMatrix__LOCATION_1.value = valueMatrix.subarray(3, 6);
-      shaderAttributes.instanceModelMatrix__LOCATION_2.value = valueMatrix.subarray(6, 9);
-      shaderAttributes.instanceTranslation.value = valueTranslation;
+
+      attribute.userData.shaderValues.instanceModelMatrix__LOCATION_0 = valueMatrix.subarray(0, 3);
```
Pessimistress: You should be able to set `attribute.value` to an array of 16 here and derive the shader attribute values using their offsets?

Pessimistress: Clarify: in `attribute.update`, if a constant value is provided and the attribute accessor contains `offset`, should it automatically extract the correct subarray from the value?

tsherif: `offset` is attribute pointer state and using a constant attribute involves disabling the attribute pointer.

Pessimistress: But you can use the information to extract the subarray from the value in https://github.com/uber/deck.gl/blob/master/modules/core/src/lib/base-attribute.js#L63, something like

```
if (constant) {
	this.value = value.slice(this.offset / value.BYTES_PER_ELEMENT, this.size);
}
```

Pessimistress: As discussed, I don't think users of attributes should be accessing/manipulating arbitrary content in the `userData` field. Eventually, shader attributes handling should be strictly internal to the `Attribute` class.

tsherif: Oh yeah, that should work. Is value always a typed array?

Pessimistress: Hmm no

tsherif:
@Pessimistress Consolidated this fix with the transitions fixes in #3028

Issue #3042, opened 2019-04-29
OrbitView does not auto-resize


Pessimistress:
#### Description
OrbitView does not resize with the deck canvas.

#### Repro Steps
- Run examples/website/point-cloud
- Resize the browser window
- The viewport does not resize with the canvas

#### Environment (please complete the following information):
 - Framework Version: deck.gl 7.0.3
 - Browser Version: Chrome 73.0
 - OS: Mac OS X 10.14.3

Pessimistress:
Fixed in 7.0.4

PR #3043, opened 2019-04-29
Fix OrbitView and OrthographicView resize bug

Pessimistress:
For #3042

When we merged viewport props the `width` and `height` in `viewState` ended up overriding the values provided by the `ViewManager`.

#### Change List
- Use base `View` class' `_getViewport` implementation instead

### DIFF
--- a/modules/core/src/views/orbit-view.js
+++ b/modules/core/src/views/orbit-view.js
@@ -79,27 +79,19 @@ class OrbitViewport extends Viewport {
 }
 
 export default class OrbitView extends View {
+  constructor(props) {
+    super(
+      Object.assign({}, props, {
+        type: OrbitViewport
+      })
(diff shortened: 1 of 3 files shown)


modules/core/src/views/orbit-view.js
```diff
@@ -79,28 +79,20 @@ class OrbitViewport extends Viewport {
 }
 
 export default class OrbitView extends View {
+  constructor(props) {
+    super(
+      Object.assign({}, props, {
+        type: OrbitViewport
```
jianhuang01: is this ViewportType?

Pessimistress: https://github.com/uber/deck.gl/blob/master/modules/core/src/views/view.js#L152

PR #3050, opened 2019-04-30
Add Jupyter widget to pydeck library

ajduberstein:
Addresses #2929.
#### Background
Adds Python support for a Jupyter widget, supporting #3035.


bindings/python/pydeck/pydeck/_version.py
```diff
@@ -3,5 +3,6 @@
 # Copyright (c) Uber Technologies, Inc.
 # Distributed under the terms of the Modified BSD License.
 
+# TODO should this be the same version number as the deck.gl release?
```
ibgreen: If we need to republish this with every deck.gl release to ensure we capture bug fixes etc it would make sense to align the version numbers.

bindings/python/pydeck/pydeck/widget/widget.py
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+"""
+DeckGLWidget renders JSON into a deck.gl plot
```
ibgreen: I think this could be a bit more articulate... ...accepts JSON encoded deck.gl layers and layer properties and renders them based on provided view states and properties.

bindings/python/pydeck/pydeck/widget/widget.py
```diff
@@ -0,0 +1,27 @@
...
+It requires a mapbox API key to be set as an environment variable
```
ibgreen: Some of our layers are/can be used as non-geospatial. In such cases, it should be possible to use this library without a mapbox key.

ajduberstein:
Successfully modularized the widget&#8212;as of this commit, we have a separate Jupyter widget JS folder and Python folder, as well as a separate directory of JSON bindings.

I still need to get CI tests green again, set up some additional tests to make sure the widget loads successfully in a notebook, and modify the pydeck widget to make its API slightly more conventional.

image

modules/jupyter-widget/package.json
```diff
@@ -23,19 +24,23 @@
...
   "dependencies": {
-    "@deck.gl/json": "7.0.0-beta.2",
-    "@deck.gl/aggregation-layers": "7.0.0-beta.2",
-    "@deck.gl/layers": "7.0.0-beta.2",
     "@jupyter-widgets/base": "^1.1.10",
     "@types/mapbox-gl": "^0.51.4",
-    "@deck.gl/core": "7.0.0-beta.2",
-    "mapbox-gl": "^0.53.1"
+    "mapbox-gl": "^0.53.1",
+    "@deck.gl/aggregation-layers": "7.0.0",
```
Pessimistress: Why lock the package versions?

modules/jupyter-widget/package.json
```diff
@@ -23,19 +24,23 @@
   ],
   "sideEffects": false,
   "scripts": {
-    "clean": "rm -fr dist dist-es6 && mkdir -p dist/es5 dist/esm dist/es6",
+    "clean": "rm -fr dist && mkdir -p dist/es5 dist/esm dist/es6",
     "build-es6": "BABEL_ENV=es6 babel src --out-dir dist/es6 --source-maps --ignore 'node_modules/'",
     "build-esm": "BABEL_ENV=esm babel src --out-dir dist/esm --source-maps --ignore 'node_modules/'",
     "build-es5": "BABEL_ENV=es5 babel src --out-dir dist/es5 --source-maps --ignore 'node_modules/'",
```
Pessimistress: Are these scripts necessary?

ajduberstein:
May-07-2019 16-41-34

Verifying support for different layers but otherwise necessary features now supported:
1) Visualizations remain as long the kernel is active. Still need to persist state to disk if kernel shuts down.
2) Multiple visualizations are now available in one notebook, up to the number of WebGL contexts supported by the browser
3) More documentation and demos, though still more are needed.

I still have to work to get Travis CI functional.

Issue #2084, opened 2018-07-17
GPUGridLayer:  Feature parity with existing GridLayer


1chandu:
#### Background 
GPUGridLayer supports aggregation on GPU and CPU and included in 6.0 release as experimental. But some of the existing features of GridLayer (upper, lower percentiles, separate aggregation parameter for elevation/color etc) need to be added before we make it official.
#### To Do List 
  - [x] Add label and assign to milestone 
  - [x] Coding
      - [x] Make `GPUGridLayer` feature parity with GridLayer
        - [X] Verify and fix world space aggregation precession.
        - [X] Update all aggregation layers and utility methods to use new multiple weights Aggregator API.
        - [x] Integrate GPU and CPU aggregations into one layer (either existing GridLayer or new composite layer that switches between GridLayer and GPUGridLayer based on fallback cases)
        - [x] Make getColorWeight/getElevationWeight  props to be accessors and constant. (for propTypes, set type to accessor and default value to be constant), for constant set up attributes as constants instead of providing full array.
        - [x] Make sure GPUGridCellLayer is feature parity with current column layer, especially elevation calculation has been changed. (If possible extend GPUGridCellLayer from ColumLayer).
        - [x] Picking: for full list of objects in picked cell, use CPU Layer's aggregation methods.
        - [x] Add new props (`getColorWeight`, `getElevationWeight`, `colorAggregation` and `elevationAggregation` to `HexagonLayer` and `GridLayer`).
        - [x] Renames and exports : `GridLayer` -> `CPUGridLayer` and `NewGridLayer` -> `GridLayer`
        - [x] Documentation about subtle differences between CPU/GPU grid layer
        - [x] Document breaking change for GridLayer (subclassing will fail)
        - [x] Docs for new-grid layer. (CPUGridLayer and GridLayer docs, remove GPUGridLayer doc)

  - [x] Doc update
  - [x] Whats new update 
  - [x] Test
     - [x] Add tests for new layer.
     - [x] Add tests for all utility methods.
     - [x] Update render-tests with new layer

PR #3051, opened 2019-05-01
 GPUGridLayer PART-1: Fix world space aggregation

1chandu:
For #2084 

#### Background
Fix issues found when enabling world space aggregation (GridLayer)
- Fix rounding issue in shader instructions when calculating domain (min/max) values.
- Fix Model creation/update when  64-bit aggregation is toggled.

#### Change List
- GPU Grid Aggregation PART-1: Aggregator: fix world space aggregation

### DIFF
--- a/modules/aggregation-layers/src/utils/gpu-grid-aggregation/aggregate-all-vs-64.glsl.js
+++ b/modules/aggregation-layers/src/utils/gpu-grid-aggregation/aggregate-all-vs-64.glsl.js
@@ -26,6 +26,7 @@ in vec2 position;
 uniform vec2 gridSize;
 
 out vec2 vTextureCoord;
+#define ROUNDING_ERROR 0.00001
 void main(void) {
   // Map each position to single pixel
   vec2 pos = vec2(-1.0, -1.0);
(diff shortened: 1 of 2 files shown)


modules/aggregation-layers/src/utils/gpu-grid-aggregation/aggregate-all-vs-64.glsl.js
```diff
@@ -36,7 +37,7 @@ void main(void) {
 
   gl_Position = vec4(pos, 0.0, 1.0);
 
-  float yIndex = floor(float(gl_InstanceID) / gridSize[0]);
+  float yIndex = floor((float(gl_InstanceID) / gridSize[0]) + ROUNDING_ERROR);
```
tsherif: Have you tried making this an integer division. Then you wouldn't even need the `floor`

```
uniform uvec2 gridSize;
...
float yIndex = float(gl_InstanceID / gridSize[0]);
```

modules/aggregation-layers/src/utils/gpu-grid-aggregation/aggregate-all-vs-64.glsl.js
```diff
@@ -26,6 +26,7 @@ in vec2 position;
 uniform vec2 gridSize;
 
 out vec2 vTextureCoord;
+#define ROUNDING_ERROR 0.00001
```
tsherif: This kind of factor is usually called "epsilon".

modules/aggregation-layers/src/utils/gpu-grid-aggregation/aggregate-all-vs-64.glsl.js
1chandu: fixed.

modules/aggregation-layers/src/utils/gpu-grid-aggregation/aggregate-all-vs-64.glsl.js
1chandu: Not needed anymore.

Issue #3015, opened 2019-04-22
Enable stroke for H3HexagonLayer


xintongxia:
#### Proposed feature
Enable `stroke` for `H3HexagonLayer`, subtask for #937

- `hightPrecision` mode, rendered by `SolidPolygonLayer` [#3024]
  - when flat, enable `stroke`, the stroke width could be controlled by users.
  - when `extruded`, enable `wireframe`
- non `hightPrecision` mode, rendered by `ColumnLayer`
  - when flat, enable `stroke`, the stroke width could be controlled by users.
  - when `extruded`, enable `wireframe`[#3053]

#### To Do List 
  - [ ] Add label and assign to milestone 
  - [ ] Coding
  - [ ] Doc update
  - [ ] What’s new update 
  - [ ] Test

PR #3053, opened 2019-05-01
Enable wireframe for ColumnLayer 

kaushikb11:
For #3015 

- [x] Enable stroke feature for Column Layer.

- [x] Fix the Vertical lines when the stroke is enabled for Column Layer.

- [x] Update Docs

Peek 2019-05-01 04-09


modules/layers/src/column-layer/cylinder-geometry.js
```diff
@@ -0,0 +1,14 @@
+import TruncatedConeGeometry from './stroke-truncated-cone-geometry';
+import {uid} from '@luma.gl/core';
+
+export default class CylinderGeometry extends TruncatedConeGeometry {
```
ibgreen: Nit: `ColumnGeometry`?

modules/layers/src/column-layer/stroke-truncated-cone-geometry.js
```diff
@@ -0,0 +1,121 @@
+import {Geometry, uid} from '@luma.gl/core';
+
+const INDEX_OFFSETS = {
+  x: [2, 0, 1],
+  y: [0, 1, 2],
+  z: [1, 2, 0]
+};
+
+export default class TruncatedConeGeometry extends Geometry {
```
ibgreen: Nit: `class StrokeTruncatedConeGeometry`?

Not possible to inherit from https://github.com/uber/luma.gl/blob/master/modules/core/src/geometries/truncated-cone-geometry.js

and just transform/change some of the results?

ibgreen: Or `ColumnOutlineGeometry`, `ColumnWireframeGeometry` etc?

modules/layers/src/column-layer/column-layer-vertex.glsl.js
```diff
@@ -45,6 +48,7 @@ varying vec4 vColor;
 
 void main(void) {
 
+  vec4 colors = isWireframe ? instanceLineColors : instanceColors;
```
Pessimistress: Nit: singular `color`

modules/layers/src/column-layer/column-layer-vertex.glsl.js
```diff
@@ -74,10 +78,10 @@ void main(void) {
   vec3 normals_commonspace = project_normal(vec3(rotationMatrix * normals.xy, normals.z));
 
   if (extruded) {
-    vec3 lightColor = lighting_getLightColor(instanceColors.rgb, project_uCameraPosition, position_commonspace.xyz, normals_commonspace);
-    vColor = vec4(lightColor, instanceColors.a * opacity) / 255.0;
+    vec3 lightColor = lighting_getLightColor(colors.rgb, project_uCameraPosition, position_commonspace.xyz, normals_commonspace);
```
Pessimistress: Lighting should be disabled on the wireframe (it's lines without normals).

modules/layers/src/column-layer/column-layer.js
```diff
@@ -115,28 +125,50 @@ export default class ColumnLayer extends Layer {
...
+        gl,
+        Object.assign({}, this.getShaders(), {
+          id: `${id}-top`,
+          geometry: this.getGeometry(diskResolution),
+          isInstanced: true,
+          shaderCache: this.context.shaderCache
+        })
+      );
+    }
+    if (extruded) {
```
Pessimistress: `wireframe`. We do want stroke in non-extruded mode.

modules/layers/src/column-layer/column-layer.js
```diff
@@ -148,24 +180,47 @@ export default class ColumnLayer extends Layer {
...
+      extruded,
+      coverage,
+      elevationScale
+    });
+    const numInstances = this.getNumInstances();
+    if (strokeModel && wireframe) {
+      strokeModel.setInstanceCount(numInstances);
+      strokeModel.setUniforms(renderUniforms);
+      strokeModel.setDrawMode(GL.LINES);
+      strokeModel.setUniforms({isWireframe: true}).draw();
```
Pessimistress: `drawMode` and this uniform can be set once in `getModels`

modules/layers/src/column-layer/cylinder-geometry.js
```diff
@@ -0,0 +1,14 @@
...
+  constructor(props = {}) {
```
Pessimistress: The subclassing seems excessive. We can just have one custom geometry that is always a cylinder.

modules/layers/src/column-layer/column-layer.js
```diff
@@ -115,28 +125,50 @@ export default class ColumnLayer extends Layer {
...
-      })
-    );
+  _getModels(gl) {
+    const {id, filled, extruded, diskResolution} = this.props;
+
+    let filledModel;
+    let strokeModel;
+
+    if (filled) {
+      filledModel = new Model(
```
Pessimistress: What are the difference between these two models? Can we reuse the same one and draw with different draw modes?

kaushikb11: I thought of defining both the models by their traits. It seemed more readable. But yes, I could reuse the same one model with different `draw modes` with the conditions `wireframe` and `filled`.

modules/layers/src/column-layer/column-layer.js
kaushikb11: Planning to reuse the same one model, the way you have advised. So, `drawMode` could be set in `draw` itself for various conditions.

modules/layers/src/column-layer/column-layer.js
xintongxia: @kaushikb11 

You can with different draw modes, if your geometry is correctly filled and stroked.

modules/layers/src/column-layer/stroke-truncated-cone-geometry.js
xintongxia: @kaushikb11 

As suggested, you can simplify this Geometry and make it as a Cylinder, could support both stroked mode or filled mode.

kaushikb11: Noted. Working on it. @xintongxia

modules/layers/src/column-layer/column-layer.js
kaushikb11: I had to have separate models as they had different `drawModes` and the `indices` generated for each model was a bit different.

modules/layers/src/column-layer/column-layer.js
```diff
@@ -148,24 +180,47 @@ export default class ColumnLayer extends Layer {
...
+      offset,
+      extruded,
+      coverage,
+      elevationScale
+    });
+    const numInstances = this.getNumInstances();
+    if (strokeModel && wireframe) {
+      strokeModel.setInstanceCount(numInstances);
+      strokeModel.setUniforms(renderUniforms);
+      strokeModel.setDrawMode(GL.LINES);
```
1chandu: If drawMode and `isWireframe` are always constants, you can pass them during Model construction and avoid setting them for every draw, same for `filledModel`.

kaushikb11:
Screenshot from 2019-05-03 02-38-34

modules/layers/src/column-layer/column-layer.js
```diff
@@ -39,9 +40,11 @@ const defaultProps = {
   elevationScale: {type: 'number', min: 0, value: 1},
   extruded: true,
   fp64: false,
-
+  wireframe: false,
+  filled: true,
   getPosition: {type: 'accessor', value: x => x.position},
   getColor: {type: 'accessor', value: DEFAULT_COLOR},
```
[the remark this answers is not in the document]
xintongxia: Check [`ScatterplotLayer`](https://github.com/uber/deck.gl/blob/master/modules/layers/src/scatterplot-layer/scatterplot-layer.js#L51) to see how to deprecate `getColor`

modules/layers/src/column-layer/column-layer.js
```diff
@@ -91,52 +102,75 @@ export default class ColumnLayer extends Layer {
...
       return;
     }
-
     const {diskResolution} = this.props;
+    // const {filledModel, strokeModel} = this.state;
     log.assert(vertices.length >= diskResolution);
 
-    const {model} = this.state;
-    const geometry = this.getGeometry(this.props.diskResolution);
+    const geometry = this.getGeometry(diskResolution);
```
xintongxia: `geometry` is not used.

modules/layers/src/column-layer/column-layer.js
```diff
@@ -91,52 +102,75 @@ export default class ColumnLayer extends Layer {
...
+
+    let filledModel;
+    let strokeModel;
+
+    if (filled) {
+      filledModel = new Model(
+        gl,
+        Object.assign({}, this.getShaders(), {
+          id: `${id}-top`,
+          geometry: this.getGeometry(diskResolution, drawLines.filled),
```
xintongxia: not necessary to declare an extra variable `drawLines`. use a string `filled` is sufficient and more readable.

modules/layers/src/column-layer/column-layer.js
```diff
@@ -148,24 +182,63 @@ export default class ColumnLayer extends Layer {
         i++;
       }
     }
-    model.setProps({geometry});
+
+    const geometries = Object.keys(drawLines).map(mode => {
```
xintongxia: It is not going to have a lot of modes, since we only have `stroke` and `fill` right now, it is better to explicitly `stroke*` and `fill*`.

modules/layers/src/column-layer/column-layer.js
```diff
@@ -148,24 +182,63 @@ export default class ColumnLayer extends Layer {
...
+      }
+    });
+    // const strokeGeometry = this.getGeometry(diskResolution, drawLines.stroke);
+    // const filledGeometry = this.getGeometry(diskResolution, drawLines.filled);
+    // if (filledModel) {
+    //   filledModel.setProps({filledGeometry});
+    // }
+    // if (strokeModel) {
+    //   strokeModel.setProps({strokeGeometry});
+    // }
```
xintongxia: clean up if not used

modules/layers/src/column-layer/column-layer-vertex.glsl.js
```diff
@@ -74,10 +78,14 @@ void main(void) {
   vec3 normals_commonspace = project_normal(vec3(rotationMatrix * normals.xy, normals.z));
 
   if (extruded) {
-    vec3 lightColor = lighting_getLightColor(instanceColors.rgb, project_uCameraPosition, position_commonspace.xyz, normals_commonspace);
-    vColor = vec4(lightColor, instanceColors.a * opacity) / 255.0;
+    if (isWireframe) {
```
Pessimistress: Use `if (extruded && !isWireFrame)` to save a switch case.

modules/layers/src/column-layer/column-layer.js
```diff
@@ -91,52 +102,75 @@ export default class ColumnLayer extends Layer {
...
     }
-
     if (props.vertices !== oldProps.vertices) {
       this._updateVertices(props.vertices);
     }
   }
 
-  getGeometry(diskResolution) {
-    return new CylinderGeometry({
+  getGeometry(diskResolution, outline = false) {
```
Pessimistress: Avoid using boolean arguments. It is usually not great for readability.
You can do something like:

```js
const FILL_MODE = 1;
const OUTLINE_MODE = 2;

getGeometry(diskResolution, mode = FILL_MODE) {
  ...
}
```

modules/layers/src/column-layer/column-layer.js
xintongxia: @kaushikb11 

this PR is for enabling `wireframe` feature for `ColumnLayer`. So the conditional drawing here should be 

```

if (filled) {
  fillModel = ...
}

if (wireframe) {
  wireframeModel = ...
}


```

modules/layers/src/column-layer/column-geometry.js
```diff
@@ -0,0 +1,133 @@
...
+      attributes: {...attributes, ...props.attributes}
+    });
+  }
+}
+
+// Primitives inspired by TDL http://code.google.com/p/webglsamples/,
+// copyright 2011 Google Inc. new BSD License
+// (http://www.opensource.org/licenses/bsd-license.php).
+/* eslint-disable max-statements, complexity */
+function tesselateTruncatedCone(props) {
```
xintongxia: This is good. But we can have a followup PR to simplify the geometry. `ColumnGeometry` is a cylinder of which has the same size plane along vertical axis, so it is not necessary to have a vertical segmentation (`nvertical` is not needed).

kaushikb11: Sounds good. Noted.

modules/layers/src/column-layer/column-layer.js
```diff
@@ -39,11 +42,13 @@ const defaultProps = {
...
-
+  wireframe: false,
+  filled: true,
   getPosition: {type: 'accessor', value: x => x.position},
-  getColor: {type: 'accessor', value: DEFAULT_COLOR},
+  getFillColor: {type: 'accessor', value: DEFAULT_COLOR},
+  getLineColor: {type: 'accessor', value: DEFAULT_COLOR},
   getElevation: {type: 'accessor', value: 1000},
-
+  getColor: {deprecatedFor: ['getFillColor', 'getLineColor']},
```
xintongxia: move this prop to the last prop.

modules/layers/src/column-layer/column-layer.js
```diff
@@ -89,55 +101,85 @@ export default class ColumnLayer extends Layer {
...
     const {diskResolution} = this.props;
     log.assert(vertices.length >= diskResolution);
+    const {filledModel, strokeModel} = this.state;
 
-    const {model} = this.state;
-    const geometry = this.getGeometry(this.props.diskResolution);
-    const positions = geometry.attributes.POSITION;
+    const strokeGeometry = this.getGeometry(diskResolution, OUTLINE_MODE);
+    const filledGeometry = this.getGeometry(diskResolution, FILL_MODE);
+    const positions = filledGeometry.attributes.POSITION;
```
xintongxia: Should also update `strokeGeometry` positions

modules/layers/src/column-layer/column-layer.js
xintongxia: @kaushikb11 

`drawMode` and `isWireframe` could be set in `_getModels`

modules/layers/src/column-layer/column-layer.js
```diff
@@ -89,83 +101,147 @@ export default class ColumnLayer extends Layer {
...
-    const positions = geometry.attributes.POSITION;
-    let i = 0;
-    for (let loopIndex = 0; loopIndex < 3; loopIndex++) {
-      for (let j = 0; j <= diskResolution; j++) {
-        const p = vertices[j] || vertices[0]; // auto close loop
-        // replace x and y in geometry
-        positions.value[i++] = p[0];
-        positions.value[i++] = p[1];
-        i++;
+    const strokeGeometry = this.getGeometry(diskResolution, OUTLINE_MODE, GL.LINES);
```
xintongxia: update `fillGeometry` positions should happen when `fillModel` exists.

Let's be consistent, use `fill` instead of `filled`, use `stroke` instead of `outline`, `OUTLINE_MODE` -> `STROKE_MODE`

modules/layers/src/column-layer/column-layer.js
```diff
@@ -20,14 +20,17 @@
...
+import ColumnGeometry from './column-geometry';
 const {fp64LowPart} = fp64;
 const defaultMaterial = new PhongMaterial();
 
 import vs from './column-layer-vertex.glsl';
 import fs from './column-layer-fragment.glsl';
 
 const DEFAULT_COLOR = [255, 0, 255, 255];
+const FILL_MODE = 'filled';
+const OUTLINE_MODE = 'stroke';
```
xintongxia: Move these two constant modes to `ColumnGeometry` and import from there.

Also call it `STROKE_MODE`

modules/layers/src/column-layer/column-layer.js
```diff
@@ -89,83 +101,147 @@ export default class ColumnLayer extends Layer {
...
     }
   }
 
-  getGeometry(diskResolution) {
-    return new CylinderGeometry({
+  getGeometry(diskResolution, mode = FILL_MODE, drawMode = GL.TRIANGLES) {
+    return new ColumnGeometry({
       radius: 1,
       topCap: false,
       bottomCap: true,
```
xintongxia: if mode is `STROKE_MODE`, shouldn't draw `bottomCap`

modules/layers/src/column-layer/column-layer.js
```diff
@@ -89,83 +101,147 @@ export default class ColumnLayer extends Layer {
...
+    const {id, filled, wireframe, diskResolution} = this.props;
+
+    let filledModel;
+    let strokeModel;
+
+    if (filled) {
+      filledModel = new Model(
+        gl,
+        Object.assign({}, this.getShaders(), {
+          id: `${id}-top`,
```
xintongxia:
```
id: `${id}-${FILL_MODE}`,
```

modules/layers/src/column-layer/column-layer.js
```diff
@@ -89,83 +101,147 @@ export default class ColumnLayer extends Layer {
...
+          isInstanced: true,
+          shaderCache: this.context.shaderCache
+        })
+      );
+    }
+    if (wireframe) {
+      strokeModel = new Model(
+        gl,
+        Object.assign({}, this.getShaders(), {
+          id: `${id}-side`,
```
xintongxia: ```id: `${id}-${STROKE_MODE}`,```

docs/layers/column-layer.md
```diff
@@ -172,6 +211,12 @@ The elevation of each cell in meters.
 * If a number is provided, it is used as the elevation for all objects.
 * If a function is provided, it is called on each object to retrieve its elevation.
 
+## Sub Layers
+
+The PolygonLayer renders the following sublayers:
```
xintongxia: Remove this. There is no `Sub Layers` in `ColumnLayer`. `Sub Layers` means deck.gl layers.

modules/layers/src/column-layer/column-layer-vertex.glsl.js
```diff
@@ -73,11 +77,11 @@ void main(void) {
 
   vec3 normals_commonspace = project_normal(vec3(rotationMatrix * normals.xy, normals.z));
 
-  if (extruded) {
-    vec3 lightColor = lighting_getLightColor(instanceColors.rgb, project_uCameraPosition, position_commonspace.xyz, normals_commonspace);
-    vColor = vec4(lightColor, instanceColors.a * opacity) / 255.0;
+  if (extruded && !isWireframe) {
```
xintongxia: Why you need this `&& !isWireframe`?

modules/layers/src/column-layer/column-layer.js
```diff
@@ -105,69 +119,122 @@ export default class ColumnLayer extends Layer {
...
-    const positions = geometry.attributes.POSITION;
-    let i = 0;
-    for (let loopIndex = 0; loopIndex < 3; loopIndex++) {
-      for (let j = 0; j <= diskResolution; j++) {
-        const p = vertices[j] || vertices[0]; // auto close loop
-        // replace x and y in geometry
-        positions.value[i++] = p[0];
-        positions.value[i++] = p[1];
-        i++;
+    const fillGeometry = this.getGeometry(diskResolution, FILL_MODE);
```
xintongxia: No need to create a geometry if the model is empty.

modules/layers/src/column-layer/column-layer-vertex.glsl.js
kaushikb11: @xintongxia Lighting should be disabled on the wireframe

Pessimistress:
@xintongxia 

> Why you need this `&& !isWireframe`?

Wireframe mode should not use lighting.

modules/layers/src/column-layer/column-layer.js
```diff
@@ -89,83 +101,147 @@ export default class ColumnLayer extends Layer {
 
   updateState({props, oldProps, changeFlags}) {
     super.updateState({props, oldProps, changeFlags});
-    if (props.fp64 !== oldProps.fp64 || props.diskResolution !== oldProps.diskResolution) {
+
+    const regenerateModels =
```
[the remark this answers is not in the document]
kaushikb11: Should I also check for `filled` prop update for regenerating?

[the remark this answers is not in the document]
kaushikb11: Should I also check for an update of `extruded` prop?

modules/layers/src/column-layer/column-layer.js
```diff
@@ -105,69 +120,121 @@ export default class ColumnLayer extends Layer {
...
+            positions.value[i++] = p[0];
+            positions.value[i++] = p[1];
+            i++;
+          }
+        }
+        model.setProps({geometry});
       }
-    }
-    model.setProps({geometry});
+    });
```
xintongxia: I would prefer to explicitly say `fillModel`, and `strokeModel`, which is more clear. 

```
const {fillModel, strokeModel} = this.state;

if (fillModel) {
  // get the geometry
  // update vertices
}

if (strokeModel) {
  // get the geometry
  // update vertices
}
```

`update vertices` could be extracted into a function.

modules/layers/src/column-layer/column-layer.js
```diff
@@ -89,14 +99,19 @@ export default class ColumnLayer extends Layer {
 
   updateState({props, oldProps, changeFlags}) {
     super.updateState({props, oldProps, changeFlags});
+
     const regenerateModels =
-      props.fp64 !== oldProps.fp64 || props.diskResolution !== oldProps.diskResolution;
+      props.fp64 !== oldProps.fp64 ||
+      props.diskResolution !== oldProps.diskResolution ||
+      props.filled !== oldProps.filled ||
+      props.wireframe !== oldProps.wireframe;
```
xintongxia: After a second thought, based on the documentation, ```Requires the `extruded` prop to be true.``` the wireframe mode should only be enabled when both `wireframe` and `extruded` set to true.

modules/layers/src/column-layer/column-layer.js
```diff
@@ -105,69 +120,121 @@ export default class ColumnLayer extends Layer {
...
+        Object.assign({}, this.getShaders(), {
+          id: `${id}-${FILL_MODE}`,
+          uniforms: {isWireframe: false},
+          geometry: this.getGeometry(diskResolution, FILL_MODE),
+          isInstanced: true,
+          shaderCache: this.context.shaderCache
+        })
+      );
+    }
+    if (wireframe) {
```
xintongxia: ```if (wireframe && extruded)```

modules/layers/src/column-layer/column-geometry.js
```diff
@@ -0,0 +1,136 @@
+import {Geometry, uid} from '@luma.gl/core';
+
+const INDEX_OFFSETS = {
+  x: [2, 0, 1],
+  y: [0, 1, 2],
+  z: [1, 2, 0]
+};
+
+export const FILL_MODE = 'fill';
+export const STROKE_MODE = 'stroke';
```
xintongxia: I am thinking to it is more accurate to call it `WIREFRAME_MODE`. Also WireframeModel, WireframeGeometry...

modules/layers/src/column-layer/column-layer.js
kaushikb11: Cool, so I should also add `extruded` prop to updateState.

modules/layers/src/column-layer/column-geometry.js
kaushikb11: `STROKE_MODE` to `WIREFRAME_MODE`?

modules/layers/src/column-layer/column-layer.js
Pessimistress: We need stroke even without extrusion.

modules/layers/src/column-layer/column-layer.js
Pessimistress: The documentation does not sound accurate to me. It was copied from the PolygonLayer which has stroke mode (path with width) when flat and wireframe mode (1-pixel line) when extruded. If you plan to implement an equivalent stroke mode, cool. If not, I prefer to have some sort of outline when flat over nothing.

modules/layers/src/column-layer/column-layer.js
```diff
@@ -105,69 +120,121 @@ export default class ColumnLayer extends Layer {
...
+    const numInstances = this.getNumInstances();
+
+    models.forEach(model => {
+      if (model) {
+        model.setInstanceCount(numInstances);
+        model.setUniforms(renderUniforms);
+      }
+    });
+
+    if (strokeModel && wireframe) {
```
Pessimistress: `strokeModel` should always be available if `wireframe` is `true`?

modules/layers/src/column-layer/column-layer.js
```diff
@@ -105,69 +120,121 @@ export default class ColumnLayer extends Layer {
...
+      angle: (angle / 180) * Math.PI,
+      offset,
+      extruded,
+      coverage,
+      elevationScale
+    });
+    const numInstances = this.getNumInstances();
+
+    models.forEach(model => {
+      if (model) {
```
Pessimistress: `model` is always truthy

modules/layers/src/column-layer/column-layer.js
```diff
@@ -105,69 +120,121 @@ export default class ColumnLayer extends Layer {
...
+      radius,
+      angle: (angle / 180) * Math.PI,
+      offset,
+      extruded,
+      coverage,
+      elevationScale
+    });
+    const numInstances = this.getNumInstances();
+
+    models.forEach(model => {
```
Pessimistress: use for loop instead of `forEach` for perf.

modules/layers/src/column-layer/column-layer.js
```diff
@@ -105,69 +120,121 @@ export default class ColumnLayer extends Layer {
...
-    for (let loopIndex = 0; loopIndex < 3; loopIndex++) {
-      for (let j = 0; j <= diskResolution; j++) {
-        const p = vertices[j] || vertices[0]; // auto close loop
-        // replace x and y in geometry
-        positions.value[i++] = p[0];
-        positions.value[i++] = p[1];
-        i++;
+    const modes = [FILL_MODE, STROKE_MODE];
+
+    modes.forEach((mode, index) => {
```
Pessimistress: Use for loop

modules/layers/src/column-layer/column-layer.js
kaushikb11: It could be `undefined` as well, right?

Pessimistress: As far as I can tell, no.

```
models: [fillModel, strokeModel].filter(Boolean),
```

docs/layers/column-layer.md
```diff
@@ -133,6 +140,16 @@ to scale all hexagon elevations without updating the data.
...
 
+##### `wireframe` (Boolean, optional)
+
+* Default: `false`
+
+Whether to generate a line wireframe of the column. The outline will have
+"horizontal" lines closing the top and bottom polygons and a vertical line
+(a "strut") for each vertex on the polygon.
+
+Requires the `extruded` prop to be true.
```
xintongxia: It is supposed to be working with non extruded mode?

kaushikb11: Not until the stroke is being implemented. I am doing a follow-up PR for the update in H3Hexagon docs, will edit this as well in the PR.

PR #3054, opened 2019-05-01
GPUAggregator: Fix Texture/Buffer resource leaks.

1chandu:
For #3039 

#### Background
GPUAggregator: Couple of resources are not properly deleted, collect and delete these resources.

#### Change List
- GPUAggregator: Fix Buffer/Texture leaks

### DIFF
--- a/modules/aggregation-layers/src/utils/gpu-grid-aggregation/gpu-grid-aggregator.js
+++ b/modules/aggregation-layers/src/utils/gpu-grid-aggregation/gpu-grid-aggregator.js
@@ -36,11 +36,7 @@ import AGGREGATE_TO_GRID_FS from './aggregate-to-grid-fs.glsl';
 import AGGREGATE_ALL_VS_FP64 from './aggregate-all-vs-64.glsl';
 import AGGREGATE_ALL_FS from './aggregate-all-fs.glsl';
 import TRANSFORM_MEAN_VS from './transform-mean-vs.glsl';
-import {
-  getFloatTexture,
-  getFramebuffer,
-  getFloatArray,
-} from './gpu-grid-aggregator-utils.js';
+import {getFloatTexture, getFramebuffer, getFloatArray} from './gpu-grid-aggregator-utils.js';
 
 export default class GPUGridAggregator {
   // Decode and return aggregation data of given pixel.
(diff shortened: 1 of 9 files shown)


modules/aggregation-layers/src/utils/gpu-grid-aggregation/gpu-grid-aggregator.js
```diff
@@ -610,12 +639,7 @@ export default class GPUGridAggregator {
       isInstanced: true,
       instanceCount: 0,
       attributes: {
-        position: new Buffer(gl, {
-          data: new Float32Array([0, 0]),
-          accessor: {
-            size: 2
-          }
-        })
+        position: this.state.resources['all-aggregation-buffer']
```
Pessimistress: You should be able to use `{constant: true, value: [0, 0]}`

modules/aggregation-layers/src/utils/gpu-grid-aggregation/gpu-grid-aggregator.js
```diff
@@ -152,11 +151,15 @@ export default class GPUGridAggregator {
 
     positionsBuffer && positionsBuffer.delete();
     position64Buffer && position64Buffer.delete();
-    this.deleteResources(framebuffers);
-    this.deleteResources(maxMinFramebuffers);
-    this.deleteResources(minFramebuffers);
-    this.deleteResources(maxFramebuffers);
-    this.deleteResources(meanTextures);
+    this.deleteResources([
+      framebuffers,
```
Pessimistress: Can we track some of these with `resources`?

Pessimistress:
Can you upgrade luma.gl to 7.0.1 which includes https://github.com/uber/luma.gl/pull/1087

modules/aggregation-layers/src/utils/gpu-grid-aggregation/gpu-grid-aggregator.js
1chandu: Yep, changed to [0, 0]

modules/aggregation-layers/src/utils/gpu-grid-aggregation/gpu-grid-aggregator.js
1chandu: Some of these are used in several places, so keeping separate objects make it much easy to read and less error prone.

1chandu:
> Can you upgrade luma.gl to 7.0.1 which includes [uber/luma.gl#1087](https://github.com/uber/luma.gl/pull/1087)

@Pessimistress Upgraded to 7.0.1, no new issues found in tests.

PR #3075, opened 2019-05-07
[React] Add the ability to use react-map-gl controls

Pessimistress:
#### Background

All react-map-gl control components require a React context provider rendered by the MapGL component. Currently, if we render these components outside of `StaticMap`, they crash; if we render them as children of `StaticMap`, they are not interactive, because they are stuck behind the deck.gl canvas.

This PR proposes a new prop `ContextProvider` for the `DeckGL` React component. If provided, deck will supply the React context to all child components.

#### Change List
- Add `ContextProvider` prop
- Add `NavigationControl` to layer-browser

Screen Shot 2019-05-07 at 11 52 26 AM

### DIFF
--- a/docs/api-reference/react/deckgl.md
+++ b/docs/api-reference/react/deckgl.md
@@ -57,6 +57,18 @@ A [Context.Provider](https://reactjs.org/docs/context.html#contextprovider) comp
 - `container` (DOMElement) - the DOM element containing the deck canvas
 - `eventManager` ([EventManager](https://github.com/uber-web/mjolnir.js/blob/master/docs/api-reference/event-manager.md))
 
+```jsx
+/// Example using react-map-gl controls with deck.gl
+import DeckGL from '@deck.gl/react';
+import {_MapContext as MapContext, NavigationControl} from 'react-map-gl';
+
+<DeckGL ... ContextProvider={MapContext.Provider}>
+  <div style={NAVIGATION_CONTROL_STYLES}>
+    <NavigationControl onViewStateChange={...} />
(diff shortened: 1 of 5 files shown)


docs/api-reference/react/deckgl.md
```diff
@@ -47,6 +47,15 @@ const App = (data) => (
 
 `DeckGL` accepts all [Deck](/docs/api-reference/deck.md#properties) properties, with these additional semantics:
 
+### React Context
+
+##### `ContextProvider` (React.Component, optional)
```
xintongxia: Do you think should mention react-map-gl MapContext here?

modules/react/src/utils/position-children-under-views.js
```diff
@@ -50,8 +52,26 @@ export default function positionChildrenUnderViews({children, viewports, deck})
       viewState
     });
 
-    const style = {position: 'absolute', left: x, top: y, width, height};
+    const style = {
+      position: 'absolute',
+      zIndex: childStyle && childStyle.zIndex,
+      pointerEvents: 'none',
```
xintongxia: Why you need set `pointerEvents` to `none`?

Pessimistress: If this container is on top of the map it will block pointer input to the deck canvas.

docs/api-reference/react/deckgl.md
Pessimistress: I can put an example snippet here, but this feature is not limited to using react-map-gl.

PR #3085, opened 2019-05-10
Ensure UMD bundles are ES5

bclinkinbeard:
For #3076

I tried to verify this but wasn't able to. Running `npm run build` doesn't seem to create the dist bundle.

### DIFF
--- a/scripts/bundle.config.js
+++ b/scripts/bundle.config.js
@@ -68,7 +68,12 @@ const config = {
         // Compile ES2015 using babel
         test: /\.js$/,
         loader: 'babel-loader',
-        include: /src/
+        include: /src/,
+        options: {
+          presets: [
+            ['@babel/preset-env', {forceAllTransforms: true}]
+          ]
+        }
       }
     ]
   },


ibgreen:
> Running npm run build doesn't seem to create the dist bundle

`npm run bootstrap` should create them.

Pessimistress:
> npm run bootstrap should create them

No, we avoid this to save dev time. You can run `npx lerna run prepublishOnly`

Pessimistress:
Is this going to break all of our Codepen and Observable examples with custom layers?

bclinkinbeard:
Only if they rely on the UMD build being ES6. This is (or at least should
be) separate from the CommonJS bundles.

Maybe more descriptive names would be good?

Pessimistress:
The pre-bundled version was technically never intended to be used in production. For example, it does not include the `@babel/runtime` polyfill. The target use case is prototyping/sandbox environments such as Codepen. If you are already transpiling/bundling your own code, why not create your own vendor bundle as well?

bclinkinbeard:
> The target use case is prototyping/sandbox environments such as Codepen.
Unless those environments need the bundle to be native ES6 for some reason that should continue to work after this change.

> why not create your own vendor bundle as well?
That's unfortunately not an option right now. I'm using deck in a prototype/sandbox environment as well, and the existing build infrastructure outputs ES5 which means I cannot extend native ES6 classes from the bundle. Both ES5 and ES6 users should be able to use a ES5 bundle, but a ES6 bundle limits what ES5 users can do.

This may not be a super common scenario, but if you're not comfortable modifying the existing bundle can we add a ES5 bundle?

ibgreen:
Ideally, fully transpiling the bundle to ES5 would not break existing examples... ES6 code examples should still be able to import and use the fully transpiled ES5 code... What potential issues do we see?

Pessimistress:
> What potential issues do we see?

The prototyping environments do not transpile user code. You will no longer be able to do things like

```js
class CustomLayer extends deck.CompositeLayer {
  ...
}
```

Because an ES6 class cannot extend an ES5 one (`Class extends value undefined is not a constructor or null` error).

bclinkinbeard:
Just curious, do you know when you expect to merge this change and publish an updated bundle?

PR #3091, opened 2019-05-13
Prep Attribute class for partial update

Pessimistress:
For #1576

#### Change List
- Avoid calculating buffer props unless needed
- Only upload changed subarray on attribute update

### DIFF
--- a/modules/core/src/lib/attribute.js
+++ b/modules/core/src/lib/attribute.js
@@ -171,6 +171,11 @@ export default class Attribute extends BaseAttribute {
 
       this.constant = false;
       this.value = new ArrayType(this.size * allocCount);
+
+      if (this.buffer && this.buffer.byteLength < this.value.byteLength) {
+        this.buffer.reallocate(this.value.byteLength);
+      }
+
       state.needsUpdate = true;
       state.allocedInstances = allocCount;
       return true;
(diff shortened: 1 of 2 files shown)


modules/core/src/lib/attribute.js
```diff
@@ -170,6 +170,11 @@ export default class Attribute extends BaseAttribute {
 
       this.constant = false;
       this.value = new ArrayType(this.size * allocCount);
+
+      if (this.buffer && this.buffer.byteLength < this.value.byteLength) {
```
Pessimistress: @ibgreen @tsherif This is the part I'm hoping to get some ideas - in the incremental update use case (e.g. data loaded by chunks), every time we reallocate the buffer, we need to re-upload the entire value array.

tsherif: Is it possible to know ahead of time in that incremental case what the total size of the data will be?

Pessimistress: Generally, no.

Alternatively, every time overflow happens, we could over-allocate the buffer instead of setting it to the exact new size.

tsherif: That makes sense to me. Something like doubling the size of the buffer on each overflow.

Pessimistress: How sensitive should we be regarding GPU memory usage?

tsherif: It could be a problem on mobile, but I think we have some leeway because we don't use textures much. It would be helpful to profile some typical use cases and see what our current memory usage is like.

PR #3095, opened 2019-05-14
GPU GridLayer PART-3: Add NewGridLayer.

1chandu:
For #2084  

#### Background

- Adding `NewGridLayer`, a composite layer that does some prop conversion and either renders `GridLayer` (cpu aggregation) or `GPUGridLayer`.
    - After all testing, `GridLayer` becomes `CPUGridLayer` and `NewGridLayer` becomes `GridLayer`.
- GPUGridCellLayer
    - Fix lighting.
    - Match elevation calculations with `ColumnLayer`.
- Tested `NewGridLayer`'s CPU and GPU paths with existing `GridLayer` golden image and in `layer-browser` example.
- Testing and docs will be added in future PRs , tracked in #2084 

#### Change List
- Add NewGridLayer, fixes to `gpu-grid-cell-layer`.


modules/aggregation-layers/src/grid-layer/grid-layer.js
```diff
@@ -349,3 +351,4 @@ export default class GridLayer extends CompositeLayer {
 
 GridLayer.layerName = 'GridLayer';
 GridLayer.defaultProps = defaultProps;
+GridLayer.DEFAULT_GETVALUE = DEFAULT_GETVALUE;
```
Pessimistress: Why is this needed?

modules/aggregation-layers/src/new-grid-layer/new-grid-layer.js
```diff
@@ -0,0 +1,155 @@
...
+import GPUGridLayer from '../gpu-grid-layer/gpu-grid-layer';
+import GridLayer from '../grid-layer/grid-layer';
+
+const defaultProps = Object.assign({}, GPUGridLayer.defaultProps, GridLayer.defaultProps);
+
+// Function to convert from getWeight/aggregation props to getValue prop for Color and Elevation
+function getMean(pts, accessor) {
+  const filtered = pts.map(item => accessor(item)).filter(pt => Number.isFinite(pt));
+
+  return filtered.length ? filtered.reduce((accu, curr) => accu + curr, 0) / filtered.length : null;
```
Pessimistress: Perf: avoid defining inline function in frequently used utility functions.

```js
const filtered = pts.map(accessor).filter(Number);
filtered.reduce(sumReducer, 0);
```

modules/aggregation-layers/src/new-grid-layer/new-grid-layer.js
```diff
@@ -0,0 +1,155 @@
...
+    this.setState(newState);
+  }
+
+  renderLayers() {
+    let gridLayer = null;
+
+    if (this.state.useGPUAggregation) {
+      gridLayer = new GPUGridLayer(
+        this.props,
+        this.getSubLayerProps({
```
Pessimistress: Lacking support for update triggers?

modules/aggregation-layers/src/new-grid-layer/new-grid-layer.js
```diff
@@ -0,0 +1,155 @@
...
+        this.props,
+        buildColorElevationProps,
+        this.getSubLayerProps({
+          id: 'CPU',
+          // Note: data has to be passed explicitly like this to avoid being empty
+          data: this.props.data
+        })
+      );
+    }
+    return [gridLayer];
```
Pessimistress: You can return a single layer without an array.

Pessimistress:
Add a basic unit test like [this](https://github.com/uber/deck.gl/blob/master/test/modules/layers/polygon-layer.spec.js)

modules/aggregation-layers/src/grid-layer/grid-layer.js
1chandu: `getColorValue` and `getElevationValue` are function props in the current API. When user provides values for these, I can not map them to new API (like, `getColorValue` to `getColorWeight` and `colorAggregation`), hence I have to fallback to CPU Aggregation. By defining a constant, `DEFAULT_GETVALUE` for default value, I can check if the user provided a custom prop or we are using a default value. For the case of default value, (which is `points => points.length`) I am converting them to new props.

Pessimistress: Maybe we can just use `GridLayer.defaultProps. getColorValue.value`?

modules/aggregation-layers/src/new-grid-layer/new-grid-layer.js
1chandu: Nothing happening in this layer, no attribute calculations or aggregation. It basically checking if aggregation can happen on GPU or not. `GridLayer` and `GPUGridLayer` have update triggers setup.

Pessimistress: You still need to pass the update triggers to the sublayers. Here the update triggers in `this.props` is overwritten by `getSubLayerProps`.

modules/aggregation-layers/src/grid-layer/grid-layer.js
1chandu: Yeah, that works, updated.

modules/aggregation-layers/src/new-grid-layer/new-grid-layer.js
1chandu: Addressed.

modules/aggregation-layers/src/new-grid-layer/new-grid-layer.js
1chandu: Done.

modules/aggregation-layers/src/new-grid-layer/new-grid-layer.js
1chandu: Done, also update the document page for `getSublayerProps` , removed `Optional` for `updateTriggers` argument. Seems like we always need to pass this to avoid overriding.

modules/aggregation-layers/src/gpu-grid-layer/gpu-grid-cell-layer-vertex.glsl.js
```diff
@@ -89,45 +87,44 @@ float linearScale(vec2 domain, vec2 range, float value) {
...
-
   vec2 colorDomain = vec2(colorData.maxMinCount.a, colorData.maxMinCount.r);
   vec4 color = quantizeScale(colorDomain, colorRange, colors.r);
 
-  // TODO: discard when noRender is true
-  float finalCellSize = noRender ? 0.0 : project_size(cellSize);
-
   float elevation = 0.0;
 
   if (extruded > 0.5) {
```
Pessimistress: Use `uniform bool extruded`

modules/aggregation-layers/src/gpu-grid-layer/gpu-grid-layer.js
```diff
@@ -187,7 +187,7 @@ export default class GPUGridLayer extends CompositeLayer {
 
     // return props to the sublayer constructor
     return super.getSubLayerProps({
-      id: 'grid-cell',
+      id: 'gpu-grid-cell',
```
Pessimistress: Instead of overriding the `getSubLayerProps()` method, pass the props directly to `SubLayerClass`:

```js
return new SubLayerClass(
  defaultSubLayerProps,
  this.getSubLayerProps({id, updateTriggers}),
  nonOverridableProps
);
```

It is 1) more performant and 2) allows for user override with the [_subLayerProps prop](https://deck.gl/#/documentation/deckgl-api-reference/layers/composite-layer?section=_sublayerprops-object-experimental).

See https://github.com/uber/deck.gl/blob/master/modules/aggregation-layers/src/hexagon-layer/hexagon-layer.js#L356

modules/aggregation-layers/src/gpu-grid-layer/gpu-grid-cell-layer-vertex.glsl.js
1chandu: updated.

modules/aggregation-layers/src/gpu-grid-layer/gpu-grid-layer.js
1chandu: updated.

PR #3103, opened 2019-05-16
Improve H3HexagonLayer perf

Pessimistress:
#### Background

The current implementation of `H3HexagonLayer` updates the instanced geometry (regenerating attributes and creating new WebGL buffers) whenever the hexagon at the center of the viewport changes. This is excessive especially at fine resolutions.

This PR introduces a distance threshold within which we consider all hexagons having the same shape. This significantly reduces the number of attribute updates needed when interacting with the viewport.

#### Change List
- Use the distance between the current center hex and the last center hex to determine whether geometry update is needed

### DIFF
--- a/modules/geo-layers/src/h3-layers/h3-hexagon-layer.js
+++ b/modules/geo-layers/src/h3-layers/h3-hexagon-layer.js
@@ -1,6 +1,20 @@
-import {h3ToGeoBoundary, h3GetResolution, h3ToGeo, geoToH3, h3IsPentagon} from 'h3-js';
+import {
+  h3ToGeoBoundary,
+  h3GetResolution,
+  h3ToGeo,
+  geoToH3,
+  h3IsPentagon,
+  h3Distance,
+  edgeLength,
+  UNITS
+} from 'h3-js';
 import {CompositeLayer, createIterable} from '@deck.gl/core';
 import {ColumnLayer, PolygonLayer} from '@deck.gl/layers';
+
+// There is a cost to updating the instanced geometries when using highPrecision: false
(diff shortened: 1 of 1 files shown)


modules/geo-layers/src/h3-layers/h3-hexagon-layer.js
```diff
@@ -1,7 +1,21 @@
...
+  edgeLength,
+  UNITS
+} from 'h3-js';
 import {CompositeLayer, createIterable} from '@deck.gl/core';
 import {ColumnLayer, PolygonLayer} from '@deck.gl/layers';
 
+// There is a cost to updating the instanced geometries when using highPrecision: false
+// This constant defines the distance between two hexagons that leads to "significant
+// distortion." Smaller value makes the column layer more sensitive to viewport change.
+const UPDATE_THRESHOLD = 10000; // 10km
```
nrabinowitz: Nit: When using unit constants, it's usually good to have the unit in the name. In this case, I'd use `km` instead of `m`, and go with
```
const UPDATE_THRESHOLD_KM = 10;
```
Then you don't need the comment and the unit info is available at the point of use.

modules/geo-layers/src/h3-layers/h3-hexagon-layer.js
```diff
@@ -54,6 +68,7 @@ export default class H3HexagonLayer extends CompositeLayer {
       }
       this.setState({
         resolution,
+        meanEdgeLength: resolution >= 0 ? edgeLength(resolution, UNITS.m) : 0,
```
nrabinowitz: If `km` is the better granularity, you can use `UNITS.km` here.

PR #3107, opened 2019-05-16
post-processing effect rendering pipeline

jianhuang01:
For #2962 

#### Change List
- new PostProcessEffect and ScreenPass classes
- new PostProcessEffect rendering logic in DeckRenderer

### DIFF
--- a/modules/core/src/effects/post-process-effect.js
+++ b/modules/core/src/effects/post-process-effect.js
@@ -36,11 +36,12 @@ export default class PostProcessEffect extends Effect {
     };
   }
 
-  finalize() {
+  cleanup() {
     if (this.passes) {
       for (const pass of this.passes) {
-        pass.finalize();
+        pass.cleanup();
       }
+      this.passes = null;
     }
   }
 }
--- a/modules/core/src/effects/post-process-effect.js
+++ b/modules/core/src/effects/post-process-effect.js
@@ -18,7 +18,7 @@ export default class PostProcessEffect extends Effect {
     }
   }
 
-  render(gl, params) {
+  render(params) {
     const {target = null} = params;
     let switchBuffer = false;
     for (let index = 0; index < this.passes.length; index++) {
@@ -54,13 +54,10 @@ function createPasses(gl, module, id, props) {
 
     return new ScreenPass(
       gl,
-      Object.assign(
-        {
-          id: idn,
-          model: getModel(gl, module, fs, idn, props)
-        },
-        props
-      )
+      Object.assign({
+        id: idn,
+        model: getModel(gl, module, fs, idn, props)
+      })
     );
   });
 }
--- a/modules/core/src/effects/post-process-effect.js
+++ b/modules/core/src/effects/post-process-effect.js
@@ -14,7 +14,7 @@ export default class PostProcessEffect extends Effect {
 
   prepare(gl) {
     if (!this.passes) {
-      this.passes = normalizePasses(gl, this.module, this.id, this.props);
+      this.passes = createPasses(gl, this.module, this.id, this.props);
     }
   }
 
@@ -37,29 +37,27 @@ export default class PostProcessEffect extends Effect {
   }
 }
 
-function normalizePasses(gl, module, id, props) {
+function createPasses(gl, module, id, props) {
   if (module.filter || module.sampler) {
     const fs = getFragmentShaderForRenderPass(module);
     const pass = new ScreenPass(gl, {
       id,
-      model: getModel(gl, module, fs, id, props),
-      uniforms: null
+      model: getModel(gl, module, fs, id, props)
     });
     return [pass];
   }
 
   const passes = module.passes || [];
-  return passes.map(pass => {
+  return passes.map((pass, index) => {
     const fs = getFragmentShaderForRenderPass(module, pass);
-    const idn = `${id}-${passes.length + 1}`;
+    const idn = `${id}-${index}`;
 
     return new ScreenPass(
       gl,
       Object.assign(
         {
           id: idn,
-          model: getModel(gl, module, fs, idn, props),
-          uniforms: pass.uniforms
+          model: getModel(gl, module, fs, idn, props)
         },
         props
       )
--- a/modules/core/src/effects/post-process-effect.js
+++ b/modules/core/src/effects/post-process-effect.js
@@ -2,7 +2,6 @@ import Effect from '../lib/effect';
 import ScreenPass from '../passes/screen-pass';
 /* eslint-disable import/no-extraneous-dependencies */
 import {normalizeShaderModule} from '@luma.gl/shadertools';
-import {ClipSpace} from '@luma.gl/core';
 
 export default class PostProcessEffect extends Effect {
   constructor(module, props = {}) {
@@ -39,19 +38,21 @@ export default class PostProcessEffect extends Effect {
   cleanup() {
     if (this.passes) {
       for (const pass of this.passes) {
-        pass.cleanup();
+        pass.delete();
       }
       this.passes = null;
     }
   }
 }
 
-function createPasses(gl, module, id, props) {
+function createPasses(gl, module, id, moduleProps) {
   if (module.filter || module.sampler) {
     const fs = getFragmentShaderForRenderPass(module);
     const pass = new ScreenPass(gl, {
       id,
-      model: getModel(gl, module, fs, id, props)
+      module,
+      fs,
+      moduleProps
     });
     return [pass];
   }
@@ -61,23 +62,13 @@ function createPasses(gl, module, id, props) {
     const fs = getFragmentShaderForRenderPass(module, pass);
     const idn = `${id}-${index}`;
 
-    return new ScreenPass(
-      gl,
-      Object.assign({
-        id: idn,
-        model: getModel(gl, module, fs, idn, props)
-      })
-    );
+    return new ScreenPass(gl, {
+      id: idn,
+      module,
+      fs,
+      moduleProps
+    });
   });
-}
-
-function getModel(gl, module, fs, id, props) {
-  const model = new ClipSpace(gl, {id, fs, modules: [module]});
-
-  const uniforms = Object.assign(module.getUniforms(), module.getUniforms(props));
-
-  model.setUniforms(uniforms);
-  return model;
 }
 
 const FILTER_FS_TEMPLATE = func => `\
--- a/modules/core/src/effects/post-process-effect.js
+++ b/modules/core/src/effects/post-process-effect.js
@@ -92,7 +92,7 @@ void main() {
   vec2 texCoord = coordinate;
 
   gl_FragColor = texture2D(texture, texCoord);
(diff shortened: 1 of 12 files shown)


modules/core/src/effects/post-process-effect.js
```diff
@@ -0,0 +1,123 @@
...
+      switchBuffer = !switchBuffer;
+    }
+    return {
+      inputBuffer: switchBuffer ? params.outputBuffer : params.inputBuffer,
+      outputBuffer: switchBuffer ? params.inputBuffer : params.outputBuffer
+    };
+  }
+}
+
+function normalizePasses(gl, module, id, props) {
```
Pessimistress: Nit: `createPasses`

modules/core/src/effects/post-process-effect.js
```diff
@@ -0,0 +1,123 @@
...
+      model: getModel(gl, module, fs, id, props),
+      uniforms: null
+    });
+    return [pass];
+  }
+
+  const passes = module.passes || [];
+  return passes.map(pass => {
+    const fs = getFragmentShaderForRenderPass(module, pass);
+    const idn = `${id}-${passes.length + 1}`;
```
Pessimistress: All passes have the same id. Do you mean to use the pass index here?

modules/core/src/effects/post-process-effect.js
```diff
@@ -0,0 +1,123 @@
...
+
+    return new ScreenPass(
+      gl,
+      Object.assign(
+        {
+          id: idn,
+          model: getModel(gl, module, fs, idn, props),
+          uniforms: pass.uniforms
+        },
+        props
```
Pessimistress: Are user props allowed to overwrite `id` `model` or `uniforms`?

modules/core/src/passes/layers-pass.js
```diff
@@ -1,10 +1,13 @@
...
+import {clear, setParameters, withParameters, Framebuffer} from '@luma.gl/core';
 
 export default class LayersPass extends Pass {
-  render(props) {
-    return this.drawLayers(props);
+  render(params) {
+    const gl = this.gl;
+    const {outputBuffer = Framebuffer.getDefaultFramebuffer(gl)} = params;
+
+    return withParameters(gl, {framebuffer: outputBuffer}, () => this.drawLayers(params));
```
Pessimistress: Can we not change the render behavior when no postProcessEffects are used?

modules/core/src/effects/post-process-effect.js
```diff
@@ -0,0 +1,123 @@
...
+    const fs = getFragmentShaderForRenderPass(module, pass);
+    const idn = `${id}-${passes.length + 1}`;
+
+    return new ScreenPass(
+      gl,
+      Object.assign(
+        {
+          id: idn,
+          model: getModel(gl, module, fs, idn, props),
+          uniforms: pass.uniforms
```
Pessimistress: Do you need to manipulate these uniforms later? Just set them on the model?

Pessimistress:
What is your strategy to delete the models/buffers when they are no longer used?

modules/core/src/effects/post-process-effect.js
jianhuang01: User props only overwrites uniforms, id is for internal debugging

modules/core/src/passes/layers-pass.js
jianhuang01: when there is no PostProcessEffects, the outputBuffer is the default frame buffer(screen), the behavior is the same as the old logic, the benefit with this solution is that LayersPass don't need know about which surface for it to render.

modules/core/src/effects/post-process-effect.js
jianhuang01: not need for now, will delete, the uniforms are already set into model.

modules/core/src/effects/post-process-effect.js
jianhuang01: it's a bug, will fix it.

modules/core/src/effects/post-process-effect.js
Pessimistress: Then the argument in `Object.assign` should be ordered to reflect that.

jianhuang01:
> What is your strategy to delete the models/buffers when they are no longer used?

the models and passes will be delete when user delete their references to PostProcessEffect objects and set deck.effects = [].
The two frame buffers are created when user adds any PostProcessEffect, and will live all the time after.

Pessimistress:
All resources must be explicitly deleted by calling `.delete()`. See [past memory leak fixes](https://github.com/uber/deck.gl/pulls?utf8=%E2%9C%93&q=is%3Apr+sort%3Aupdated-desc+is%3Aclosed+memory+leak)

jianhuang01:
> All resources must be explicitly deleted by calling `.delete()`. See [past memory leak fixes](https://github.com/uber/deck.gl/pulls?utf8=%E2%9C%93&q=is%3Apr+sort%3Aupdated-desc+is%3Aclosed+memory+leak)

Done, thanks for reminding.

modules/core/src/lib/deck-renderer.js
```diff
@@ -88,15 +101,56 @@ export default class DeckRenderer {
...
+    if (this.offscreenBuffer) {
+      this.offscreenBuffer.delete();
+      this.offscreenBuffer = null;
+    }
+  }
+
   // Private
   prepareEffects(effects) {
     const effectProps = {};
+    const postProcessEffects = [];
```
Pessimistress: Perf: instead of building a new array on each render, you can just use a boolean flag here:

```js
let hasPostProcessEffects = false;
...
if (hasPostProcessEffects) {
  this.prepareRenderBuffers();
}
```

And `renderPostProcessEffects(postProcessEffects)` becomes `postRender(effects)`. This has the following advantages:

- Performing `instanceof` check again in a for loop is very cheap, while garbage collection has a cost.
- Introducing a generic lifecycle method, instead of something specific for `PostProcessEffect`, allows this class to accommodate future effect types without constantly adding new APIs.

modules/core/src/lib/effect-manager.js
```diff
@@ -36,11 +36,24 @@ export default class EffectManager {
...
+  }
+
   // Private
   setEffects(effects = []) {
+    this.cleanup();
     this.effects = effects;
   }
 
+  cleanup() {
+    if (this.effects) {
```
Pessimistress: `this.effects` is never empty.

modules/core/src/lib/effect-manager.js
```diff
@@ -36,11 +36,24 @@ export default class EffectManager {
...
+      for (const effect of this.effects) {
+        effect.cleanup();
```
Pessimistress: It's good practice to always remove the reference to an instance after its resources are released. This way there is no ambiguity about its state from the owner class' perspective.

modules/core/src/passes/screen-pass.js
```diff
@@ -0,0 +1,47 @@
...
+export default class ScreenPass extends Pass {
+  render(params) {
+    const gl = this.gl;
+
+    withParameters(gl, {framebuffer: params.outputBuffer, clearColor: [0, 0, 0, 0]}, () =>
+      this._renderPass(gl, params)
+    );
+  }
+
+  cleanup() {
```
Pessimistress: - A pass should always be "deleted" if it's never recycled.
- From resource management perspective, whoever created the model should be deleting them. Since `model` is provided externally to the `ScreenPass`, deleting it here can cause unexpected side effects.

test/modules/core/effects/post-process-effect.spec.js
```diff
@@ -0,0 +1,33 @@
...
+};
+
+test('PostProcessEfect#constructor', t => {
+  const effect = new PostProcessEffect(testModule);
+
+  t.ok(effect, 'post-processing effect created');
+  t.end();
+});
+
+test('PostProcessEfect#prepare', t => {
```
Pessimistress: Spelling

test/modules/core/effects/post-process-effect.spec.js
```diff
@@ -0,0 +1,33 @@
...
+  const effect = new PostProcessEffect(testModule);
+  effect.prepare(gl);
+  t.ok(effect.passes, 'post-processing pass created');
+  t.equal(effect.passes.length, 1, 'post-processing pass length is right');
```
Pessimistress: Clean up before existing the test case.

modules/core/src/effects/post-process-effect.js
```diff
@@ -0,0 +1,127 @@
...
+import ScreenPass from '../passes/screen-pass';
+/* eslint-disable import/no-extraneous-dependencies */
+import {normalizeShaderModule} from '@luma.gl/shadertools';
+import {ClipSpace} from '@luma.gl/core';
+
+export default class PostProcessEffect extends Effect {
+  constructor(module, props = {}) {
+    super(props);
+    this.id = `${module.name}-pass`;
+    normalizeShaderModule(module);
```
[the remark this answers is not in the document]
jianhuang01: need generate the default uniform object and getUniform() function from effect shader module, luma.gl doesn't handle it internally.

modules/core/src/passes/screen-pass.js
jianhuang01: > * A pass should always be "deleted" if it's never recycled.

Do you mean we need change the cleanup() function to delete() or something else?

> * From resource management perspective, whoever created the model should be deleting them. Since `model` is provided externally to the `ScreenPass`, deleting it here can cause unexpected side effects.

Maybe "whoever owns the model should be deleting them" more clearer? Although PostProcessEffect creates the model, but the actual ownership is already transferred to ScreenPass class, from this perspective deleting model here is a good fit.

Pessimistress: > the actual ownership is already transferred to ScreenPass class

There is no "transfer" happening, you **think** `ScreenPass` owns the model because you **know** what's going on. This is very dangerous when a different maintainer take over this part of the code, because nothing inside the `ScreenPass` class indicates what is expected in `props`. If you want `ScreenPass` to own the model, create it inside the `ScreenPass` constructor.

modules/core/src/passes/screen-pass.js
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  delete() {
+    this.model.delete();
+    this.model = null;
+  }
+
+  // Private method
+
+  getModel(gl, module, fs, id, userProps) {
```
Pessimistress: Use `_` prefix for private method names

modules/core/src/lib/effect-manager.js
```diff
@@ -36,11 +36,23 @@ export default class EffectManager {
...
   setEffects(effects = []) {
+    this.cleanup();
     this.effects = effects;
   }
 
+  cleanup() {
+    for (const effect of this.effects) {
+      effect.cleanup();
+    }
+    this.effects = [];
```
Pessimistress: Nit: use `this.effects.length = 0;` to avoid creating new object.

PR #3108, opened 2019-05-16
Fix react integration of mapbox layers

Pessimistress:
`MapboxLayer` was overwriting `<DeckGL>`'s `_customRender`, which handles viewport update for the child `<Map>` component.


xintongxia:
> `MapboxLayer` was overwriting `<DeckGL>`'s `_customRender`, which handles viewport update for the child `<Map>` component.

Add this comment to the code.
I am wondering when and why will this be overwritten by `MapboxLayer`?

PR #3114, opened 2019-05-18
Add gouraud/phong/pbr lighting modes to scenegraph-layer

georgios-uber:


georgios-uber:
Will rewrite this to only support PBR. gouraud/phong makes this too complex.

PR #3115, opened 2019-05-19
GPU GridLayer PART-4: Add support for picking

1chandu:
For #2084 

#### Background
- `GPUGridLayer` : Adding support for picking.
    - `onHover` will read data from aggregated textures, will only display, `colorValue`, `elevationValue` and `count` and `position` of selected grid cell.
    - For all other modes, data is aggregated on CPU (and saved in `state`). Based current selected cell `index`, corresponding `hash-key` is generated to access full aggregation results (including array of all aggregated `points`) from cpu Aggregation data.  CPU Aggregation is triggered only during picking and it is previously not calculated.
- `Picking Tests`
    - Add picking tests for `NewGridLayer`, for both cpu and gpu aggregaiton paths, and results are matched.
    - In addition to `pickInfo` count, update the tests to compare, number of aggregated points for each cell, results match in CPU and GPU paths. (this uses accessing CPU results from GPU layer).

OnHoover CPU result:
Screen Shot 2019-05-18 at 7 36 41 PM

OnHoover GPU result:
Screen Shot 2019-05-18 at 7 36 33 PM

#### Change List
- GPU GridLayer PART-4: Add supporting for picking

### DIFF
--- a/modules/aggregation-layers/src/gpu-grid-layer/gpu-grid-cell-layer.js
+++ b/modules/aggregation-layers/src/gpu-grid-layer/gpu-grid-cell-layer.js
@@ -37,8 +37,6 @@ const defaultProps = {
   elevationScale: {type: 'number', min: 0, value: 1},
   extruded: true,
   fp64: false,
-  pickable: false, // TODO: add picking support (read from aggregated texture)
-
   material: defaultMaterial
 };
 
@@ -48,6 +46,7 @@ export default class GPUGridCellLayer extends Layer {
   }
 
   initializeState() {
+    const {gl} = this.context;
     const attributeManager = this.getAttributeManager();
     attributeManager.addInstanced({
       colors: {
@@ -61,27 +60,9 @@ export default class GPUGridCellLayer extends Layer {
         noAlloc: true
       }
     });
-  }
-
-  updateState({props, oldProps, changeFlags}) {
-    super.updateState({props, oldProps, changeFlags});
-    // Re-generate model if geometry changed
-    if (props.fp64 !== oldProps.fp64) {
-      const {gl} = this.context;
-      if (this.state.model) {
-        this.state.model.delete();
-      }
-      const model = this._getModel(gl);
-      this._setupUniformBuffer(model);
-      this.setState({model});
-      this.state.attributeManager.invalidateAll();
-    }
-    if (props.colorBuffer !== oldProps.colorBuffer) {
-      this.state.attributeManager.invalidate('colors');
-    }
-    if (props.elevationBuffer !== oldProps.elevationBuffer) {
-      this.state.attributeManager.invalidate('elevations');
-    }
+    const model = this._getModel(gl);
+    this._setupUniformBuffer(model);
+    this.setState({model});
   }
 
   _getModel(gl) {
@@ -98,6 +79,7 @@ export default class GPUGridCellLayer extends Layer {
 
   draw({uniforms}) {
     const {
+      data,
       cellSize,
       offset,
       extruded,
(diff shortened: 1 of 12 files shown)


modules/aggregation-layers/src/new-grid-layer/new-grid-layer.js
```diff
@@ -4,7 +4,9 @@ import GPUGridAggregator from '../utils/gpu-grid-aggregation/gpu-grid-aggregator
 import GPUGridLayer from '../gpu-grid-layer/gpu-grid-layer';
 import GridLayer from '../grid-layer/grid-layer';
 
-const defaultProps = Object.assign({}, GPUGridLayer.defaultProps, GridLayer.defaultProps);
+const defaultProps = Object.assign({}, GPUGridLayer.defaultProps, GridLayer.defaultProps, {
+  pickable: false
```
Pessimistress: Ditto

modules/aggregation-layers/src/grid-layer/grid-aggregator.js
```diff
@@ -29,11 +29,12 @@ const R_EARTH = 6378000;
  * @param {function} getPosition - position accessor
  * @returns {object} - grid data, cell dimension
  */
-export function pointToDensityGridData(points, cellSize, getPosition) {
+export function pointToDensityGridDataCPU(points, cellSize, getPosition) {
   const {gridHash, gridOffset} = _pointsToGridHashing(points, cellSize, getPosition);
   const layerData = _getGridLayerDataFromGridHash(gridHash, gridOffset);
 
   return {
+    gridHash,
```
Pessimistress: Nit: this name is not very self-explanatory. Call it `pointsByGrid`?

modules/aggregation-layers/src/gpu-grid-layer/gpu-grid-layer.js
```diff
@@ -162,10 +165,73 @@ export default class GPUGridLayer extends CompositeLayer {
...
+    const yIndex = Math.floor(index / gridSize[0]);
+    const xIndex = index - yIndex * gridSize[0];
+    const yPos = yIndex * cellSize[1] + gridOrigin[1];
+    const xPos = xIndex * cellSize[0] + gridOrigin[0];
+    return [xPos, yPos];
+  }
+
+  getPickingInfo({info, mode}) {
+    const {index} = info;
+    const isPicked = info.picked && index > -1;
```
Pessimistress: When may we have `picked: true` and `index: -1`? Add a comment?

modules/aggregation-layers/src/gpu-grid-layer/gpu-grid-cell-layer.js
```diff
@@ -39,7 +39,7 @@ const defaultProps = {
   elevationScale: {type: 'number', min: 0, value: 1},
   extruded: true,
   fp64: false,
-  pickable: false, // TODO: add picking support (read from aggregated texture)
+  pickable: true,
```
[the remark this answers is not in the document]
1chandu: Removed setting this prop here and at other places. I was trying disable pick method for this composite layer, but figured it will be handled by base `CompositeLayer` class.

modules/aggregation-layers/src/gpu-grid-layer/gpu-grid-cell-layer.js
```diff
@@ -87,6 +87,9 @@ export default class GPUGridCellLayer extends Layer {
     if (props.elevationBuffer !== oldProps.elevationBuffer) {
       this.state.attributeManager.invalidate('elevations');
     }
+    if (props.numInstances !== oldProps.numInstances) {
```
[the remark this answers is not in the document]
1chandu: Refactored the code, this is not needed, removed.

modules/aggregation-layers/src/gpu-grid-layer/gpu-grid-layer.js
1chandu: Checking just `index >= 0` should be enough, same as how `Layer` class does.

PR #3121, opened 2019-05-20
Handle partial update in Tesselator

Pessimistress:
For #1576 

#### Change List
- Make `Tesselator` class support partial update
- Unit tests

### DIFF
--- a/modules/core/src/utils/tesselator.js
+++ b/modules/core/src/utils/tesselator.js
@@ -148,7 +148,9 @@ export default class Tesselator {
     const {attributes, _attributeDefs, typedArrayManager, fp64} = this;
     for (const name in _attributeDefs) {
       const def = _attributeDefs[name];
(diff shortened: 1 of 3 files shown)


modules/core/src/utils/tesselator.js
```diff
@@ -105,35 +105,51 @@ export default class Tesselator {
...
-      bufferLayout[dataIndex] = count;
-    });
+    }
 
     // allocate attributes
     const {attributes, _attributeDefs, typedArrayManager, fp64} = this;
     for (const name in _attributeDefs) {
       const def = _attributeDefs[name];
+      // if partial update, copy the old array into the new one
+      def.copy = Boolean(dataRange);
```
xintongxia: Does the comment apply to line 152?

Pessimistress: Yes, I'll expand the comment

PR #3131, opened 2019-05-21
Fix coordinate origin z bug

Pessimistress:
This is a breaking change if a vec3 is provided to `coordinateOrigin` while `coordinateSystem` is `COORDINATE_SYSTEM.METER_OFFSETS` or `COORDINATE_SYSTEM.LNGLAT_OFFSETS`.

`LNGLAT`, `LNGLAT_AUTO_OFFSET` and `IDENTITY` modes are not affected.

### DIFF
--- a/modules/core/src/shaderlib/project/project.glsl.js
+++ b/modules/core/src/shaderlib/project/project.glsl.js
@@ -42,7 +42,7 @@ uniform vec2 project_uViewportSize;
 uniform float project_uDevicePixelRatio;
 uniform float project_uFocalDistance;
 uniform vec3 project_uCameraPosition;
-uniform vec2 project_uCoordinateOrigin;
+uniform vec3 project_uCoordinateOrigin;
 
 const float TILE_SIZE = 512.0;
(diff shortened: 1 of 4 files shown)


modules/core/src/shaderlib/project/project.glsl.js
```diff
@@ -42,7 +42,7 @@ uniform vec2 project_uViewportSize;
 uniform float project_uDevicePixelRatio;
 uniform float project_uFocalDistance;
 uniform vec3 project_uCameraPosition;
-uniform vec2 project_uCoordinateOrigin;
+uniform vec3 project_uCoordinateOrigin;
```
jianhuang01: Do we need change the code which set the project_uCoordinateOrigin uniform? cause it is vec3 now.

Pessimistress: https://github.com/uber/deck.gl/pull/3131/files#diff-c3a9d768f351000a98d9a6d74a57f38fR104

PR #3137, opened 2019-05-24
GPU GridLayer PART 5: Add new color/elevation props to Grid and Hexagon layers

1chandu:
For #2084

#### Background
Following props are needed to perform GPU Aggregation. For parity add these props to `GridLayer` and `HexagonLayer`
- `getColorWeight`, `colorAggregation`, `getElevationWeight` and `elevationAggregation`

#### Change List
- Add new color/elevation props to Grid and Hexagon layers

### DIFF
--- a/docs/layers/grid-layer.md
+++ b/docs/layers/grid-layer.md
@@ -218,6 +218,26 @@ You should pass in the function defined outside the render function so it doesn'
  }
 ```
 
+
+##### `getColorWeight` (Function, optional) ![transition-enabled](https://img.shields.io/badge/transition-enabled-green.svg?style=flat-square")
+
+* Default: `point => 1`
+
+`getColorWeight` is the accessor function to get the weight of a point used to calcuate the color value for a cell.
+It takes the data prop array element and returns the weight, for example, to use `SPACE` field, `getColorWeight` should be set to `point => point.SPACES`.
+By default `getColorWeight` returns `1`.
+
+Note: similar to `getColorValue`, grid layer compares whether `getColorWeight` has changed to recalculate the value for each bin that its color based on.
+
+
+##### `colorAggregation` (String, optional)
+
+* Default: 'SUM'
+
+`colorAggregation` defines, operation used to aggregate all data point weights to calculate a cell's color value. Valid values are 'SUM', 'MEAN', 'MIN' and 'MAX'. 'SUM' is used when an invalid value is provided.
+
+Note: `getColorWeight` and `colorAggregation` together define how color value of cell is determined, same can be done by setting `getColorValue` prop. But to enable gpu aggregation, former props must be provided instead of later.
+
 ##### `getElevationValue` (Function, optional) ![transition-enabled](https://img.shields.io/badge/transition-enabled-green.svg?style=flat-square")
 
 * Default: `points => points.length`
@@ -228,6 +248,27 @@ By default `getElevationValue` returns the length of the points array.
 
 Note: grid layer compares whether `getElevationValue` has changed to recalculate the value for each cell for its elevation.
 You should pass in the function defined outside the render function so it doesn't create a new function on every rendering pass.
+
+
+##### `getElevationWeight` (Function, optional) ![transition-enabled](https://img.shields.io/badge/transition-enabled-green.svg?style=flat-square")
+
+* Default: `point => 1`
+
+`getElevationWeight` is the accessor function to get the weight of a point used to calcuate the elevation value for a cell.
+It takes the data prop array element and returns the weight, for example, to use `SPACE` field, `getElevationWeight` should be set to `point => point.SPACES`.
+By default `getElevationWeight` returns `1`.
+
+Note: similar to `getElevationValue`, grid layer compares whether `getElevationWeight` has changed to recalculate the value for each bin that its color based on.
+
+
+##### `elevationAggregation` (String, optional)
+
+* Default: 'SUM'
+
+`elevationAggregation` defines, operation used to aggregate all data point weights to calculate a cell's elevation value. Valid values are 'SUM', 'MEAN', 'MIN' and 'MAX'. 'SUM' is used when an invalid value is provided.
+
+Note: `getElevationWeight` and `elevationAggregation` together define how elevation value of cell is determined, same can be done by setting `getColorValue` prop. But to enable gpu aggregation, former props must be provided instead of later.
+
 
 ##### `onSetColorDomain` (Function, optional)
 
--- a/docs/layers/grid-layer.md
+++ b/docs/layers/grid-layer.md
@@ -224,8 +224,6 @@ You should pass in the function defined outside the render function so it doesn'
 * Default: `point => 1`
 
 `getColorWeight` is the accessor function to get the weight of a point used to calcuate the color value for a cell.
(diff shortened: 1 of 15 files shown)


modules/aggregation-layers/src/grid-layer/grid-layer.js
```diff
@@ -334,8 +353,8 @@ export default class GridLayer extends CompositeLayer {
         getFillColor: this._onGetSublayerColor.bind(this),
         getElevation: this._onGetSublayerElevation.bind(this),
         transitions: transitions && {
-          getFillColor: transitions.getColorValue,
-          getElevation: transitions.getElevationValue
+          getFillColor: transitions.getColorValue || transitions.getColorWeight,
+          getElevation: transitions.getElevationValue || transitions.getElevationWeight
```
1chandu: > ```
> getFillColor: transitions.getColorValue || transitions.getColorWeight,
> getElevation: transitions.getElevationValue || transitions.getElevationWeight
> ```
When using new props, color value of cell is defined together by `getColorWeight` and `colorAggregation`.  But given transition key must be prop name just using `getColorWeight`.  Another option is, we can let users use `getColorValue` as key even when `getColorWeight` and `colorAggregation` are specified or use a more generic key as `colorValue` (especially for CompositeLayers) that can be mapped to a sub-layer's attribute/accessor name.

docs/layers/grid-layer.md
```diff
@@ -218,6 +218,26 @@ You should pass in the function defined outside the render function so it doesn'
  }
 ```
 
+
+##### `getColorWeight` (Function, optional) ![transition-enabled](https://img.shields.io/badge/transition-enabled-green.svg?style=flat-square")
+
+* Default: `point => 1`
+
+`getColorWeight` is the accessor function to get the weight of a point used to calcuate the color value for a cell.
+It takes the data prop array element and returns the weight, for example, to use `SPACE` field, `getColorWeight` should be set to `point => point.SPACES`.
```
Pessimistress: Move this to the section below, where you explain that these two props must be used together, and provide a few examples.

Ex 1. Use color to encode the total count of points:

```js
colorAggregation: 'SUM',
getColorWeight: point => 1
```

is equivalent to:

```js
getColorValue: points => points.length
```

Ex 2. Use color to encode the average number of spaces:

```js
colorAggregation: 'MEAN',
getColorWeight: point => point.spaces
```

is equivalent to:

```js
getColorValue: points => points.reduce((sum, p) => sum += p.spaces, 0) / points.length
```

docs/layers/grid-layer.md
```diff
@@ -229,6 +249,27 @@ By default `getElevationValue` returns the length of the points array.
 Note: grid layer compares whether `getElevationValue` has changed to recalculate the value for each cell for its elevation.
 You should pass in the function defined outside the render function so it doesn't create a new function on every rendering pass.
 
+
+##### `getElevationWeight` (Function, optional) ![transition-enabled](https://img.shields.io/badge/transition-enabled-green.svg?style=flat-square")
+
+* Default: `point => 1`
+
+`getElevationWeight` is the accessor function to get the weight of a point used to calcuate the elevation value for a cell.
+It takes the data prop array element and returns the weight, for example, to use `SPACE` field, `getElevationWeight` should be set to `point => point.SPACES`.
```
Pessimistress: It would be nice to provide examples as the combination of these two props to showcase what they do, see comment on `getColorWeight`

modules/aggregation-layers/src/grid-layer/grid-layer.js
```diff
@@ -89,6 +93,21 @@ export default class GridLayer extends CompositeLayer {
     }
   }
 
+  updateGetValueFuncs(props) {
+    let {getColorValue, getElevationValue} = props;
+    const {colorAggregation, getColorWeight, elevationAggregation, getElevationWeight} = this.props;
+
+    // If a custom `getColorValue` is provided (other than default) use it
+    if (getColorValue === defaultProps.getColorValue.value) {
```
Pessimistress: If the default value is never used, just set to `null` for clarity.

modules/aggregation-layers/src/hexagon-layer/hexagon-layer.js
```diff
@@ -85,6 +93,21 @@ export default class HexagonLayer extends CompositeLayer {
     }
   }
 
+  updateGetValueFuncs(props) {
+    let {getColorValue, getElevationValue} = props;
+    const {colorAggregation, getColorWeight, elevationAggregation, getElevationWeight} = this.props;
+
+    // If a custom `getColorValue` is provided (other than default) use it
+    if (getColorValue === defaultProps.getColorValue.value) {
+      getColorValue = getValueFunc(colorAggregation, getColorWeight);
```
Pessimistress: This is run on every layer update, and `getValueFunc` creates new functions whenever called. Check if the relevant props have changed before updating the state.

docs/layers/grid-layer.md
1chandu: Added examples, will make same changes to `HexagonLayer` docs, in Part6 PR.

docs/layers/grid-layer.md
1chandu: added examples

modules/aggregation-layers/src/grid-layer/grid-layer.js
1chandu: ok, using `null`.

modules/aggregation-layers/src/hexagon-layer/hexagon-layer.js
1chandu: done.

docs/layers/grid-layer.md
```diff
@@ -218,6 +218,86 @@ You should pass in the function defined outside the render function so it doesn'
...
+##### `colorAggregation` (String, optional)
+
+* Default: 'SUM'
+
+`colorAggregation` defines, operation used to aggregate all data point weights to calculate a cell's color value. Valid values are 'SUM', 'MEAN', 'MIN' and 'MAX'. 'SUM' is used when an invalid value is provided.
+
+Note: `getColorWeight` and `colorAggregation` together define how color value of cell is determined, same can be done by setting `getColorValue` prop. But to enable GPU aggregation, former props must be provided instead of later.
+
+###### Example1 : Using count of data elements that fall into a cell to encode the its color
+```js
```
Pessimistress: Header: Using `getColorValue`

docs/layers/grid-layer.md
```diff
@@ -218,6 +218,86 @@ You should pass in the function defined outside the render function so it doesn'
...
 ```
 
+
+##### `getColorWeight` (Function, optional) ![transition-enabled](https://img.shields.io/badge/transition-enabled-green.svg?style=flat-square")
+
+* Default: `point => 1`
+
+`getColorWeight` is the accessor function to get the weight of a point used to calcuate the color value for a cell.
+
+Note: similar to `getColorValue`, grid layer compares whether `getColorWeight` has changed to recalculate the value for each bin that its color based on.
```
Pessimistress: Why does this need to be called out? Does these props not behave like other accessors (i.e. recalculation is triggered by updateTriggers)?

1chandu: It will get triggered by update triggers, and behaves similar to other accessor props, nothing special. I am following the existing documentation of `getColorValue`, this can be removed.

PR #3139, opened 2019-05-24
Metrics-tracking Deck property

tsherif:
Simple Deck member prop that tracks performance statistics over a 60-frame window.

### DIFF
--- a/docs/api-reference/deck.md
+++ b/docs/api-reference/deck.md
@@ -379,6 +379,10 @@ Notes:
 
 ## Member Variables
 
+#### metrics
+
+A map of various performance statistics for the last 60 frames of rendering.
+
 
 ## Remarks
 
--- a/docs/api-reference/deck.md
+++ b/docs/api-reference/deck.md
@@ -384,9 +384,10 @@ Notes:
 A map of various performance statistics for the last 60 frames of rendering. Metrics gathered in deck.gl are the following:
 - 'fps': average number of frames rendered per second
 - 'updateAttributesTime': time spent updating layer attributes
-- 'layersRedrawn': number of layers drawn
-- 'setPropsTime': time spent setting layer properties
-- 'drawCount': number of times the scene was rendered
+- 'setPropsTime': time spent setting deck properties
+- 'framesRedrawn': number of times the scene was rendered
+- 'pickTime': total time spent on picking operations
+- 'pickCount': number of times a pick operation was performed
 - 'gpuTime': total time spent on GPU processing
 - 'gpuTimePerFrame': average time spent on GPU processing per frame
 - 'cpuTime': total time spent on CPU processing
--- a/docs/api-reference/deck.md
+++ b/docs/api-reference/deck.md
@@ -381,7 +381,21 @@ Notes:
 
 #### metrics
 
-A map of various performance statistics for the last 60 frames of rendering.
+A map of various performance statistics for the last 60 frames of rendering. Statistics generally fall into two categories, count and time, and the relevancy of data provided for
+statistic will depend on it's category (e.g. "time"-based values are only relevant for statistics that include timing). Data provided for each statistic are:
+- `time`: total time spent on an operation (relevant for time statistics)
+- `count`: number or amount of some metric (relevant for count statistics)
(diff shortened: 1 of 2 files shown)


docs/api-reference/deck.md
```diff
@@ -377,6 +377,13 @@ Notes:
 * The query methods offer more flexibility for developers to handle events compared to the built-in hover and click callbacks.
 
 
+## Member Variables
+
+#### metrics
+
+A map of various performance statistics for the last 60 frames of rendering.
```
Pessimistress: Document the names of stats, and what each field means?

docs/api-reference/deck.md
```diff
@@ -377,6 +377,27 @@ Notes:
...
+- `count`: number or amount of some metric (relevant for count statistics)
+- `average`: average time spent per operation (relevant for time statistics)
+- `hz`: number of operations per second (relevant for time statistics)
+
+Some metrics gathered in deck.gl include:
+- 'Update Attributes': time spent updating attributes
+- 'Redraw Layers': number of layers drawn
+- 'setProps Time': time spent setting properties
+- 'Redraw Count': number of times the scene was rendered
+- 'pickObject Time': time spent picking an object
```
Pessimistress: If the stat is called `pickObject Time`, the value should be a single number that is the time. If the value is an object, the stat should just be called `pickObject`?

tsherif: The stat names are just from grepping where stats are currently used. I agree that it's confusing, but I can't break the current API for this update.

tsherif:
@Pessimistress this actually worked out pretty well. Decided to pull some stats in from luma as well.

docs/api-reference/deck.md
```diff
@@ -377,6 +377,26 @@ Notes:
...
 
+## Member Variables
+
+#### metrics
+
+A map of various performance statistics for the last 60 frames of rendering. Metrics gathered in deck.gl are the following:
+- 'fps': average number of frames rendered per second
+- 'updateAttributesTime': time spent updating layer attributes
+- 'layersRedrawn': number of layers drawn
+- 'setPropsTime': time spent setting layer properties
```
Pessimistress: This tracks the time spent in `deck.setProps` so it should be "deck properties"?

modules/core/src/lib/deck.js
```diff
@@ -766,6 +779,63 @@ export default class Deck {
...
+      .get('CPU Time')
+      .addTime(
+        animationLoopStats.get('CPU Time') ? animationLoopStats.get('CPU Time').lastTiming : 0
+      );
+  }
+
+  /* eslint-disable complexity */
+  _getMetrics() {
+    this.metrics.fps = this.stats.get('frameRate') ? this.stats.get('frameRate').getHz() : 0;
+    this.metrics.setPropsTime = this.stats.get('setProps Time')
```
Pessimistress: I think `stats.get` always returns a `Stat` instance.

docs/api-reference/deck.md
```diff
@@ -377,6 +377,26 @@ Notes:
...
+- 'drawCount': number of times the scene was rendered
```
Pessimistress: `framesRedrawn`?

modules/core/src/lib/deck.js
tsherif: You are absolutely correct. Forgot how my own API works :p

modules/core/src/lib/deck.js
```diff
@@ -766,6 +783,40 @@ export default class Deck {
...
+    this.stats.get('CPU Time').addTime(animationLoopStats.get('CPU Time').lastTiming);
+  }
+
+  _getMetrics() {
+    this.metrics.fps = this.stats.get('frameRate').getHz();
+    this.metrics.setPropsTime = this.stats.get('setProps Time').time;
+    this.metrics.updateAttributesTime = this.stats.get('Update Attributes').time;
+    this.metrics.framesRedrawn = this.stats.get('Redraw Count').count;
+    this.metrics.pickTime =
+      this.stats.get('pickObject Time').time +
```
Pessimistress: Let's just merge these three stats. I don't think anyone is using one in particular.

Issue #3138, opened 2019-05-24
Add billboard mode to PathLayer


Pessimistress:
#### Target Use case
PathLayer does not work well for arbitrary 3D paths, due to its extrusion always facing up (z+).

Related: #2302 

@georgios-uber 

#### Proposed feature
Add a `billboard` prop that extrudes the paths in screen space.

#### To Do List 
  - [x] Add label and assign to milestone
  - [x] Coding
  - [x] Doc update
  - [x] What’s new update
  - [x] Test

PR #3140, opened 2019-05-24
Add billboard mode to PathLayer

Pessimistress:
For #3138 

billboard: false (default)

image

billboard: true

image

This new prop is not supported in the deprecated fp64 mode.

There are precision issues at high zoom levels, will address in a follow up PR.

#### Change List
- Add `billboard` prop to path layer (screenspace extrusion)
- Docs

### DIFF
--- a/modules/layers/src/path-layer/path-layer-vertex.glsl.js
+++ b/modules/layers/src/path-layer/path-layer-vertex.glsl.js
@@ -38,6 +38,7 @@ uniform float widthMinPixels;
 uniform float widthMaxPixels;
 uniform float jointType;
 uniform float miterLimit;
+uniform bool billboard;
 
 uniform float opacity;
 
@@ -49,6 +50,7 @@ varying vec2 vPathPosition;
 varying float vPathLength;
 
 const float EPSILON = 0.001;
+const vec3 ZERO_OFFSET = vec3(1.0);
 
 float flipIfTrue(bool flag) {
   return -(float(flag) * 2. - 1.);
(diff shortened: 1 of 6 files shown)


modules/layers/src/path-layer/path-layer-vertex.glsl.js
```diff
@@ -58,18 +60,18 @@ float flipIfTrue(bool flag) {
 vec3 lineJoin(
   vec3 prevPoint, vec3 currPoint, vec3 nextPoint,
   float relativePosition, bool isEnd, bool isJoint,
-  float width
+  vec2 width
```
tsherif: Good god, this function...

modules/layers/src/path-layer/path-layer-vertex.glsl.js
```diff
@@ -58,18 +60,18 @@ float flipIfTrue(bool flag) {
...
 ) {
-  vec2 deltaA = currPoint.xy - prevPoint.xy;
-  vec2 deltaB = nextPoint.xy - currPoint.xy;
+  vec2 deltaA = (currPoint.xy - prevPoint.xy) / width;
+  vec2 deltaB = (nextPoint.xy - currPoint.xy) / width;
 
   float lenA = length(deltaA);
   float lenB = length(deltaB);
```
tsherif: Perf note unrelated to this PR: It seems like this logic mostly compares length and doesn't really care about the value, so you can get a bit of a win by avoiding the square root until it's necessary. I.e.:
```cpp
float lenA2 = dot(deltaA, deltaA);
float lenB2 = dot(deltaB, deltaB);
lenA2 = lenA2 > EPSILON2 ? lenA2 : 0.0;
lenB2 = lenB2 > EPSILON2 ? lenB2 : 0.0;

// Later...
if (cornerPosition < 0.0) {
  float L = sqrt(L2);
  offsetScale = min(offsetScale, L / max(cosHalfA, EPSILON));
}
```

modules/layers/src/path-layer/path-layer-vertex.glsl.js
Pessimistress: Yep.

modules/layers/src/path-layer/path-layer-vertex.glsl.js
Pessimistress: Will do in a follow up PR.

PR #3166, opened 2019-05-31
new post-processing effect rendering test

jianhuang01:
For #2962

#### Change List
- new test case with `noise` and `vignette` effects


test/render/test-cases.js
```diff
@@ -1470,5 +1480,41 @@ export const TEST_CASES = [
...
+    effects: [effect1, effect2],
+    viewState: {
+      latitude: 37.78,
+      longitude: -122.45,
+      zoom: 12,
+      pitch: 0,
+      bearing: 0
+    },
+    layers: [
+      new GeoJsonLayer({
```
Pessimistress: Can we use a less complex layer, e.g. ScatterplotLayer?

PR #3174, opened 2019-06-03
Address different publishing requirements

ajduberstein:
- Change versioning scheme for release of `0.1dev1`
- Add warning if Mapbox token is not set
- Add an additional notebook
- Re-introduce support for updating pre-existing visuals

### DIFF
--- a/bindings/python/pydeck/README.md
+++ b/bindings/python/pydeck/README.md
@@ -3,35 +3,22 @@
 The pydeck module is a set of experimental Python bindings for deck.gl
 
 For a quick example, install pydeck and try the following in a Jupyter notebook:
-```python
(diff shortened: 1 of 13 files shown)


ajduberstein:
Supports updates without re-rendering the entire canvas now
time

ajduberstein:
Add an example for Conway's game of life

Conway

ibgreen:
This looks amazing! 

Now, the way you are pushing the interactive updates between python and deck, we really need to find a way to open a direct connection between kernel and browser. 

I.e. I strongly suspect that serializing python `data` to HTML and sending it to the browser 30 times per second, even though it works, will make the computer run really hot, start the fans and drain the battery in no time.

PR #3179, opened 2019-06-03
upgrade website to react-map-gl v5

Pessimistress:
The website currently uses react-map-gl v3, which depends on a mapbox version that is going to be deprecated.

This change upgrades react-map-gl to v5 and simplifies the demo setup.

We do lose the transition between demos by dropping the `<MapGL><DeckGL/></MapGL>` pattern.


examples/website/icon/app.js
```diff
@@ -155,23 +155,20 @@ export class App extends Component {
...
-    const {viewState, controller = true, baseMap = true} = this.props;
+    const {mapStyle = 'mapbox://styles/mapbox/dark-v9'} = this.props;
 
     return (
       <DeckGL
         layers={this._renderLayers()}
         initialViewState={INITIAL_VIEW_STATE}
-        viewState={viewState}
-        controller={controller}
+        controller={{dragRotate: false}}
```
xintongxia: why this is needed?

website/src/components/demos/layer-demo-base.js
```diff
@@ -107,7 +104,14 @@ export default function createLayerDemoClass(settings) {
 
       return (
         <div>
-          <DeckGL pickingRadius={5} viewState={viewState} layers={layers} />
+          <DeckGL pickingRadius={5}
```
xintongxia: formatting not correct

website/src/components/demos/line-demo.js
```diff
@@ -31,10 +31,6 @@ export default class LineDemo extends Component {
     };
   }
 
-  static get viewport() {
-    return INITIAL_VIEW_STATE;
-  }
-
   static get mapStyle() {
     return MAPBOX_STYLES.DARK;
```
xintongxia: Does this style have be defined in every demo?

Pessimistress: Yes.

examples/website/icon/app.js
Pessimistress: This demo should only be viewed top-down.

Issue #3173, opened 2019-06-03
7.1 Release Checklist


1chandu:
- [x] Verify whats-new (@Pessimistress)
- [x] Verify upgrade-guide (@Pessimistress)
- [x] Verify all 7.1 issues are closed or move out (@tgorkin)
- [x] Verify all API changes are documented
- [x] Documentation: Update source file urls to point to 7.1-release branch
- [x] Website (docs and example) testing: [staging](http://pessimistress.github.io/deck.gl/#/documentation/overview/whats-new)
    - [x] Testing on Chrome (@jianhuang01)
    - [x] Testing on Safari (@xintongxia)
    - [x] Testing on Firefox (@1chandu) 
    - [x] Testing on Edge + IE11 (@tsherif)
    - [x] Testing on iOS (@Pessimistress)
    - [x] Testing on Android (@Pessimistress)

Testing procedure
- access staged doc website
- go through all examples
- verify functionality in control panels
- look for console errors/warnings
- add results of testing as comments on this issue

Pessimistress:
Mobile:
- [ ] What's new images are squashed
- [ ] IconLayer example: picking does not work in iOS Safari
- [ ] TripsLayer thumbnail on the "examples" page goes to the brushing example

1chandu:
Firefox :


 - [ ] Postprocessing doc page (Console Errors/Warnings)
  Error: WebGL warning: getContext: Disallowing antialiased backbuffers due to blacklisting. dist.min.js:1:76341
Error: WebGL warning: texImage2D: Alpha-premult and y-flip are deprecated for non-DOM-Element uploads.

 - [ ] Bitmap layer doesn't render anything.

 - [ ] ColumnLayer (doc page):  `stroked` doesn't render the stroke line (with extuded off)

 - [ ] ColumnLayer (doc page) : crash while changing `diskResolution` and `radius`
Error: ColumnLayer-program  Bad uniform edgeDistance bundle.js:1:76706
ColumnLayer-program  Bad uniform edgeDistance NaN

 - [ ] ColumnLayer (doc page) : `discResolution` controller changes between 0 and 1 instead of actual prop limits.

Pessimistress:
The PathLayer issue is just z-fighting. We can disable depth test for this demo.

tsherif:
Edge:
- [ ] Icon layer example not working. Only console error is about NPOT texture

IE:
[...]

PR #3180, opened 2019-06-03
Website improvements

Pessimistress:
For #3173 	#3160 

#### Change List
- Use meaningful props for non-extruded mode in ColumnLayer embedded demo
- Fix what's new image styling
- Add new showcase
- Improve layer parameter controls in embedded demos (use prop type instead of name parsing)


website/src/utils/layer-params.js
```diff
@@ -18,38 +15,30 @@ export function propToParam(key, value) {
...
+    case 'color':
       return {...param, type: 'color'};
-    }
-    if (/mapping|domain|range/i.test(key)) {
-      return {...param, type: 'json'};
-    }
-    break;
-  default:
+    case 'object':
+      if (/mapping|domain|range/i.test(key)) {
```
xintongxia: formatting

PR #3187, opened 2019-06-04
Website: fix crashes in GridCellLayer and TileLayer Demos

1chandu:
For #3173 

#### Background
- Remove deprecated API usage thats causing crashes.
- Fix mapbox token variable.

#### Change List
- Fix crashes in GridCellLayer and TileLayer Demos

### DIFF
--- a/website/src/components/demos/geo-layer-demos.js
+++ b/website/src/components/demos/geo-layer-demos.js
@@ -95,7 +95,9 @@ export const TileLayerDemo = createLayerDemoClass({
     lineWidthMinPixels: 1,
 
     getTileData: ({x, y, z}) => {
-      const mapSource = `https://a.tiles.mapbox.com/v4/mapbox.mapbox-streets-v7/${z}/${x}/${y}.vector.pbf?access_token=${MapboxAccessToken}`;
+      const mapSource = `https://a.tiles.mapbox.com/v4/mapbox.mapbox-streets-v7/${z}/${x}/${y}.vector.pbf?access_token=${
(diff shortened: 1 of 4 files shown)


website/src/components/demos/geo-layer-demos.js
```diff
@@ -73,29 +73,9 @@ export const TileLayerDemo = createLayerDemoClass({
   formatTooltip: f => JSON.stringify(f.properties),
   allowMissingData: true,
   props: {
-    stroked: false,
     opacity: 1,
-
-    getLineColor: [192, 192, 192],
-    getFillColor: [140, 170, 180],
-
-    getLineWidth: f => {
```
Pessimistress: Why do we need to remove this? The demo will end up not matching the sample code in the documentation.

1chandu: Ok, fixed it.
Problem was, `propType` for these props was undefined causing a crash. Fixed it in `layer-params.js`.
But side question: in `tile-layer.js` should we change

```
const defaultProps = {
  renderSubLayers: {type: 'function', value: props => new GeoJsonLayer(props)},
  ...
  maxCacheSize: null
};
```
to

```
const defaultProps = Object.assign( {}, GeoJsonLayer.defaultProps,  {
  renderSubLayers: {type: 'function', value: props => new GeoJsonLayer(props)},
  ...
  maxCacheSize: null
};

```

Pessimistress: No, because `TileLayer` is not required to use GeoJsonLayer as its sublayer.

PR #3192, opened 2019-06-05
Support 3D paths with TripsLayer

Pessimistress:
For https://github.com/uber/deck.gl/issues/3084

#### Change List
- Add `getTimestamps` prop to `TripsLayer`
- Documentation


xintongxia:
Add a render test if possible?

docs/layers/trips-layer.md
```diff
@@ -40,8 +40,9 @@ const App = ({data, viewport}) => {
   const layer = new TripsLayer({
     id: 'trips-layer',
     data,
+    getPath: d => d.waypoints.map(p => p.coordinates),
```
xintongxia: maybe update the comment describing the data format to 3D instead of 2D?

modules/geo-layers/src/trips-layer/trips-layer.js
```diff
@@ -59,16 +72,66 @@ if(vTime > currentTime || vTime < currentTime - trailLength) {
...
+
   draw(params) {
-    const {trailLength, currentTime} = this.props;
+    const {trailLength, currentTime, getTimestamps} = this.props;
 
     params.uniforms = Object.assign({}, params.uniforms, {
       trailLength,
-      currentTime
+      currentTime,
+      isPath3D: Boolean(getTimestamps)
```
xintongxia: this uniform will be also removed in v8

docs/layers/trips-layer.md
Pessimistress: Where do we describe the data format as 2d?

xintongxia: before layer construction, line 27 - line 39

Pessimistress: This format is allowed. We also point users to the PathLayer doc which describes all the path formats that are supported.

srikanth526:
I don't see elevation in the trips layer sample for 3D. @Pessimistress @SymbolixAU. I just need trips layer for the bird migration so I just need elevation property for it..
{
    "travel_type": "car",
    "instruction": "Drive northwest on Steuart Street.",
    "length": 0.049,
    "begin_shape": [-122.394181, 37.793991],
    "end_shape": [-122.394845, 37.794448],
    "begin_time": 0,
    "end_time": 11
  }

PR #3195, opened 2019-06-05
GPUGridLayer: make it exclusive for WebGL2

1chandu:
For #3173 

#### Background
GPUGridLayer is only supported when using `WebGL2`, add a warning to use `GridLayer` when not supported.

#### Change List
- GPUGridLayer: make it exclusive for WebGL2

### DIFF
--- a/docs/layers/cpu-grid-layer.md
+++ b/docs/layers/cpu-grid-layer.md
@@ -8,11 +8,13 @@
 
 # CPUGridLayer
 
-The Grid Layer renders a grid heatmap based on an array of points.
+The `CPUGridLayer` renders a grid heatmap based on an array of points.
 It takes the constant cell size, aggregates input points into cells. Aggregation is performed on CPU. The color
 and height of the cell is scaled by number of points it contains.
 
-CPUGridLayer is a [CompositeLayer](/docs/api-reference/composite-layer.md).
+`CPUGridLayer` is one of the sublayers for [GridLayer](/docs/layers/grid-layer.md), and is provided to customize CPU Aggregation for advanced use cases. For any regular use case, [GridLayer](/docs/layers/grid-layer.md) is recommended.
+
+`CPUGridLayer` is a [CompositeLayer](/docs/api-reference/composite-layer.md).
 
 ```js
 import DeckGL from '@deck.gl/react';
--- a/docs/layers/gpu-grid-layer.md
+++ b/docs/layers/gpu-grid-layer.md
@@ -12,7 +12,9 @@ The `GPUGridLayer` renders a grid heatmap based on an array of points.
 It takes the constant cell size, aggregates input points into cells. When possible aggregation is performed on GPU. The color
 and height of the cell is scaled by number of points it contains.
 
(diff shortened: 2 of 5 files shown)


modules/aggregation-layers/src/gpu-grid-layer/gpu-grid-layer.js
```diff
@@ -233,6 +240,9 @@ export default class GPUGridLayer extends CompositeLayer {
   }
 
   renderLayers() {
+    if (!GPUGridAggregator.isSupported(this.context.gl)) {
```
xintongxia: I wonder is there other solution, which only need check once instead of checking in multiple places?

1chandu: Addressed, added a state variable. I was actually looking for a way, so layer-manager doesn't call updates on this, but I can only do that for sublayers by returning `null`, couldn't find a way to do it for current layer.

docs/layers/gpu-grid-layer.md
```diff
@@ -6,13 +6,13 @@
   <img src="https://img.shields.io/badge/lighting-yes-blue.svg?style=flat-square" alt="lighting" />
 </p>
 
-# GPUGridLayer
+# GPUGridLayer (WebGL2)
 
 The `GPUGridLayer` renders a grid heatmap based on an array of points.
```
Pessimistress: I wonder if we should add a sentence to the top of GPUGridLayer/CPUGridLayer docs that directs non-advanced users to use the GridLayer. It can be quite confusing since all three classes are presented as peers in the documentation.

Pessimistress: You may also consider adding this tag: https://github.com/uber/deck.gl/blob/master/website/contents/pages.js#L357

1chandu: Done.

docs/layers/cpu-grid-layer.md
```diff
@@ -8,11 +8,13 @@
 
 # CPUGridLayer
 
-The Grid Layer renders a grid heatmap based on an array of points.
+The `CPUGridLayer` renders a grid heatmap based on an array of points.
 It takes the constant cell size, aggregates input points into cells. Aggregation is performed on CPU. The color
 and height of the cell is scaled by number of points it contains.
 
-CPUGridLayer is a [CompositeLayer](/docs/api-reference/composite-layer.md).
+`CPUGridLayer` is one of the sublayers for [GridLayer](/docs/layers/grid-layer.md), and is provided to customize CPU Aggregation for advanced use cases. For any regular use case, [GridLayer](/docs/layers/grid-layer.md) should be used.
```
Pessimistress: <del>should be used</del> is recommended?

modules/aggregation-layers/src/gpu-grid-layer/gpu-grid-layer.js
```diff
@@ -60,12 +60,18 @@ const defaultProps = {
 export default class GPUGridLayer extends CompositeLayer {
   initializeState() {
     const {gl} = this.context;
+    let isSupported = true;
+    if (!GPUGridAggregator.isSupported(gl)) {
```
Pessimistress: Nit:

```js
const isSupported = GPUGridAggregator.isSupported(gl);
if (!isSupported) {
  ...
}
```

Issue #3199, opened 2019-06-05
How to display video frames with deck.gl ?


jrabary:
I'm new to deck.gl and wondering how to display video frames with it. 
I would like to build an app like the ascii demo but instead draw the actual frames.
Is the BitmapLayer suitable for this task or do I need to write a custom layers ?
Any hints will be helpful.
Regards

Pessimistress:
The `BitmapLayer` does not support video right now. I think this is a reasonable use case and we should consider supporting it. @xintongxia 

I put together an example here: https://codepen.io/Pessimistress/pen/pmmbRj

jrabary:
@Pessimistress Thanks for the example. Is there something missing it the code ? The image in the canvas is not updated when the video is playing.

Pessimistress:
Are you using Safari? WebGL1 does not have good support for non-power-of-two texture sizes (and the demo video demensions are not power of two). You will need to resize your source video.

tsherif:
To get an non-power-of-two texture to work in WebGL 1, just change the wrap mode to `CLAMP_TO_EDGE` (and turn off mipmaps, which was already done here), i.e. in the codepen example, change the texture construction to: 

```js
this.setState({
  bitmapTexture: new luma.Texture2D(gl, {
    width: 1,
    height: 1,
    parameters: {
      [gl.TEXTURE_MIN_FILTER]: gl.NEAREST,
      [gl.TEXTURE_MAG_FILTER]: gl.NEAREST,
      
      // THIS IS THE NEW PART
      [gl.TEXTURE_WRAP_S]: gl.CLAMP_TO_EDGE,
      [gl.TEXTURE_WRAP_T]: gl.CLAMP_TO_EDGE
    },
    mipmaps: false          
  })
});
```

Pessimistress:
Thanks @tsherif ! Codepen is updated.

jrabary:
Thank you. It's working great!

jrabary:
@Pessimistress Just to understand what's going on.  Does the fact the we set `_animate = true` forces deckGL to call layer.draw every timestep ?

Pessimistress:
That is correct. If you want more control over this, you can manually trigger redraw:

```js
let animation = null;

const bitmaplayer

function animate() {
  videoBitmapLayer.setNeedsRedraw(true);
  animation = requestAnimationFrame(animate);
};

video.onplay = () => animate();
video.onpause = () => cancelAnimationFrame(animation);
```

PR #3203, opened 2019-06-05
Add video support in BitmapLayer

xintongxia:
#### Background
for #3199 

#### Change List
- support video
- update doc 
- add video test app

### DIFF
--- a/modules/layers/src/bitmap-layer/bitmap-layer.js
+++ b/modules/layers/src/bitmap-layer/bitmap-layer.js
@@ -18,7 +18,7 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-/* global Image, HTMLCanvasElement */
+/* global Image, HTMLCanvasElement, HTMLVideoElement */
 import GL from '@luma.gl/constants';
 import {Layer} from '@deck.gl/core';
 import {Model, Geometry, Texture2D, fp64} from '@luma.gl/core';
@@ -30,8 +30,14 @@ import vs from './bitmap-layer-vertex';
 import fs from './bitmap-layer-fragment';
 
 const DEFAULT_TEXTURE_PARAMETERS = {
+  [GL.TEXTURE_MIN_FILTER]: GL.LINEAR,
+  [GL.TEXTURE_MAG_FILTER]: GL.LINEAR,
+  [GL.TEXTURE_WRAP_S]: GL.CLAMP_TO_EDGE,
+  [GL.TEXTURE_WRAP_T]: GL.CLAMP_TO_EDGE
+};
+
+const DEFAULT_MIPMAP_TEXTURE_PARAMETERS = {
   [GL.TEXTURE_MIN_FILTER]: GL.LINEAR_MIPMAP_LINEAR,
-  // GL.LINEAR is the default value but explicitly set it here
   [GL.TEXTURE_MAG_FILTER]: GL.LINEAR,
   [GL.TEXTURE_WRAP_S]: GL.CLAMP_TO_EDGE,
   [GL.TEXTURE_WRAP_T]: GL.CLAMP_TO_EDGE
@@ -180,9 +186,34 @@ export default class BitmapLayer extends Layer {
     );
   }
 
-  draw({uniforms}) {
+  draw(opts) {
+    const {uniforms} = opts;
     const {bitmapTexture, model} = this.state;
-    const {desaturate, transparentColor, tintColor} = this.props;
+    const {image, desaturate, transparentColor, tintColor} = this.props;
+
+    // Update video frame
+    if (
+      bitmapTexture &&
+      image instanceof HTMLVideoElement &&
+      image.readyState > HTMLVideoElement.HAVE_METADATA
+    ) {
+      const sizeChanged =
+        bitmapTexture.width !== image.videoWidth || bitmapTexture.height !== image.videoHeight;
+      if (sizeChanged) {
+        // note clears image and mipmaps when resizing
+        bitmapTexture.resize({width: image.videoWidth, height: image.videoHeight});
+        bitmapTexture.setImageData({
+          data: image,
+          parameters: DEFAULT_TEXTURE_PARAMETERS
+        });
+        bitmapTexture.generateMipmap();
+      } else {
+        bitmapTexture.setImageData({
+          data: image,
+          parameters: DEFAULT_TEXTURE_PARAMETERS
+        });
+      }
+    }
 
     // // TODO fix zFighting
     // Render the image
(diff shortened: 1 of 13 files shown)


examples/website/video/app.js
```diff
@@ -0,0 +1,62 @@
...
+  render() {
+    return (
+      <div>
+        <DeckGL
+          layers={this._renderLayers()}
+          initialViewState={INITIAL_VIEW_STATE}
+          views={new OrthographicView({})}
+          controller={true}
+          // force deck to redraw on every frame
+          _animate={true}
```
xintongxia: I don't think this prop is supposed to depend on user to to pass in? Is there a better way, can VideoLayer enable the animation by itself?

Pessimistress:
Why do we need a new layer? Can we just support video with the bitmap layer?

ibgreen:
> Why do we need a new layer? Can we just support video with the bitmap layer?

That was also my first though. An argument could be made from the discovery point-of-view. It may not be obvious to people that video textures are possible and how to use them.

It should make for a nice demo though. Maybe keep the new layer in the demo folder until it matures?

examples/website/video/app.js
Pessimistress: You can trigger layer redraw by calling `layer.setNeedsRedraw` from inside the layer.

Pessimistress:
> It may not be obvious to people that video textures are possible and how to use them.

The support for video sources can just be added to the documentation. It's possible for the layer to watch whether the video is playing and update itself, so no additional changes are required from the user's perspective.

Pessimistress:
@xintongxia Can you add unit tests for this layer?

modules/layers/src/bitmap-layer/bitmap-layer.js
```diff
@@ -29,10 +29,9 @@ const {fp64LowPart} = fp64;
 import vs from './bitmap-layer-vertex';
 import fs from './bitmap-layer-fragment';
 
-const DEFAULT_TEXTURE_PARAMETERS = {
-  [GL.TEXTURE_MIN_FILTER]: GL.LINEAR_MIPMAP_LINEAR,
-  // GL.LINEAR is the default value but explicitly set it here
-  [GL.TEXTURE_MAG_FILTER]: GL.LINEAR,
+export const DEFAULT_TEXTURE_PARAMETERS = {
+  [GL.TEXTURE_MIN_FILTER]: GL.NEAREST,
```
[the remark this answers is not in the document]
tsherif: I would say yes. You'll likely see textures looking pixelated, more artifacts at acute angles and it might take a performance hit for bigger textures (because of the mipmaps).

I'm not sure what you mean by handling the parameter fallback.

Would it make sense to have different defaults for NPOT textures?

modules/layers/src/bitmap-layer/bitmap-layer.js
```diff
@@ -180,9 +179,20 @@ export default class BitmapLayer extends Layer {
...
+    const {uniforms} = opts;
     const {bitmapTexture, model} = this.state;
-    const {desaturate, transparentColor, tintColor} = this.props;
+    const {image, desaturate, transparentColor, tintColor} = this.props;
+
+    // Update video frame
+    if (
+      bitmapTexture &&
+      image instanceof HTMLVideoElement &&
+      image.readyState > HTMLVideoElement.HAVE_METADATA
```
tsherif: Maybe check if the video is playing before updating?

modules/layers/src/bitmap-layer/bitmap-layer.js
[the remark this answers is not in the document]
Pessimistress: Would it make sense for `Texture2D` to automatically disable certain parameters if the size is NPOT?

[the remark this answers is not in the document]
tsherif: It looks like it should already be doing that: https://github.com/uber/luma.gl/blob/master/modules/webgl/src/classes/texture.js#L673

Maybe that logic's not working?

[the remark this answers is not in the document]
Pessimistress: So when we first create the texture, we don't have the dimensions yet (video may still be loading). We just set `width: 1, height: 1` so the default NPOT logic does not kick in. When we call `resize` later the parameters are never checked again.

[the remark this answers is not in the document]
xintongxia: luma does check and replace the parameters for NPOT, if I understand correctly. So here in BitmapLayer, I can keep the original params.

Pessimistress:
Name the PR to match your changes.

modules/layers/src/bitmap-layer/bitmap-layer.js
```diff
@@ -228,6 +239,16 @@ export default class BitmapLayer extends Layer {
...
         })
       });
+    } else if (image instanceof HTMLVideoElement) {
+      // Initialize an empty texture while we wait for the video to load
+      this.setState({
+        bitmapTexture: new Texture2D(gl, {
+          width: 1,
+          height: 1,
+          parameters: DEFAULT_TEXTURE_PARAMETERS,
+          mipmaps: false
```
Pessimistress: the default parameters wouldn't work with mipmaps turned off.

modules/layers/src/bitmap-layer/bitmap-layer.js
```diff
@@ -180,9 +179,21 @@ export default class BitmapLayer extends Layer {
...
+    const {image, desaturate, transparentColor, tintColor} = this.props;
+
+    // Update video frame
+    if (
+      bitmapTexture &&
+      image instanceof HTMLVideoElement &&
+      image.readyState > HTMLVideoElement.HAVE_METADATA
+    ) {
+      bitmapTexture.resize({width: image.videoWidth, height: image.videoHeight});
+      bitmapTexture.setParameters(DEFAULT_TEXTURE_PARAMETERS);
```
Pessimistress: Only call `setParameters` if the dimensions have changed.

modules/layers/src/bitmap-layer/bitmap-layer.js
```diff
@@ -180,9 +179,21 @@ export default class BitmapLayer extends Layer {
...
-    const {desaturate, transparentColor, tintColor} = this.props;
+    const {image, desaturate, transparentColor, tintColor} = this.props;
+
+    // Update video frame
+    if (
+      bitmapTexture &&
+      image instanceof HTMLVideoElement &&
+      image.readyState > HTMLVideoElement.HAVE_METADATA
+    ) {
+      bitmapTexture.resize({width: image.videoWidth, height: image.videoHeight});
```
Pessimistress: Looks like `resize` forces `mipmaps` to false? https://github.com/uber/luma.gl/blob/master/modules/webgl/src/classes/texture.js#L198

modules/layers/src/bitmap-layer/bitmap-layer.js
Pessimistress: true?

modules/layers/src/bitmap-layer/bitmap-layer.js
```diff
@@ -180,9 +179,34 @@ export default class BitmapLayer extends Layer {
...
+        bitmapTexture.width !== image.videoWidth || bitmapTexture.height !== image.videoHeight;
+      if (sizeChanged) {
+        // note clears image and mipmaps when resizing
+        bitmapTexture.resize({width: image.videoWidth, height: image.videoHeight, mipmaps: true});
+        bitmapTexture.setImageData({
+          data: image,
+          paramters: DEFAULT_TEXTURE_PARAMETERS
+        });
+      } else {
+        bitmapTexture.setImageData({
```
Pessimistress: subImageData

xintongxia: what is the difference between `subImageData` and `setImageData`?

Pessimistress: `subImageData` skips some allocation operations and upload the data directly.

PR #3205, opened 2019-06-06
Add viewport helper functions, add documentation, support saving state across kernel restarts

ajduberstein:
Add additional functionality to make the Python library more widely usable

- Support using a Pandas data frame for data ingestion (without actually including Pandas in the library dependencies)
- Add auto-zoom helper function which generate a viewport
- Add demo notebooks with a saved state


ajduberstein:
Added an interactive example of the brightness of lights at night in Chengdu over 20 years.

lights_at_night

bindings/python/pydeck/pydeck/data_utils/viewport_helpers.py
```diff
@@ -0,0 +1,164 @@
...
+    if proportion == 1:
+        return points
+    # Compute the medioid of the data
+    centroid = geometric_mean(points)
+    # Retain the closest n*proportion points
+    n_to_keep = math.floor(proportion * len(points))
+    return k_nearest_neighbors(points, centroid, n_to_keep)
+
+
+def bbox_to_zoom_level(bbox):
```
Pessimistress: Use [this function](https://github.com/uber-common/viewport-mercator-project/blob/master/src/web-mercator-utils.js#L41) to project lngLat coordinates onto the WebMercator plane. A unit on the WebMercator plane can be considered equivalent to a pixel. The proper zoom level should be dependent on the width and height of the viewport.

bindings/python/pydeck/pydeck/bindings/layer.py
```diff
@@ -72,25 +33,128 @@ def __init__(
         elevation_range=None,
         elevation_scale=None,
         extruded=None,
-        upper_percentile=100,
+        upper_percentile=None,
```
Pessimistress: Will it make sense to automatically generate this list from each layer's `defaultProps` list?

ibgreen: It seems to me that the best approach would be for the pydeck layer to not specify defaults, and instead just omit props when generating the JSON payload if they were not specified in python. 

That way the default prop values implemented in deck.gl (and described in deck.gl docs) will always take effect.

Since different layers could potentially have different defaults for a similarly named prop, trying to collect a superset of all props here might not work.

Maybe setting default to `None` here and not adding them to JSON payload in that case?

ajduberstein: If `None` is passed, the default within deck.gl is actually used. The `layer` object in this file is converted to JSON, and if a property doesn't have a specified value, it doesn't get passed into the deck object.

Issue #3209, opened 2019-06-06
H3HexagonLayer: Add support for coverage prop.


1chandu:
Add support for `coverage` prop in `H3HexagonLayer`

- [x] ColumnLayer: Forward `coverage` prop to `ColumnLayer`
- [x] PolygonLayer: Scale polygon vertices relative to the center.

#### To Do List 
  - [x] Add label and assign to milestone 
  - [x] Coding
  - [x] Doc update
  - [ ] What’s new update 
  - [x] Test

PR #3210, opened 2019-06-06
H3Hexagon: add support for coverage.

1chandu:
For #3209 

#### Background
- Forward `coverage` prop to `ColumnLayer`
- Scale polygon vertices when using `PolygonLayer`.

#### Change List
- H3Hexagon: add support for coverage.

H3Coverage

### DIFF
--- a/modules/geo-layers/src/h3-layers/h3-hexagon-layer.js
+++ b/modules/geo-layers/src/h3-layers/h3-hexagon-layer.js
@@ -18,8 +18,8 @@ import {ColumnLayer, PolygonLayer} from '@deck.gl/layers';
 const UPDATE_THRESHOLD_KM = 10;
 
 // normalize longitudes w.r.t center `refLng`, when not provided first vertex
-export function normalizeLongitudes(vertices, refLng = null) {
-  refLng = refLng || vertices[0][0];
+export function normalizeLongitudes(vertices, refLng) {
+  refLng = refLng === undefined ? vertices[0][0] : refLng;
   for (const pt of vertices) {
     const deltaLng = pt[0] - refLng;
     if (deltaLng > 180) {
@@ -57,12 +57,12 @@ function h3ToPolygon(hexId, coverage = 1) {
   const vertices = h3ToGeoBoundary(hexId, true);
 
   if (coverage !== 1) {
-    // updates array object elements of vertices array.
+    // scale and normalize vertices w.r.t to center
     scalePolygon(hexId, vertices, coverage);
+  } else {
+    // normalize w.r.t to start vertex
+    normalizeLongitudes(vertices);
   }
-
-  // normalize with respect start vertex
-  normalizeLongitudes(vertices);
 
   return vertices;
 }
--- a/modules/geo-layers/src/h3-layers/h3-hexagon-layer.js
+++ b/modules/geo-layers/src/h3-layers/h3-hexagon-layer.js
@@ -17,7 +17,7 @@ import {ColumnLayer, PolygonLayer} from '@deck.gl/layers';
 // distortion." Smaller value makes the column layer more sensitive to viewport change.
 const UPDATE_THRESHOLD_KM = 10;
 
-// normalize longitudes w.r.t center `refLng`, when not provided first vertex
+// normalize longitudes w.r.t center (refLng), when not provided first vertex
 export function normalizeLongitudes(vertices, refLng) {
   refLng = refLng === undefined ? vertices[0][0] : refLng;
   for (const pt of vertices) {
--- a/modules/geo-layers/src/h3-layers/h3-hexagon-layer.js
+++ b/modules/geo-layers/src/h3-layers/h3-hexagon-layer.js
@@ -23,12 +23,13 @@ function getHexagonCentroid(getHexagon, object, objectInfo) {
   return [lng, lat];
 }
 
-function scalePolygon(hexId, vertices, factor) {
+export function scalePolygon(hexId, vertices, factor) {
   const [lat, lng] = h3ToGeo(hexId);
+  const vertexCount = vertices.length;
+  log.assert(vertices[0] === vertices[vertexCount - 1]);
   // `h3ToGeoBoundary` returns same array object for first and last vertex (closed polygon),
   // hence skip scaling the last vertex
-  log.assert(vertices[0] === vertices[vertices.length - 1]);
-  for (let i = 0; i < vertices.length - 1; i++) {
+  for (let i = 0; i < vertexCount - 1; i++) {
     vertices[i][0] = lerp(lng, vertices[i][0], factor);
     vertices[i][1] = lerp(lat, vertices[i][1], factor);
   }
@@ -54,16 +55,14 @@ function h3ToPolygon(hexId, coverage = 1) {
   return vertices;
 }
 
-function mergeTriggers(currentTrigger, newTrigger) {
+function mergeTriggers(getHexagon, coverage) {
   let trigger;
-  if (currentTrigger === undefined || currentTrigger === null) {
-    trigger = newTrigger.coverage;
-  } else if (Array.isArray(currentTrigger)) {
-    trigger = [].concat(currentTrigger, newTrigger.coverage);
-  } else if (typeof currentTrigger === 'object') {
-    trigger = Object.assign({}, currentTrigger, newTrigger);
+  if (getHexagon === undefined || getHexagon === null) {
+    trigger = coverage;
+  } else if (typeof getHexagon === 'object') {
+    trigger = Object.assign({}, getHexagon, {coverage});
   } else {
-    trigger = [currentTrigger, newTrigger.coverage];
+    trigger = {getHexagon, coverage};
   }
   return trigger;
 }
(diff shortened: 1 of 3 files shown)


modules/geo-layers/src/h3-layers/h3-hexagon-layer.js
```diff
@@ -22,9 +23,27 @@ function getHexagonCentroid(getHexagon, object, objectInfo) {
   return [lng, lat];
 }
 
-function h3ToPolygon(hexId) {
+function scalePolygon(hexId, vertices, factor) {
+  const [lat, lng] = h3ToGeo(hexId);
+  // get distinct vertex objects,
+  // `h3ToGeoBoundary` returns same array object for first and last vertex (closed polygon)
+  const uniqueVertices = [...new Set(vertices)];
```
Pessimistress: - Don't use spread operator in this repo
- This is unnecessary. the construction of new objects is more expensive than transforming one additional point. `h3ToPolygon` should consistently return the same format regardless of `coverage`.

modules/geo-layers/src/h3-layers/h3-hexagon-layer.js
```diff
@@ -22,9 +23,27 @@ function getHexagonCentroid(getHexagon, object, objectInfo) {
...
+  for (const vertex of uniqueVertices) {
+    const scaledPt = lerp([lng, lat], vertex, factor);
```
Pessimistress: Do not create new array in for loop. You can simply call `lerp` twice.

modules/geo-layers/src/h3-layers/h3-hexagon-layer.js
```diff
@@ -36,6 +55,18 @@ function h3ToPolygon(hexId) {
   return vertices;
 }
 
+function mergeTriggers(currentTrigger, newTrigger) {
+  let trigger;
+  if (!currentTrigger) {
+    trigger = newTrigger;
+  } else if (Array.isArray(currentTrigger)) {
+    trigger = [...currentTrigger, newTrigger];
```
Pessimistress: Perf: use `concat`

modules/geo-layers/src/h3-layers/h3-hexagon-layer.js
```diff
@@ -36,6 +55,18 @@ function h3ToPolygon(hexId) {
...
+  } else {
+    trigger = [currentTrigger, newTrigger];
```
Pessimistress: `currentTrigger` can be an object.

modules/geo-layers/src/h3-layers/h3-hexagon-layer.js
1chandu: `h3ToGeoBoundary` is returning same object (Array object) for first and last vertex , changing one changes the other, I will have WA here (just skip the last), and will open bug for h3-js.

Pessimistress: I see. I don't think this is necessarily a bug for h3-js, but indeed a catch.

1chandu:
@Pessimistress , addressed comments.

deck.gl 7.0 supported the coverage for non `highPrecision` mode (we were forwarding the prop), any concerns cherry-picking this into 7.1-release branch ?

modules/geo-layers/src/h3-layers/h3-hexagon-layer.js
```diff
@@ -22,9 +23,26 @@ function getHexagonCentroid(getHexagon, object, objectInfo) {
   return [lng, lat];
 }
 
-function h3ToPolygon(hexId) {
+function scalePolygon(hexId, vertices, factor) {
+  const [lat, lng] = h3ToGeo(hexId);
+  // `h3ToGeoBoundary` returns same array object for first and last vertex (closed polygon),
+  // hence skip scaling the last vertex
+  log.assert(vertices[0] === vertices[vertices.length - 1]);
```
Pessimistress: Perf: save `vertices.length` in a variable:

```
const n = vertices.length;
for (let i = 1; i < n; i++) {
  ...
}
```

modules/geo-layers/src/h3-layers/h3-hexagon-layer.js
```diff
@@ -36,6 +54,20 @@ function h3ToPolygon(hexId) {
   return vertices;
 }
 
+function mergeTriggers(currentTrigger, newTrigger) {
```
Pessimistress: What's the point of creating a `newTrigger` object if you're just going to use `newTrigger.cover`?

modules/geo-layers/src/h3-layers/h3-hexagon-layer.js
```diff
@@ -36,6 +54,20 @@ function h3ToPolygon(hexId) {
...
+  let trigger;
+  if (currentTrigger === undefined || currentTrigger === null) {
+    trigger = newTrigger.coverage;
+  } else if (Array.isArray(currentTrigger)) {
+    trigger = [].concat(currentTrigger, newTrigger.coverage);
```
Pessimistress: `Object.assign({}, <array>)` works, you don't have to handle it as a special case.

modules/geo-layers/src/h3-layers/h3-hexagon-layer.js
```diff
@@ -36,6 +54,20 @@ function h3ToPolygon(hexId) {
...
+  } else if (typeof currentTrigger === 'object') {
+    trigger = Object.assign({}, currentTrigger, newTrigger);
+  } else {
+    trigger = [currentTrigger, newTrigger.coverage];
```
Pessimistress: `{currentTrigger, coverage}`

modules/geo-layers/src/h3-layers/h3-hexagon-layer.js
```diff
@@ -22,9 +23,26 @@ function getHexagonCentroid(getHexagon, object, objectInfo) {
   return [lng, lat];
 }
 
-function h3ToPolygon(hexId) {
+function scalePolygon(hexId, vertices, factor) {
```
Pessimistress: Add a test for this function?

modules/geo-layers/src/h3-layers/h3-hexagon-layer.js
1chandu: Either I have to pass value and do {coverage: value} or pass an object and access it as object.coverage. Changed it to latter former case.

modules/geo-layers/src/h3-layers/h3-hexagon-layer.js
```diff
@@ -22,9 +23,27 @@ function getHexagonCentroid(getHexagon, object, objectInfo) {
...
+  }
+}
+
+function h3ToPolygon(hexId, coverage = 1) {
   const vertices = h3ToGeoBoundary(hexId, true);
   const refLng = vertices[0][0];
+
+  if (coverage !== 1) {
+    // updates array object elements of vertices array.
+    scalePolygon(hexId, vertices, coverage);
```
Pessimistress: This breaks if the hexagon crosses the 180th meridian. You need to set the refLng to that of the center, normalize it before scaling.

modules/geo-layers/src/h3-layers/h3-hexagon-layer.js
```diff
@@ -22,9 +23,27 @@ function getHexagonCentroid(getHexagon, object, objectInfo) {
   return [lng, lat];
 }
 
-function h3ToPolygon(hexId) {
+export function scalePolygon(hexId, vertices, factor) {
+  const [lat, lng] = h3ToGeo(hexId);
+  const vertexCount = vertices.length;
+  log.assert(vertices[0] === vertices[vertexCount - 1]);
```
Pessimistress: This assert will break the layer of h3-js makes a change in the future. It won't be the user's fault and it won't provide any useful error message. Can we just adapt the for loop based on whether these two points are the same object?

test/modules/geo-layers/h3-layers.spec.js
```diff
@@ -50,6 +54,121 @@ test('H3HexagonLayer', t => {
   t.end();
 });
 
+test.only('H3HexagonLayer#mergeTriggers', t => {
```
Pessimistress: test.only

modules/geo-layers/src/h3-layers/h3-hexagon-layer.js
1chandu: Done.

modules/geo-layers/src/h3-layers/h3-hexagon-layer.js
1chandu: Done.

modules/geo-layers/src/h3-layers/h3-hexagon-layer.js
```diff
@@ -33,9 +28,57 @@ function h3ToPolygon(hexId) {
...
+function h3ToPolygon(hexId, coverage = 1) {
+  const vertices = h3ToGeoBoundary(hexId, true);
+
+  if (coverage !== 1) {
+    // updates array object elements of vertices array.
+    scalePolygon(hexId, vertices, coverage);
+  }
+
+  // normalize with respect start vertex
+  normalizeLongitudes(vertices);
```
1chandu: I wonder we should move this into `PolygonLayer` or it's sublayers.

PR #3218, opened 2019-06-07
Add layer shader hooks rfc

Pessimistress:
Related to #2524 	

#### Background

See RFC content.

### DIFF
--- a/dev-docs/RFCs/v7.x/layer-shader-hook-rfc.md
+++ b/dev-docs/RFCs/v7.x/layer-shader-hook-rfc.md
@@ -1,4 +1,4 @@
-# RFC: Using Binary Data
+# RFC: Layer Shader Hooks
 
 * Authors: Xiaoji Chen
 * Date: June 7, 2019
@@ -21,14 +21,14 @@ While the second approach is a little more robust by dynamically adding custom c
 - The subclass is difficult to read, as it does not contain the full shader code.
 - The injection usually depends on certain assumptions about the base layer implementation, e.g. attribute names and varyings, which is prone to breakage in a future release.
 - Because each layer's attributes are different, the injection code cannot be ported to other layers easily.
-- The standard injection points (`decl`, `main-start`, `main-end`) is very limiting. If the subclass wishes to manipulate intermediate values in the shader, e.g. `offset` in `ScatterplotLayer`'s vertex shader before projection is performed, custom code must be injected using key strings in the base shader, which is even more fragile.
+- The standard injection points (`decl`, `main-start`, `main-end`) are very limiting. If the subclass wishes to manipulate intermediate values in the shader, e.g. `offset` in `ScatterplotLayer`'s vertex shader before projection is performed, custom code must be injected using key strings in the base shader, which is even more fragile.
 
 ## Target Use Cases
 
 Publicly documented shader hooks make it possible to:
 
 - Standardize layer shaders and make them easier to tweak for people who are less familiar with WebGL.
-- Write one set of injection that works for many layers.
+- Write one set of injections that works for many layers.
 - Write custom layers that remain working for future releases of the same major version.
 - As a result of the above, package and distribute layers and layer extensions as reusable components.
 
@@ -65,7 +65,7 @@ Shader hook names follow the pattern `DECKGL_<action>_<target>`.
 
 Modify the commonspace dimensions associated with the current vertex, before it's projected.
 
-Sementic meanings:
+Semantic meanings:
 
 - ScatterplotLayer/IconLayer/PointCloudLayer/ColumnLayer: the offset from center
 - PathLayer/ArcLayer/LineLayer: the extrusion from the current vertex
@@ -75,9 +75,9 @@ Sementic meanings:
 
 #### vs:DECKGL_USE_POSITION(vec3 worldPosition)
 
-Observe the world space position associated with a point-link object.
+Observe the world space position associated with a point-like object.
 
-Sementic meanings:
+Semantic meanings:
 
 - SolidPolygonLayer (top): the vertex position (from `getPolygon`)
 - BitmapLayer: the vertex position (from `bounds`)
@@ -89,7 +89,7 @@ Sementic meanings:
 
 Observe the world space source and target positions associated with an edge-like object.
 
-Sementic meanings:
+Semantic meanings:
 
 - ArcLayer/LineLayer: the source/target positions (from `getSourcePosition` and `getTargetPosition`)
 - PathLayer: the start/end of the current leg (from `getPath`)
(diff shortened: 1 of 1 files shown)


dev-docs/RFCs/v7.x/layer-shader-hook-rfc.md
```diff
@@ -0,0 +1,107 @@
+# RFC: Using Binary Data
```
tsherif: Wrong title

dev-docs/RFCs/v7.x/layer-shader-hook-rfc.md
```diff
@@ -0,0 +1,107 @@
...
+- Our [brushing example](https://github.com/uber/deck.gl/tree/7.1-release/examples/website/brushing/brushing-layers) and [TripsLayer](https://github.com/uber/deck.gl/blob/7.1-release/modules/geo-layers/src/trips-layer/trips-layer.js) use shader injection.
+ 
+While the second approach is a little more robust by dynamically adding custom code to the base shader, the following issues remain:
+
+- The subclass is difficult to read, as it does not contain the full shader code.
+- The injection usually depends on certain assumptions about the base layer implementation, e.g. attribute names and varyings, which is prone to breakage in a future release.
+- Because each layer's attributes are different, the injection code cannot be ported to other layers easily.
+- The standard injection points (`decl`, `main-start`, `main-end`) is very limiting. If the subclass wishes to manipulate intermediate values in the shader, e.g. `offset` in `ScatterplotLayer`'s vertex shader before projection is performed, custom code must be injected using key strings in the base shader, which is even more fragile.
```
tsherif: are very limiting

dev-docs/RFCs/v7.x/layer-shader-hook-rfc.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Target Use Cases
+
+Publicly documented shader hooks make it possible to:
+
+- Standardize layer shaders and make them easier to tweak for people who are less familiar with WebGL.
+- Write one set of injection that works for many layers.
```
tsherif: injections

dev-docs/RFCs/v7.x/layer-shader-hook-rfc.md
```diff
@@ -0,0 +1,107 @@
...
+  + Coordinates are presumed to be in common space.
+  + Non-commonspace coordinates must contain modifiers in their names, e.g. `worldPosition`, `pixelSize`.
+
+### Standard Layer Hooks
+
+#### vs:DECKGL_FILTER_SIZE(inout vec3 size)
+
+Modify the commonspace dimensions associated with the current vertex, before it's projected.
+
+Sementic meanings:
```
tsherif: Semantic

dev-docs/RFCs/v7.x/layer-shader-hook-rfc.md
```diff
@@ -0,0 +1,107 @@
...
+  + `FILTER`: if the hook can modify the input value(s).
+  + `USE`: if the hook does not expect any output.
+- Targets:
+  + Colors are normalized to the WebGL color space.
+  + Coordinates are presumed to be in common space.
+  + Non-commonspace coordinates must contain modifiers in their names, e.g. `worldPosition`, `pixelSize`.
+
+### Standard Layer Hooks
+
+#### vs:DECKGL_FILTER_SIZE(inout vec3 size)
```
jianhuang01: how about DECKGL_FILTER_COMMON_SPACE_DIMENSION? DECKGL_FILTER_SIZE is pretty vague.

dev-docs/RFCs/v7.x/layer-shader-hook-rfc.md
```diff
@@ -0,0 +1,107 @@
...
+
+Semantic meanings:
+
+- ScatterplotLayer/IconLayer/PointCloudLayer/ColumnLayer: the offset from center
+- PathLayer/ArcLayer/LineLayer: the extrusion from the current vertex
+- SimpleMeshLayer: the local position
+- SolidPolygonLayer/BitmapLayer: (not applicable)
+
+
+#### vs:DECKGL_USE_POSITION(vec3 worldPosition)
```
jianhuang01: How about DECKGL_USE_WORLD_POSITION?

dev-docs/RFCs/v7.x/layer-shader-hook-rfc.md
```diff
@@ -0,0 +1,107 @@
...
+
+Semantic meanings:
+
+- SolidPolygonLayer (top): the vertex position (from `getPolygon`)
+- BitmapLayer: the vertex position (from `bounds`)
+- PathLayer/ArcLayer/LineLayer: (not applicable)
+- All other layers: the object center (from `getPosition`)
+
+
+#### vs:DECKGL_USE_POSITIONS(vec3 sourceWorldPosition, vec3 targetWorldPosition)
```
jianhuang01: DECKGL_USE_WORLD_POSITIONS?

jianhuang01:
@Pessimistress For shadow feature, only need this VS hook
vs:DECKGL_FILTER_GL_POSITION(inout vec4 position, vec4 position_commonspace), the FS hook is already covered in your proposal.

dev-docs/RFCs/v7.x/layer-shader-hook-rfc.md
```diff
@@ -0,0 +1,107 @@
...
+
+Semantic meanings:
+
+- ArcLayer/LineLayer: the source/target positions (from `getSourcePosition` and `getTargetPosition`)
+- PathLayer: the start/end of the current leg (from `getPath`)
+- SolidPolygonLayer (side): the two vertices that form the current edge (from `getPolygon`)
+- All other layers: (not applicable)
+
+
+#### vs:DECKGL_FILTER_COLOR(inout vec4 color, vec3 position, vec3 normal)
```
jianhuang01: is the position.w always `1` in all layers? so we skip it here?

Pessimistress: `w` is not always `1`. But your lighting module ignores it.

dev-docs/RFCs/v7.x/layer-shader-hook-rfc.md
Pessimistress: All shader names default to common space if not specified, see the "naming conventions" section.

Pessimistress:
@jianhuang01 	what is the first `position`?

jianhuang01:
> @jianhuang01 what is the first `position`?

gl_Position

Pessimistress:
Add unit tests for this layer.

tsherif:
> One thing that makes the current system somewhat "fragile" is the fixed parameter lists being passed in to each hook. Discovering that another argument is needed by an extension could lead to breaking changes.

Keep in mind that injections are based on the hook name, not the arguments, so if you add arguments to a hook, injections will still be inserted correctly and just won't use the new arguments. There's no risk of breakage in the case you mentioned.

In general, controlling the interface between the main shader and the hooks was a design goal of this system. To me, it increases maintainability in the same way that a function with specific arguments that operates on local variables is more maintainable than one that just operates on global variables. 

> An approach that could mitigate this is to use interface modules.

This could potentially work in conjunction with the hook system. It would be adding another layer of indirection though, which I think would make shaders that use them harder to debug. They are also essentially global variables, so I would worry about interactions between the hook code and the main shader code.

PR #3220, opened 2019-06-10
Layer browser perf and bug fixes

Pessimistress:
#### Change List
- Separate state handling and layer generation into two components (no longer re-renders examples on view state change and/or hover)
- Fix `OrbitView` usage
- Fix `MapView` bug with `position`

### DIFF
--- a/examples/layer-browser/src/map.js
+++ b/examples/layer-browser/src/map.js
@@ -1,0 +1,169 @@
+import React, {PureComponent} from 'react';
+import {StaticMap, _MapContext as MapContext, NavigationControl} from 'react-map-gl';
+import autobind from 'react-autobind';
+
+import DeckGL from '@deck.gl/react';
+import {COORDINATE_SYSTEM, View} from '@deck.gl/core';
+
+import LayerInfo from './components/layer-info';
+
+/* eslint-disable no-process-env */
+const MapboxAccessToken =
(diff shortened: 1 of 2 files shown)


examples/layer-browser/src/map.js
```diff
@@ -0,0 +1,169 @@
...
+
+  _onViewStateChange({viewState, viewId}) {
+    if (viewId === 'infovis') {
+      this.setState({orbitViewState: viewState});
+      return;
+    }
+    if (viewState.pitch > 60) {
+      viewState.pitch = 60;
+    }
+    viewState.position = [0, 0, 50];
```
xintongxia: Why z is `50`?

examples/layer-browser/src/map.js
```diff
@@ -0,0 +1,169 @@
...
+
+const NAVIGATION_CONTROL_STYLES = {
+  margin: 10,
+  position: 'absolute',
+  zIndex: 1
+};
+
+const VIEW_LABEL_STYLES = {
+  zIndex: 10,
+  // position: 'relative',
```
xintongxia: remove?

examples/layer-browser/src/map.js
Pessimistress: Arbitrary number for testing first person view.

PR #3224, opened 2019-06-11
Data Partial Update

Pessimistress:
For #1576 

#### Change List
- Expose the partial update feature with a `_dataDiff` prop
- Support partial update in all primitive layers
- Add a test application
- Documentation

### DIFF
--- a/docs/api-reference/layer.md
+++ b/docs/api-reference/layer.md
@@ -235,6 +235,39 @@ Used to override the default shallow comparison of the `data` object.
 
 As an illustration, the app could set this to e.g. 'lodash.isequal', enabling deep comparison of the data structure. This particular examples would obviously have considerable performance impact and should only be used as a temporary solution for small data sets until the application can be refactored to avoid the need.
 
+##### `_dataDiff` (Function, optional) **Experimental**
+
+This function is called when the data changes, either shallowly when `dataComparator` is not supplied or because `dataComparator` returns `false`, to retrieve the indices of the changed objects. By default, when data changes, the attributes of all objects are recalculated. If this prop is supplied, only the attributes of the specified objects will be updated. This can lead to significant performance improvement if a few rows in a large data table need to change frequently:
+
+```js
+let data = [...];
+
+function updateData(index, item) {
+  // make a shallow copy
+  data = data.slice();
+  data.splice(index, 1, item);
+  const layer = new ScatterplotLayer({
+    data,
+    // Only update the attributes at `index``
+    _dataDiff: (oldData, newData) => [{startRow: index, endRow: index + 1}],
+    ...
+  });
+  deck.setProps({layers: [layer]});
+}
+```
+
+The function receives two arguments:
+
+- `oldData` - the data prop before the update
+- `newData` - the new data prop
+
+And is expected to return an array of "ranges", in the form of `{startRow, endRow}`:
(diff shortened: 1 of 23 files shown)


docs/api-reference/layer.md
```diff
@@ -235,6 +235,39 @@ Used to override the default shallow comparison of the `data` object.
 
 As an illustration, the app could set this to e.g. 'lodash.isequal', enabling deep comparison of the data structure. This particular examples would obviously have considerable performance impact and should only be used as a temporary solution for small data sets until the application can be refactored to avoid the need.
 
+##### `_dataDiff` (Function, optional) **Experimental**
+
+This function is called when the data changes, either shallowly when `dataComparator` is not supplied or because `dataComparator` returns `false`, to retrieve the indices of the changed objects. By default, when data changes, the attributes of all objects are recalculated. If this prop is supplied, only the attributes of the specified objects will be updated. This can lead to significant performance improvement if a few rows in a large data table need to change frequently:
```
1chandu: Nit:  `the attributes of all objects are recalculated` to `the attributes of all instances are recalculated` ?  Or clarify `object` is a rendering instance.

modules/layers/src/path-layer/path-layer.js
```diff
@@ -145,13 +145,18 @@ export default class PathLayer extends Layer {
...
-        fp64: this.use64bitPositions()
+        fp64: this.use64bitPositions(),
+        dataChanged: changeFlags.dataChanged
       });
       this.setState({
         numInstances: pathTesselator.instanceCount,
         bufferLayout: pathTesselator.bufferLayout
       });
-      attributeManager.invalidateAll();
+      if (!changeFlags.dataChanged) {
```
1chandu: I am assuming this `if` should test if any thing else changed other than `changeFlags.dataChanged`, what if `changeFlags.dataChanged` is true and also some other flag, like fp64 too? Shouldn't we exclusively check for all other cases?

modules/layers/src/solid-polygon-layer/solid-polygon-layer.js
```diff
@@ -264,15 +264,20 @@ export default class SolidPolygonLayer extends Layer {
...
+        dataChanged: changeFlags.dataChanged
       });
 
       this.setState({
         numInstances: polygonTesselator.instanceCount,
         bufferLayout: polygonTesselator.bufferLayout
       });
 
-      this.getAttributeManager().invalidateAll();
+      if (!changeFlags.dataChanged) {
```
1chandu: Same as above.

1chandu:
Also some unit tests to cover new code changes.

docs/api-reference/layer.md
Pessimistress: We do not have the concept "instance" in our documentation. From the user's perspective, an "object" is an item in the data array.

modules/layers/src/path-layer/path-layer.js
Pessimistress: If `changeFlags.dataChanged` is `true`, the base layer `updateState` calls `attributeManager.invalidateAll()`. If `changeFlags.dataChanged` is a range, then the user has specifically requested a partial update, and therefore responsible for making sure that the rest of the data are not changed (see remarks in documentation).

1chandu: 👍

PR #3230, opened 2019-06-12
Add unit tests for controllers

Pessimistress:


test/modules/core/controllers/test-controller.js
```diff
@@ -0,0 +1,195 @@
...
+    }
+    // reset
+    onViewStateChangeCalled = 0;
+    onStateChangeCalled = 0;
+
+    controller.setProps(Object.assign({}, defaultProps, testCase.props));
+    for (const event of testCase.events) {
+      controller.handleEvent(event);
+    }
+    t.is(onViewStateChangeCalled, testCase.viewStateChanges, `${testCase.title} onViewStateChange`);
```
xintongxia: This is good enough, but I wonder if it is possible to check the specific events in sequence which are supposed to be triggered in each case.

modules/core/src/controllers/controller.js
```diff
@@ -265,9 +268,6 @@ export default class Controller {
       return false;
     }
     const pos = this.getCenter(event);
-    if (!this.isDragging()) {
```
xintongxia: which this is removed?

Pessimistress: This is moved to `onPan`.

Issue #3233, opened 2019-06-12
CompositeLayer: Allow to override `updateTriggers` using `_subLayerProps`.


1chandu:
#### Target Use case

Using `_subLayerProps` one can provide a new Layer type or just the props to be used for a Sublayer. Some customizations might also need to provide additional update triggers.

#### Proposed feature

Allow `updateTriggers` to be specified using `_subLayerProps`.

Example: use customized `ColumnLayer` with a additional updateTrigger (required for the added customization)

```js
import {ColumnLayer} from '@deck.gl/layers';
import {H3HexagonLayer} from '@deck.gl/geo-layers';

class CustomizedColumnLayer extends ColumnLayer {
  // customization that adds new accessor prop `getCoverage` and hence requires new updateTrigger
  // following code show how to provide this update trigger
  // ...
}

new H3HexagonLayer({
  // ...other props
  _subLayerProps: {
    'hexagon-cell': {
      type: CustomizedColumnLayer,
      getCoverage: data.getCoverage,
      updateTriggers: {getCoverage: getCoverageTrigger}
    }
  }
});
```

#### To Do List 
  - [x] Add label and assign to milestone 
  - [x] Coding
  - [x] Doc update
  - [ ] What’s new update 
  - [x] Test

PR #3234, opened 2019-06-12
CompositeLayer: add ability to override sublayer updateTriggers

1chandu:
For #3233 

#### Background
Allow to override `updateTriggers` using `_subLayerProps`.
Add unit tests and update docs.

#### Change List
- CompositeLayer: add ability to override sublayer updateTriggers

### DIFF
--- a/docs/api-reference/composite-layer.md
+++ b/docs/api-reference/composite-layer.md
@@ -78,6 +78,30 @@ new GeoJsonLayer({
 });
 ```
 
+Example: use customized `ColumnLayer` with a additional updateTrigger (required for the added customization)
+
+```js
+import {ColumnLayer} from '@deck.gl/layers';
+import {H3HexagonLayer} from '@deck.gl/geo-layers';
+
+class CustomizedColumnLayer extends ColumnLayer {
+  // customization that adds new accessor prop `getCoverage` and hence requires new updateTrigger
+  // following code show how to provide this update trigger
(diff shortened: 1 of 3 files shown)


docs/api-reference/composite-layer.md
```diff
@@ -78,6 +78,30 @@ new GeoJsonLayer({
 });
 ```
 
+Example: use customized `ColumnLayer` with a additional updateTrigger (required for the added customization)
```
Pessimistress: This example may be confusing because it's missing a lot of context. Can you just add an overriding `updateTrigger` e.g. `getIcon` to the old example?

test/modules/core/lib/composite-layer.spec.js
```diff
@@ -81,3 +104,82 @@ test('CompositeLayer#getSubLayerProps', t => {
 
   t.end();
 });
+
+test.only('CompositeLayer#getSubLayerProps(override)', t => {
```
Pessimistress: Keep an eye on the coverage please.

1chandu: Oops, my bad. I wonder we can add lint rule for this. Will update.

docs/api-reference/composite-layer.md
1chandu: Yes, incremental example is good idea, updated.

Issue #2726, opened 2019-02-27
onViewStateChange return abnormal interactionState property when zoom or pan the map


hijiangtao:
According to codes in master branch, nothing is passed to `updateViewport` method when trigger the mouse wheel event, so `interactionState` is an empty object all the time.

However, other event like pinch will [pass](https://github.com/uber/deck.gl/blob/master/modules/core/src/controllers/controller.js#L314) the property as expected:

```
return this.updateViewport(newControllerState, NO_TRANSITION_PROPS, {isDragging: true});
```

According to react-map-gl [documents](https://uber.github.io/react-map-gl/#/Documentation/api-reference/interactive-map?section=transitions), onViewStateChange should receive three parameters, which are viewState, interactionState and oldViewState. I suspect deck.gl may follow the same data format (I console.log the result when viewState has changed, see the result below), but there's still some TODO that needs to be implemented.

```
{
	interactionState: {},
	oldViewState: {width: 1169, height: 840, latitude: 4.462305801893527, longitude: 108.44194451498399, zoom: 2.3687270141337544, …},
	viewId: "default-view",
	viewState: {width: 1169, height: 840, latitude: 7.4917905251536565, longitude: 108.06177999147857, zoom: 2.5395902402201074, …},
}
```

deck.gl may need to change this [line](https://github.com/uber/deck.gl/blob/master/modules/core/src/controllers/controller.js#L338) code to:

```
return this.updateViewport(newControllerState, NO_TRANSITION_PROPS, {isZooming: true});
```

On the other hand, we may want to notice developers that `isPanning` is `true` when corresponding event is fired, so we may also add a property called `isPanning` to this line https://github.com/uber/deck.gl/blob/master/modules/core/src/controllers/controller.js#L288

```
return this.updateViewport(newControllerState, NO_TRANSITION_PROPS, {
  isDragging: true,
  isPanning: true,
});
```

jianhuang01:
@hijiangtao Thanks for notifying us, actually I don't think there is any bug on these two lines, our controller doc is [here](http://deck.gl/#/documentation/deckgl-api-reference/controllers/controller?section=constructor), if you think there is bug, follow our bug report template, please.

hijiangtao:
@jianhuang01 Actually, in my opinion, it can be handled in a more elegant way, since deck.gl provides the `interactionState` property, and react-map-gl is more friendly about this situation.

Thanks for replying, I will close it since you didn't think it's a bug of deck.gl.

PR #3235, opened 2019-06-12
Align controller behavior with react-map-gl

Pessimistress:
For #2726 

#### Change List
- Add more flags to `interactionState`
- Tests


Pessimistress:
> we could extend the controller from react-map-gl.

This has been discussed before. deck.gl cannot use react-map-gl as a dependency.

PR #3244, opened 2019-06-14
Fix picking rect calculation

Pessimistress:
#### Background

The bug: {y: 0, radius: 0} yields an invalid picking rect. This is because we are actually picking the pixel above (x, y) when `pickingRadius` is `0`.

#### Change List
- Fix how `getPickingRect` and `getClosestObject` handle `pickingRadius: 0`
- Tests

### DIFF
--- a/modules/core/src/lib/deck-picker.js
+++ b/modules/core/src/lib/deck-picker.js
@@ -176,6 +176,7 @@ export default class DeckPicker {
         deviceRadius,
         deviceRect
       });
+
       // Only exclude if we need to run picking again.
       // We need to run picking again if an object is detected AND
       // we have not exhausted the requested depth.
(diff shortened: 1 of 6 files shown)


test/modules/core/lib/pick-layers.spec.js
```diff
@@ -54,12 +54,12 @@ const NEW_GRID_LAYER_PICK_METHODS = {
...
       parameters: {
         x: 300,
-        y: 300
+        y: 209
       },
       results: {
         count: 1,
         // point count in the aggregated cell for each pickInfo object
-        cellCounts: [1]
+        cellCounts: [8]
```
Pessimistress: @1chandu Not sure why `cellCounts` changed...

modules/core/src/lib/deck-picker.js
```diff
@@ -318,19 +319,17 @@ export default class DeckPicker {
   // Calculate a picking rect centered on deviceX and deviceY and clipped to device
   // Returns null if pixel is outside of device
   getPickingRect({deviceX, deviceY, deviceRadius, deviceWidth, deviceHeight}) {
-    const valid = deviceX >= 0 && deviceY >= 0 && deviceX < deviceWidth && deviceY < deviceHeight;
+    // Create a box of size `radius * 2 + 1` centered at [deviceX, deviceY]
+    const x = Math.max(0, deviceX - deviceRadius);
+    const y = Math.max(0, deviceY - deviceRadius - 1);
```
1chandu: Why `y` is calculated differently than `x` ?

test/modules/core/lib/pick-layers.spec.js
1chandu: `y` is changed from `300` to `209`, so must be picking a different cell?

modules/core/src/lib/deck-picker.js
Pessimistress: device coordinates starts from bottom-left.

test/modules/core/lib/pick-layers.spec.js
Pessimistress: See PR description - we were mistakenly picking the pixel above the specified position. After the fix, I shifted the test case params one pixel up so they are picking the same pixel as before.
Technically speaking this test case is picking a different subpixel (device pixel ratio is 2). From visual inspection I believe they belong to the same cell. Unless picking results differ by pixel there is something else wrong.

1chandu:
Looks good to me, the changes for grid layer picking test cases can be reverted, I am ok to keep them as is too, since the result between CPU and GPU paths is still the same.
Changing picking location may not change picking result, if changed location still points to same cell.

PR #3250, opened 2019-06-18
Add an HTML page writer for pydeck

ajduberstein:
Allows a user to write pydeck visualizations to a static HTML page

#### Background
Provides failover in case a widget can't be installed (e.g., because of user permissions on a notebook server)


bindings/python/pydeck/pydeck/io/templates/js.j2
```diff
@@ -0,0 +1,76 @@
+requirejs.config({
```
ajduberstein: This page is the crux of the frontend for iframe / standalone HTML page support. Here by "standalone" I mean "outside of a Jupyter environment."

Standalone pydeck uses a [Jinja template](http://jinja.pocoo.org/) and RequireJS to request dependencies at the time of page load.

You'll notice some code re-use here from the jupyter-widget module. Not sure of the best way of getting these functions here, having a linter apply itself to these files, etc.

Screen diff tests are applied to make sure the frontend isn't broken. See [the screenshot utilities](https://github.com/uber/deck.gl/pull/3250/files#diff-1577f2377d3b7a458f9c6a7fce61c4f7) below, which use the Python wrapper around Puppeteer (which is used in the browser tests for deck.gl)

modules/jupyter-widget/src/widget.js
```diff
@@ -10,10 +10,10 @@ import {
   waitForElementToDisplay
```
ajduberstein: Minimal changes here–essentially cleaning up the file to conform to patterns elsewhere in the repo (e.g., using import instead of require)

scripts/bindings-precommit-hook.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
ajduberstein: Adds a Python commit hook that only runs if the Python bindings are modified. Extensible to other bindings.

PR #3251, opened 2019-06-19
Improve unit test coverage

Pessimistress:
Now that we can construct Deck instance under Node, add tests for `DeckGL` React component and `MapboxLayer`


1chandu:
coverage +1.3% 👍

Issue #3112, opened 2019-05-17
GoogleMapsOverlay doesn't resize in fullscreen mode


donmccurdy:
#### Description

Google Maps API includes a built-in fullscreen mode:

![Screen Shot 2019-05-17 at 11 07 40 AM](https://user-images.githubusercontent.com/1848368/57947537-01ab3080-7894-11e9-9c47-f9f247217446.png)

If the map was initialized with element A, this fullscreen mode promotes element B, the first child of A, to `document.fullscreenElement`. To be specific, A is the element created by the user, and B would have been generated by the Maps API during initialization.

Because the current resize logic looks only at `map.getDiv()`, which returns A, the size is not correctly updated when entering fullscreen mode:

https://github.com/uber/deck.gl/blob/63837b6a4ac5a00970028321622910205843636f/modules/google-maps/src/utils.js#L85-L87

It's also possible to do a manual fullscreen mode by promoting some parent of A, or the whole page, to `document.fullscreenElement`. I haven't tested that but would assume it works just fine, but this would require the user to implement the fullscreen enter/exit logic themselves rather than using Maps API's.

#### Environment (please complete the following information):
 - Framework Version: deck.gl 7.0.2
 - Browser Version: Chrome 74
 - OS: macOS 10.14.4

Pessimistress:
I'm looking at [this demo](http://bl.ocks.org/Pessimistress/b5e9b86cd0f69ea69f20150734d4e2b6) -- going into and out of fullscreen mode seems to be ok. How can I reproduce the issue?

donmccurdy:
Hm, now I can't reproduce it either. 🤔 

Will reopen if i see this again.

donmccurdy:
Ok, figured out how to reproduce this. The `map.getDiv()` element expands to the full viewport size when its child becomes `document.fullscreenElement` only if its CSS styles allow that. If the container had a less flexible size, the deck.gl canvas on the fullscreen element will be constrained. Changing your demo to:

```css
#container {width: 100vw; height: 75vh;}
```

You can now see the issue in fullscreen mode: http://bl.ocks.org/donmccurdy/ca7d8488c1cdc42da5bb8826a23ddb37.

![Screen Shot 2019-06-18 at 2 48 06 PM](https://user-images.githubusercontent.com/1848368/59722270-1d2d9200-91d8-11e9-94ba-10240c8e0f32.png)

[...]

PR #3255, opened 2019-06-19
Fix fullscreen bug in GoogleMapsOverlay

Pessimistress:
For #3112


Pessimistress:
@donmccurdy I was able to reproduce the bug using `examples/pure-js/google-maps` and verify this fix.

PR #3262, opened 2019-06-20
Improve test coverage for Grid layers and utils

1chandu:
For #

#### Background
Improve test coverage for Grid layers and utils

#### Change List
- Improve test coverage for Grid layers and utils


test/modules/aggregation-layers/index.js
```diff
@@ -22,10 +22,12 @@ import './contour-layer/contour-layer.spec';
...
 import './gpu-grid-layer/gpu-grid-layer.spec';
+import './gpu-grid-layer/gpu-grid-cell-layer.spec';
 import './cpu-grid-layer/cpu-grid-layer.spec';
 import './grid-aggregator.spec';
 import './hexagon-layer.spec';
 import './hexagon-aggregator.spec';
 import './grid-layer.spec';
 import './screen-grid-layer.spec';
 import './utils/scale-utils.spec';
+import './utils/aggregation-operation-utils.spec';
```
Pessimistress: This needs to be removed from `test/browser.js`

modules/aggregation-layers/src/utils/color-utils.js
```diff
@@ -29,6 +29,9 @@ export const defaultColorRange = [
 
 // Converts a colorRange array to a flat array with 4 components per color
 export function colorRangeToFlatArray(colorRange, ArrayType, defaultValue) {
+  if (colorRange.constructor === ArrayType) {
```
Pessimistress: If the intention is to support a flat array from the user, we should relax this check to be `Arraybuffer.isView` or even `Numer.isFinite(colorRange[0])`

test/modules/aggregation-layers/index.js
1chandu: Any specific reason? Its testing pure JS code only.

Pessimistress: `test/browser.js` already imports `modules/index.js` so this ends up being imported twice.

Issue #3269, opened 2019-06-21
S2Layer token to id conversion bug


Pessimistress:
@shannonmoeller

#### Description

S2's [token](https://github.com/google/s2-geometry-library-java/blob/master/src/com/google/common/geometry/S2CellId.java#L480) represents a 64-bit long which JavaScript fails to handle:

> (6003298303284871168).toString(16)
< "5350000000000000"
> parseInt("5350000000000000", 16)
< 6003298303284871000

#### Proposal

Bug fix:

- Make sure token > id [conversion](https://github.com/uber/deck.gl/blob/master/modules/geo-layers/src/s2-layer/s2-utils.js#L28) returns correct result

New feature to unblock users:

- if `getS2Token` returns a string, treat it as a token (existing behavior)
- if `getS2Token` returns a [Long](https://www.npmjs.com/package/long) instance, it is treated as an id
- if `getS2Token` returns string with a `/`, it's treated as a quad key

Pessimistress:
Fixed in 7.1.6

PR #3270, opened 2019-06-21
S2Layer bug fix

Pessimistress:
For #3269 

#### Change List
- Fix token to id conversion bug
- Support additional token formats (see documentation update)


modules/geo-layers/package.json
```diff
@@ -31,6 +31,7 @@
   },
   "dependencies": {
     "h3-js": "^3.4.3",
+    "long": "^3.2.0",
```
Pessimistress: This is already a dependency of `s2-geometry`.

PR #3273, opened 2019-06-21
Update lerna.json

ajduberstein:
Ignore jupyter-widget module

For #

#### Change List
-

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -8,7 +8,7 @@
   ],
   "command": {
     "publish": {
-      "ignore": "modules/jupyter-widget/*",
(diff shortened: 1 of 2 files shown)


lerna.json
```diff
@@ -7,7 +7,9 @@
     "modules/*"
   ],
   "command": {
-    "publish": {},
+    "publish": {
+      "ignore": "modules/jupyter-widget/*",
```
Pessimistress: I don't think this is a valid field. https://github.com/lerna/lerna#lernajson

I'd say add `ignoreChanges`, then set `private: true` in jupyter-widget's package.json

PR #3295, opened 2019-06-27
Implement layer shader hooks

Pessimistress:
[RFC](https://github.com/uber/deck.gl/blob/3797277f5aa230ca78250c78452c06f8867b8192/dev-docs/RFCs/v7.x/layer-shader-hook-rfc.md)

#### Change List
- Add standard shader hooks to all layers
- Documentation

### DIFF
--- a/modules/core/src/shaderlib/misc/geometry.js
+++ b/modules/core/src/shaderlib/misc/geometry.js
@@ -18,13 +18,20 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-export default `\
-highp float random(vec2 co) {
-  highp float a = 12.9898;
-  highp float b = 78.233;
-  highp float c = 43758.5453;
-  highp float dt= dot(co.xy ,vec2(a,b));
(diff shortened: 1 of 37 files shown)


Pessimistress:
For some reason upgrading luma from `7.2.0-alpha.1` to `7.2.0-alpha.2` is causing node tests to fail. A constant attribute buffer created using the default VAO is not deleted when the TripsLayer is finalized. @tsherif I can't locate any commit between the two releases that could cause this. Ideas?

modules/core/src/shaderlib/misc/geometry.js
```diff
@@ -18,13 +18,20 @@
...
-  highp float c = 43758.5453;
-  highp float dt= dot(co.xy ,vec2(a,b));
-  highp float sn= mod(dt,3.14);
-  return fract(sin(sn) / c) - .5;
-}
+const vs = `
+struct VertexGeometry {
+  vec3 worldPosition;
+  vec3 worldPositionAlt;
+  vec4 position;
```
tsherif: Would suggest ordering members in decreasing order of size. It's a convention that tends to pack things tightly (in C and std140 at least, not 100% sure for GLSL in general).

PR #3302, opened 2019-06-28
Support loaders.gl data attributes

Pessimistress:
For [RFC](https://github.com/uber/deck.gl/blob/c21a75a428d5126dab7d4f89af21b5f67916203e/dev-docs/RFCs/v7.x/data-loading-rfc.md)

#### Change List
- Add `onDataLoad` callback (needs audit)
- In addition to `Buffer` and typed array, `Attribute` class supports externally provided attribute descriptors (`{value, size, stride, ...}`)
- `AttributeManager` supports external buffers provided via `data.attributes`.

### DIFF
--- a/modules/core/src/lib/attribute-manager.js
+++ b/modules/core/src/lib/attribute-manager.js
@@ -232,7 +232,11 @@ export default class AttributeManager {
     for (const attributeName in this.attributes) {
       const attribute = this.attributes[attributeName];
 
-      if (attribute.setExternalBuffer(buffers[attributeName], this.numInstances)) {
+      if (
+        attribute.setExternalBuffer(
+          buffers[attributeName] || (data.attributes && data.attributes[attributeName])
+        )
+      ) {
         // Attribute is using external buffer from the props
       } else if (attribute.setGenericValue(props[attribute.getAccessor()])) {
         // Attribute is using generic value from the props
--- a/modules/core/src/lib/attribute.js
+++ b/modules/core/src/lib/attribute.js
@@ -11,6 +11,7 @@ import typedArrayManager from '../utils/typed-array-manager';
 
 const DEFAULT_STATE = {
   isExternalBuffer: false,
+  lastExternalBuffer: null,
   needsUpdate: true,
   needsRedraw: false,
   updateRanges: range.FULL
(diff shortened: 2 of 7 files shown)


modules/core/src/lib/attribute.js
```diff
@@ -287,36 +288,40 @@ export default class Attribute extends BaseAttribute {
...
           throw new Error('Attribute prop must be typed array');
         }
-        if (state.auto && buffer.length <= numInstances * this.size) {
-          throw new Error('Attribute prop array must match length and size');
-        }
 
         const ArrayType = glArrayFromType(this.type || GL.FLOAT);
-        if (buffer instanceof ArrayType) {
-          this.update({constant: false, value: buffer});
+        if (attribute.value instanceof ArrayType) {
```
Pessimistress: The catch here is that if an external attribute overwrote the default `size` `offset` etc. it will not be automatically reverted when the external buffer is unset. Arguably this is an advanced feature and the user should not reuse the same layer with both automatically and externally generated buffer...

modules/core/src/lib/attribute.js
```diff
@@ -327,6 +332,25 @@ export default class Attribute extends BaseAttribute {
     return false;
   }
 
+  hasAlias(name) {
```
tsherif: This feels risky and very loaders-specific. It assumes a certain mapping between loaders attributes and layer attributes.

modules/core/src/lib/attribute-manager.js
```diff
@@ -224,6 +224,8 @@ export default class AttributeManager {
     // keep track of whether some attributes are updated
     let updated = false;
 
+    const externalAttributes = this._updateExternalAttributes(data);
```
tsherif: It seems like a lot to add a whole new concept of "external attributes" to make the API slightly prettier.

Pessimistress: A little background: we are trying to address multiple use cases here

- Support loaders.gl output format (good to have)
- Provide an official API for users to provide their own buffers to layers. This is currently supported (but undocumented) by checking attribute names in the layer props.
  + Users have to look into the layer code to understand what attributes there are. The "attribute alias" feature added in this PR is trying to relax this constraint.
  + The existing feature only accepts typed arrays, which means that users have to pack the external buffer exactly the same as that by the layer itself. For example, an external buffer for position that is in XYXYXY instead of XYZXYZXYZ cannot be used, because there is no way to set the stride.

The second use case is much bigger, and fundamental to our future GPGPU work.

PR #3305, opened 2019-07-01
Implement layer extensions RFC

Pessimistress:

#### Change List
- Add `LayerExtensions` interface
- Handle `extensions` prop in base Layer class: lifecycle callbacks, `extensionsChanged` flag
- Update all layers to use new `getShaders` API
- Tests

### DIFF
--- a/modules/core/src/lifecycle/props.js
+++ b/modules/core/src/lifecycle/props.js
@@ -19,7 +19,7 @@ export function diffProps(props, oldProps) {
     newProps: props,
     oldProps,
     propTypes: getPropTypes(props),
-    ignoreProps: {data: null, updateTriggers: null}
+    ignoreProps: {data: null, updateTriggers: null, extensions: null}
   });
 
   // Now check if any data related props have changed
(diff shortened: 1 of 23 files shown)


modules/core/src/lifecycle/props.js
```diff
@@ -173,6 +174,29 @@ function diffUpdateTriggers(props, oldProps) {
...
+  const {extensions} = props;
+
+  if (extensions === oldExtensions) {
+    return false;
+  }
+  if (extensions.length !== oldExtensions.length) {
+    return true;
+  }
+  for (let i = 0; i < extensions.length; i++) {
+    if (!extensions[i].equals(oldExtensions[i])) {
```
tsherif: Probably not a major concern, but to we care about the case when the it's the same extensions in a different order?

Pessimistress: Theoretically order matters because extensions can overwrite the changes done by a previous extension... But that's another can of worms.

PR #3308, opened 2019-07-02
 Move fp64 support to an extension

Pessimistress:

#### Change List
- Layers:
  + Remove `fp64` prop from most places except `GPUGridAggregator` and the layers that uses GPU aggregation
  + Remove ArcLayer's fp64 shader (breaking change)
  + Remove PathLayer's fp64 shader (was not entirely working)
- Core:
  + Remove default registration of `fp64` and `project64` modules (for effective tree shaking)
  + Add `fp64LowPart` util as an alternative to importing `fp64.fp64LowPart`
- Extensions:
  + Add `Fp64Extension` class
- Tests
- Update docs


modules/core/src/utils/shader.js
```diff
@@ -30,6 +30,16 @@ export function mergeShaders(target, source) {
...
   if ('modules' in source) {
     result.modules = (target.modules || []).concat(source.modules);
+
+    // Hack: prject32 and project64 cannot co-exist
+    if (source.modules.some(module => module === 'project64' || module.name === 'project64')) {
+      const index = result.modules.findIndex(
+        module => module === 'project32' || module.name === 'project32'
+      );
+      if (index >= 0) {
+        result.modules.splice(index, 1);
```
1chandu: Nit: log.warn to fix the source.

test/modules/extensions/fp64.spec.js
```diff
@@ -0,0 +1,24 @@
...
+test('Fp64Extension', t => {
+  const layer = new ScatterplotLayer({
+    id: 'fp64-test',
+    data: [
+      {position: [-122.453, 37.782], timestamp: 120, entry: 13567, exit: 4802},
+      {position: [-122.454, 37.781], timestamp: 140, entry: 14475, exit: 5493}
+    ],
+    getPosition: d => d.position,
+    coordinateSystem: COORDINATE_SYSTEM.LNGLAT_DEPRECATED,
+    extensions: [new Fp64Extension()]
```
1chandu: Add test case , that changes `extensions` prop. (different PR is ok)

modules/core/src/utils/shader.js
Pessimistress: This is going to happen with every layer that uses this extension. What are we warning the user about?

Pessimistress:
> Where are we setting changeFlags.extensionsChanged to true, other than in init?

https://github.com/uber/deck.gl/blob/master/modules/core/src/lifecycle/props.js#L39

modules/core/src/utils/shader.js
1chandu: I see, add that information to the comment ?

1chandu:
> > Where are we setting changeFlags.extensionsChanged to true, other than in init?
> 
> https://github.com/uber/deck.gl/blob/master/modules/core/src/lifecycle/props.js#L39

👍 looking for that in this diff.

PR #3309, opened 2019-07-02
Add Brushing extension

Pessimistress:

#### Change List
- Add `BrushingExtension`
- Update the brushing example


modules/extensions/src/brushing/brushing.js
```diff
@@ -0,0 +1,92 @@
...
+              divisor: 1
+            }
+          }
+        }
+      });
+    }
+
+    // Trigger redraw when mouse moves
+    // TODO - expose this in a better way
+    extension.onMouseMove = () => {
```
tsherif: Would it be worth it to just make capturing UI events part of the extension API? E.g. 
```js
this.onEvents({
  pointermove: ...,
  pointerleave: ...
});
```

modules/core/src/lib/deck.js
```diff
@@ -525,6 +525,10 @@ export default class Deck {
       _pickRequest.radius = this.props.pickingRadius;
     }
 
+    if (this.layerManager) {
```
tsherif: Since `_requestPick` is no longer just a picking method, maybe it should be renamed.

modules/extensions/src/brushing/brushing.js
Pessimistress: The `LayerExtension` interface is supposed to mirror that of `Layer`. Layers can implement methods such as `onHover` `onDrag` but they are only called when the event is interacting with an object rendered by this layer. If we add a new API for generic event handling, it has to be supported in both layers and extensions.

PR #3318, opened 2019-07-08
Typed array manager

Pessimistress:
For data streaming

#### Change List
- Split `TypedArrayManager` off of `Tesselator`
- Use central typed array pool in the `Attribute` class
- Tests

### DIFF
--- a/modules/core/src/utils/typed-array-manager.js
+++ b/modules/core/src/utils/typed-array-manager.js
@@ -1,20 +1,25 @@
 export class TypedArrayManager {
-  constructor({overAlloc = 1, poolSize = 100} = {}) {
+  constructor({overAlloc = 2, poolSize = 100} = {}) {
     this.overAlloc = overAlloc;
     this.poolSize = poolSize;
 
     this._pool = [];
   }
 
-  allocate(typedArray, count, {size, type, padding = 0, copy = false}) {
+  allocate(typedArray, count, {size = 1, type, padding = 0, copy = false}) {
+    const Type = type || (typedArray && typedArray.constructor) || Float32Array;
+
     const newSize = count * size + padding;
-    if (typedArray && newSize <= typedArray.length) {
-      return typedArray;
+    if (ArrayBuffer.isView(typedArray)) {
+      if (newSize <= typedArray.length) {
+        return typedArray;
+      }
+      if (newSize * typedArray.BYTES_PER_ELEMENT <= typedArray.buffer.byteLength) {
+        return new Type(typedArray.buffer, 0, newSize);
+      }
     }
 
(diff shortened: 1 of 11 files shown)


modules/core/src/utils/typed-array-manager.js
```diff
@@ -0,0 +1,66 @@
+export class TypedArrayManager {
+  constructor({overAlloc = 1, poolSize = 5} = {}) {
```
tsherif: 5 seems like a very small limit to me. I'm thinking let it grow to the size the application needs?

modules/core/src/utils/typed-array-manager.js
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  _allocate(Type = Float32Array, size) {
+    // Check if available in pool
+    const typeName = Type.name;
+    const pool = this._pool[typeName];
+    if (pool) {
+      const i = pool.findIndex(array => array.length >= size);
+      if (i >= 0) {
+        return pool.splice(i, 1)[0];
```
tsherif: This linear search and splicing could become costly if the pool gets big. We should think about a better data structure.

Also, I'd make it `pool.splice(i, 1)[0].subarray(0, size);` so the length of the returned array is as expected.

modules/core/src/utils/typed-array-manager.js
```diff
@@ -0,0 +1,66 @@
...
+    }
+    const typeName = typedArray.constructor.name;
+    let pool = this._pool[typeName];
+    if (!pool) {
+      pool = [];
+      this._pool[typeName] = pool;
+    }
+    // Save the released array into the pool
+    if (pool.length < this.poolSize) {
+      pool.push(typedArray);
```
tsherif: I'd store `typedArray.buffer` in the pool. This would allow you to use the same underlying memory for other types, and also allow you to return different sized arrays but maintain the same buffer.

modules/core/src/utils/typed-array-manager.js
Pessimistress: With a few hundred items the search is extremely cheap. I'll add a comment for future consideration.

tsherif: I had an idea for this. Make it that all arrays returned have byte lengths that are powers of 2. Each element in the pool would be an array of `ArrayBuffer`s of length `2 ** index`. Then the look up would just be 
```js
const index = Math.ceil(Math.log2(requestedByteLength));
if (pool[index] && pool[index].length > 0) {
    return new Type(pool[index].pop()).subarray(0, size);
} 
return new Type(Math.pow(2, index) / Type.BYTES_PER_ELEMENT).subarray(0, size);
```

Issue #3299, opened 2019-06-28
Is it possible for TextLayer to support text wrapping?


ForeverArt:
Hello,
I am using deck.gl to show buildings’ name on a GeoJsonLayer.
However, those names are too long to have a good visual effects or contain description that should be displayed on the second line.

So, i am wondering whether TextLayer supports text wrapping？

The following screenshot is made by mapbox studio.
I want to known if there is a way to achieve the same effect by using `TextLayer` in deck.gl.

Thanks!

<img width="857" alt="WX20190628-141306@2x" src="https://user-images.githubusercontent.com/23061097/60321270-e5c89d80-99ae-11e9-929c-6a23d4c97dbf.png">

Pessimistress:
TextLayer does not support wrapping at the moment. Per discussion with @xintongxia, we can split this feature into two parts:

- [ ] Support line breaks in text strings - we should be able to do this in a patch
- [ ] Support auto-wrapping - this will require an API audit. We can add this to the next milestone.

PR #3322, opened 2019-07-09
Support line breaks in text strings

xintongxia:
#### Background
For #3299 

#### Change List
- TextLayer - support line breaks
- add test

Screen Shot 2019-07-09 at 10 00 41 AM


modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer.js
```diff
@@ -29,8 +29,10 @@ const DEFAULT_GAMMA = 0.2;
 const DEFAULT_BUFFER = 192.0 / 256;
 
 const defaultProps = {
+  getLineLength: {type: 'accessor', value: x => x.lineLength || 0},
```
Pessimistress: - Add comments to explain what these accessors are
- Make the default a constant for better perf for those who are using this layer standalone

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer.js
```diff
@@ -29,8 +29,10 @@ const DEFAULT_GAMMA = 0.2;
...
   getShiftInQueue: {type: 'accessor', value: x => x.shift || 0},
```
Pessimistress: Looks like the format of the return value has changed?

modules/layers/src/text-layer/text-layer.js
```diff
@@ -311,10 +348,11 @@ export default class TextLayer extends CompositeLayer {
       }),
       {
         data,
-
         getIcon: d => d.text,
+        getLineLength: d => this.getLineLength(d),
```
Pessimistress: Either `d => d.lineLengths[d.lineIndex]` or `this.getLineLength`. Why wrap functions that already work?

I personally prefer defining these accessors inline. It's a lot easier to read.

modules/layers/src/text-layer/text-layer.js
```diff
@@ -188,14 +188,27 @@ export default class TextLayer extends CompositeLayer {
...
       if (text) {
         const letters = Array.from(text);
-        const offsets = [0];
+
         let offsetLeft = 0;
+        let offsetTop = 0;
+        let lineHeight = 0;
+
+        // offsetX, offsetY
+        const offsets = [[0, 0]];
```
Pessimistress: My understanding is that there is one offset value for each character. Why create a shared array instead of just assigning the value to each datum?

Pessimistress:
Add unit tests for the new functionality.

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer.js
xintongxia: What do you mean by saying `the format of the return value has changed`?

Pessimistress: When you call `getShiftInQueue` you are expecting it to return an array. And the default value set here is a number.

examples/layer-browser/src/examples/core-layers.js
```diff
@@ -407,7 +407,7 @@ const HexagonLayerExample = {
 
 const TextLayerExample = {
   layer: TextLayer,
-  getData: () => dataSamples.texts,
+  getData: () => dataSamples.texts.slice(1, 3),
```
Pessimistress: This is for debug only? Remove?

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer.js
```diff
@@ -29,8 +29,16 @@ const DEFAULT_GAMMA = 0.2;
...
+  // each queue can have one or multiple line(s)
+  // each line can have one or multiple object(s)
+  // length of the line where the given object is
+  getLineLength: {type: 'accessor', value: x => x.lineLength || 0},
+  // offset from the left, top position of the queue
+  getOffsets: {type: 'accessor', value: x => x.shift || 0},
+  // length of the queue
   getLengthOfQueue: {type: 'accessor', value: x => x.len || 1},
+  // height of the queue
+  getHeightOfQueue: {type: 'accessor', value: x => x.height || 1},
```
Pessimistress: Since `getLengthOfQueue` and `getHeightOfQueue` are always called together, just merge them into one accessor. Prop comparison has a cost.

"Queue" was a really bad name. More intuitive terms here may be
- character
- column
- row
- paragraph

modules/layers/src/text-layer/utils.js
```diff
@@ -64,3 +68,75 @@ export function buildMapping({
     canvasHeight: nextPowOfTwo(yOffset + (row + 1) * rowHeight)
   };
 }
+
+export function updateRange({array, startIndex, endIndex, data}) {
+  for (let i = startIndex; i < Math.min(endIndex, array.length); i++) {
```
Pessimistress: Don't put upper limit calculation in a loop.

modules/layers/src/text-layer/utils.js
```diff
@@ -64,3 +68,75 @@ export function buildMapping({
...
+  let lineHeight = 0;
+
+  // width and height of the text
+  const size = [0, 0];
+
+  letters.forEach((letter, i) => {
+    if (letter === '\n') {
+      size[0] = Math.max(offsetLeft, size[0]);
+
+      updateRange({
```
Pessimistress: This retro update is very hard to read. Just break your lines like

```js
letters.split('\n').forEach(...)
```

And you can access the line length as you build the data.

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer.js
```diff
@@ -104,11 +110,16 @@ export default class MultiIconLayer extends IconLayer {
...
+      const [width, height] = getParagraphSize(object);
+      const [rowWidth] = getRowSize(object);
+      const [offsetX, offsetY] = getOffsets(object);
+      const anchorX = getAnchorX(object);
+
+      // For a multi-line object, offset in x-direction needs consider
+      // the row offset in the paragraph and the object offset in the row
+      const rowOffset = ((1 - anchorX) * (width - rowWidth)) / 2;
+      value[i++] = ((anchorX - 1) * width) / 2 + rowOffset + rect.width / 2 + offsetX || 0;
+      value[i++] = ((getAnchorY(object) - 1) * height) / 2 + rect.height / 2 + offsetY || 0;
```
Pessimistress: Nit: define `anchorY` so that these two lines have comparable structures.

modules/layers/src/text-layer/text-layer.js
```diff
@@ -179,53 +179,40 @@ export default class TextLayer extends CompositeLayer {
     const {data, getText} = this.props;
     const {iconMapping} = this.state;
     const {startRow, endRow} = dataRange;
+    const {iterable, objectInfo} = createIterable(data, startRow, endRow);
 
     const transformedData = [];
 
-    const {iterable, objectInfo} = createIterable(data, startRow, endRow);
     for (const object of iterable) {
+      const transformCharacter = ({text, index, offsetLeft, offsetTop, size, rowSize, len}) => {
```
Pessimistress: Perf: This is a large amount of deconstructing and reconstructing small objects in a private function.

modules/layers/src/text-layer/text-layer.js
```diff
@@ -311,10 +298,10 @@ export default class TextLayer extends CompositeLayer {
       }),
       {
         data,
-
         getIcon: d => d.text,
-        getShiftInQueue: d => this.getLetterOffset(d),
-        getLengthOfQueue: d => this.getTextLength(d)
+        getRowSize: d => d.rowSize,
+        getOffsets: d => this.getLetterOffsets(d),
```
Pessimistress: Just `d => [d.offsetLeft, d.offsetTop]`

modules/layers/src/text-layer/utils.js
```diff
@@ -0,0 +1,141 @@
...
+  let offsetTop = 0;
+  let index = 0;
+
+  rows.forEach(row => {
+    const {characters, rowWidth, rowHeight} = transformRow(row, index, iconMapping);
+
+    characters.forEach(datum => {
+      datum.offsetTop = offsetTop;
+      datum.size = size;
+      datum.rowSize = [rowWidth, rowHeight];
```
Pessimistress: This object should be shared among all characters in the same row?

modules/layers/src/text-layer/utils.js
```diff
@@ -0,0 +1,141 @@
...
+      datum.len = paragraph.length;
```
Pessimistress: Is this still used?

modules/layers/src/text-layer/utils.js
```diff
@@ -0,0 +1,141 @@
...
+      text: character,
+      index: indexInParagraph + i,
+      offsetLeft
+    };
+
+    const frame = iconMapping[character];
+
+    if (frame) {
+      offsetLeft += frame.width;
+      rowHeight = Math.max(rowHeight, frame.height);
```
Pessimistress: `rowHeight` should be a fixed number. We need a new prop `lineHeight` and then the row height will be calculated as `lineHeight * fontHeight`.

modules/layers/src/text-layer/utils.js
```diff
@@ -0,0 +1,141 @@
...
+
+export function transformRow(row, indexInParagraph, iconMapping) {
+  let offsetLeft = 0;
+  let rowHeight = 0;
+
+  let characters = Array.from(row);
+  characters = characters.map((character, i) => {
+    const datum = {
+      text: character,
+      index: indexInParagraph + i,
```
Pessimistress: Is this still used?

docs/layers/text-layer.md
```diff
@@ -129,6 +129,12 @@ Specifies a list of characters to include in the font. By default, only characte
 
 css `font-weight`.
 
+##### `` (Number, optional)
+
+* Default: 1.0.
+
+css `line-height`.
```
Pessimistress: Explain or link to css spec

docs/layers/text-layer.md
```diff
@@ -129,6 +129,12 @@ Specifies a list of characters to include in the font. By default, only characte
 
 css `font-weight`.
 
+##### `` (Number, optional)
```
Pessimistress: Incomplete?

Issue #3321, opened 2019-07-09
onClick does not apply pickingRadius


kouak:
#### Description

When passing a `onHover` prop to a layer, the pickingRadius is applied and the event fires when the cursor is in range of items.

However, when passing a `onClick` callback, users must have the mouse pointer right over items to trigger the callback.

#### Repro Steps
https://codepen.io/kouak/pen/dBwOWo

Zoom in, open the codepen console, watch the onHover callback firing when the mouse approaches plots. Try to click on the same mouse position.

#### Environment (please complete the following information):
 - Framework Version: 7.0 (with 6.4 working as intended)
 - Browser Version: Chrome 73
 - OS: Linux 

#### Logs
N/A

kouak:
I believe this regression was introduced in #3041.

I think the [pickObject call here](https://github.com/uber/deck.gl/blob/7.0-release/modules/core/src/lib/deck.js#L734-L737) should include `this.props.pickingRadius`.

I tried this fix in my project by editing the `node_modules` folder with success, however I'm not familiar enough with the event system to assess the effects of this fix.

If this looks fine to the maintainers, please let me know, I'll submit a PR ASAP.

Thanks !

Pessimistress:
That's indeed a bug, thanks for reporting!

Your fix sounds legit to me. Feel free to open a PR and I'll patch.

Pessimistress:
Patched in 7.1.8. Thank you for contributing!

PR #3327, opened 2019-07-10
Fix missing pickingRadius for onClick handlers (#3321)

kouak:
For #3321

### DIFF
--- a/modules/core/src/lib/deck.js
+++ b/modules/core/src/lib/deck.js
@@ -770,7 +770,7 @@ export default class Deck {
     this._lastPointerDownInfo = this.pickObject({
       x: pos.x,
       y: pos.y,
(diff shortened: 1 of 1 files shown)


modules/core/src/lib/deck.js
```diff
@@ -769,7 +769,8 @@ export default class Deck {
     const pos = event.offsetCenter;
     this._lastPointerDownInfo = this.pickObject({
       x: pos.x,
-      y: pos.y
+      y: pos.y,
+      pickingRadius: this.props.pickingRadius
```
Pessimistress: This should be `radius`

kouak: Fixed thanks.

PR #3334, opened 2019-07-11
Improve prop comparison perf

Pessimistress:
#### Background

Our `props` object is created by inheriting the `defaultProps` object (one per `Layer` class), and then assign user-specified values on top. Currently, we compare two props objects by iterating through all key-value pairs, including the default prop values even if they are not supplied by the user. This PR switches to only examining the props explicitly set by the user.

#### Benchmark

Before:
├─ compareProps#empty: 65.1K iterations/s ±7.02%
├─ compareProps#real world sample: 55.6K iterations/s ±7.31%

After:
├─ compareProps#empty: 15.2M iterations/s ±0.92%
├─ compareProps#real world sample: 1.08M iterations/s ±1.23%

#### Change List
- `compareProps` util
- Creation of `oldProps` object when an async prop is loaded
- Benchmark and unit tests

### DIFF
--- a/modules/core/src/lifecycle/props.js
+++ b/modules/core/src/lifecycle/props.js
@@ -50,6 +50,11 @@ export function diffProps(props, oldProps) {
  *   if unequal, returns a string explaining what changed.
  */
 /* eslint-disable max-statements, max-depth, complexity */
+/*
+ * Note: for better performance, this function assumes that both oldProps and newProps
+   inherit the same prototype (defaultProps). That is, if neither object contains own
+   property <key>, assume `oldProps.<key>` and `newProps.<key>` are equal.
+ */
 export function compareProps({
   newProps,
   oldProps,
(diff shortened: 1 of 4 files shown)


modules/core/src/lifecycle/props.js
```diff
@@ -73,41 +78,31 @@ export function compareProps({
...
-  for (const key in newProps) {
+  // Test if any old props have been dropped
+  for (const key of Object.keys(oldProps)) {
     if (!(key in ignoreProps)) {
-      if (!(key in oldProps)) {
-        return `${triggerName}.${key} added: undefined -> ${newProps[key]}`;
+      if (!(key in newProps)) {
+        return `${triggerName}.${key} dropped`;
+      }
+      if (!Object.hasOwnProperty.call(newProps, key)) {
```
tsherif: Just curious, what does this do? Does it compare further up the prototype chain?

Pessimistress: The first for loop already checked all the own properties of `newProps`, and in this second for loop we only want to check props that are in `oldProps` but not `newProps`.

PR #3336, opened 2019-07-11
[NOT FOR REVIEW] Added Linear Interpolation in marching square (Isoline)

manishmz:
update marching square algorithm for contour isoline layer

### DIFF
--- a/modules/aggregation-layers/src/contour-layer/marching-squares-codes.js
+++ b/modules/aggregation-layers/src/contour-layer/marching-squares-codes.js
@@ -19,6 +19,17 @@ const OFFSET = {
 };
 
 // CORNERS of Cell
+// Table to map edge of intersection, pair of cell's who's weight is used as a domain for interpolation
+//          N
+//   top________topRight
+//     |    |    |
+//     |    |    |
+// W   |---------|   E  // cell
+//     |    |    |
+//     |____|____|
+//   current     right
+//          S
+//
 const CORNERS = {
   W: ['current', 'top'],
   E: ['right', 'topRight'],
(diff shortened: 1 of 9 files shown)


test/modules/aggregation-layers/contour-layer/marching-squares.spec.js
```diff
@@ -393,6 +450,15 @@ const GETVERTEX_TESTS = [
   }
 ];
 
+const GETLINEARINTERPOLATION_TESTS = [
+  {
```
1chandu: Add some corner cases as well,` [10, 80, 10] -> -5, [10, 80, 80] -> 5, [10, 80, 35] -> 0`

test/modules/aggregation-layers/contour-layer/marching-squares.spec.js
```diff
@@ -393,6 +450,15 @@ const GETVERTEX_TESTS = [
...
+    min: 10,
+    max: 80,
+    threshold: 50,
+    li: 0.571
```
1chandu: is this in [-.5, .5] ?

1chandu: nit: change `li` to `offset` for readability.

manishmz: No its 0 to 1

manishmz: I have another function named getLinearOffsetValue which convert the range [0 , 1]
to [-0.5, 0.5]
```
function getLinearOffsetValue(offset, weights, threshold) {
  const linearInterpolation = getLinearInterpolation(
    weights[offset[0]],
    weights[offset[1]],
    threshold
  );
  // linearInterpolation is between 0 and 1, to match with (-0.5, 0.5) subtracting with 0.5
  return linearInterpolation - 0.5;
}
```

manishmz:
linearInterPolation2

modules/aggregation-layers/src/utils/gpu-grid-aggregation/gpu-grid-aggregator.js
```diff
@@ -341,7 +341,10 @@ export default class GPUGridAggregator {
   validateProps(aggregationParams, opts) {
     const {changeFlags, projectPoints, gridTransformMatrix} = aggregationParams;
     log.assert(
-      changeFlags.dataChanged || changeFlags.viewportChanged || changeFlags.cellSizeChanged
+      changeFlags.dataChanged ||
```
1chandu: This feature (generating smooth edges) should only operate on data produced by aggregation, you shouldn't be doing any changes to this file. Any reason you have to make these changes? If not needed remove.

modules/aggregation-layers/src/contour-layer/marching-squares-codes.js
```diff
@@ -149,6 +157,37 @@ export const ISOLINES_CODE_OFFSET_MAP = {
   15: []
 };
 
+export const LINEAR_INTERPOLATION_CODE_OFFSET_MAP = {
```
1chandu: Is this map same as `ISOLINES_CODE_OFFSET_MAP` ? if same remove, if not can you explain how it differs and update the comments to understand what it is mapping.

test/modules/aggregation-layers/contour-layer/marching-squares.spec.js
```diff
@@ -393,6 +450,33 @@ const GETVERTEX_TESTS = [
   }
 ];
```
1chandu: Did you add any new tests similar to `GETVERTEX_TESTS` that operate using computed threshold. The key of this feature is how vertices are adjusted based on neighbor cell weights, we should have these unit tests for different cases, and validate functionality, otherwise it is hard to find these problem looking at rendered images.

modules/aggregation-layers/src/utils/gpu-grid-aggregation/gpu-grid-aggregator.js
manishmz: I added this change to handle Linear Interpolation checkbox in the example, after changing the value for linear interpolation the contour layer was not refreshing, so I added this flag to handle this.

When we will remove the Linear Interpolation checkbox we will remove this change, It just a temporary change.

modules/aggregation-layers/src/contour-layer/marching-squares.js
```diff
@@ -162,12 +189,32 @@ export function getVertices(opts) {
...
+      }
+      const vX = refVertexX + xOffset * cellSize[0];
+      const vY = refVertexY + yOffset * cellSize[1];
       lines.push([vX, vY, vZ]);
     });
   });
   return lines;
 }
+
+function getLinearOffsetValue(offset, weights, threshold) {
```
1chandu: I would recommend merging `getLinearInterpolation` method into `getLienarOffsetValue`, and change function name to something like ,  `getsmoothOffset` or just `getOffset` which will internally check if interpolation is on/off and return the correct offset.

1chandu:
@manishmz , overall looking good, but it is crucial for this feature to have detailed unit tests to verify vertex generation.

modules/aggregation-layers/src/contour-layer/marching-squares-codes.js
manishmz: We have four weights for a single cell in below format 
```
weights = {
top: 10,
right: 12,
current: 8,
topRight: 14
}
```
So this map will return the keys array `(top, topRight, current, right)` bypassing cell code value. using these keys, I am extracting the weights and calculate the Linear interpolation for that code.

I also change the description of this map.

modules/aggregation-layers/src/contour-layer/marching-squares.js
manishmz: I will merge these two methods. 

Before merging the PR we are removing the interpolation check, right?

1chandu: If we have vertex generation test cases, and if we validate all cases using unit tests, we can make the prop experimental and land the PR. Effectively all this new functionality is enabled only when this prop is set.

modules/aggregation-layers/src/utils/gpu-grid-aggregation/gpu-grid-aggregator.js
1chandu: What you really need is check for this flag in `_shouldRebuildContours`, and when linear interpolation mode changed, this method should return true and it will re trigger contour generation.

test/modules/aggregation-layers/contour-layer/marching-squares.spec.js
```diff
@@ -390,6 +391,185 @@ const GETVERTEX_TESTS = [
     cellSize: [12, 24],
     vertices: [[[106, 236], [106, 224], [112, 212], [118, 212], [118, 224], [112, 236]]],
     type: CONTOUR_TYPE.ISO_BANDS
+  },
+  // linear interpolation flag set true
```
1chandu: Lets reduce the code duplication here, you can have each test case defined in following way :
```
{
  gridOrigin: [..]
  // all other props needed for both original and smooth cases
  vertices: ...
  smoothVertices: ...
 smoothEnabled: [true, false]
}
```
And update the for loop, so each test case is run twice, once with smoothEnabled[0] and smoothEnabled[1], that way we can also visually compare how vertices change when these flag is on/off.

test/modules/aggregation-layers/contour-layer/marching-squares.spec.js
```diff
@@ -390,6 +391,185 @@ const GETVERTEX_TESTS = [
...
+  },
+
+  // saddle cases
+  {
+    gridOrigin: [100, 200],
+    code: 5,
+    meanCode: 1,
+    x: 0,
+    y: 0,
+    vertices: [[105, 223.33333333333334], [110, 230], [115, 210], [115, 210]],
```
1chandu: use Number.toFixed() to may be 4 digits?

modules/aggregation-layers/src/contour-layer/marching-squares-codes.js
```diff
@@ -18,6 +18,14 @@ const OFFSET = {
   SW: [-HALF, -HALF]
 };
 
+// CORNERS of Cell
+const CORNERS = {
```
1chandu: Add comment similar to other enums defined above, something like :
`Table to map edge of intersection, pair of cell's who's weight is used as a domain for interpolation`.

modules/aggregation-layers/src/contour-layer/marching-squares-codes.js
```diff
@@ -149,6 +157,37 @@ export const ISOLINES_CODE_OFFSET_MAP = {
...
 
+export const LINEAR_INTERPOLATION_CODE_OFFSET_MAP = {
+  // key is equal to the code of 4 vertices (invert the code specified in wiki)
+  // value can be an array or an Object
+  // Array defines the cell corners (top, current, topRight, right)
+  // Object : to handle saddle cases, whose output depends on mean value of all 4 corners
+  //  key: code of mean value (0 or 1)
+  //  value: Array , as above defines
+
+  0: [],
```
1chandu: This table has same format as `ISOLINES_CODE_OFFSET_MAP`, reduce the duplication and potential errors, this can be converted into a function, which takes either `OFFSET` or `CORNERS` has an argument.

modules/aggregation-layers/src/contour-layer/marching-squares.js
```diff
@@ -56,6 +60,7 @@ export function getCode(opts) {
   // TOP
   if (isLeftBoundary || isTopBoundary) {
     codes.top = 0;
+    weights.top = 0;
```
1chandu: I think this may be a problem, when processing the boundary, we don't want to perform interpolation , we want the intersection point to be on the boundary, please investigate, you can use the demo to verify, with and without interpolation.

modules/aggregation-layers/src/contour-layer/marching-squares.js
```diff
@@ -104,20 +112,38 @@ export function getCode(opts) {
...
+  const {
+    gridOrigin,
+    cellSize,
+    x,
+    y,
+    code,
+    meanCode,
+    type = CONTOUR_TYPE.ISO_LINES,
+    weights,
+    isLI = false
```
1chandu: change `isLI` to `smooth` or `smoothEnabled`

modules/aggregation-layers/src/contour-layer/marching-squares.js
```diff
@@ -162,12 +188,35 @@ export function getVertices(opts) {
...
-    xyOffsets.forEach(offset => {
-      const vX = refVertexX + offset[0] * cellSize[0];
-      const vY = refVertexY + offset[1] * cellSize[1];
+  offsets.forEach((xyOffsets, i) => {
+    xyOffsets.forEach((offset, j) => {
+      let xOffset;
+      let yOffset;
+      if (isLI) {
+        const li = getSmoothOffset(linearOffsets[i][j], weights, thresholdData.threshold);
+        xOffset = !offset[0] ? offset[0] + li : offset[0];
```
1chandu: Add a comment here, why are we doing `!offset[0]`?

manishmz: The offset x and y axis contains value 0, 0.5, - 0.5. As I have to add the interpolation to offset having value 0 only. So I added this checks `!offset[0]?` to check whether the value is 0 or other than 0.
example - consider the offset [0, 0.5]. In this case, the Interpolation should be added to offset[0] and not to offset[1].

1chandu:
This is not complete, having few concerns on the code and we need unit tests to verify the functionality, it is hard to verify all cases looking at rendered images. Closing for now.

Issue #3197, opened 2019-06-05
instanceColors accept RGB format


xintongxia:
#### Target Use case
[3d tiles](https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/master/specification/TileFormats/PointCloud#point-colors) PointCloud colors could be in RGB format. Could save resources by allowing user pass colors in rgb directly.

#### To Do List 
  - [ ] Add label and assign to milestone 
  - [ ] Coding
  - [ ] Doc update
  - [ ] What’s new update 
  - [ ] Test

PR #3338, opened 2019-07-11
Add support for RGB color format

Pessimistress:
For #3197 

There is a small save on resources, but this is most useful when using external buffers.

#### Change List
- Add `colorFormat` prop
- Fix `positionFormat` support in `PolygonLayer`
- Documentation

### DIFF
--- a/docs/api-reference/layer.md
+++ b/docs/api-reference/layer.md
@@ -300,7 +300,8 @@ Default `'XYZ'`.
 
 One of `'RGBA'`, `'RGB'`.
 
-Setting it to `'RGB'` will make the layer ignore the alpha channel and assume `a: 255`.
+Setting it to `'RGB'` will make the layer ignore the alpha channel of the colors returned by accessors, and instead assume all objects to be opaque.
+The layer's overall transparency controlled by `opacity` is still applied.
 
 Default `'RGBA'`.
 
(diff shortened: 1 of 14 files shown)


docs/api-reference/layer.md
```diff
@@ -288,6 +288,21 @@ And is expected to return an array of "ranges", in the form of `{startRow, endRo
...
+
+Default `'XYZ'`.
+
+##### `colorFormat` (String, optional)
+
+One of `'RGBA'`, `'RGB'`.
+
+Setting it to `'RGB'` will make the layer ignore the alpha channel and assume `a: 255`.
+
+Default `'RGBA'`.
```
xintongxia: 3D tiles also have [rgb565](https://github.com/uber-web/loaders.gl/blob/master/modules/3d-tiles/src/parsers/parse-3d-tile-point-cloud.js#L215) color format. Do you think it could also be provided in deck.gl?

Pessimistress: No, that cannot be done without changing attribute generation.

docs/api-reference/layer.md
```diff
@@ -288,6 +288,21 @@ And is expected to return an array of "ranges", in the form of `{startRow, endRo
...
 
+##### `positionFormat` (String, optional)
+
+One of `'XYZ'`, `'XY'`.
+
+This prop is currently only effective in `PathLayer`, `SolidPolygonLayer` and `PolygonLayer`.
+
+Default `'XYZ'`.
+
+##### `colorFormat` (String, optional)
```
tsherif: It might simplify things to make these numbers, e.g. `positionDimensions` and `colorChannels`. The setting the size would just be `size: this.props.colorChannels` (with a default of 4). Also leaves the door open for passing fewer components (if that might ever be useful?).

Pessimistress: Unfortunately `positionFormat` has been a public API since v6. I can change the usage to `size: this.props.colorFormat.length` though.

PR #3342, opened 2019-07-11
Attribute allocation bug fixes

Pessimistress:
#### Change List
- Fix a bug where `Attribute.allocate` may skip allocation if previously set external buffer or constant value is large enough
- Make BitmapLayer attributes use pre-allocated typed arrays
- Unit tests

### DIFF
--- a/modules/core/src/lib/attribute.js
+++ b/modules/core/src/lib/attribute.js
@@ -422,7 +422,7 @@ export default class Attribute extends BaseAttribute {
   _validateAttributeUpdaters() {
     const state = this.userData;
 
-    // Check that either 'accessor' or 'update' is a valid function
+    // Check that 'update' is a valid function
     const hasUpdater = state.noAlloc || typeof state.update === 'function';
     if (!hasUpdater) {
       throw new Error(`Attribute ${this.id} missing update or accessor`);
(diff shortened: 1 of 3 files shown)


modules/core/src/lib/attribute.js
```diff
@@ -101,10 +100,6 @@ export default class Attribute extends BaseAttribute {
     return needsRedraw;
   }
 
-  getInstanceCount() {
```
xintongxia: This is not used anywhere else?

modules/core/src/lib/attribute.js
```diff
@@ -436,8 +423,7 @@ export default class Attribute extends BaseAttribute {
     const state = this.userData;
 
     // Check that either 'accessor' or 'update' is a valid function
```
xintongxia: comment is not applied

modules/core/src/lib/attribute.js
```diff
@@ -200,15 +192,15 @@ export default class Attribute extends BaseAttribute {
...
-        if (shouldCopy) {
+        if (shouldCopy && oldValue) {
           // Upload the full existing attribute value to the GPU, so that updateBuffer
           // can choose to only update a partial range.
           // TODO - copy old buffer to new buffer on the GPU
           this.buffer.subData(oldValue);
         }
       }
 
-      this.setNeedsUpdate(true, {startRow: instanceCount});
```
xintongxia: Why redraw is not needed?

modules/core/src/lib/attribute.js
Pessimistress: No.

modules/core/src/lib/attribute.js
Pessimistress: Redraw and update are two different things. `allocate` is only called during update. The line was excessive.

Issue #3294, opened 2019-06-27
Minimap controller event propagation


serdaroquai:
Hi, thank you for this great project,

In my use case I  have two views, a 'main' one that spans entire window and a 'mini' one on top of the main one. Both manage their own view states.

```
const views = [
        new MapView({ 
            id: "main",
            viewState,
            controller: true,
        }),
        new MapView({
            id: "mini",
            x: "68%", y: "2%", width: "30%",height: "30%",
            clear: true, 
            viewState: miniViewState,
            controller: true
        })
    ];
```
Since both views overlap, problem is controller of main view picks up interaction events from controller of mini. Therefore I can not move around mini view without modifying main view

I tried two approaches so far, first one is by supplying a custom onViewStateChange function to Deck and using the passed in interactionState parameter. It works mostly, however I can not catch mouse wheel zooms (interactionState.isZooming is always undefined) with this approach.

Second approach is I provide a custom controller that extends MapController for the mini view in order to stop propagation of mouse events. Problem is by the time mini view receives the mouse event, main views controller has already processed it.

Any pointers, tips, helps etc is much appreciated, thank you,
(using deck 7.0.5 with react 16.8.6)

Pessimistress:
Try 7.1.10

serdaroquai:
Unfortunately 7.1.10 did not solve the issue. I have two views one overlapping the other and would like to control two view states independently. So when I am hovering above the topmost view  I expect the controller to modify the viewstate belonging to that view. 

After the upgrade, regardless of which view I try to pan,  function onViewStateChange() only logs viewstate updates with the id "main", and is not called for view "minimap" even though minimap is on top. 

Only time onViewStateChange() receives an info with viewId equals "minimap" is at the end of panning when  `interactionState: {isDragging: false}` 

Here is the code I use:

```
    const views = [
        new MapView({ 
            id: "main",
            viewState: viewState,
            controller: true,
        }),
        new MapView({
            id: "minimap",
            x: "68%", y: "2%", width: "30%",height: "30%",
            clear: true, 
            viewState: miniViewState,
            controller: true
        })
    ];

    function onViewStateChange(info) {
        console.log(info);
        const {viewId, interactionState, viewState, oldViewState} = info;
        if ("main" === viewId) {
            setViewState({...viewState, id:"main"});
        } else if ("minimap" === viewId) {
            setMinimapViewState({...viewState, id:"minimap"}); 
        }
    }
```
```
            <DeckGL
[...]

PR #3345, opened 2019-07-12
Fix controller event propagation

Pessimistress:
For #3294 

#### Change List
- Make views on top capture controller events
- Fix test/apps/minimap

### DIFF
--- a/modules/core/src/controllers/controller.js
+++ b/modules/core/src/controllers/controller.js
@@ -110,10 +110,17 @@ export default class Controller {
     return [offsetCenter.x - x, offsetCenter.y - y];
   }
 
-  isPointInBounds(pos) {
+  isPointInBounds(pos, event) {
     const {width, height} = this.controllerStateProps;
+    if (event && event.handled) {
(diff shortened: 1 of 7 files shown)


modules/core/src/controllers/controller.js
```diff
@@ -110,10 +110,17 @@ export default class Controller {
     return [offsetCenter.x - x, offsetCenter.y - y];
   }
 
-  isPointInBounds(pos) {
+  isPointInBounds(pos, event) {
```
xintongxia: update doc to mention the event will not be forwarded if point in bounds

test/apps/minimap/app.js
```diff
@@ -18,34 +18,60 @@ const DATA_URL = {
...
+        }
+      }
     };
+
+    this._onViewStateChange = this._onViewStateChange.bind(this);
   }
 
   componentDidMount() {
-    this._animate();
+    // this._animate();
```
xintongxia: remove? or add comment

PR #3348, opened 2019-07-12
Add data filter example to the website

Pessimistress:
data-filter-example

#### Change List
- Add data filter example


examples/website/data-filter/app.js
```diff
@@ -0,0 +1,181 @@
...
+          />
+        )}
+      </Fragment>
+    );
+  }
+}
+
+export function renderToDOM(container) {
+  render(<App />, container);
+  require('d3-request').csv(DATA_URL, (error, response) => {
```
xintongxia: Cannot use `loaders.gl`?

examples/website/data-filter/range-input.js
```diff
@@ -0,0 +1,110 @@
...
+      newValueMin = min;
+    }
+    this.props.onChange({
+      value: [newValueMin, newValueMin + span]
+    });
+
+    this._animationFrame = requestAnimationFrame(this._animate);
+  }
+
+  _renderThumbValue({$thumbIndex, $value}) {
```
xintongxia: what does `$` represent?

Pessimistress: That's the naming convention that baseui uses.

examples/website/data-filter/app.js
Pessimistress: Loaders' CSV module only supports async iterators.

website/src/components/demos/data-filter-demo.js
```diff
@@ -0,0 +1,66 @@
+import React, {Component} from 'react';
+import {readableInteger} from '../../utils/format-utils';
+import {MAPBOX_STYLES, DATA_URI} from '../../constants/defaults';
+import {App} from 'website-examples/data-filter/app';
+
+export default class HighwayDemo extends Component {
```
ajduberstein: Should the class name be changed?

Issue #3341, opened 2019-07-11
Shadow feature for directional lights


jianhuang01:
#### Proposed feature

Add shadows for multiple directional lights

#### To Do List 
  - [x] Add label and assign to milestone 
  - [x] Coding
  - [x] Doc update
  - [x] Test

PR #3350, opened 2019-07-15
shadow effect RFC

jianhuang01:
For #3341  

#### Change List
- RFC doc

### DIFF
--- a/dev-docs/RFCs/v7.x/shadow-for-directional-lights.md
+++ b/dev-docs/RFCs/v7.x/shadow-for-directional-lights.md
@@ -11,7 +11,7 @@ Deck.gl has new basic lighting effect since 7.0, having shadow effects for direc
 
 ## Proposal
 ### API
-`LightWithShadowEffect` class is the public interface to create shadow effects, it extends from LightingEffect and has one param.
+`LightWithShadowEffect` class is the public interface to create shadow effects, it extends from `LightingEffect` and has one param.
 * `lights`(Object): collection of light sources
 
 New layer prop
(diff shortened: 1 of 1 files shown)


dev-docs/RFCs/v7.x/shadow-for-directional-lights.md
```diff
@@ -0,0 +1,59 @@
...
+      ...})]
+});
+```
+### Workflow
+* User creates LightWithShadowEffect which adds shadow modules into default shader modules
+* LightWithShadowEffect is passed into deck and create shadow passes based on light directions, there is one shadow pass for each directional light, for now totally two lights are supported
+* For each shadow pass, layers are rendered to a shadow map
+* After all the effects are processed, there is a regular layer rendering pass, shadow maps are used to render the final shadows.
+
+## Future Ideas
```
Pessimistress: Limitations

dev-docs/RFCs/v7.x/shadow-for-directional-lights.md
```diff
@@ -0,0 +1,59 @@
...
+## Abstract
+Deck.gl has new basic lighting effect since 7.0, shadow effects for directional lights can improve rendering quality.
+
+## Proposal
+### API
+`LightWithShadowEffect` class is the public interface to create shadow effects, it extends from LightingEffect and has one param.
+* `lights`(Object): collection of light sources
+
+New layer prop
+* `enableShadow`(Boolean): when this prop is true, layer casts and renders shadows
```
Pessimistress: Add to sample code

jianhuang01: will do it in the code PR.

dev-docs/RFCs/v7.x/shadow-for-directional-lights.md
```diff
@@ -0,0 +1,61 @@
...
+});
+```
+### Workflow
+* User creates LightWithShadowEffect which adds shadow modules into default shader modules
+* LightWithShadowEffect is passed into deck and create shadow passes based on light directions, there is one shadow pass for each directional light, for now totally two lights are supported
+* For each shadow pass, layers are rendered to a shadow map
+* After all the effects are processed, there is a regular layer rendering pass, shadow maps are used to render the final shadows.
+
+## Limitations
+ With current design user need create LightWithShadowEffect from the beginning, can't create after app is initialized, so the default modules can be modified before layer shaders are assembled
```
Pessimistress: With the current design, the user needs to create LightWithShadowEffect before Deck's initialization, so that the default modules can be modified before layer shaders are assembled.

PR #3358, opened 2019-07-17
Update documents, color properties descriptions

manishmz:
Updated layer documents color properties descriptions

### DIFF
--- a/docs/layers/arc-layer.md
+++ b/docs/layers/arc-layer.md
@@ -130,7 +130,7 @@ Method called to retrieve the target position of each object.
 
 * Default: `[0, 0, 0, 255]`
 
-The rgba color at the source, in `r, g, b, [a]`. Each component is in the 0-255 range.
+The rgba color at the source, is `[r, g, b, [a]]`. `r, g, b, a` is in 0-255 range and `a` is 255 by default.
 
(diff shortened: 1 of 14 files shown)


1chandu:
Instead of updating the existing text, I think it is good to just add one line specifying `a` affects the transparency:
```
`a` affects the transparency, default value `255` when supplying custom value for `a`, lower value makes it transparent and higher value makes it opaque
```

docs/layers/arc-layer.md
```diff
@@ -130,7 +130,7 @@ Method called to retrieve the target position of each object.
 
 * Default: `[0, 0, 0, 255]`
 
-The rgba color at the source, in `r, g, b, [a]`. Each component is in the 0-255 range.
+The rgba color at the source, is `[r, g, b, [a]]`. `r, g, b, a` is in 0-255 range and `a` is 255 by default.
```
Pessimistress: `...in the format of [r, g, b, [a]]. Each channel is a number between 0-255...`

1chandu: @manishmz can you make this change and push?

1chandu:
Closing in favor of : #3993

PR #3359, opened 2019-07-18
Improve core module test coverage

Pessimistress:


modules/core/src/lib/view-manager.js
```diff
@@ -134,29 +134,6 @@ export default class ViewManager {
...
-   * - [longitude, latitude] to [x, y]
-   * - [longitude, latitude, Z] => [x, y, z]
-   * Note: By default, returns top-left coordinates for canvas/SVG type render
-   *
-   * @param {Array} lngLatZ - [lng, lat] or [lng, lat, Z]
-   * @param {Object} opts - options
-   * @param {Object} opts.topLeft=true - Whether projected coords are top left
-   * @return {Array} - [x, y] or [x, y, z] in top left coords
-   */
-  project(xyz, opts = {topLeft: true}) {
```
Pessimistress: This is 1) broken (there is no `viewport.contains`) and 2) does not serve any reasonable purpose. One lngLat can be projected to multiple screen coordinates in a multi-view scenario. It doesn't make sense to return only one point. This method cannot be fixed without fundamentally changing what it does. Since it's not used anywhere, might as well remove it for now.

PR #3360, opened 2019-07-18
Deprecation warning for IE

tsherif:


modules/core/src/lib/deck.js
```diff
@@ -122,6 +122,8 @@ const defaultProps = {
 /* eslint-disable max-statements */
 export default class Deck {
   constructor(props) {
+    assert(!isIE11(), 'IE 11 not supported');
```
Pessimistress: Use a console log here instead of a hard failure, so that it doesn't break existing apps.

tsherif:
@Pessimistress better?

PR #3378, opened 2019-07-25
Rename frustum plane properties

Pessimistress:


Pessimistress:
> Wasn't there also an example that uses these fields?

That is the test app, already included in this PR

PR #3386, opened 2019-07-29
[HeatmapLayer] add debounce to zoom updates

Pessimistress:
#### Change List
- Make zoom more responsive by debouncing texture updates
- Fix bug in bounds clipping

### DIFF
--- a/modules/aggregation-layers/src/heatmap-layer/heatmap-layer.js
+++ b/modules/aggregation-layers/src/heatmap-layer/heatmap-layer.js
@@ -18,6 +18,7 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
+/* global setTimeout */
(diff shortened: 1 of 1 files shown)


modules/aggregation-layers/src/heatmap-layer/heatmap-layer.js
```diff
@@ -358,19 +348,16 @@ export default class HeatmapLayer extends CompositeLayer {
...
 
-    const commonBounds = this._worldToCommonBounds(worldBounds, {
-      scaleToAspect: true,
-      width: textureSize * RESOLUTION,
-      height: textureSize * RESOLUTION
-    });
+    const {scale} = this.context.viewport;
+    const commonBounds = [
+      normalizedCommonBounds[0].map(x => x * scale),
+      normalizedCommonBounds[1].map(x => x * scale)
```
Pessimistress: @1chandu This is just unnecessary complexity. Can we make the bounds object flat?

PR #3387, opened 2019-07-29
Add shadow effect to LightingEffect class

jianhuang01:
For #3341

#### Change List
- doc update
- Merge LightWithShadow class with LightingEffect class
- Merge unit tests


modules/core/src/effects/lighting/lighting-effect.js
```diff
@@ -15,6 +23,7 @@ const DefaultDirectionalLightProps = [
     direction: [1, 8, -2.5]
   }
 ];
+const DefaultShadowColor = [0, 0, 0, 200 / 255];
```
Pessimistress: Use all uppercase for constants.

PR #3393, opened 2019-08-01
Calc right scale for shadow map

jianhuang01:
For #3341

#### Change List
- calc right scale by unproject 4 screen corners
- remove default shadow module in cleanup function

### DIFF
--- a/modules/core/src/effects/lighting/lighting-effect.js
+++ b/modules/core/src/effects/lighting/lighting-effect.js
@@ -62,12 +62,14 @@ export default class LightingEffect extends Effect {
     if (this.directionalLights.some(light => light.shadow)) {
       this.shadow = true;
       this._addShadowModule();
-      this._createLightMatrix();
     }
   }
 
   prepare(gl, {layers, viewports, onViewportActive, views, pixelRatio}) {
     if (!this.shadow) return {};
+
+    // create light matrix every frame to make sure always updated from light source
+    this._createLightMatrix();
 
     if (this.shadowPasses.length === 0) {
       this._createShadowPasses(gl, pixelRatio);
@@ -87,9 +89,9 @@ export default class LightingEffect extends Effect {
         onViewportActive,
         views,
         effectProps: {
-          shadow_lightId: i,
+          shadowLightId: i,
           dummyShadowMaps: this.dummyShadowMaps,
-          shadow_matrices: this.lightMatrices
+          shadowMatrices: this.lightMatrices
         }
       });
       shadowMaps.push(shadowPass.shadowMap);
@@ -98,9 +100,9 @@ export default class LightingEffect extends Effect {
     return {
       shadowMaps,
       dummyShadowMaps: this.dummyShadowMaps,
-      shadow_lightId: 0,
+      shadowLightId: 0,
       shadowColor: this.shadowColor,
-      shadow_matrices: this.lightMatrices
+      shadowMatrices: this.lightMatrices
     };
   }
 
@@ -131,6 +133,7 @@ export default class LightingEffect extends Effect {
   }
 
   _createLightMatrix() {
+    this.lightMatrices = [];
     for (const light of this.directionalLights) {
       const viewMatrix = new Matrix4().lookAt({
         eye: new Vector3(light.direction).negate()
--- a/modules/core/src/effects/lighting/lighting-effect.js
+++ b/modules/core/src/effects/lighting/lighting-effect.js
@@ -62,13 +62,12 @@ export default class LightingEffect extends Effect {
     if (this.directionalLights.some(light => light.shadow)) {
       this.shadow = true;
       this._addShadowModule();
+      this._createLightMatrix();
     }
   }
 
   prepare(gl, {layers, viewports, onViewportActive, views, pixelRatio}) {
     if (!this.shadow) return {};
-
-    this._createLightMatrix();
 
     if (this.shadowPasses.length === 0) {
       this._createShadowPasses(gl, pixelRatio);
@@ -90,7 +89,7 @@ export default class LightingEffect extends Effect {
         effectProps: {
           shadow_lightId: i,
           dummyShadowMaps: this.dummyShadowMaps,
-          shadow_viewProjectionMatrices: this.lightMatrices
+          shadow_matrices: this.lightMatrices
         }
       });
       shadowMaps.push(shadowPass.shadowMap);
@@ -101,7 +100,7 @@ export default class LightingEffect extends Effect {
       dummyShadowMaps: this.dummyShadowMaps,
       shadow_lightId: 0,
       shadowColor: this.shadowColor,
-      shadow_viewProjectionMatrices: this.lightMatrices
+      shadow_matrices: this.lightMatrices
     };
   }
 
@@ -124,28 +123,20 @@ export default class LightingEffect extends Effect {
       dummyShadowMap.delete();
     }
     this.dummyShadowMaps.length = 0;
+
+    if (this.shadow) {
+      this._removeShadowModule();
+      this.shadow = false;
+    }
   }
 
   _createLightMatrix() {
-    const projectionMatrix = new Matrix4().ortho({
-      left: -1,
-      right: 1,
-      bottom: -1,
-      top: 1,
-      near: 0,
-      far: 2
-    });
-
-    this.lightMatrices = [];
     for (const light of this.directionalLights) {
-      const viewMatrix = new Matrix4()
-        .lookAt({
-          eye: new Vector3(light.direction).negate()
-        })
-        // arbitrary number that covers enough grounds
-        .scale(1e-3);
-      const viewProjectionMatrix = projectionMatrix.clone().multiplyRight(viewMatrix);
-      this.lightMatrices.push(viewProjectionMatrix);
+      const viewMatrix = new Matrix4().lookAt({
+        eye: new Vector3(light.direction).negate()
+      });
+
+      this.lightMatrices.push(viewMatrix);
     }
   }
 
--- a/modules/core/src/effects/lighting/lighting-effect.js
+++ b/modules/core/src/effects/lighting/lighting-effect.js
@@ -99,7 +99,6 @@ export default class LightingEffect extends Effect {
     return {
       shadowMaps,
       dummyShadowMaps: this.dummyShadowMaps,
-      shadowLightId: 0,
       shadowColor: this.shadowColor,
       shadowMatrices
     };
--- a/modules/core/src/effects/lighting/lighting-effect.js
+++ b/modules/core/src/effects/lighting/lighting-effect.js
@@ -35,7 +35,6 @@ export default class LightingEffect extends Effect {
 
     this.shadowColor = DEFAULT_SHADOW_COLOR;
     this.shadowPasses = [];
-    this.lightMatrices = [];
     this.dummyShadowMaps = [];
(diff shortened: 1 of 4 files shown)


modules/core/src/effects/lighting/lighting-effect.js
```diff
@@ -62,14 +62,13 @@ export default class LightingEffect extends Effect {
     if (this.directionalLights.some(light => light.shadow)) {
       this.shadow = true;
       this._addShadowModule();
+      this._createLightMatrix();
```
Pessimistress: Light props (direction, position, etc.) may change after the effect is created. Can you move this back and add a comment so that future maintainers remember.

modules/core/src/shaderlib/shadow/shadow.js
```diff
@@ -118,13 +119,40 @@ function _getViewportCenterPosition({viewport, center}) {
   return new Matrix4(viewport.viewProjectionMatrix).invert().transformVector4(center);
 }
 
+function _getViewProjectionMatrix({viewport, shadowMatrix}) {
```
Pessimistress: No need to add underscore. If the function is not exported, it is not exposed to the user.

modules/core/src/shaderlib/shadow/shadow.js
```diff
@@ -118,13 +119,40 @@ function _getViewportCenterPosition({viewport, center}) {
...
+  const viewMatrix = shadowMatrix.clone().translate(new Vector3(viewport.center).negate());
+
+  const topLeft = viewport.projectPosition(viewport.unproject([0, 0]));
```
Pessimistress: Perf: make a utility that performs unprojection to the common space from:

https://github.com/uber/deck.gl/blob/master/modules/core/src/viewports/viewport.js#L144

Just skip the `unprojectPosition` part.

modules/core/src/effects/lighting/lighting-effect.js
```diff
@@ -90,7 +89,7 @@ export default class LightingEffect extends Effect {
         effectProps: {
           shadow_lightId: i,
           dummyShadowMaps: this.dummyShadowMaps,
-          shadow_viewProjectionMatrices: this.lightMatrices
+          shadow_matrices: this.lightMatrices
```
Pessimistress: Use camel case since this is not matching any shader uniform name.

modules/core/src/shaderlib/shadow/shadow.js
```diff
@@ -118,13 +119,40 @@ function _getViewportCenterPosition({viewport, center}) {
...
+  const pos2 = viewMatrix.transformVector3([bottomLeft[0], bottomLeft[1], 0]);
+  const pos3 = viewMatrix.transformVector3([bottomRight[0], bottomRight[1], 0]);
+
+  const projectionMatrix = new Matrix4().ortho({
+    left: Math.min(pos0[0], pos1[0], pos2[0], pos3[0]),
+    right: Math.max(pos0[0], pos1[0], pos2[0], pos3[0]),
+    bottom: Math.min(pos0[1], pos1[1], pos2[1], pos3[1]),
+    top: Math.max(pos0[1], pos1[1], pos2[1], pos3[1]),
+    // Near plane could be too close to cover tall objects, scale up by 2.0
+    near: Math.min(-pos0[2], -pos1[2], -pos2[2], -pos3[2]) * 2.0,
```
Pessimistress: Does this mean the shadow map is not covering the closest corner?

modules/core/src/shaderlib/shadow/shadow.js
```diff
@@ -135,19 +163,24 @@ function createShadowUniforms(opts = {}, context = {}) {
   const viewProjectionMatrices = [];
   const projectCenters = [];
 
-  for (let i = 0; i < opts.shadow_viewProjectionMatrices.length; i++) {
-    const viewProjectionMatrix = opts.shadow_viewProjectionMatrices[i]
+  for (let i = 0; i < opts.shadow_matrices.length; i++) {
+    const viewProjectionMatrix = _getMemoizedViewProjectionMatrix({
+      shadowMatrix: opts.shadow_matrices[i],
```
Pessimistress: The memoized function is still recomputing every time if you have more than one light.

test/modules/core/shaderlib/shadow/shadow.spec.js
```diff
@@ -57,8 +43,25 @@ test('shadow#getUniforms', t => {
     `Shadow projection center in LNG_LAT mode is correct!`
   );
   t.deepEqual(
-    uniforms[`shadow_uViewProjectionMatrices[0]`],
-    viewProjectionMatrix.clone().translate(new Vector3(viewport.center).negate()),
+    uniforms[`shadow_uViewProjectionMatrices[0]`].toArray(),
```
Pessimistress: A test case is meaningless without explaining why the expected output is there. This function is not returning *correct* result, just *expected* result, if the correctness of the expected value cannot be independently validated. Next time you tweak the implementation, you will have to come here and change these tests. That is to say, these tests do not contribute to you writing more robust code.

A better way to test this would be to pick a few points inside the test viewport, e.g. the center, a near corner, a far corner, both on the ground and with altitude (to test the "high building" use case). Transform those points with your projection matrix and projection center, and then check if they are indeed visible in the shadow texture. Next time you tweak the calculation of these matrices, you will have solid test cases to prove that they still work, or work better.

modules/core/src/shaderlib/shadow/shadow.js
jianhuang01: it covers the closest ground corner but may not cover the closest 3d object top.

modules/core/src/effects/lighting/lighting-effect.js
```diff
@@ -131,6 +133,7 @@ export default class LightingEffect extends Effect {
   }
 
   _createLightMatrix() {
+    this.lightMatrices = [];
```
Pessimistress: This is only used in the current frame. No need to save it on the instance.

modules/core/src/effects/lighting/lighting-effect.js
```diff
@@ -124,28 +125,21 @@ export default class LightingEffect extends Effect {
       dummyShadowMap.delete();
     }
     this.dummyShadowMaps.length = 0;
+
+    if (this.shadow) {
+      this._removeShadowModule();
+      this.shadow = false;
```
Pessimistress: If an effect is removed then added back, it no longer works?

modules/core/src/shaderlib/shadow/shadow.js
```diff
@@ -114,40 +116,75 @@ color = shadow_filterShadowColor(color);
...
 }
 
+function getViewProjectionMatrices({viewport, shadowMatrices}) {
+  const projectionMatrices = [];
+
+  for (const shadowMatrix of shadowMatrices) {
+    const viewMatrix = shadowMatrix.clone().translate(new Vector3(viewport.center).negate());
+    const pixelUnprojectionMatrix = viewport.pixelUnprojectionMatrix;
+
+    const topLeft = pixelsToWorld([0, 0], pixelUnprojectionMatrix);
```
Pessimistress: These can be done outside of the for loop

test/modules/core/shaderlib/shadow/shadow.spec.js
```diff
@@ -5,6 +5,67 @@ import {shadow} from '@deck.gl/core/shaderlib';
 import {Matrix4, Vector3} from 'math.gl';
 import {PROJECT_COORDINATE_SYSTEM} from '@deck.gl/core/shaderlib/project/constants';
 
+const TEST_CASE1 = [
+  {
+    xyz: [83095, 202499, 0],
```
Pessimistress: Use lnglat coordinates and test with multiple viewports (z <12 and z > 12)

jianhuang01: it is done, TEST_CASE1 is for zoom == 10, TEST_CASE2 is for zoom == 15

test/modules/core/shaderlib/shadow/shadow.spec.js
```diff
@@ -94,14 +145,23 @@ test('shadow#getUniforms', t => {
 
   t.deepEqual(
     uniforms[`shadow_uProjectCenters[0]`].toArray(),
-    [0.00003672678690236353, -0.00003393860333744669, 16.320487093036718, 1.0000000000047367],
+    [0.27978817346138385, 0.2796375220259506, 0.057706067456820165, 1.0000000000047367],
```
Pessimistress: Drop this - this is specific to your projection matrix implementation.

test/modules/core/shaderlib/shadow/shadow.spec.js
Pessimistress: You need to document how these numbers are calculated.

jianhuang01: add some comments for test cases

test/modules/core/shaderlib/shadow/shadow.spec.js
```diff
@@ -5,52 +5,50 @@ import {shadow} from '@deck.gl/core/shaderlib';
 import {Matrix4, Vector3} from 'math.gl';
 import {PROJECT_COORDINATE_SYSTEM} from '@deck.gl/core/shaderlib/project/constants';
 
+// unproject viewport corners to ground and use them as reference points to test tight fitting bounding box
```
Pessimistress: Define test viewports with the test cases so that future maintainers know where they come from.

modules/core/src/effects/lighting/lighting-effect.js
```diff
@@ -99,9 +99,9 @@ export default class LightingEffect extends Effect {
     return {
       shadowMaps,
       dummyShadowMaps: this.dummyShadowMaps,
-      shadow_lightId: 0,
+      shadowLightId: 0,
```
Pessimistress: Why does this need to be passed to the module if it never changes?

modules/core/src/shaderlib/shadow/shadow.js
```diff
@@ -114,40 +116,89 @@ color = shadow_filterShadowColor(color);
...
+function getViewportCenterPosition({viewport, center}) {
   return new Matrix4(viewport.viewProjectionMatrix).invert().transformVector4(center);
 }
 
+function getViewProjectionMatrices({viewport, shadowMatrices}) {
+  const projectionMatrices = [];
+
+  const pixelUnprojectionMatrix = viewport.pixelUnprojectionMatrix;
+
+  const topLeftGround = pixelsToWorld([0, 0], pixelUnprojectionMatrix);
```
Pessimistress: Make a utility function that unprojects and add the missing z.

modules/core/src/shaderlib/shadow/shadow.js
```diff
@@ -114,40 +116,89 @@ color = shadow_filterShadowColor(color);
...
+  const topRightGround = pixelsToWorld([viewport.width, 0], pixelUnprojectionMatrix);
+  const bottomLeftGround = pixelsToWorld([0, viewport.height], pixelUnprojectionMatrix);
+  const bottomRightGround = pixelsToWorld(
+    [viewport.width, viewport.height],
+    pixelUnprojectionMatrix
+  );
+  const topLeftNear = pixelsToWorld([0, 0, 0.0], pixelUnprojectionMatrix);
```
Pessimistress: z should be -1 if it's on the near plane

PR #3395, opened 2019-08-02
Remove side effects from core module

Pessimistress:
#### Background

Per discussion in #3213, `sideEffects` is not sufficient to prevent tree shaking in some use cases.

#### Change List
- Wrap global registration in a function

Tested with build-bundle script.

### DIFF
--- a/modules/core/package.json
+++ b/modules/core/package.json
@@ -24,10 +24,7 @@
     "src",
     "dist.min.js"
   ],
-  "sideEffects": [
-    "./dist/**/init.js",
-    "./src/lib/init.js"
-  ],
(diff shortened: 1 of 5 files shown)


Pessimistress:
> I could see some issues in that one might for instance want to call loaders.gl load on something before creating deck and now one does get different loaders before and after deck is instantiated.

That's fair; I have pushed an alternative approach. It should make no change to the current behavior.

SterlingMcCall:
Due to the fact that this was/is a rather inconvenient bug for those affected, would it be possible to release this fix as version 7.1.11 so we don't have to wait until the release of 7.2.0?

Issue #3396, opened 2019-08-05
7.2 Release Checklist


tgorkin:
- [x] Verify whats-new (@tgorkin )
- [x] Verify upgrade-guide (@Pessimistress  )
- [x] Verify all issues in milestone are closed or move out (@tgorkin )
- [x] Verify all API changes are documented (@Pessimistress )
- [x] Website (docs and example) testing: [staging](https://1chandu.github.io/deck.gl/#/)
    - [x] Testing on Chrome (@ajduberstein)
    - [x] Testing on Safari (@jianhuang01)
    - [x] Testing on Firefox (@1chandu)
    - [x] Testing on Edge + IE11 (@tsherif)
    - [x] Testing on iOS (@Pessimistress )
    - [x] Testing on Android (@Pessimistress)
- [x] Cut 7.2-release branch after all fixes
- [x] Update website doc urls to point to `7-2-release` branch  (@1chandu )
- [x] Update 7.2 badge to point to 7.2 branch on Overview page. (@1chandu)
- [x] Update luma.gl dependency to 7.2 prod (@1chandu )
- [x] Publish 7.2 prod version.  (@1chandu )

Testing procedure
- access staged doc website
- go through all examples
- verify functionality in control panels
- look for console errors/warnings
- add results of testing as comments on this issue

jianhuang01:
Safari Testing
1. TripsLayer example is extremely slow
2. Geojson (paths) example is down 
3. Heatmap layer document page is down
5. Data Filter example doesn't show data.

tsherif:
Edge: 👍 
IE: Site does not load at all. There's untranspiled ES6 in the site bundle.js.

tsherif:
IE: Offending dependency (or at least one of them) is [kdbush](https://www.npmjs.com/package/kdbush). Not sure what it's used for, but finding e.g. this line untranspiled in the bundle: https://github.com/mourner/kdbush/blob/master/src/index.js#L14

ajduberstein:
Chrome testing: 💯

A few warnings but all the examples are functional.

`deck: e.componentName not specified` on main page
Screen grid layer produces a warning 95 times, `[.WebGL-0x7fef75911400]RENDER WARNING: Render count or primcount is 0.`

tgorkin:
Doc version badges on https://1chandu.github.io/deck.gl/#/documentation/overview/introduction need to be updated.  Need to add 7.1 and 7.2.
[...]

PR #3410, opened 2019-08-07
Fix PathLayer/trips demo perf regression

Pessimistress:
#### Background

For #3396 

In https://github.com/uber/deck.gl/pull/3211/files#diff-f1f8e62816742b25869f5b5a7c79e7cfR150 we removed the check for very small neighbor segments, and replaced it with explicit end cap flags. Although this reduced ambiguity for end caps, when a path contains two adjacent identical points, the shader produces a big joint triangle which kills rasterization.

This was not caught in testing because we have clean data in layer-browser/render tests. The trips demo has a bug that produces duplicate vertices at navigation points.

#### Change List
- Restore check for neighbor size
- Fix trips demo data decoder

### DIFF
--- a/test/modules/core/lib/pick-layers.spec.js
+++ b/test/modules/core/lib/pick-layers.spec.js
@@ -315,7 +315,7 @@ const TEST_CASES = [
             height: 400
           },
           results: {
-            count: 4
+            count: 3
           }
         },
         {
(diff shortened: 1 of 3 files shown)


test/modules/core/lib/pick-layers.spec.js
```diff
@@ -315,7 +315,7 @@ const TEST_CASES = [
             height: 400
           },
           results: {
-            count: 4
+            count: 3
```
Pessimistress: This reverts the change in https://github.com/uber/deck.gl/pull/3211

PR #3414, opened 2019-08-07
Fix sunlight with shadows

Pessimistress:
image

#### Change List
- Fix SunLight direction
- Fix SunLight shadow
- Add shadows to geojson example

### DIFF
--- a/modules/core/src/effects/lighting/suncalc.js
+++ b/modules/core/src/effects/lighting/suncalc.js
@@ -42,7 +42,7 @@ export function getSunlightDirection(timestamp, latitude, longitude) {
   const azimuthN = azimuth + Math.PI;
 
   // solar position to light direction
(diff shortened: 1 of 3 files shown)


modules/core/src/effects/lighting/suncalc.js
```diff
@@ -42,7 +42,7 @@ export function getSunlightDirection(timestamp, latitude, longitude) {
   const azimuthN = azimuth + Math.PI;
 
   // solar position to light direction
-  return [-Math.sin(azimuthN), -Math.cos(azimuthN), -Math.sin(altitude)];
+  return [-Math.sin(azimuthN), Math.cos(azimuthN), -Math.sin(altitude)];
```
xintongxia: unit test is not impacted by this change?

Pessimistress: I think only `getSolarPosition` is tested.

PR #3419, opened 2019-08-08
Fix set external buffer

xintongxia:
#### Background
External color buffers got cleared when zoom in/out map.

#### Change List
- setExternalBuffer


test/modules/core/lib/attribute.spec.js
```diff
@@ -510,9 +510,9 @@ test('Attribute#setExternalBuffer', t => {
   t.is(attribute.getBuffer(), buffer, 'external buffer is set');
   t.notOk(attribute.needsUpdate(), 'attribute is updated');
 
-  t.notOk(
+  t.ok(
```
Pessimistress: Add spy to `attribute.update` to make sure it's not called in this case.

PR #3420, opened 2019-08-08
Upgrade to loaders.gl@1.2.0

Pessimistress:
#### Change List
- Use `ImageLoader` instead of `HTMLImageLoader` (web worker compatible)
- Support `ImageBitmap` in `IconManager` and `BitmapLayer`


modules/layers/src/bitmap-layer/bitmap-layer.js
```diff
@@ -251,6 +240,13 @@ export default class BitmapLayer extends Layer {
           mipmaps: false
         })
       });
+    } else if (image) {
```
xintongxia: add a comment, what kind of format could be supported.

PR #3440, opened 2019-08-13
Use HTMLImageLoader for browser compatibility

Pessimistress:
For https://github.com/uber/deck.gl/issues/3436

Revert back to HTMLImageLoader for now to unblock. In the long term:

- loaders.gl should detect browser support before using `options` in `createImageBitmap`
- To remove the need for `flipY` option in `createImageBitmap`, we need to flip all textures by setting `unpackFlipY: false` for `Texture2D` (only works for DOM elements as data source) and update the shaders. This will be a breaking change. We may also want to introduce a layer prop to revert the behavior.

#### Change List
- Fix texture loading in FF

### DIFF
--- a/modules/core/src/lib/init.js
+++ b/modules/core/src/lib/init.js
@@ -19,7 +19,7 @@
 // THE SOFTWARE.
 
 import {registerLoaders} from '@loaders.gl/core';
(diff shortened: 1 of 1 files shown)


Pessimistress:
https://github.com/uber-web/loaders.gl/issues/317

PR #3442, opened 2019-08-13
More unit test for SunLight class

jianhuang01:
#### Change List
- Unit test


test/modules/core/effects/lighting/sunlight.spec.js
```diff
@@ -60,3 +63,43 @@ test('Sunlight#azimuth and altitude', t => {
...
+    getPolygon: f => f,
+    getFillColor: (f, {index}) => [index, 0, 0]
+  });
+
+  layer.context = {viewport};
+
+  const projectedLight = sunLight.getProjectedLight({layer});
+  t.deepEqual(
+    projectedLight.direction,
+    [0.8448592259153318, -0.5349886806145961, -0.8670767876117542],
```
1chandu: Nit: to avoid any precision issues you can use `equal` from math.gl or just use `Number.toFixed`.

PR #3443, opened 2019-08-13
[POC] generic prop transition

Pessimistress:
For https://github.com/uber/deck.gl/blob/master/dev-docs/RFCs/v7.x/property-transition-rfc.md

#### Change List

- Update prop transition RFC
- Drop support for luma-style function uniforms
- Support uniform transition in primitive layers
- Support `elevationScale` and `coverage` transitions in HexagonLayer
- Update HexagonLayer example

#### TODO

- Update all composite layers to map prop transition settings
- Update layer docs to mark transition-enabled props
- Tests

### DIFF
--- a/examples/website/3d-heatmap/app.js
+++ b/examples/website/3d-heatmap/app.js
@@ -1,4 +1,3 @@
-/* global window */
 import React, {Component} from 'react';
 import {render} from 'react-dom';
 import {StaticMap} from 'react-map-gl';
@@ -95,7 +94,7 @@ export class App extends Component {
         material,
 
         transitions: {
-          elevationScale: 1000
+          elevationScale: 3000
         }
       })
     ];
--- a/examples/website/3d-heatmap/app.js
+++ b/examples/website/3d-heatmap/app.js
@@ -72,50 +72,6 @@ export class App extends Component {
     this.state = {
       elevationScale: elevationScale.min
     };
-
-    this.startAnimationTimer = null;
-    this.intervalTimer = null;
-
-    this._startAnimate = this._startAnimate.bind(this);
-    this._animateHeight = this._animateHeight.bind(this);
-  }
-
-  componentDidMount() {
-    this._animate();
-  }
-
-  componentWillReceiveProps(nextProps) {
-    if (nextProps.data && this.props.data && nextProps.data.length !== this.props.data.length) {
-      this._animate();
-    }
-  }
-
-  componentWillUnmount() {
-    this._stopAnimate();
-  }
-
-  _animate() {
-    this._stopAnimate();
-
-    // wait 1.5 secs to start animation so that all data are loaded
-    this.startAnimationTimer = window.setTimeout(this._startAnimate, 1500);
-  }
-
-  _startAnimate() {
-    this.intervalTimer = window.setInterval(this._animateHeight, 20);
-  }
-
-  _stopAnimate() {
-    window.clearTimeout(this.startAnimationTimer);
-    window.clearTimeout(this.intervalTimer);
-  }
-
-  _animateHeight() {
-    if (this.state.elevationScale === elevationScale.max) {
-      this._stopAnimate();
-    } else {
-      this.setState({elevationScale: this.state.elevationScale + 1});
-    }
   }
 
   _renderLayers() {
@@ -128,7 +84,7 @@ export class App extends Component {
         coverage,
         data,
         elevationRange: [0, 3000],
-        elevationScale: this.state.elevationScale,
+        elevationScale: data && data.length ? 50 : 0,
         extruded: true,
         getPosition: d => d,
         onHover: this.props.onHover,
@@ -136,7 +92,11 @@ export class App extends Component {
         pickable: Boolean(this.props.onHover),
         radius,
         upperPercentile,
-        material
+        material,
+
+        transitions: {
+          elevationScale: 1000
+        }
       })
     ];
   }
--- a/modules/core/src/lib/layer-manager.js
+++ b/modules/core/src/lib/layer-manager.js
@@ -20,6 +20,7 @@
 
 import assert from '../utils/assert';
 import {_ShaderCache as ShaderCache} from '@luma.gl/core';
+import {Timeline} from '@luma.gl/addons';
 import seer from 'seer';
 import Layer from './layer';
 import {LIFECYCLE} from '../lifecycle/constants';
@@ -45,7 +46,6 @@ const INITIAL_CONTEXT = Object.seal({
   layerManager: null,
   deck: null,
   gl: null,
-  time: -1,
 
   // Settings
   useDevicePixels: true, // Exposed in case custom layers need to adjust sizes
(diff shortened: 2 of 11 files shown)


modules/core/src/lib/layer.js
```diff
@@ -679,23 +699,21 @@ export default class Layer extends Component {
...
+    const currentProps = this.props;
     if (!uniforms.picking_uActive) {
-      this.updateTransition();
+      // Overwrite this.props during redraw to use in-transition prop values
+      this.props = this.updateTransition();
     }
 
+    const {opacity} = this.props;
+    // apply gamma to opacity to make it visually "linear"
+    uniforms.opacity = Math.pow(opacity, 1 / 2.2);
```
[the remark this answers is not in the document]
Pessimistress: Opacity can potentially be in transition.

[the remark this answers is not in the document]
tsherif: I understand that, but why is it gamma corrected here? I don't remember seeing gamma correction anywhere else in deck.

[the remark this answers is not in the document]
Pessimistress: This is existing logic copied from the current `_updateBaseUniforms` method. @ibgreen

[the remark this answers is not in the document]
ibgreen: Yes, we do apply this gamma to opacity before passing it to the shader. 

This is just the classic gamma correction to make the opacity perceptually linear on typical non-linear computer screens.

Try playing with an opacity slider in a layer (perhaps in the layer browser) with and without this correction...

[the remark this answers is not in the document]
tsherif: We can discuss in detail elsewhere, since it's unrelated to this PR, but:
- gamma correction is a color correction. It isn't meant for opacity (see [e.g.](https://learnopengl.com/Advanced-Lighting/Gamma-Correction)).
- If you aren't transitioning to 0 or 1, you'll end up on the wrong value.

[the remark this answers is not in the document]
ibgreen: > gamma correction is a color correction. It isn't meant for opacity

I think I get what you are saying. 

To be clear, this was not added because of an incorrect assumption that this is the "blessed or correct way" to handle opacity. 

It was added after experimentation because it gives noticeable improvements in the behavior of this prop. 

We essentially use it as a brightness correction heuristic (which actually seems consistent with the linked article). 

Have nothing against replacing it with some other adjustment, but I would not be in favor of simply removing the correction as it solves a problem.

> If you aren't transitioning to 0 or 1, you'll end up on the wrong value.

If that is the case it would be a bug in this PR. The gamma adjustment should not affect the prop value being transitioned, only the derived uniform values being passed to the shader.

examples/website/3d-heatmap/app.js
```diff
@@ -128,15 +83,19 @@ export class App extends Component {
         coverage,
         data,
         elevationRange: [0, 3000],
-        elevationScale: this.state.elevationScale,
+        elevationScale: data && data.length ? 50 : 0,
```
ibgreen: One almost wants a way to specify that "transitions" should start at data load time, so this could be dropped...

modules/aggregation-layers/src/heatmap-layer/triangle-layer.js
```diff
@@ -65,7 +65,11 @@ export default class TriangleLayer extends Layer {
   draw({uniforms}) {
     const {model} = this.state;
     const {texture, maxTexture, colorTexture, intensity, threshold} = this.props;
-    model.setUniforms({texture, maxTexture, colorTexture, intensity, threshold}).draw();
+    model
```
ibgreen: I know you prefer Object.assign, which is fine. Since ES2018 now finally supports spread operator `...` as part of the JS standard, other repos like loaders.gl are adopting it.

modules/core/src/lib/layer-manager.js
```diff
@@ -208,7 +209,7 @@ export default class LayerManager {
   // Update layers from last cycle if `setNeedsUpdate()` has been called
   updateLayers(animationProps = {}) {
     if ('time' in animationProps) {
-      this.context.time = animationProps.time;
+      this.context.timeline.setTime(animationProps.time);
```
ibgreen: I presume I am wrong, but I was under the impression that timelines could be added directly to animation loop and have time synced automatically?

modules/core/src/lib/layer.js
```diff
@@ -324,6 +325,13 @@ export default class Layer extends Component {
         attributeManager.invalidateAll();
       }
     }
+
+    if (!this.isComposite && changeFlags.transitionsChanged) {
```
ibgreen: Add some comment on how we treat (or not) composite layers, i.e. the `if (!this.isComposite)` part

modules/core/src/lib/layer.js
```diff
@@ -324,6 +325,13 @@ export default class Layer extends Component {
...
+      for (const key in changeFlags.transitionsChanged) {
+        // prop changed and transition is enabled
+        this.internalState.transitions.add(key, oldProps[key], props[key], props.transitions[key]);
```
ibgreen: Are these `transitions` used for both attributes and uniforms? Or should we call this `propTransitions`?

modules/core/src/lib/layer.js
```diff
@@ -417,12 +426,24 @@ export default class Layer extends Component {
...
     if (attributeManager) {
-      attributeManager.updateTransition(this.context.time);
+      attributeManager.updateTransition(this.context.timeline.getTime());
+    }
+
+    const {transitions} = this.internalState;
+    if (transitions.size > 0) {
+      // clone props
+      const propsInTransition = transitions.update();
+      const props = Object.create(this.props);
```
ibgreen: Nits:
- Generates a bunch of GC objects during transitions.

modules/core/src/lib/layer.js
```diff
@@ -903,7 +928,7 @@ ${flags.viewportChanged ? 'viewport' : ''}\
     this.state = {};
     // TODO deprecated, for backwards compatibility with older layers
     this.state.attributeManager = attributeManager;
-
+    this.internalState.transitions = new UniformTransitionManager(this.context.timeline);
```
ibgreen: call this `uniformTransitions` to distinguish from attribute transitions?

modules/core/src/lib/layer.js
```diff
@@ -679,23 +699,21 @@ export default class Layer extends Component {
 
   // Calculates uniforms
   drawLayer({moduleParameters = null, uniforms = {}, parameters = {}}) {
+    const currentProps = this.props;
     if (!uniforms.picking_uActive) {
-      this.updateTransition();
+      // Overwrite this.props during redraw to use in-transition prop values
+      this.props = this.updateTransition();
```
ibgreen: Unrelated: The fact that `layer.props` can be changed is interesting. Wonder if applications could use that to subvert having to create new layers...

modules/core/src/lib/uniform-transition-manager.js
```diff
@@ -0,0 +1,67 @@
+import {lerp} from 'math.gl';
+import {normalizeTransitionSettings} from './attribute-transition-utils';
+
+export default class UniformTransitionManager {
+  constructor(timeline) {
+    this.transitions = new Map();
+    this.timeline = timeline;
+  }
+
+  get size() {
```
ibgreen: Nit: Do we need this method? I only see this being used to check if `empty()` or rather `active()`, which migh be more self-documenting at call sites.

modules/core/src/lib/layer.js
```diff
@@ -679,23 +699,21 @@ export default class Layer extends Component {
...
     }
 
+    const {opacity} = this.props;
+    // apply gamma to opacity to make it visually "linear"
```
ibgreen: Nit: Seems like this gamma conversion should be specified in a table somewhere, rather than being a special case here, could even be in the prop types...

modules/core/src/lib/layer.js
Pessimistress: The prop will contain both attribute and uniform transitions.

examples/website/3d-heatmap/app.js
Pessimistress: In the website data is passed in from the parent, so `onDataLoad` will not fire.

PR #3452, opened 2019-08-15
Improve shadow for mesh layer

jianhuang01:
#### Background
Mesh layer has a sampler which blocks the binding of shadow map sampler array. Because of that the shadow map texture sampling always returns [0, 0, 0, 0]. Made the full mesh layer rendered with shadow.

#### Change List
- fix the shadow map sampler issue
- new bounding box for identity mode
- add shadow to mesh example
- unit test

### DIFF
--- a/examples/website/mesh/app.js
+++ b/examples/website/mesh/app.js
@@ -56,7 +56,7 @@ const dirLight = new DirectionalLight({
 
 const lightingEffect = new LightingEffect({ambientLight, dirLight});
 
-const landCover = [
+const background = [
   [[-1000.0, -1000.0, -40], [1000.0, -1000.0, -40], [1000.0, 1000.0, -40], [-1000.0, 1000.0, -40]]
 ];
 
@@ -73,8 +73,8 @@ class Example extends PureComponent {
         getOrientation: d => d.orientation
       }),
       new SolidPolygonLayer({
-        id: 'land',
-        data: landCover,
+        id: 'background',
+        data: background,
         opacity: 1,
         extruded: false,
         coordinateSystem: COORDINATE_SYSTEM.IDENTITY,
--- a/examples/website/mesh/app.js
+++ b/examples/website/mesh/app.js
@@ -1,6 +1,14 @@
 import React, {PureComponent} from 'react';
 import {render} from 'react-dom';
-import DeckGL, {COORDINATE_SYSTEM, SimpleMeshLayer, OrbitView} from 'deck.gl';
+import DeckGL, {
+  COORDINATE_SYSTEM,
+  SimpleMeshLayer,
+  OrbitView,
+  SolidPolygonLayer,
+  DirectionalLight,
+  LightingEffect,
+  AmbientLight
+} from 'deck.gl';
 
 import {OBJLoader} from '@loaders.gl/obj';
 import {registerLoaders} from '@loaders.gl/core';
@@ -34,6 +42,24 @@ const SAMPLE_DATA = (([xCount, yCount], spacing) => {
   return data;
 })([10, 10], 120);
 
+const ambientLight = new AmbientLight({
+  color: [255, 255, 255],
+  intensity: 1.0
+});
+
+const dirLight = new DirectionalLight({
+  color: [255, 255, 255],
+  intensity: 1.0,
+  direction: [-10, -2, -15],
+  _shadow: true
+});
+
+const lightingEffect = new LightingEffect({ambientLight, dirLight});
+
+const landCover = [
+  [[-1000.0, -1000.0, -40], [1000.0, -1000.0, -40], [1000.0, 1000.0, -40], [-1000.0, 1000.0, -40]]
+];
+
 class Example extends PureComponent {
   render() {
     const layers = [
@@ -45,15 +71,30 @@ class Example extends PureComponent {
         getPosition: d => d.position,
         getColor: d => d.color,
         getOrientation: d => d.orientation
+      }),
+      new SolidPolygonLayer({
+        id: 'land',
+        data: landCover,
+        opacity: 1,
+        extruded: false,
+        coordinateSystem: COORDINATE_SYSTEM.IDENTITY,
+        getPolygon: f => f,
+        getFillColor: [0, 0, 0, 0]
       })
     ];
 
     return (
       <DeckGL
-        views={new OrbitView()}
+        views={
+          new OrbitView({
+            near: 0.1,
+            far: 2
+          })
+        }
         initialViewState={INITIAL_VIEW_STATE}
         controller={true}
         layers={layers}
+        effects={[lightingEffect]}
       />
     );
   }
--- a/examples/website/mesh/app.js
+++ b/examples/website/mesh/app.js
@@ -72,6 +72,7 @@ class Example extends PureComponent {
         getColor: d => d.color,
         getOrientation: d => d.orientation
       }),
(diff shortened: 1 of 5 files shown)


modules/core/src/shaderlib/shadow/shadow.js
```diff
@@ -81,7 +88,7 @@ vec4 shadow_filterShadowColor(vec4 color) {
     float shadowAlpha = 0.0;
     for (int i = 0; i < max_lights; i++) {
       if(i < int(shadow_uLightCount)) {
-        shadowAlpha += shadow_getShadowWeight(shadow_vPosition[i], shadow_uShadowMap[i]) * shadow_uColor.a / shadow_uLightCount;
+        shadowAlpha += shadow_getShadowWeight(shadow_vPosition[i], i) * shadow_uColor.a / shadow_uLightCount;
```
Pessimistress: if statements are generally considered bad for shader performance. Since you only have two lights, might as well do

```glsl
shadowAlpha += shadow_getShadowWeight(shadow_vPosition[0], shadow_uShadowMap0);
shadowAlpha += shadow_getShadowWeight(shadow_vPosition[1], shadow_uShadowMap1);
shadowAlpha *= a;
```

examples/website/mesh/app.js
```diff
@@ -34,6 +42,24 @@ const SAMPLE_DATA = (([xCount, yCount], spacing) => {
...
+const dirLight = new DirectionalLight({
+  color: [255, 255, 255],
+  intensity: 1.0,
+  direction: [-10, -2, -15],
+  _shadow: true
+});
+
+const lightingEffect = new LightingEffect({ambientLight, dirLight});
+
+const landCover = [
```
Pessimistress: This is not really "land", call it background or something?

modules/core/src/shaderlib/shadow/shadow.js
```diff
@@ -133,16 +140,30 @@ function getViewportCenterPosition({viewport, center}) {
...
-    [viewport.width, 0], // top right ground
-    [0, viewport.height], // bottom left ground
-    [viewport.width, viewport.height], // bottom right ground
-    [0, 0, -1.0], // top left near
-    [viewport.width, 0, -1.0], // top right near
-    [0, viewport.height, -1.0], // bottom left near
-    [viewport.width, viewport.height, -1.0] // bottom right near
-  ].map(pixel => screenToCommonSpace(pixel, pixelUnprojectionMatrix));
+  let corners = [];
+  if (viewport.isGeospatial) {
```
Pessimistress: `farDepth = viewports.isGeospatial ? 0 : 1` then only define `corners` once

modules/core/src/shaderlib/shadow/shadow.js
```diff
@@ -133,16 +140,30 @@ function getViewportCenterPosition({viewport, center}) {
...
+      [0, viewport.height], // bottom left ground
+      [viewport.width, viewport.height], // bottom right ground
+      [0, 0, -1.0], // top left near
+      [viewport.width, 0, -1.0], // top right near
+      [0, viewport.height, -1.0], // bottom left near
+      [viewport.width, viewport.height, -1.0] // bottom right near
+    ].map(pixel => screenToCommonSpace(pixel, pixelUnprojectionMatrix));
+  } else {
+    corners = [
+      [0, 0, 1.0], // top left far
```
Pessimistress: 1.0 and 1 are the same thing in js

modules/core/src/shaderlib/shadow/shadow.js
```diff
@@ -202,9 +223,9 @@ function createShadowUniforms(opts = {}, context = {}) {
     uniforms[`shadow_uProjectCenters[${i}]`] = projectCenters[i];
 
     if (opts.shadowMaps && opts.shadowMaps.length > 0) {
-      uniforms[`shadow_uShadowMap[${i}]`] = opts.shadowMaps[i];
+      uniforms[`shadow_uShadowMap${i}`] = opts.shadowMaps[i];
     } else {
-      uniforms[`shadow_uShadowMap[${i}]`] = opts.dummyShadowMaps[0];
+      uniforms[`shadow_uShadowMap${i}`] = opts.dummyShadowMaps[0];
```
Pessimistress: Why do we need multiple dummy textures if only one is used?

modules/core/src/shaderlib/shadow/shadow.js
jianhuang01: here it is a little different, without the depth z by default our API assumes the target z to be 0, not depth z to be zero.

Pessimistress: You can set it to undefined in that case

modules/core/src/shaderlib/shadow/shadow.js
```diff
@@ -79,11 +80,8 @@ vec4 shadow_filterShadowColor(vec4 color) {
   }
   if (shadow_uUseShadowMap) {
     float shadowAlpha = 0.0;
-    for (int i = 0; i < max_lights; i++) {
-      if(i < int(shadow_uLightCount)) {
-        shadowAlpha += shadow_getShadowWeight(shadow_vPosition[i], shadow_uShadowMap[i]) * shadow_uColor.a / shadow_uLightCount;
-      }
-    }
+    shadowAlpha += shadow_getShadowWeight(shadow_vPosition[0], shadow_uShadowMap0) * shadow_uColor.a / shadow_uLightCount;
+    shadowAlpha += shadow_getShadowWeight(shadow_vPosition[1], shadow_uShadowMap1) * shadow_uColor.a / shadow_uLightCount;
```
Pessimistress: Perf: Apply constant multiplier ` * shadow_uColor.a / shadow_uLightCount` only once

modules/core/src/shaderlib/shadow/shadow.js
```diff
@@ -79,11 +80,9 @@ vec4 shadow_filterShadowColor(vec4 color) {
   }
   if (shadow_uUseShadowMap) {
     float shadowAlpha = 0.0;
-    for (int i = 0; i < max_lights; i++) {
-      if(i < int(shadow_uLightCount)) {
-        shadowAlpha += shadow_getShadowWeight(shadow_vPosition[i], shadow_uShadowMap[i]) * shadow_uColor.a / shadow_uLightCount;
-      }
-    }
+    shadowAlpha += shadow_getShadowWeight(shadow_vPosition[0], shadow_uShadowMap0);
+    shadowAlpha += shadow_getShadowWeight(shadow_vPosition[1], shadow_uShadowMap1);
```
Pessimistress: Do you need to check `shadow_uLightCount` here?

jianhuang01: no, if the count is less than one, the shadow_uUseShadowMap won't be true, this code snippet won't be executed.

Pessimistress: What if the count is 1? What is in `shadow_uShadowMap1`?

examples/website/mesh/app.js
```diff
@@ -45,15 +71,30 @@ class Example extends PureComponent {
         getPosition: d => d.position,
         getColor: d => d.color,
         getOrientation: d => d.orientation
+      }),
+      new SolidPolygonLayer({
```
Pessimistress: Add a comment, see https://github.com/uber/deck.gl/blob/master/examples/website/geojson/app.js#L80

PR #3453, opened 2019-08-15
Bug fix: OrbitView with orbitAxis='Z'

Pessimistress:
#### Change List

- Fix view matrix in OrbitViewport
- Tests


xintongxia:
Does this mean I can make the skybox example?

Pessimistress:
> Does this mean I can make the skybox example?

I thought the skybox use case is in MapView?

Issue #3456, opened 2019-08-16
GoogleMapsOverlay click on map markers is broken


pulges:
Adding a layer via GoogleMapsOverlay to google map brakes clicking on googles own markers (landmarks, shops, etc.). These markers that appear when map is zoomed in close enough.
Console displays error: "Uncaught TypeError: Cannot read property 'x' of undefined".

Pessimistress:
Fixed in 8.2.4

PR #3457, opened 2019-08-16
Fixes google marker click error

pulges:
It fixes the issue #3456 "GoogleMapsOverlay click on map markers is broken"

I debugged the problem and this line: https://github.com/uber/deck.gl/blob/master/modules/google-maps/src/utils.js#L138 expects `.pixel` property to be present in `event`.

The problem is that `.pixel`is an undocumented feature and is not present when user clicks on a marker. https://developers.google.com/maps/documentation/javascript/reference/map#MouseEvent

If `.pixel` is not present, I found a way to calculate x and y from native mouseEvent properties passed in google maps event, but this relies also on an undocumented feature.

Only documented google property is `latLng` but it does not always return position of mouse.
For example it returns position of google POI center if clicked on it. Thus I use it as a fallback here in case all previous fails.

### DIFF
--- a/modules/google-maps/src/utils.js
+++ b/modules/google-maps/src/utils.js
@@ -1,4 +1,4 @@
-/* global document, google */
+/* global document, google, window */
 import {Deck} from '@deck.gl/core';
 
 /**
@@ -39,7 +39,7 @@ export function createDeckInstance(map, overlay, deck) {
   // Register event listeners
   for (const eventType in eventListeners) {
     eventListeners[eventType] = map.addListener(eventType, evt =>
-      handleMouseEvent(deck, eventType, evt)
+      handleMouseEvent(deck, eventType, evt, map)
     );
   }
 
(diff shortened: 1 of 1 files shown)


Pessimistress:
Thank you for the quick fix! The solution is hacky but otherwise looks fine if this is what we have to do. cc @donmccurdy Any suggestions?

donmccurdy:
Thanks – @pulges I believe you're correct that `event.latLng` is the only guaranteed property on the mouse event from Google Maps API. However, you might be able to calculate the pixel coordinates much more easily using the map's Projection object.

See `fromLatLngToContainerPixel` or `fromLatLngToDivPixel` here: https://developers.google.com/maps/documentation/javascript/reference/overlay-view#MapCanvasProjection.fromLatLngToContainerPixel

If that works, I'd suggest using this helper function and just ignoring `.pixel` entirely.

Pessimistress:
If you are always projecting latLng back to pixels, you can also use

```js
deck.getViewports()[0].project([longitude, latitude])
```

pulges:
I tested using just `event.latLng`, but it did not work as expected in our use-case. We are using deck to draw a lot of clickable points on map (more than google tools can handle), that are often located near google markers. As google gives `latLng` of the marker, not mouse position, it makes clicking/selecting our points really difficult and sometimes impossible (google markers center can be out of dots bounds).

Looking at @Pessimistress suggestion I'm wondering, what is the reason behind listening the mouse events through google map, not directly from the deck instance?

I'm asking this as I found out that the mousemove events (and actually no other events too)  are not fired when mouse moves over a google marker. Thus hover states get stuck too until user moves out of markers range.

donmccurdy:
> ...what is the reason behind listening the mouse events through google map, not directly from the deck instance?

Mouse events are captured by Maps API, and passed through to deckgl with synthetic events. This cannot currently be changed, as capturing mouse events with deckgl would prevent the map from functioning.

I'm pretty surprised to hear that mousemove events would stop when moving over a point-of-interest (POI). Would you be able to share a demo of this problem? If it helps, you can disable POIs entirely: https://stackoverflow.com/a/7510891/1314762.

And just to confirm, you mean the points of interest that come on the map by default, not something you've added with `google.maps.Marker`?

SterlingMcCall:
@Pessimistress This PR seems to be neglected. The most simplistic fix would be to simply add
```the issue that `'click'`

SterlingMcCall:
@Pessimistress This PR seems to be neglected. The most simplistic fix would be to stop the exception from being thrown by simply adding `if (!event.pixel) { return; }` to the `'click'` case within `handleMouseEvent()`.

Is it possible that quick fix could be implemented for the time being, and then this larger discussion be continued if there is to be a more robust solution? As it currently stands, Google maps POIs cannot be clicked when using deck.gl and that is problematic.

Pessimistress:
I believe the only hold up here is to use `fromLatLngToContainerPixel` instead of the custom `getXYFromLatLng`.

FYI @pulges you can ignore the CLA; it is no longer required.

Pessimistress:
Closing in favor of #4810

PR #3473, opened 2019-08-20
[Feat] Hexagon/Grid, getColorValue, getElevationValue based on updateTriggers

heshan0131:
#### Background
Currently in both Hexagon and Grid layer. diff prop of `getColorValue`, `getElevationValue`, `getColorWeight`, `getElevationWeight` are based on shallow comparison. This diff moved them to be checked using updateTriggers.
#### Change List
- [Breaking] Update props above without passing updateTriggers will no longer trigger layer update.

### DIFF
--- a/docs/layers/cpu-grid-layer.md
+++ b/docs/layers/cpu-grid-layer.md
@@ -193,19 +193,14 @@ It takes an array of points inside each cell as arguments, returns a number. For
 You can pass in `getColorValue` to color the cells by avg/mean/max of a specific attributes of each point.
 By default `getColorValue` returns the length of the points array.
 
-Note: grid layer compares whether `getColorValue` has changed to recalculate the value for each bin that its color based on.
-You should pass in the function defined outside the render function so it doesn't create a new function on every rendering pass.
+`getColorValue` is compared using [`updateTriggers`](https://deck.gl/#/documentation/deckgl-api-reference/layers/layer?section=_datadiff-function-optional-experimental)
 
 ```js
  class MyGridLayer {
-    getColorValue (points) {
-        return points.length;
-    }
-
     renderLayers() {
       return new CPUGridLayer({
         id: 'grid-layer',
-        getColorValue: this.getColorValue // instead of getColorValue: (points) => { return points.length; }
+        getColorValue: points => points.length
         data,
         cellSize: 500
       });
@@ -220,6 +215,7 @@ You should pass in the function defined outside the render function so it doesn'
 
 `getColorWeight` is the accessor function to get the weight of a point used to calculate the color value for a cell.
 
+`getColorWeight` is compared using [`updateTriggers`](https://deck.gl/#/documentation/deckgl-api-reference/layers/layer?section=_datadiff-function-optional-experimental).
 
 ##### `colorAggregation` (String, optional)
 
@@ -233,28 +229,24 @@ Note: `getColorWeight` and `colorAggregation` together define how color value of
 
 * Using `getColorValue`
 ```js
-function getCount(points) {
-  return points.length;
-}
+
 ...
 const layer = new CPUGridLayer({
   id: 'my-grid-layer',
   ...
-  getColorValue: getCount,
+  getColorValue: points => points.length,
   ...
 });
 ```
 
 * Using `getColorWeight` and `colorAggregation`
 ```js
-function getWeight(point) {
-  return 1;
-}
+
 ...
 const layer = new CPUGridLayer({
   id: 'my-grid-layer',
   ...
-  getColorWeight: getWeight,
+  getColorWeight: point => 1,
   colorAggregation: 'SUM'
   ...
 });
@@ -278,14 +270,11 @@ const layer = new CPUGridLayer({
 
 * Using `getColorWeight` and `colorAggregation`
 ```js
-function getWeight(point) {
-  return point.SPACES;
-}
 ...
 const layer = new CPUGridLayer({
   id: 'my-grid-layer',
   ...
-  getColorWeight: getWeight,
+  getColorWeight: point => point.SPACES,
   colorAggregation: 'SUM'
   ...
 });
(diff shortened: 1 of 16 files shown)


modules/aggregation-layers/src/cpu-grid-layer/cpu-grid-layer.js
```diff
@@ -81,44 +78,42 @@ export default class CPUGridLayer extends CompositeLayer {
...
-  updateGetValueFuncs(oldProps, props) {
+  updateGetValueFuncs(oldProps, props, changeFlags) {
+    const {getFillColor, getElevation} = this.getDimensionUpdaters();
     let {getColorValue, getElevationValue} = props;
     const {colorAggregation, getColorWeight, elevationAggregation, getElevationWeight} = this.props;
-    const {colorChanged, elevationChanged} = this.colorElevationPropsChanged(oldProps, props);
 
-    if (colorChanged && getColorValue === null) {
+    const getColorValueChanged = this.needUpdateDimensionStep(
+      getFillColor[0],
```
Pessimistress: Should this be a map instead of an array?

modules/aggregation-layers/src/cpu-grid-layer/cpu-grid-layer.js
```diff
@@ -181,15 +178,31 @@ export default class CPUGridLayer extends CompositeLayer {
...
+    return dimensionStep.triggers.some(t => {
+      if (dimensionStep.updateTriggers && dimensionStep.updateTriggers[t]) {
+        // check based on updateTriggers change first
+        return (
+          changeFlags.updateTriggersChanged &&
+          (changeFlags.updateTriggersChanged.all || changeFlags.updateTriggersChanged[t])
+        );
+      }
+      // fallback to direct comparison
+      return oldProps[t] !== props[t];
```
Pessimistress: This is inconsistent with the rest of the layers - we ignore shallow change in all accessors and exclusively rely on update triggers.

modules/aggregation-layers/src/cpu-grid-layer/cpu-grid-layer.js
heshan0131: Made it a map inside cpuAggregationManager

heshan0131:
> Seems to me that there are a lot of duplicated code among the aggregation layers. Maybe we can have a common class that the grid layers and the hexagon layer all extend?

- created a `cpuAggregationManager` class to manage aggregation in hexagon and grid layer.

modules/aggregation-layers/src/cpu-grid-layer/cpu-grid-layer.js
heshan0131: direct comparison only applies to layer props that are not an accessor. For all accessors in dimensionUpdater, we only check updateTriggers

docs/layers/cpu-grid-layer.md
```diff
@@ -193,19 +193,14 @@ It takes an array of points inside each cell as arguments, returns a number. For
 You can pass in `getColorValue` to color the cells by avg/mean/max of a specific attributes of each point.
 By default `getColorValue` returns the length of the points array.
 
-Note: grid layer compares whether `getColorValue` has changed to recalculate the value for each bin that its color based on.
-You should pass in the function defined outside the render function so it doesn't create a new function on every rendering pass.
+`getColorValue` is compared using [`updateTriggers`](https://deck.gl/#/documentation/deckgl-api-reference/layers/layer?section=_datadiff-function-optional-experimental)
```
Pessimistress: - Wrong link
- Why do we need to call out `getColorValue` specifically? Can we just link to the [accessors section](https://github.com/uber/deck.gl/blob/master/docs/developer-guide/using-layers.md#accessors) like any other prop?

docs/layers/cpu-grid-layer.md
```diff
@@ -220,6 +215,7 @@ You should pass in the function defined outside the render function so it doesn'
 
 `getColorWeight` is the accessor function to get the weight of a point used to calculate the color value for a cell.
 
+`getColorWeight` is compared using [`updateTriggers`](https://deck.gl/#/documentation/deckgl-api-reference/layers/layer?section=_datadiff-function-optional-experimental).
```
Pessimistress: Ditto

modules/aggregation-layers/src/cpu-grid-layer/cpu-grid-layer.js
```diff
@@ -63,305 +66,52 @@ const defaultProps = {
   material: defaultMaterial
 };
 
-const COLOR_PROPS = ['getColorValue', 'colorAggregation', 'getColorWeight'];
-const ELEVATION_PROPS = ['getElevationValue', 'elevationAggregation', 'getElevationWeight'];
-
 export default class CPUGridLayer extends CompositeLayer {
   initializeState() {
+    const cpuAggregationManager = new CPUAggregationManager({
```
Pessimistress: Nit: call it `CPUAggregator` corresponding to `GPUAggregator`?

modules/aggregation-layers/src/cpu-grid-layer/cpu-grid-layer.js
```diff
@@ -63,305 +66,52 @@ const defaultProps = {
...
-    const isColorValueInDomain = cv >= colorDomain[0] && cv <= colorDomain[colorDomain.length - 1];
-
-    // if cell value is outside domain, set alpha to 0
-    const color = isColorValueInDomain ? colorScaleFunc(cv) : [0, 0, 0, 0];
-
-    // add alpha to color if not defined in colorRange
-    color[3] = Number.isFinite(color[3]) ? color[3] : 255;
-
-    return color;
+    return this.state.cpuAggregationManager.dimensionUpdaters.fillColor.attributeAccessor(cell);
```
Pessimistress: Nit: make a method `cpuAggregationManager.getAccessor('fillColor')`?

docs/layers/cpu-grid-layer.md
heshan0131: i am no difference, previously we did call out `getColorValue` say that it is shallow comparison...Maybe should just add to upgrade guide instead

Pessimistress: +1 to upgrade guide. The way our docs are organized it is hard to track which version a change is introduced in :(

docs/upgrade-guide.md
```diff
@@ -73,6 +73,9 @@ The `project64` shader module is no longer registered by default. If you were pr
     return {vs, fs, modules: [project64]};
   }
   ```
+#### CPU Grid layer and Hexagon layer updateTriggers
```
heshan0131: @1chandu

PR #3481, opened 2019-08-22
IconLayer: check texture width and height before drawing

xintongxia:
For #3449


modules/layers/src/icon-layer/icon-layer.js
```diff
@@ -180,7 +180,7 @@ export default class IconLayer extends Layer {
     const {viewport} = this.context;
 
     const iconsTexture = iconManager.getTexture();
-    if (iconsTexture) {
+    if (iconsTexture && iconsTexture.width && iconsTexture.height) {
```
Pessimistress: Use `texture.loaded`.

Issue #3480, opened 2019-08-22
HeatmapLayer throws GL_BLEND with floating-point color attachments requires the EXT_float_blend extension


christianbalderrama:
#### Description

Implementing the `HeatmapLayer` doesn't reflect the heatmap and throws warning in the console see logs below.

#### Repro Steps
```
    const layers = [
      new HeatmapLayer({
        id: "heatmaplayer",
        data: [{COORDINATES: [-122.42177834, 37.78346622], WEIGHT: 10}],
        getPosition: d => d.COORDINATES,
        getWeight: d => d.WEIGHT,
      }),
    ];
    return (
      <MapGL ...>
        <DeckGL
            viewState={this.state.viewState}
            initialViewState={this.state.viewState}
            controller={{type: MapController}}
            layers={layers}
            onContextMenu={event => event.preventDefault()}
        />
      </MapGL>
    );
```

#### Environment:
 - Framework Versions:
```
    "@deck.gl/aggregation-layers": "^7.2.2",
    "@deck.gl/core": "^7.2.2",
    "@deck.gl/layers": "^7.2.2",
    "@deck.gl/mapbox": "^7.2.2",
    "@deck.gl/react": "^7.2.2",
```
 - Browser Version: Chrome 75.0.3770.142
 - OS: Xubuntu 18.04 LTS x64

#### Logs
```
2TWFwOjE=:1 [.WebGL-0xc881cd8b300]GL ERROR :GL_INVALID_OPERATION : glDrawArrays: GL_BLEND with floating-point color attachments requires the EXT_float_blend extension
```

christianbalderrama:
![Screenshot_2019-08-22_09-12-44](https://user-images.githubusercontent.com/18398769/63478592-0435ab80-c4bd-11e9-8de0-7f1d67f197bb.png)

This console warning shows in the official documentation and does not reflect the heatmaps as well.

1chandu:
`HeatmapLayer` is `WebGL2` only for now (support for WebGL1 in future releases), but it should print an error message when run under `WebGL2`.

One possibility is you are using `WebGL2`, but `EXT_float_blend` is not supported, if so we have to add explicit check for it.  Can you set `luma.log.priority=1` in your chrome browser console and refresh the heatmap documentation page, and copy paste logs here? It should log what kind of WebGL context it is using.

[...]

PR #3483, opened 2019-08-23
HeatmapLayer/GPUAggregator: fix WebGL feature checking

1chandu:
For #3480 

Requires: https://github.com/uber/luma.gl/pull/1210

#### Background
Fix `feature` checking in Heatmap Layer and GPUGridAggregator.

#### Change List
- HeatmapLayer/GPUAggregator: fix WebGL feature checking

### DIFF
--- a/modules/aggregation-layers/src/heatmap-layer/heatmap-layer.js
+++ b/modules/aggregation-layers/src/heatmap-layer/heatmap-layer.js
@@ -27,7 +27,7 @@ import {
   scaleToAspectRatio,
   getTextureCoordinates
 } from './heatmap-layer-utils';
-import {Buffer, Transform, getParameter, isWebGL2} from '@luma.gl/core';
+import {Buffer, Transform, getParameter, FEATURES, hasFeatures} from '@luma.gl/core';
 import {CompositeLayer, AttributeManager, COORDINATE_SYSTEM, log} from '@deck.gl/core';
 import TriangleLayer from './triangle-layer';
 import {getFloatTexture} from '../utils/resource-utils';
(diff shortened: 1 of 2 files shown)


modules/aggregation-layers/src/heatmap-layer/heatmap-layer.js
```diff
@@ -55,13 +55,26 @@ const defaultProps = {
...
+
 export default class HeatmapLayer extends CompositeLayer {
   initializeState() {
     const {gl} = this.context;
     const textureSize = Math.min(SIZE_2K, getParameter(gl, gl.MAX_TEXTURE_SIZE));
     this.state = {textureSize, supported: true};
-    if (!isWebGL2(gl)) {
-      log.error(`HeatmapLayer ${this.id} is not supported on this browser, requires WebGL2`)();
+    if (!hasFeatures(gl, REQUIRED_FEATURES)) {
+      const info = REQUIRED_FEATURES.reduce(
```
Pessimistress: Call `GPUAggregator.isSupported`?

modules/aggregation-layers/src/heatmap-layer/heatmap-layer.js
```diff
@@ -55,13 +55,26 @@ const defaultProps = {
...
+        (acc, feature) =>
+          `${acc} ${feature}: ${hasFeatures(gl, feature) ? 'SUPPORTED' : 'NOT SUPPORTED'} `,
```
Pessimistress: - this is technically debug info. I suggest moving this into `hasFeature` and use `log.log(NON_ZERO_PRIORITY, ...)`
- `map` and then `join` will be more readable.

modules/aggregation-layers/src/heatmap-layer/heatmap-layer.js
1chandu: Though the checks are same, HeatmapLayer doesn't use GPUGridAggregator, in future it is possible these feature list will change.

1chandu:
luma.gl now logs, when feature is checked and not supported. Updated `HeatmapLayer` to just log single statement when not supported.

PR #3510, opened 2019-08-30
Re-introduce support for live updates and introduce 2-way API communication

ajduberstein:
For #2929 

#### Background
`Deck.update()` which refreshes the data in a Jupyter widget broke in a refactor–this PR corrects that.

### DIFF
--- a/modules/jupyter-widget/src/widget.js
+++ b/modules/jupyter-widget/src/widget.js
@@ -3,7 +3,7 @@ import {DOMWidgetModel, DOMWidgetView} from '@jupyter-widgets/base';
 
 import {MODULE_NAME, MODULE_VERSION} from './version';
 
-import {loadCss, hideMapboxCSSWarning} from './utils';
+import {loadCss, hideMapboxCSSWarning, initDeck, updateDeck} from './utils';
 
 const MAPBOX_CSS_URL = 'https://api.tiles.mapbox.com/mapbox-gl-js/v1.2.1/mapbox-gl.css';
 
@@ -72,7 +72,9 @@ export class DeckGLView extends DOMWidgetView {
           container: containerDiv,
           jsonInput: JSON.parse(this.model.get('json_input'))
         },
-        this,
+        x => {
+          this.jsonDeck = x;
+        },
         this.handleClick.bind(this)
       );
       this.model.set('initialized', true);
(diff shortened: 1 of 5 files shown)


ajduberstein:
Support for selecting data is now available. Below is a quick demo of clicking on an item in a map and pushing it from JS to Python.

2 way communication

modules/jupyter-widget/src/widget.js
```diff
@@ -53,35 +55,49 @@ export class DeckGLModel extends DOMWidgetModel {
...
+  }
 }
 
 function updateDeck(inputJSON, {jsonConverter, deckConfig}) {
   const results = jsonConverter.convertJsonToDeckProps(inputJSON);
   deckConfig.setProps(results);
 }
 
-export function initDeck({mapboxApiKey, container, jsonInput}) {
+export function initDeck({mapboxApiKey, container, jsonInput}, context, handleClick) {
```
ibgreen: Nit: Should we put this function in a separate file?

modules/jupyter-widget/src/widget.js
```diff
@@ -103,9 +119,11 @@ export function initDeck({mapboxApiKey, container, jsonInput}) {
           latitude: 0,
           longitude: 0,
           zoom: 1,
+          onClick: handleClick,
           container,
           onLoad: () => updateDeck(jsonInput, {jsonConverter, deckConfig})
         });
+        context.jsonDeck = {jsonConverter, deckConfig};
```
ibgreen: This function modifying `context`, instead of returning these values and having the caller add them to `context` raises flags. 

Modification of arguments is a subtle source of bugs and should be avoided whenever possible unless justified by a strong reason or a specific design choice. I see evidence of neither here.

modules/jupyter-widget/src/widget.js
```diff
@@ -53,35 +55,49 @@ export class DeckGLModel extends DOMWidgetModel {
...
+        },
+        this,
+        this.handleClick.bind(this)
+      );
+      this.model.set('initialized', true);
+    }
   }
 
   valueChanged() {
-    updateDeck(this.model.get('json_input'), this.deck);
```
ibgreen: This invocation looks confusing at it passes in all of deck and not at all clear what is being used. 

I'd expand the arguments and pass them in one by one, or make `updateDeck` a method on `this.deck`.

modules/jupyter-widget/src/widget.js
```diff
@@ -53,35 +55,49 @@ export class DeckGLModel extends DOMWidgetModel {
...
   }
+
+  handleClick(e) {
+    console.table(e); // eslint-disable-line
+    this.model.set('selected_data', e.object.points);
+    this.model.save_changes();
+  }
 }
 
 function updateDeck(inputJSON, {jsonConverter, deckConfig}) {
```
ibgreen: For readability's sake, instead of destructuring the two latter params, jus make them top level args and call explicitly?

PR #3527, opened 2019-09-05
Add update trigger tests to generated tests

Pessimistress:
#### Change List
- Make updateTrigger tests with `generateLayerTests`
- Fix an issue in `testLayers` where non-incremental prop update is not respected
- Remove unsupported `getLineDashArray` from `H3HexagonLayer`'s default props

### DIFF
--- a/modules/test-utils/src/generate-layer-tests.js
+++ b/modules/test-utils/src/generate-layer-tests.js
@@ -20,6 +20,8 @@
 import {experimental} from '@deck.gl/core';
 const {count} = experimental;
 
+function noop() {}
+
 function defaultAssert(condition, comment) {
   if (!condition) {
     throw new Error(comment);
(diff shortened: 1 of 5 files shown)


modules/test-utils/src/generate-layer-tests.js
```diff
@@ -135,24 +138,51 @@ function makeAltPropTestCase(propName, propTypes, defaultProps) {
...
+          updateProps: {
+            updateTriggers: {
+              [propName]: 'function+trigger'
+            }
+          },
+          onBeforeUpdate,
+          onAfterUpdate
+        }
+      ];
+    }
```
1chandu: Nit: Do we care about adding test case, where `updateTrigger` doesn't change and `callCount` should be 0.

Pessimistress: I don't think that needs to be tested per layer. Theoretically that should be covered by the unit tests of `diffProps` and `AttributeManager`.

PR #3530, opened 2019-09-05
attribute spring transition

rolyatmax:
This PR is the first pass at implementing spring transitions for attributes. In the process, I've refactored the `GPUInterpolationTransition` class to simplify it and bring it in line with the new `GPUSpringTransition` class (please see [this comment](https://github.com/uber/deck.gl/pull/3530#issuecomment-530472320) regarding the resulting duplicated code). I've also made some changes to the attribute transition test app to incorporate the new spring transition types.

### Work _not_ implemented in this PR

This PR does not implement spring transitions for uniforms. Also, because the spring transition takes an unscaled step every time the transform is run, it is not currently affected by the application's `Timeline` (we plan to address this in a follow-up). Finally, the `onStart` and `onEnd` events are not yet implemented for spring transitions as they rely on implementing a reliable `isTransitioning` check.

### Follow-up work to do (https://github.com/uber/deck.gl/issues/3574)

 - support switching between transition types (perhaps lift the transitioning buffer management to `AttributeTransitionManager`?)
 - implement spring transition type for uniforms
 - scale the spring's step increments with Timeline
 - implement `isTransitioning` check (and `onStart` / `onEnd` events) for spring transitions (render an "isTransitioning flag" per vertex to a 1x1 framebuffer)

### Screenshot

attribute-spring-transition

### DIFF
--- a/modules/core/src/lib/attribute-transition-utils.js
+++ b/modules/core/src/lib/attribute-transition-utils.js
@@ -60,11 +60,7 @@ export function getAttributeTypeFromSize(size) {
 }
 
 export function cycleBuffers(buffers) {
-  const firstBuffer = buffers[0];
-  for (let i = 0; i < buffers.length - 1; i++) {
-    buffers[i] = buffers[i + 1];
-  }
-  buffers[buffers.length - 1] = firstBuffer;
+  buffers.push(buffers.shift());
   return buffers;
 }
 
--- a/modules/core/src/lib/attribute-transition-utils.js
+++ b/modules/core/src/lib/attribute-transition-utils.js
@@ -1,4 +1,4 @@
-import {Buffer} from '@luma.gl/core';
+import BaseAttribute from '../lib/base-attribute';
 import {padArray} from '../utils/array-utils';
 
 const noop = () => {};
@@ -9,6 +9,7 @@ const DEFAULT_TRANSITION_SETTINGS = {
   onStart: noop,
   onEnd: noop,
   onInterrupt: noop,
+  onUpdate: noop,
   enter: x => x
 };
 
@@ -25,6 +26,54 @@ export function normalizeTransitionSettings(userSettings, layerSettings = {}) {
   return Object.assign({}, DEFAULT_TRANSITION_SETTINGS, layerSettings, userSettings);
 }
 
+// NOTE: NOT COPYING OVER OFFSET OR STRIDE HERE BECAUSE:
+// (1) WE DON'T SUPPORT INTERLEAVED BUFFERS FOR TRANSITIONS
+// (2) BUFFERS WITH OFFSETS ALWAYS CONTAIN VALUES OF THE SAME SIZE
+// (3) THE OPERATIONS IN THE SHADER ARE PER-COMPONENT (addition and scaling)
+export function getSourceBufferAttribute(gl, attribute) {
+  const {size, value, normalized, constant} = attribute;
+  // The Attribute we pass to Transform as a sourceBuffer must have {divisor: 0}
+  // so we create a copy of the attribute (with divisor=0) to use when running
+  // transform feedback
+  if (constant) {
+    // don't pass normalized here because the `value` from a normalized attribute is
+    // already normalized
+    return new BaseAttribute(gl, {constant, value, size});
+  }
+  const buffer = attribute.getBuffer();
+  return new BaseAttribute(gl, {divisor: 0, constant, buffer, size, normalized});
+}
+
+export function getAttributeTypeFromSize(size) {
+  switch (size) {
+    case 1:
+      return 'float';
+    case 2:
+      return 'vec2';
+    case 3:
+      return 'vec3';
+    case 4:
+      return 'vec4';
+    default:
+      throw new Error(`No defined attribute type for size "${size}"`);
+  }
+}
+
+export function cycleBuffers(buffers) {
+  const firstBuffer = buffers[0];
+  for (let i = 0; i < buffers.length - 1; i++) {
+    buffers[i] = buffers[i + 1];
+  }
+  buffers[buffers.length - 1] = firstBuffer;
+  return buffers;
+}
+
+export function getAttributeBufferLength(attribute, numInstances) {
+  const {doublePrecision, userData, value, size} = attribute;
+  const multiplier = doublePrecision ? 2 : 1;
+  return (userData.noAlloc ? value.length : numInstances * size) * multiplier;
+}
+
 // This helper is used when transitioning attributes from a set of values in one buffer layout
 // to a set of values in a different buffer layout. (Buffer layouts are used when attribute values
 // within a buffer should be grouped for drawElements, like the Polygon layer.) For example, a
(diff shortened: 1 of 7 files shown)


modules/core/src/transitions/gpu-spring-transition.js
```diff
@@ -0,0 +1,240 @@
...
+export default class GPUSpringTransition {
+  constructor({gl, attribute, transitionSettings}) {
+    this.type = 'spring';
+    this.attribute = attribute;
+    this.transitionSettings = transitionSettings;
+    // this is the attribute we return during the transition
+    // `attribute.userData` is the original options passed when constructing the attribute.
+    // This ensures that we set the proper `doublePrecision` flag and shader attributes.
+    this.attributeInTransition = new Attribute(gl, attribute.userData);
+    this.attributeInTransition.update({
```
Pessimistress: `constant` will be set to `false` when you update this attribute with a buffer.

modules/core/src/transitions/gpu-spring-transition.js
```diff
@@ -0,0 +1,240 @@
...
+
+    const data = new Float32Array(toLength);
+    const usage = GL.DYNAMIC_COPY;
+    buffers.next.initialize({data, usage});
+
+    return buffers;
+  }
+}
+
+const ATTRIBUTE_MAPPING = {
```
Pessimistress: Move this to a `getAttributeType` helper in `attribute-transition-utils`?

modules/core/src/transitions/gpu-spring-transition.js
```diff
@@ -0,0 +1,240 @@
...
+
+    // TODO: fire an onEnd() event here if the transition has just ended
+
+    return true;
+  }
+
+  cancel() {
+    this.transitionSettings.onInterrupt();
+    this.transform.delete();
+    if (this.buffers.previous) this.buffers.previous.delete();
```
Pessimistress: Nit: just use a for loop to delete & reset to null

modules/core/src/transitions/gpu-spring-transition.js
```diff
@@ -0,0 +1,240 @@
...
+    },
+    varyings: ['vNext']
+  };
+}
+
+// NOTE: NOT COPYING OVER OFFSET OR STRIDE HERE BECAUSE:
+// (1) WE DON'T SUPPORT INTERLEAVED BUFFERS FOR TRANSITIONS
+// (2) BUFFERS WITH OFFSETS ALWAYS CONTAIN VALUES OF THE SAME SIZE
+// (3) THE OPERATIONS IN THE SHADER ARE PER-COMPONENT (addition and scaling)
+function getSourceBufferAttribute(gl, attribute) {
```
Pessimistress: Move to utils?

modules/core/src/transitions/gpu-spring-transition.js
rolyatmax: I mostly have this line here to show the reader the intention of this code. Additionally, it seems bad to rely on what feels like an implementation detail of `update` to make sure `constant` is set to `false`. Thoughts?

Pessimistress: It is by design that `constant` cannot be `true` if a buffer is present. You should never use this attribute instance without updating it anyways.

modules/core/src/transitions/gpu-spring-transition.js
rolyatmax: Would you prefer `buffers` be an array rather than an object with `previous`, `current`, and `next` properties?

modules/core/src/transitions/gpu-spring-transition.js
rolyatmax: I can remove this code, but I want to add a comment. It is not obvious at all to a user of the `AttributeTransitionManager` that if you pass a constant attribute, it will be converted to a regular attribute. I want to make sure it is clear to the reader that this is the intended contract of our API.

rolyatmax:
I believe this is now ready for review. I've put a lot of work into bringing the two transition types in line with one another, and as a result there are some points of duplication. I want to take care of this duplication in a follow-up refactor. I'm a bit hesitant to try to factor out an "abstract GPU transition class" at the moment given we've only got two transition types at the moment. Getting the wrong abstraction now will make adding a third type very painful. For that reason, I'd rather live with some of the duplication until we need a third type.

That said, I believe we might be storing the buffers in the wrong place. It seems like they should persist when switching to a new transition type (since the transform buffers are the only place we store an attribute's last position - once the attributeTransitionManager's `update()` is called, the attribute instance has already had its values updated). I believe lifting this state to the attributeTransitionManager would both allow us to persist the transitioning attribute's state between transition types, and would likely result in the deletion of the duplicated logic (since much of that would be lifted, too). I think it's best to save this all for a follow-up refactor, however.

Here's a list of follow-up work to do (I'll create some Issues for these):
 - support switching between transition types (perhaps lift the transitioning buffer management to `AttributeTransitionManager`?)
 - implement spring transition type for uniforms
 - scale the spring's step increments with Timeline
 - implement `isTransitioning` check (and `onStart` / `onEnd` events) for spring transitions (render an "isTransitioning flag" per vertex to a 1x1 framebuffer)

modules/core/src/transitions/gpu-spring-transition.js
rolyatmax: I went ahead and made the buffers for both transition types an array.

modules/core/src/lib/attribute-transition-utils.js
```diff
@@ -25,6 +26,54 @@ export function normalizeTransitionSettings(userSettings, layerSettings = {}) {
...
+    case 1:
+      return 'float';
+    case 2:
+      return 'vec2';
+    case 3:
+      return 'vec3';
+    case 4:
+      return 'vec4';
+    default:
+      throw new Error(`No defined attribute type for size "${size}"`);
```
Pessimistress: Nit: we generally avoid detailed error messages unless its source is the user.

modules/core/src/lib/attribute-transition-utils.js
```diff
@@ -25,6 +26,54 @@ export function normalizeTransitionSettings(userSettings, layerSettings = {}) {
...
+  }
+}
+
+export function cycleBuffers(buffers) {
+  const firstBuffer = buffers[0];
+  for (let i = 0; i < buffers.length - 1; i++) {
+    buffers[i] = buffers[i + 1];
+  }
+  buffers[buffers.length - 1] = firstBuffer;
+  return buffers;
```
Pessimistress: May be easier to just use `shift()` and `push()`?

modules/core/src/lib/attribute-transition-utils.js
```diff
@@ -33,39 +82,41 @@ export function normalizeTransitionSettings(userSettings, layerSettings = {}) {
...
   fromLength,
-  toLength,
   fromBufferLayout,
-  toBufferLayout,
-  size,
-  offset,
   getData = x => x
 }) {
+  // TODO: move the precisionMultiplier logic to the attribute when retrieving
+  // its `size` and `elementOffset`?
```
Pessimistress: `attribute.size` is used with `vertexAttribPointer` while the `size` here is used for manipulating buffer content on the CPU.

rolyatmax: I see - maybe a `getBufferContentLength` or something, since this calculation seems to be done a few times?

I'm not suggesting we do it as part of this work - just an idea.

modules/core/src/lib/attribute-transition-utils.js
rolyatmax: Hmm - what would you do instead of a detailed error message?

modules/core/src/lib/attribute-transition-utils.js
rolyatmax: Ha! Good call. Much simpler :-)

modules/core/src/lib/attribute-transition-utils.js
Pessimistress: `assert(false);`

Generally errors generated from an internal path is considered debug info and we avoid producing readable errors to reduce bundle size.

rolyatmax: I really think we should give as much detail as is necessary in all the failed `asserts()` and errors we throw. Error messages are just as much for the engineers who are contributing to Deck as they are for those who `npm install` Deck into their projects. These engineers include lots of newcomers to the project, many from other teams and other companies. I've run into several of those cryptic `assert() failed` error messages during development, and definitely burned a lot of time on some of them due to the lack of detail. 

If we think removing all these strings will result in a significant drop in bundle size, can we strip them out for the production bundle using a webpack plugin, like React does?

rolyatmax:
Issues with Travis CI right now. Tests pass locally, though, so gonna go ahead and land this one.

PR #3531, opened 2019-09-06
CSS to Device conversion: Replace custom code with luma.gl utilities methods

1chandu:
Requires https://github.com/uber/luma.gl/pull/1155

- [ ] Replace CSS to Device pixel conversion code with new luma.gl utility methods.
- [ ] Remove deprecated `screenToDevicePixels` method.


modules/core/src/lib/deck-picker.js
```diff
@@ -230,16 +224,14 @@ export default class DeckPicker {
...
+    const [deviceLeft, deviceBottom] = cssToDevicePixels(this.gl, [x, y], true);
+    const [deviceRight, deviceTop] = cssToDevicePixels(this.gl, [x + width, y + height], true);
 
     const deviceRect = {
       x: deviceLeft,
       y: deviceTop,
-      width: deviceRight - deviceLeft,
-      height: deviceBottom - deviceTop
+      width: deviceRight - deviceLeft + 1,
+      height: deviceBottom - deviceTop + 1
```
Pessimistress: Why `+1` ? Make a comment?

test/modules/core/lib/pick-layers.spec.js
```diff
@@ -157,7 +157,7 @@ const TEST_CASES = [
             height: 100
           },
           results: {
-            count: 33
+            count: 34
```
Pessimistress: Has a breaking change been introduced?

modules/core/src/lib/deck-picker.js
1chandu: I made this change to fix a unit test, but looking deep into it, I can just adjust the test, the test allocates a 1 X 1, window and tests rect starting at (0, 0) and ending at (1, 1), which effectively needs 2 X 2 window.  The new utility methods I added in luma.gl , clamp down the final device pixels to actual bounds causing the test failure.
Updating the test to use 2 X 2 window, doesn't need this change and the test change below, but has another test failure in (Node only, browser is good, checking ...)

1chandu: Updated the PR to use newer luma utility method that returns range of valid pixels.
Old code was not correctly calculating `width` and `height`, we should add `1`, updated the comment, this change also needs update to picking test result.

test/modules/core/lib/pick-layers.spec.js
1chandu: As commented above, this change required to updates to deviceRect `width` and `height` calculations.

PR #3532, opened 2019-09-06
Fix double precision attribute transitions

Pessimistress:
#### Change List
- Handle 64bit attribute
- Fix issue in transitioning normalized buffers

### DIFF
--- a/modules/core/src/lib/attribute-transition-manager.js
+++ b/modules/core/src/lib/attribute-transition-manager.js
@@ -120,7 +120,7 @@ export default class AttributeTransitionManager {
         name: attributeName,
         timeline: this.timeline,
         attribute,
-        attributeInTransition: new Attribute(this.gl, attribute),
+        attributeInTransition: new Attribute(this.gl, attribute.userData),
         bufferLayout: attribute.bufferLayout
       });
       this.transitions[attributeName] = transition;
@@ -173,27 +173,25 @@ export default class AttributeTransitionManager {
   // get current values of an attribute, clipped/padded to the size of the new buffer
   _getNextTransitionStates(transition, settings) {
     const {attribute} = transition;
-    const {size, offset, normalized} = attribute;
+    const {size, normalized} = attribute;
+    const multiplier = attribute.doublePrecision ? 2 : 1;
 
     let toState;
     if (attribute.constant) {
-      toState = new BaseAttribute(this.gl, {constant: true, value: attribute.value, size, offset});
+      toState = new BaseAttribute(this.gl, {constant: true, value: attribute.value, size});
     } else {
       toState = new BaseAttribute(this.gl, {
         constant: false,
         buffer: attribute.getBuffer(),
         divisor: 0,
         size,
-        offset,
-        normalized,
-        // attribute's `value` does not match the content of external buffer,
-        // will need to call buffer.getData if needed
-        value: attribute.externalBuffer || attribute.doublePrecision ? null : attribute.value
+        normalized
       });
     }
     const fromState = transition.buffer || toState;
-    const toLength = attribute.userData.noAlloc ? attribute.value.length : this.numInstances * size;
-    const fromLength = (fromState instanceof Buffer && fromState.getElementCount()) || toLength;
+    const toLength =
+      (attribute.userData.noAlloc ? attribute.value.length : this.numInstances * size) * multiplier;
+    const fromLength = transition.length || toLength;
 
     // Alternate between two buffers when new transitions start.
     // Last destination buffer is used as an attribute (from state),
@@ -213,16 +211,23 @@ export default class AttributeTransitionManager {
       });
     }
 
-    transition.attributeInTransition.update({buffer});
+    transition.length = toLength;
+    transition.attributeInTransition.update({
+      buffer,
+      // Hack: Float64Array is required for double-precision attributes
+      // to generate correct shader attributes
+      value: attribute.value
+    });
 
     padBuffer({
       fromState,
       toState,
       fromLength,
       toLength,
+      size: size * multiplier,
       fromBufferLayout: transition.bufferLayout,
       toBufferLayout: attribute.bufferLayout,
-      offset: attribute.elementOffset,
+      offset: attribute.elementOffset * multiplier,
       getData: settings.enter
     });
 
@@ -244,16 +249,17 @@ export default class AttributeTransitionManager {
       Object.assign({}, this._getNextTransitionStates(transition, settings), settings)
     );
     let transform = this.transforms[transition.name];
+    const elementCount = Math.floor(transition.length / transition.attribute.size);
 
     if (transform) {
       transform.update({
         ...getBuffers(transition),
-        elementCount: this.numInstances
+        elementCount
       });
     } else {
       // Buffers must be supplied to the transform constructor
       transform = new Transform(this.gl, {
-        elementCount: this.numInstances,
+        elementCount,
         ...getShaders(transition),
         ...getBuffers(transition)
       });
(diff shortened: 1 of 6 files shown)


test/apps/attribute-transition/app.js
```diff
@@ -2,13 +2,12 @@
 /* eslint-disable no-console */
 import React, {Component} from 'react';
 import {render} from 'react-dom';
-import DeckGL, {COORDINATE_SYSTEM, ScatterplotLayer, PolygonLayer, MapController} from 'deck.gl';
+import DeckGL, {COORDINATE_SYSTEM, OrthographicView, ScatterplotLayer, PolygonLayer} from 'deck.gl';
 
 import DataGenerator from './data-generator';
 
 // Set your mapbox token here
 const MAPBOX_TOKEN = process.env.MapboxAccessToken; // eslint-disable-line
```
rolyatmax: Can drop this now, right?

modules/core/src/lib/attribute-transition-manager.js
```diff
@@ -173,27 +173,25 @@ export default class AttributeTransitionManager {
...
     if (attribute.constant) {
-      toState = new BaseAttribute(this.gl, {constant: true, value: attribute.value, size, offset});
+      toState = new BaseAttribute(this.gl, {constant: true, value: attribute.value, size});
     } else {
       toState = new BaseAttribute(this.gl, {
         constant: false,
         buffer: attribute.getBuffer(),
         divisor: 0,
         size,
-        offset,
```
rolyatmax: Why are we getting rid of `offset`? If the attribute we're transitioning has an `offset`, won't the transform feedback pass run with the wrong values (the values before the offset)?

modules/core/src/lib/attribute-transition-manager.js
```diff
@@ -173,27 +173,25 @@ export default class AttributeTransitionManager {
...
   _getNextTransitionStates(transition, settings) {
     const {attribute} = transition;
-    const {size, offset, normalized} = attribute;
+    const {size, normalized} = attribute;
+    const multiplier = attribute.doublePrecision ? 2 : 1;
 
     let toState;
     if (attribute.constant) {
-      toState = new BaseAttribute(this.gl, {constant: true, value: attribute.value, size, offset});
+      toState = new BaseAttribute(this.gl, {constant: true, value: attribute.value, size});
```
rolyatmax: Can constant attributes be normalized? Should that be passed here in the `opts`?

modules/core/src/lib/attribute-transition-manager.js
```diff
@@ -213,16 +211,23 @@ export default class AttributeTransitionManager {
       });
     }
 
-    transition.attributeInTransition.update({buffer});
+    transition.length = toLength;
```
rolyatmax: Do we have to keep track of the length like this? Isn't this the same as `transition.buffer`'s length?

modules/core/src/lib/base-attribute.js
```diff
@@ -48,7 +48,7 @@ export default class BaseAttribute {
     if (buffer) {
       this.externalBuffer = buffer;
       this.constant = false;
-      this.value = null;
+      this.value = value || null;
```
rolyatmax: Is the purpose of this change to make the hack you mentioned in this comment work?
```
transition.attributeInTransition.update({
  buffer,
  // Hack: Float64Array is required for double-precision attributes
  // to generate correct shader attributes
  value: attribute.value
});
```

If so, can we add a comment here, too, since it will seem strange (or like a bug) to other readers (and our future selves when we forget how all this works).

modules/core/src/lib/attribute-transition-manager.js
```diff
@@ -120,7 +120,7 @@ export default class AttributeTransitionManager {
         name: attributeName,
         timeline: this.timeline,
         attribute,
-        attributeInTransition: new Attribute(this.gl, attribute),
+        attributeInTransition: new Attribute(this.gl, attribute.userData),
```
rolyatmax: Why are we making this change?

modules/core/src/lib/attribute-transition-manager.js
Pessimistress: All we need is to run the transform with the same offset in from, to and feedback buffers. Always using 0 makes things easier.

modules/core/src/lib/attribute-transition-manager.js
Pessimistress: `attribute.value` is already normalized in https://github.com/uber/deck.gl/blob/master/modules/core/src/lib/base-attribute.js#L86

modules/core/src/lib/attribute-transition-manager.js
Pessimistress: We only ever expand the buffers when the sizes are too small. So they can be over-allocated.

modules/core/src/lib/attribute-transition-manager.js
Pessimistress: `attribute.userData` is the original options passed when constructing `attribute`. This ensures that we set the proper `doublePrecision` flag and shader attributes.
Added a comment.

modules/core/src/lib/attribute-transition-manager.js
rolyatmax: Hmmm - I suppose this is only true because we happen to just be using `mix` in this particular shader. If we were to use `length()` (as we would have to in order to support a `maxSpeed` setting for springs), for example, this wouldn't work, right?

My worry is that it reads like an oversight and leads the reader to believe `offset` and `stride` don't matter for transitions, when they just happen to not matter because of the specific calculations we're doing in this transition. If we really want to leave them out, maybe we leave a comment here about why we're intentionally doing that?

modules/core/src/lib/attribute-transition-manager.js
rolyatmax: Creating an attribute with `attribute.userData` as the `opts` doesn't appear to transfer the attribute's `buffer` or the `value`. Is this intentional?

modules/core/src/lib/attribute-transition-manager.js
rolyatmax: Why do we only update the buffer size when it's too small? Wouldn't it be simpler (and match reader/developer expectations) to correctly size the buffer whether it has grown or shrunk in length?

modules/core/src/lib/attribute-transition-manager.js
Pessimistress: Yes, `attributeInTransition` is supposed to use the feedback buffer.

modules/core/src/lib/attribute-transition-manager.js
Pessimistress: Allocating memory for a buffer has a performance cost. We do the same thing in the Attributes class (over-allocate then use `bufferSubData`). Generally speaking we should never expect the buffers in deck.gl sized precisely to the draw list.

modules/core/src/lib/attribute-transition-manager.js
rolyatmax: That comment you added is super helpful! Thank you!

modules/core/src/lib/attribute-transition-manager.js
Pessimistress: For attributes that are packed by the `AttributeManager`, we do not have a situation where these two matter.

For externally provided interleaved buffers, because we did not pass `stride` before, transition never worked. In fact, I do not believe interleaved buffers work at all with any attribute that contains shader attributes (i.e. positions and transform matrices), given that we are overriding `offset` and `stride` internally. @tsherif

PR #3535, opened 2019-09-06
Fix GPUGridLayer unhandled cellsize

### LINKED ISSUE #3534: GPUGridLayer crashes when cell size is 0
#### Description

GPUGridLayer crashes when cell size is 0.
#### Repro Steps

gpubugfix

 - Framework Version: [e.g. deck.gl 7.2.3]
 - Browser Version: [e.g. Chrome 76.0]
 - OS: [e.g. Ubuntu 18.04.3 LTS]

#### Logs

Screenshot from 2019-09-06 18-28-22

Navneet-Sahota:
Fix #3534

#### Background
GPU Grid Layer currently doesn't handles the case when cell size is zero. Hence, some variables get `Infinity` or `NaN` values.

### DIFF
--- a/docs/layers/gpu-grid-layer.md
+++ b/docs/layers/gpu-grid-layer.md
@@ -94,7 +94,7 @@ Inherits from all [Base Layer](/docs/api-reference/layer.md) and [CompositeLayer
 
 * Default: `1000`
 
-Size of each cell in meters
+Size of each cell in meters. Must be greater than `0`.
 
 ##### `colorRange` (Array, optional)
 
--- a/modules/aggregation-layers/src/gpu-grid-layer/gpu-grid-layer.js
+++ b/modules/aggregation-layers/src/gpu-grid-layer/gpu-grid-layer.js
@@ -44,7 +44,7 @@ const defaultProps = {
   elevationScale: {type: 'number', min: 0, value: 1},
 
   // grid
-  cellSize: {type: 'number', min: 0, max: 1000, value: 1000},
+  cellSize: {type: 'number', min: 1, max: 1000, value: 1000},
   coverage: {type: 'number', min: 0, max: 1, value: 1},
   getPosition: {type: 'accessor', value: x => x.position},
   extruded: false,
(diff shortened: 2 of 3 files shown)


Navneet-Sahota:
@1chandu Kindly review this PR

1chandu:
This would fix the crash, but we would still be performing aggregation but not saving the results any where (grid cell count becomes 0), for GPU case we will be setting up all the resources and running the model with 0 vertex count.

We should not even call into aggregator, or return even early instead of handling this case inside the aggregator methods.  

@Pessimistress , how about updating [propTypes system](https://github.com/uber/deck.gl/blob/4ff387e3d455b7f062ac8c464b889729b11e08d4/modules/core/src/lifecycle/prop-types.js#L15) to handle  `min<value<=max`, in addition to handling `min<=value<=max`

1chandu:
@Navneet-Sahota , you can also set the [default](https://github.com/uber/deck.gl/blob/4ff387e3d455b7f062ac8c464b889729b11e08d4/modules/aggregation-layers/src/gpu-grid-layer/gpu-grid-layer.js#L47) cellSize min value to 0.01, instead of 0, given it is in meters, we should be fine.

Navneet-Sahota:
@1chandu Currently, this layer crashes for a **value** less than **3**. If I set the **min value** to **0.01** then the slider behavior changes as well.
The **initial value** of the slider is **200** but when moved it will give values like 190.01, 200.01, 210.01, 220.01 and so on. This is because the **step count** which is **10**, in this case, starts from the minimum value.
Screenshot from 2019-09-07 01-12-38
Should we accept this behavior and keep **min value** to **0.01** and try to find a way to stop crashes due to miscalculation?

1chandu:
> Currently, this layer crashes for a value less than 3

Where does it crash, is this same crash when cell size is 0, if so that needs to be debugged, my understanding is it only crashes when cell size is 0.

> The initial value of the slider is 200 but when moved it will give values like 190.01, 200.01, 210.01, 220.01 and so on. This is because the step count which is 10, in this case, starts from the minimum value.

Another option is use minValue to be 1,  this will make we don't support grid cell size less than 1 meter.

PR #3540, opened 2019-09-06
refactor interpolation transition

rolyatmax:
Pulled out the refactoring work I was doing as part of the Spring Transition implementation.

This removes the interpolation-related logic from `attribute-transition-manager` and `attribute-transition-utils` and moves it into its own `gpu-interpolation-transition` file.

This also tries to simplify things a bit and removes unnecessary copies of state from some of the classes.


modules/core/src/transitions/gpu-interpolation-transition.js
```diff
@@ -0,0 +1,225 @@
...
+      aFrom: fromState,
+      aTo: toState
+    },
+    feedbackBuffers: {
+      vCurrent: buffer
+    }
+  };
+}
+
+function padBuffer({
```
Pessimistress: Why is this function moved? Shouldn't all transitions need this?

modules/core/src/transitions/gpu-interpolation-transition.js
```diff
@@ -0,0 +1,225 @@
...
+  cancel() {
+    this.transition.cancel();
+    this.transform.delete();
+    if (this.buffer) this.buffer.delete();
+    if (this._swapBuffer) this._swapBuffer.delete();
+  }
+
+  // get current values of an attribute, clipped/padded to the size of the new buffer
+  _setupBuffers(gl, settings, numInstances) {
+    const {size, normalized} = this.attribute;
```
Pessimistress: I would expect the `fromState` and `toState` preparation the same for all transition types...

modules/core/src/transitions/gpu-interpolation-transition.js
rolyatmax: I wasn't sure if it was going to be the same for all transition types. But I think you're right - we can use it in its current form.

modules/core/src/transitions/gpu-interpolation-transition.js
rolyatmax: For physics-based transitions this function will be a little different because there are essentially two "from" buffers to update - the "current value" and the "previous value". But yes, there will be logic we can share between the transition types.

modules/core/src/transitions/gpu-interpolation-transition.js
rolyatmax: Moved this back to `attribute-transition-utils` with a comment explaining its use.

modules/core/src/transitions/gpu-interpolation-transition.js
Pessimistress: Can we split off the common logic into a base class e.g. `GPUTransition`?

rolyatmax: Yeah - I think it might make sense to do that in the spring-transition work.

PR #3542, opened 2019-09-06
Fix ColumnLayer elevationScale example

### LINKED ISSUE #3541: ColumnLayer elevationScale default max limit
#### Description

ColumnLayer's `elevationScale` in the documentation is set to `5000` initially, but the input range have a default max value of `100`, so it does not move beyond 100 once changed from the initial state.
#### Repro Steps

err

- Framework Version: [e.g. deck.gl 7.2.3]
 - Browser Version: [e.g. Chrome 76.0]
 - OS: [e.g. Ubuntu 18.04.3 LTS]

#### Logs

No warnings or error is thrown

Navneet-Sahota:
Fix #3541

#### Background
ColumnLayer's `elevationScale` in the documentation is set to `5000` initially, but the input range have a default max value of `100`, so it does not move beyond 100 once changed from the initial state.
Here, I pass a `max` prop to input field with value equal to `5000` .

sol

### DIFF
--- a/website/src/utils/layer-params.js
+++ b/website/src/utils/layer-params.js
@@ -57,9 +57,6 @@ export function getLayerParams(layer, propParameters = {}) {
       paramsArray.push({name: key, ...propParameters[key]});
     } else {
       const LAYER_PROPTYPES = layer.constructor._propTypes[key];
-      if (layer.constructor.name === 'ColumnLayer') {
-        LAYER_PROPTYPES.max = 5000;
-      }
       const param = propToParam(key, LAYER_PROPTYPES, layer.props[key]);
       if (param) {
         paramsArray.push(param);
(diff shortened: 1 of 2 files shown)


website/src/utils/layer-params.js
```diff
@@ -56,7 +56,11 @@ export function getLayerParams(layer, propParameters = {}) {
     if (propParameters[key]) {
       paramsArray.push({name: key, ...propParameters[key]});
     } else {
-      const param = propToParam(key, layer.constructor._propTypes[key], layer.props[key]);
+      const LAYER_PROPTYPES = layer.constructor._propTypes[key];
+      if (layer.constructor.name === 'ColumnLayer') {
```
Pessimistress: I'm not a fan of special cases like this. Can we scale up the `getElevation` prop instead and keep the default range?

Navneet-Sahota: Yes, that seems like a more obvious option since we have access to all the props of `ColumnLayer`. Will update the code.

PR #3562, opened 2019-09-10
Add default tooltip to pydeck

ajduberstein:
For #2929

#### Change List
- Supports a tooltip in Jupyter notebook and standalone rendering pydeck
environments

### DIFF
--- a/modules/jupyter-widget/src/utils.js
+++ b/modules/jupyter-widget/src/utils.js
@@ -22,6 +22,23 @@ export function updateDeck(inputJSON, {jsonConverter, deckConfig}) {
   deckConfig.setProps(results);
 }
 
+function getTooltip(pickedInfo) {
+  if (!pickedInfo.picked) {
+    return null;
+  }
+  return {
+    html: tabularize(pickedInfo.object),
+    style: {
+      fontFamily: '"Helvetica Neue", Helvetica, Arial, sans-serif',
+      display: 'flex',
+      flex: 'wrap',
+      maxWidth: '500px',
+      flexDirection: 'column',
+      zIndex: 2
+    }
+  };
+}
+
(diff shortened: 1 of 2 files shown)


ajduberstein:
tooltip

CSS suggestions are welcome and appreciated

modules/jupyter-widget/src/utils.js
```diff
@@ -60,3 +78,30 @@ export function initDeck({mapboxApiKey, container, jsonInput}, onComplete, handl
...
+  for (const key in json) {
+    const row = document.createElement('div');
+    const header = document.createElement('div');
+    header.className = 'header';
+    header.innerHTML = key;
+    header.style.fontWeight = 500;
+    header.style.marginRight = '10px';
+    const value = document.createElement('div');
+    value.className = 'value';
+    value.innerHTML = JSON.stringify(json[key]);
```
Pessimistress: Use `innerText` instead of `innerHTML` - user data may contain contaminated content.

modules/jupyter-widget/src/utils.js
```diff
@@ -22,6 +22,23 @@ export function updateDeck(inputJSON, {jsonConverter, deckConfig}) {
   deckConfig.setProps(results);
 }
 
+function getTooltip(pickedInfo) {
```
Pessimistress: Nit: define private functions together in this file

modules/jupyter-widget/src/utils.js
```diff
@@ -60,3 +61,54 @@ export function initDeck({mapboxApiKey, container, jsonInput}, onComplete, handl
     });
   });
 }
+
+function getTooltip(pickedInfo) {
+  if (!pickedInfo.picked) {
+    return null;
+  }
+  return {
```
Pessimistress: Cache the last result so we don't repeat the calculation when cursor moves over the same object.

Issue #3521, opened 2019-09-04
Shadow effect remaining issues


Pessimistress:
After https://github.com/uber/deck.gl/pull/3504 shadow effect can be dynamically toggled.

- [x] When shadow is turned off and on in the layer browser, it is rendered incorrectly until the next redraw.
- [x] Add render tests: lnglat, identity, shadow disabled by layer prop.

PR #3564, opened 2019-09-10
add rendering test for shadow

jianhuang01:
For #3521
#### Change List
- add rendering test


test/render/test-cases.js
```diff
@@ -12,14 +12,30 @@ import {
...
+  PostProcessEffect,
+  LightingEffect,
+  AmbientLight,
+  DirectionalLight
 } from '@deck.gl/core';
 import {noise, vignette} from '@luma.gl/effects';
 
 import {Fp64Extension} from '@deck.gl/extensions';
 
+const ambientLight = new AmbientLight({
```
Pessimistress: Define these inline with the test case that uses them.

test/render/test-cases.js
```diff
@@ -840,6 +856,7 @@ export const TEST_CASES = [
   },
   {
     name: 'column-lnglat',
+    effects: [lightingEffect],
```
Pessimistress: - Do not modify an existing test case - we need verification that make sure that each feature works independently.
- Missing tests for layers with shadow disabled

test/render/test-cases.js
```diff
@@ -865,6 +869,56 @@ export const TEST_CASES = [
...
+        extruded: true,
+        pickable: true,
+        opacity: 1,
+        getPosition: h => h.centroid,
+        getFillColor: h => [48, 128, h.value * 255, 255],
+        getElevation: h => h.value * 5000
+      }),
+      new SolidPolygonLayer({
+        id: 'polygon-lnglat',
+        data: dataSamples.polygons,
```
Pessimistress: Can you use a dataset that is 3D?

PR #3590, opened 2019-09-13
Update pydeck tooltip style and modularize the Jupyter widget tooltip

ajduberstein:
For #2929

#### Change List
- Change default tooltip styling for pydeck

### DIFF
--- a/modules/jupyter-widget/src/utils.js
+++ b/modules/jupyter-widget/src/utils.js
@@ -89,24 +89,39 @@ function getTooltip(pickedInfo) {
 }
 
 function tabularize(json) {
+  // Turns a JSON object of picked info into HTML for a tooltip
   const dataTable = document.createElement('div');
   dataTable.className = 'dataTable';
 
+  // Creates rows of two columns for the tooltip
   for (const key in json) {
+    if (['undefined', 'position', 'index'].includes(key)) {
+      continue; // eslint-disable-line
+    }
     const row = document.createElement('div');
     const header = document.createElement('div');
     header.className = 'header';
     header.innerText = key;
     Object.assign(header.style, {
-      fontWeight: 500,
-      marginRight: '10px'
+      fontWeight: 700,
+      marginRight: '10px',
+      flex: 1
     });
     const value = document.createElement('div');
     value.className = 'value';
-    value.innerText = JSON.stringify(json[key]);
+
+    if (Array.isArray(json[key])) {
+      value.innerText = JSON.stringify(`Array<${json[key].length}>`);
+    } else {
+      value.innerText = JSON.stringify(json[key]);
+    }
+
     Object.assign(value.style, {
-      float: 'right',
-      margin: 'header'
+      flex: 'none',
+      maxWidth: '250px',
+      overflow: 'hidden',
+      whiteSpace: 'nowrap',
+      textOverflow: 'ellipsis'
     });
     value.style.margin = 'header';
     row.appendChild(header);
--- a/modules/jupyter-widget/src/utils.js
+++ b/modules/jupyter-widget/src/utils.js
@@ -1,3 +1,5 @@
+import getTooltip from './widget-tooltip';
+
 /* global document */
 export function loadCss(url) {
   const link = document.createElement('link');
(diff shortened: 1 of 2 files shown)


modules/jupyter-widget/src/utils.js
```diff
@@ -89,24 +89,39 @@ function getTooltip(pickedInfo) {
 }
 
 function tabularize(json) {
+  // Turns a JSON object of picked info into HTML for a tooltip
   const dataTable = document.createElement('div');
   dataTable.className = 'dataTable';
 
+  // Creates rows of two columns for the tooltip
   for (const key in json) {
+    if (['undefined', 'position', 'index'].includes(key)) {
```
Pessimistress: - How could there ever be a key named `undefined`?
- Make the black list a constant so that you don't create a new array in each iteration. It's also faster to use `Set` instead of array.

modules/jupyter-widget/src/utils.js
```diff
@@ -89,24 +89,39 @@ function getTooltip(pickedInfo) {
...
-      marginRight: '10px'
+      fontWeight: 700,
+      marginRight: '10px',
+      flex: 1
     });
     const value = document.createElement('div');
     value.className = 'value';
-    value.innerText = JSON.stringify(json[key]);
+
+    if (Array.isArray(json[key])) {
```
Pessimistress: I think it's still valuable to see the array content when it's short, for example `color: [255, 0, 0]`. Maybe check if the array is less than 4 element long and the first element is a number?

modules/jupyter-widget/src/utils.js
```diff
@@ -89,24 +89,39 @@ function getTooltip(pickedInfo) {
...
+      value.innerText = JSON.stringify(`Array<${json[key].length}>`);
+    } else {
+      value.innerText = JSON.stringify(json[key]);
```
Pessimistress: - This should be clamped in case the object is very large.
- `JSON.stringify` may [throw exceptions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#Exceptions)

modules/jupyter-widget/src/utils.js
```diff
@@ -89,24 +89,39 @@ function getTooltip(pickedInfo) {
...
+      fontWeight: 700,
+      marginRight: '10px',
+      flex: 1
     });
     const value = document.createElement('div');
     value.className = 'value';
-    value.innerText = JSON.stringify(json[key]);
+
+    if (Array.isArray(json[key])) {
+      value.innerText = JSON.stringify(`Array<${json[key].length}>`);
```
Pessimistress: `JSON.stringify` is unnecessary since it's already a string.

modules/jupyter-widget/src/widget-tooltip.js
```diff
@@ -0,0 +1,99 @@
...
+    if (EXCLUDES.has(key)) {
+      continue; // eslint-disable-line
+    }
+    const header = document.createElement('div');
+    header.className = 'header';
+    header.innerText = key;
+    const valueElement = document.createElement('div');
+    valueElement.className = 'value';
+
+    setInnerText(valueElement, json[key]);
```
Pessimistress: Nit: use the right level of abstraction:

`valueElement.innerText = toText(json[key])`

modules/jupyter-widget/src/widget-tooltip.js
```diff
@@ -0,0 +1,99 @@
...
+function setInnerText(element, jsonValue) {
+  // Set contents of table value, trimming for certain types of data
+  if (Array.isArray(jsonValue) && jsonValue.length > 4) {
+    element.innerText = `Array<${jsonValue.length}>`;
+  } else {
+    try {
+      element.innerText = JSON.stringify(jsonValue);
+    } catch (err) {
+      // eslint-disable-next-line
+      console.error(err);
```
Pessimistress: Logging a console error is too strong. This is not a mistake on the user's end, merely that jupyter-widget does not deal with a particular edge case.

I would just make the text something like `<Non-Serializable Object>` in this case.

modules/jupyter-widget/src/widget-tooltip.js
```diff
@@ -0,0 +1,98 @@
...
+  // Creates rows of two columns for the tooltip
+  for (const key in json) {
+    if (EXCLUDES.has(key)) {
+      continue; // eslint-disable-line
+    }
+    const header = document.createElement('div');
+    header.className = 'header';
+    header.innerText = key;
+
+    const valueElement = toText(json[key]);
```
Pessimistress: `valueElement` is not an element?

PR #3593, opened 2019-09-13
Bundle optimization

Pessimistress:
#### Background

Shaves ~56K in minimized bundle

#### Change List
- Cherry-pick luma exports


Pessimistress:
I prefer not asking the users to include two scripts...

PR #3599, opened 2019-09-13
Wboit experiment

tsherif:
Created a `WBOITLayer` which is the `SolidPolygonLayer` with a custom draw function. This layer won't work with any other layers or picking since it has to manipulate framebuffers to work, but should be enough to see if this technique is worth pursuing. It's hard to depth weight properly because of how far apart our near and far planes are, but I came up with a function that gives reasonable results.

Some things that would have to be addressed in deck's rendering pipeline if we wanted a production-ready version of this:
- splitting up opaque and transparent objects so they can be drawn separately
- drawing opaque/transparent objects to different framebuffers with different programs and different blend operations, but sharing the depth buffer
- tightening the near and far planes to the bounds of the scene

Screen Shot 2019-09-13 at 7 17 15 PM


Pessimistress:
Can you extend the `SolidPolygonLayer` instead of copying files over? It will be easier to review.

tsherif:
@Pessimistress @georgios-uber Some cleanup.

For the shaders, the only real difference is the output of the fragment shader.

examples/experimental/wboit/wboit-layer/wboit-layer.js
```diff
@@ -0,0 +1,161 @@
...
+import vsTop from './wboit-layer-vertex-top.glsl';
+import vsSide from './wboit-layer-vertex-side.glsl';
+import fs from './wboit-layer-fragment.glsl';
+import oitBlendVs from './oit-blend-vertex.glsl';
+import oitBlendFs from './oit-blend-fragment.glsl';
+
+export default class WBOITLayer extends SolidPolygonLayer {
+  getShaders(vs) {
+    const shaders = super.getShaders(vs);
+    shaders.vs = vs.indexOf('SHADER_NAME solid-polygon-layer-vertex-shader-side') > 0 ? vsSide : vsTop;
```
georgios-uber: 😢

PR #3607, opened 2019-09-16
Add alphaCutoff prop to IconLayer

Pessimistress:
For https://github.com/uber/deck.gl/issues/3178#issuecomment-524427608

@danmarshall We cannot make the `TextLayer` background default pickable in a minor release, as it would be a breaking change. However, you will be able to achieve the effect using

```js
new TextLayer({
  // ...
  _subLayerProps: {
    characters: {alphaCutoff: 0}
  }
});
```

image

#### Change List
- Add `alphaCutoff` prop to IconLayer
- Documentation

### DIFF
--- a/docs/layers/icon-layer.md
+++ b/docs/layers/icon-layer.md
@@ -208,11 +208,11 @@ The maximum size in pixels.
 
 If on, the icon always faces camera. Otherwise the icon faces up (z)
 
-##### `cutoff` (Number, optional)
+##### `alphaCutoff` (Number, optional)
 
 - Default: `0.05`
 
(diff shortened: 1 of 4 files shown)


docs/layers/icon-layer.md
```diff
@@ -208,6 +208,13 @@ The maximum size in pixels.
 
 If on, the icon always faces camera. Otherwise the icon faces up (z)
 
+##### `cutoff` (Number, optional)
```
xintongxia: Does this prop impact both rendering and picking?

danmarshall:
Thanks @Pessimistress ! However, this seems to "flood-fill" with the `highlightColor` when `autoHighlight` is true.

PR #3608, opened 2019-09-16
Update data for 3D tiles example

jianhuang01:
#### Change List
- update token and data
- add thumbnail


examples/website/3d-tiles/app.js
```diff
@@ -54,7 +54,7 @@ export class App extends PureComponent {
         // Update deck.gl viewState, moving the camera to the new tileset
         longitude: cartographicCenter[0],
         latitude: cartographicCenter[1],
-        zoom: zoom + 1.5, // TODO - remove adjustment when Tileset3D calculates correct zoom
+        zoom,
```
ibgreen: Not too zoomed out?

xintongxia: is it corrected?

jianhuang01: I think it is better to zoom out a little on the website, users can understand the point cloud better. In high zoom, the model is kind of too sparse, people may not notice the full picture.

jianhuang01:
@ibgreen @xintongxia 
image

examples/website/3d-tiles/app.js
jianhuang01: Trying to make the model has a higher density, easier to perceive.

PR #3612, opened 2019-09-16
Allow easier configuration of height and width in a Deck visualization

ajduberstein:
For #2929

#### Change List
-


ajduberstein:
Closed in favor of #3613

PR #3615, opened 2019-09-17
Fix pixel ratio calculation, update docs

1chandu:
For #

#### Background
- Updated docs for `useDevicePixels` prop
- minor clean up of un used `useDevicePixels` prop on `LayerManager`
- Compute pixel ratio for each frame, instead of using cached value.

#### Change List
- useDevicePixels docs, minor cleanup
- Calculate device pixel ratio on every frame.

### DIFF
--- a/docs/api-reference/deck.md
+++ b/docs/api-reference/deck.md
@@ -179,19 +179,21 @@ background-color: #29323c;
 padding: 10px;
 ```
 
-##### `useDevicePixels` (Boolean or Number, optional)
+##### `useDevicePixels` (Boolean|Number, optional)
 
-When `true` (default), it will use device's full retina/HD resolution for rendering, when `false`, it will use the same resolution as the screen window (CSS window).
+Controls the resolution of drawing buffer used for rendering.
 
-As an experimental API, a custom ratio (Number) can be set to be used instead of actual device pixel ratio. This allows using lower or higher resolution than CSS window for rendering.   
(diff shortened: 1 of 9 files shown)


docs/api-reference/deck.md
```diff
@@ -179,16 +179,20 @@ background-color: #29323c;
 padding: 10px;
 ```
 
-##### `useDevicePixels` (Boolean, optional)
+##### `useDevicePixels` (Boolean or Number, optional)
```
Pessimistress: Convention: `Boolean|Number`

docs/api-reference/deck.md
```diff
@@ -179,16 +179,20 @@ background-color: #29323c;
...
+
+When `true` (default), it will use device's full retina/HD resolution for rendering, when `false`, it will use the same resolution as the screen window (CSS window).
```
Pessimistress: Use a list:
- `true`:
- `false`:
- `<number>`:

Instead of `device's full retina/HD resolution`, just link to https://developer.mozilla.org/en-US/docs/Web/API/Window/devicePixelRatio

Is `CSS window` a thing? Don't we just use `1` when `useDevicePixels: false`?

1chandu: I will updated to `... when false, it will use the CSS pixels resolution`.

Pessimistress: `CSS pixels resolution` doesn't really mean anything. I recommend sticking with the term "device pixel ratio" which people can find plenty of explanations online.

1chandu: We need to talk about two resolutions, `CSS pixels` resolution and `Device (physical) pixels` resolution as referred in MDN docs, I updated the PR accordingly.

ajduberstein:
image

Seems to work!

PR #3618, opened 2019-09-17
Implement isTransitioning check for attribute spring-transitions

rolyatmax:
This PR adds an `isTransitioning` check to spring-transitions and implements the `onStart`, `onInterrupt`, and `onEnd` events using the `isTransitioning` check. (This is one of the blockers listed in this issue: https://github.com/uber/deck.gl/issues/3574)

It accomplishes this by having each transitioning value render an `isTransitioning` flag to a 1x1 fragment. The fragments are blended, taking the max value, and if the resulting pixel has a value greater than `0`, at least one of the transitions is still in progress.

I'd like to flag a concern here about scalability. This has not been tried with millions of values, yet, though, so I wonder if the blend step is performant enough to be run on every frame for those types of numbers.

### DIFF
--- a/modules/core/src/transitions/gpu-spring-transition.js
+++ b/modules/core/src/transitions/gpu-spring-transition.js
@@ -1,6 +1,6 @@
 /* eslint-disable complexity, max-statements, max-params */
 import GL from '@luma.gl/constants';
-import {Buffer, Transform} from '@luma.gl/core';
+import {Buffer, Transform, Framebuffer, Texture2D, readPixelsToArray} from '@luma.gl/core';
 import {
   padBuffer,
   getAttributeTypeFromSize,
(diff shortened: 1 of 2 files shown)


modules/core/src/lib/attribute-transition-utils.js
```diff
@@ -1,14 +1,24 @@
...
 import {padArray} from '../utils/array-utils';
 
+function noop() {}
+
 const DEFAULT_TRANSITION_SETTINGS = {
   interpolation: {
     duration: 0,
-    easing: t => t
+    easing: t => t,
+    onStart: noop,
```
Pessimistress: @rolyatmax this bug is addressed in https://github.com/uber/deck.gl/pull/3627

rolyatmax:
@Pessimistress rebased on latest `master` and resolved conflicts. Let me know if this looks okay to you

Issue #3574, opened 2019-09-11
Follow-ups for Spring Transition


rolyatmax:
- [ ] support switching between transition types (perhaps lift the transitioning buffer management to `AttributeTransitionManager`?)
 - [ ] implement spring transition type for uniforms
 - [ ] scale the spring's step increments with Timeline
 - [ ] implement `isTransitioning` check (and `onStart` / `onEnd` events) for spring transitions (render an "isTransitioning flag" per vertex to a 1x1 framebuffer)

Pessimistress:
- [x] P0: Update `transitions` documentation in `layer.md`
- [x] P2: Add a `Transition and Animation` article to Developer Guide. Ideally with Codepen examples. Maybe build on [this demo](https://codepen.io/vis-gl/pen/xMGzxB)?

rolyatmax:
Updated description

charlieforward9:
Transition and Animation guide complete: https://deck.gl/docs/developer-guide/animations-and-transitions

PR #3621, opened 2019-09-17
Support spring transition in UniformTransitionManager

Pessimistress:
For #3574 

#### Change List
- Add default spring settings
- Add `CPUInterpolationTransition` and `GPUInterpolationTransition` class
- Unit tests

### DIFF
--- a/modules/core/src/lib/attribute-transition-utils.js
+++ b/modules/core/src/lib/attribute-transition-utils.js
@@ -1,33 +1,31 @@
 import BaseAttribute from '../lib/base-attribute';
 import {padArray} from '../utils/array-utils';
 
-const noop = () => {};
 const DEFAULT_TRANSITION_SETTINGS = {
-  type: 'interpolation',
-  onStart: noop,
-  onEnd: noop,
-  onInterrupt: noop,
-  onUpdate: noop,
-  enter: x => x,
-  // interpolation
-  duration: 0,
-  easing: t => t,
-  // spring
-  stiffness: 0.05,
-  damping: 0.5
+  interpolation: {
+    duration: 0,
+    easing: t => t
+  },
+  spring: {
+    stiffness: 0.05,
+    damping: 0.5
+  }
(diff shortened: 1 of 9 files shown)


modules/core/src/lib/attribute-transition-utils.js
```diff
@@ -4,13 +4,17 @@ import {padArray} from '../utils/array-utils';
...
   onInterrupt: noop,
   onUpdate: noop,
-  enter: x => x
+  enter: x => x,
+  // interpolation
+  duration: 0,
+  easing: t => t,
+  // spring
+  stiffness: 0.1,
+  damping: 0.5
```
rolyatmax: These values might be a bit snappy for defaults. If it was up to me, I might drop the `stiffness` to `0.05`. Though, if you tried a bunch of values and you preferred these, I'm okay with them! :-)

modules/core/src/transitions/cpu-spring-transition.js
```diff
@@ -0,0 +1,81 @@
...
+  }
+  return Math.abs(value1 - value2);
+}
+
+export default class CPUSpringTransition extends Transition {
+  get value() {
+    return this._currValue;
+  }
+
+  update() {
```
rolyatmax: Should we add a todo about making use of the `Timeline`?

modules/core/src/transitions/transition.js
```diff
@@ -76,7 +77,6 @@ export default class Transition {
         duration
       });
       this._handle = handle;
-      this.onStart(this);
```
rolyatmax: I like this change. (It simplifies the implementation of the same `onStart` event for attribute spring transitions.) Would it be crazy to move the `this._handle = timeline.addChannel({...})` up to the `start()` method as well? If we did, we could drop all the `this._handle = null`s, which would no longer be needed to trigger the creation of a new timeline channel in this `update()` method.

PR #3622, opened 2019-09-17
Update change log

1chandu:
Update 7.2.4 change log.


1chandu:
Doc only change, merging.

PR #3626, opened 2019-09-17
Allow pydeck users to hide tooltip

ajduberstein:
For #2929

#### Change List
- Provides support for users to hide pydeck tooltip while still allowing
data selection

### DIFF
--- a/modules/jupyter-widget/src/utils.js
+++ b/modules/jupyter-widget/src/utils.js
@@ -24,12 +24,14 @@ export function updateDeck(inputJSON, {jsonConverter, deckgl}) {
   deckgl.setProps(results);
 }
 
-export function initDeck(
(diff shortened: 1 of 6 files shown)


modules/jupyter-widget/src/utils.js
```diff
@@ -24,7 +24,12 @@ export function updateDeck(inputJSON, {jsonConverter, deckgl}) {
   deckgl.setProps(results);
 }
 
-export function initDeck({mapboxApiKey, container, jsonInput}, onComplete, handleClick) {
+export function initDeck(
+  {mapboxApiKey, container, jsonInput},
+  useTooltip,
+  onComplete,
+  handleClick
```
Pessimistress: Please use named parameters for all.

PR #3627, opened 2019-09-18
Fix GPUSpringTransition

Pessimistress:
For #3574 

Broken due to default callbacks removed from `normalizeTransitionSettings` in #3621 

#### Change List
- Generalize `Transition` class, remove `duration` and `easing` references
- Store transition settings in dedicated object in `Transition` instance
- Add `end` method to `Transition`
- Use `Transition` class to handle callbacks in `GPUSpringTransition`
- Align code style in `GPUSpringTransition` and `GPUInterpolationTransition`

### DIFF
--- a/modules/core/src/transitions/cpu-spring-transition.js
+++ b/modules/core/src/transitions/cpu-spring-transition.js
@@ -69,7 +69,7 @@ export default class CPUSpringTransition extends Transition {
 
     if (delta < EPSILON && velocity < EPSILON) {
       nextValue = toValue;
-      this.end(this);
+      this.end();
     }
 
     this._prevValue = _currValue;
--- a/modules/core/src/transitions/cpu-spring-transition.js
+++ b/modules/core/src/transitions/cpu-spring-transition.js
@@ -57,27 +57,22 @@ export default class CPUSpringTransition extends Transition {
     return this._currValue;
   }
 
-  update() {
-    const updated = super.update();
-    if (updated) {
-      // TODO - use timeline
-      // const {time} = this;
+  _onUpdate() {
+    // TODO - use timeline
+    // const {time} = this;
 
-      const {fromValue, toValue, damping, stiffness} = this;
-      const {_prevValue = fromValue, _currValue = fromValue} = this;
-      let nextValue = updateSpring(_prevValue, _currValue, toValue, damping, stiffness);
-      const delta = distance(nextValue, toValue);
-      const velocity = distance(nextValue, _currValue);
+    const {fromValue, toValue, damping, stiffness} = this.settings;
+    const {_prevValue = fromValue, _currValue = fromValue} = this;
+    let nextValue = updateSpring(_prevValue, _currValue, toValue, damping, stiffness);
+    const delta = distance(nextValue, toValue);
+    const velocity = distance(nextValue, _currValue);
 
-      if (delta < EPSILON && velocity < EPSILON) {
-        nextValue = toValue;
-        this._inProgress = false;
-        this.onEnd(this);
-      }
-
-      this._prevValue = _currValue;
-      this._currValue = nextValue;
+    if (delta < EPSILON && velocity < EPSILON) {
+      nextValue = toValue;
+      this.end(this);
     }
-    return updated;
+
+    this._prevValue = _currValue;
+    this._currValue = nextValue;
   }
 }
--- a/modules/core/src/transitions/gpu-spring-transition.js
+++ b/modules/core/src/transitions/gpu-spring-transition.js
@@ -9,11 +9,13 @@ import {
   cycleBuffers
 } from '../lib/attribute-transition-utils';
 import Attribute from '../lib/attribute';
+import Transition from './transition';
 
 export default class GPUSpringTransition {
-  constructor({gl, attribute, transitionSettings}) {
+  constructor({gl, attribute, timeline}) {
+    this.gl = gl;
     this.type = 'spring';
-    this.transitionSettings = transitionSettings;
+    this.transition = new Transition(timeline);
     this.attribute = attribute;
     // this is the attribute we return during the transition - note: if it is a constant
     // attribute, it will be converted and returned as a regular attribute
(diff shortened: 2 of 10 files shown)


modules/core/src/transitions/gpu-spring-transition.js
```diff
@@ -51,79 +53,79 @@ export default class GPUSpringTransition {
...
+      value: attribute.value
     });
 
     // when an attribute changes values, a new transition is started. These
     // are properties that we have to store on this instance but can change
     // when new transitions are started, so we have to keep them up-to-date. :(
-    this.transitionSettings = transitionSettings;
-    if (this.isTransitioning()) {
-      this.transitionSettings.onInterrupt();
-    }
```
rolyatmax: Why are we removing this?

modules/core/src/transitions/gpu-spring-transition.js
```diff
@@ -9,11 +9,13 @@ import {
...
 import Attribute from '../lib/attribute';
+import Transition from './transition';
 
 export default class GPUSpringTransition {
-  constructor({gl, attribute, transitionSettings}) {
+  constructor({gl, attribute, timeline}) {
+    this.gl = gl;
     this.type = 'spring';
-    this.transitionSettings = transitionSettings;
+    this.transition = new Transition({timeline});
```
rolyatmax: I don't think the `Transition` class, as it stands, is the right place for the configuration. It tries to also fire the `on...` events by relying on `duration`, but the `duration` property only makes sense for interpolation transitions. When we implement the timeline stuff here, we'll likely need to pull out the event-firing functionality from the `Transition` class. I've got the events implemented in [this PR](https://github.com/uber/deck.gl/pull/3618) already, so I don't think this change adds any new functionality. Can we leave `Transition` out of this class until we clean up `Transition`, more effectively dividing its responsibilities into classes/functions?

Pessimistress: The `Transition` class no longer require `duration` as of #3621. Aside from callbacks, it also takes care of attaching channels to the timeline. We can continue to generalize the class, but I feel strongly about having all the logic of interpreting transition settings in one place. The fact that transition callbacks are fired in unexpected places (and the lack of tests) was the reason this was broken in the first place.

rolyatmax: It looks like the `Transition` class is trying to do these three things:
(1) create a channel on the given timeline and use it + the supplied `easing` function to determine the transition's `time`.
(2) fire the `on...` event handlers
(3) store transition configuration (e.g. `duration`, `easing`, `on...` event handlers)

Number 1 only matters for interpolation transitions. As for number 2, it _cannot_ correctly fire the `on...` event handlers for spring transitions (it cannot know when a transition has ended for spring transitions). So all this is used for in spring transitions is, number 3, storing configuration. And this last one could/should be accomplished with a plain JavaScript object (we don't need a class to store properties of a configuration).

The reason why this broke is because we were defining the defaults for transition settings in two places. I agree, we should have defaults slotted in in one spot only. I don't think the `Transition` class is the right place for that because the `Transition` class doesn't make sense as part of a spring transition, since it conflates the three functions above. Can't we just fix the defaults-in-more-than-one-place issue by putting the event defaults in `attribute-transition-utils`' `DEFAULT_TRANSITION_SETTINGS`?

Pessimistress: Ok there are two things I'm trying to fix here, maybe they don't both fit in this PR.

1. Fix the crash in `GPUSpringTransition` when `onInterrupt` and `onUpdate` are not supplied - this can indeed be fixed by adding defaults to `DEFAULT_TRANSITION_SETTINGS`.
2. Allow `GPUSpringTransition` to access `timeline` which will allow us to add integration tests. I agree with you that the role of the `Transition` class is ambiguous, and I prefer to define the class better rather than duplicate logic in various places.

rolyatmax: For posterity/visibility: @Pessimistress and I talked through this offline and we decided that we'll update this PR with these changes:
(1) Remove interpolation-related logic (the `easing` calculation)
(2) Add an `onEnd` method to the `Transition` class which tears down the channel and fires the `onEnd` user-supplied callback. This means that the `Transition` class will be completely responsible for firing _all_ user-supplied callbacks.

Note: we will land these changes before https://github.com/uber/deck.gl/pull/3618.

@Pessimistress - let me know if I've missed anything here.

modules/core/src/transitions/cpu-spring-transition.js
```diff
@@ -57,27 +57,22 @@ export default class CPUSpringTransition extends Transition {
...
-        nextValue = toValue;
-        this._inProgress = false;
-        this.onEnd(this);
-      }
-
-      this._prevValue = _currValue;
-      this._currValue = nextValue;
+    if (delta < EPSILON && velocity < EPSILON) {
+      nextValue = toValue;
+      this.end(this);
```
rolyatmax: Looks like we don't need to pass `this` here.

modules/core/src/transitions/transition.js
```diff
@@ -37,56 +30,67 @@ export default class Transition {
...
 
-    if (timeline.isFinished(handle)) {
-      timeline.removeChannel(handle);
-      this._handle = null;
-      this._inProgress = false;
-      this.onEnd(this);
+    // This only works if `settings.duration` is set
+    // Spring transition must call `end` manually
+    if (this.timeline.isFinished(this._handle)) {
+      this.end();
```
rolyatmax: To me this hints that we still don't have the right abstraction, but I'm okay with this (the comment helps) until we can find a better solution.

Pessimistress: I agree, this is mostly for backward compatibility.

Pessimistress:
On an abstract level, I think the `GPU*Transition` classes should be subclassed from `Transition`. Their current structure already closely follow that of `Transition`. As discussed offline, the ownership of these instances is not streamlined, and we will need to address the issue as a whole after the current milestone.

PR #3632, opened 2019-09-18
Update transition docs with spring info

rolyatmax:
Adding spring transition info to transition docs. I get the feeling that at some point maybe we should offer some explanation on the differences between `interpolation` and `spring` types?

Addresses a blocker in [this list of spring-transition follow-ups](https://github.com/uber/deck.gl/issues/3574).

### DIFF
--- a/docs/api-reference/layer.md
+++ b/docs/api-reference/layer.md
@@ -402,6 +402,12 @@ new Layer({
       duration: 300,
       easing: d3.easeCubicInOut,
       enter: value => [value[0], value[1], value[2], 0] // fade in
+    },
+    getRadius: {
+      type: 'spring',
+      stiffness: 0.01,
+      damping: 0.15,
+      enter: value => [0] // grow from size 0
     }
   }
 });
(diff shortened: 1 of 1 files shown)


rolyatmax:
I also wonder how we could better indicate the properties that are only for `spring`-type transitions and the properties that are only for `interpolation`-type transitions. That still feels very easy to miss in this table. 🤔

docs/api-reference/layer.md
```diff
@@ -402,23 +402,32 @@ new Layer({
...
 });
 ```
 
 | Parameter     | Type       | Default     | Description |
 | ---------     | --------   | ----------- | ----------- |
-| `duration`    | `Number`   | `0`         | Duration of the transition animation, in milliseconds |
-| `easing`      | `Function` | LINEAR (`t => t`) | Easing function that maps a value from [0, 1] to [0, 1], see [http://easings.net/](http://easings.net/)  |
-| `enter`      | `Function` | APPEARANCE (`value => value`) | Callback to get the value that the entering vertices are transitioning from. See notes below |
+| `type`        | `String`   | `interpolation` | Type of the transition (either `interpolation` or `spring`) |
+| `duration`    | `Number`   | `0`         | Duration of the transition animation, in milliseconds (only for `interpolation` transition `type`) |
```
Pessimistress: This is getting hard to read. Should we split these into separate tables, sorted by `type`?

Pessimistress:
Pushed an update to split off type-specific parameters into their own tables.

rolyatmax:
@Pessimistress Thanks for pushing that last commit! I'm happy with this if you are. Time to land?

PR #3643, opened 2019-09-19
deck layer visibility on zoom less than 3

divyanshupathania:
Bug fix for deck layer visibility on zoom level below 3
On google maps when zoom is changed to 2 or 1, the deck layer does not render
as canvas is shifted to extreme right.

### DIFF
--- a/modules/google-maps/src/utils.js
+++ b/modules/google-maps/src/utils.js
@@ -106,8 +106,14 @@ export function getViewState(map, overlay) {
   const nw = projection.fromContainerPixelToLatLng(nwContainerPx);
   const nwDivPx = projection.fromLatLngToDivPixel(nw);
 
+  // checking the number of maps visible on container
+  var mapCount = width / projection.getWorldWidth()
+  if (mapCount >= 1) {
+    nwDivPx.x = nwDivPx.x - projection.getWorldWidth()
+  }
+
   // Compute fractional zoom.
-  const scale = (topRight.x - bottomLeft.x) / width;
+  var scale = (bottomLeft.y - topRight.y) / height;
   const zoom = Math.log2(scale) + map.getZoom() - 1;
 
   // Compute fractional center.
--- a/modules/google-maps/src/utils.js
+++ b/modules/google-maps/src/utils.js
@@ -107,13 +107,13 @@ export function getViewState(map, overlay) {
   const nwDivPx = projection.fromLatLngToDivPixel(nw);
 
   // checking the number of maps visible on container
-  var mapCount = width / projection.getWorldWidth()
+  const mapCount = width / projection.getWorldWidth()
   if (mapCount >= 1) {
     nwDivPx.x = nwDivPx.x - projection.getWorldWidth()
   }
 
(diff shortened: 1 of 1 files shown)


modules/google-maps/src/utils.js
```diff
@@ -106,8 +106,14 @@ export function getViewState(map, overlay) {
   const nw = projection.fromContainerPixelToLatLng(nwContainerPx);
   const nwDivPx = projection.fromLatLngToDivPixel(nw);
 
+  // checking the number of maps visible on container
+  var mapCount = width / projection.getWorldWidth()
+  if (mapCount >= 1) {
+    nwDivPx.x = nwDivPx.x - projection.getWorldWidth()
```
Pessimistress: What happens when `mapCount > 2`? Which copy does `fromLatLngToDivPixel` project to?

modules/google-maps/src/utils.js
```diff
@@ -106,8 +106,14 @@ export function getViewState(map, overlay) {
...
+  }
+
   // Compute fractional zoom.
-  const scale = (topRight.x - bottomLeft.x) / width;
+  var scale = (bottomLeft.y - topRight.y) / height;
```
Pessimistress: Use `const`

modules/google-maps/src/utils.js
donmccurdy: If the change above were omitted, and the change below (compute scale from Y axis, not X) were kept, what is the result? deckgl would draw on the leftmost visible map? A demo might be helpful – trying to be sure I understand what each of these changes is doing.

divyanshupathania: @Pessimistress the above code just check that how many maps are visible on the container screen, so in case of zoom greater than 1 mapCount is always less then 1. It is for handling layers on zoom level 1. The fromLatLngToDivPixel projects to the map which is more visible on container

@donmccurdy, yes you are right if above code is omitted and scale computation is done using Y axis, then deck layers are visible on zoom level 2. It is just to handle the zoom level 1. It will handle the left property of canvas, so that it remains on the map which is more visible.

donmccurdy: The zoom levels are not coupled to mapCount in the way you appear to be assuming – the resolution of the screen and user agent zooming will both affect the layout as well. On the lowest zoom level you might get anywhere from 1 visible map to many visible maps:

Screen Shot 2019-09-24 at 8 35 39 AM

^That's an extreme case, of course, but the logic above — shifting the canvas by a single world width — does not look robust enough to me. If the change to `scale` alone improves things, it might be better to do that first and consider which map to render over with another PR? Or if you want to try to implement it here, I would suggest trying to use the centermost map, and assuming there could be any number of maps.

owap:
Looks like the build is failing due to lint issues.  According to Travis:

> Not all files using prettier code style. This may be fixed by running `npm run lint fix`

Thanks for your hard work on the issue, we're looking forward to having this resolved!

Pessimistress:
Closing in favor of #4247

PR #3644, opened 2019-09-19
Aggregate tiles by zoom level

jesusbotella:
WIP.

This PR kicks off the project to improve deck.gl's TileLayer performance.

The first step, that is what the PR has so far, is a tile aggregation by zoom level to avoid rendering that many layers.


jesusbotella:
Sorry, this was intended to be a private PR 😓

PR #3645, opened 2019-09-19
icon-layer-triangle-issue

ellie-farjoud:
Replaced TRIANGLE_FAN with TRIANGLES, since IE11 is not supporting TRIANGLE_FAN

### DIFF
--- a/modules/layers/src/icon-layer/icon-layer.js
+++ b/modules/layers/src/icon-layer/icon-layer.js
@@ -200,7 +200,7 @@ export default class IconLayer extends Layer {
   }
 
   _getModel(gl) {
-    const positions = [-1, -1, 1, -1, 1, 0, 1, 1, 0, 1, -1, 1, -1, -1, 1, -1, -1, 1];
+    const positions = [-1, -1, 0, -1, 1, 0, 1, 1, 0, 1, 1, 0, 1, -1, 0, -1, -1, 0];
 
     return new Model(
       gl,
--- a/modules/layers/src/icon-layer/icon-layer.js
+++ b/modules/layers/src/icon-layer/icon-layer.js
@@ -200,14 +200,14 @@ export default class IconLayer extends Layer {
   }
 
   _getModel(gl) {
-    const positions = [-1, -1, 0, -1, 1, 0, 1, 1, 0, 1, -1, 0];
+    const positions = [-1, -1, 1, -1, 1, 0, 1, 1, 0, 1, -1, 1, -1, -1, 1, -1, -1, 1];
 
     return new Model(
       gl,
       Object.assign({}, this.getShaders(), {
         id: this.props.id,
         geometry: new Geometry({
-          drawMode: GL.TRIANGLE_FAN,
+          drawMode: GL.TRIANGLES,
           attributes: {
             positions: new Float32Array(positions)
           }


ibgreen:
There are quite a few layers that use `TRIANGLE_FAN` @tsherif

ibgreen:
@tsherif I wonder if we could handle this on a luma/polyfill level - i.e. if we get a `TRIANGLE_FAN` on IE11 we generated an expanded buffer and replace with `TRIANGLE` under the hood.

That way we could keep the deck.gl code clean...

modules/layers/src/icon-layer/icon-layer.js
```diff
@@ -200,14 +200,14 @@ export default class IconLayer extends Layer {
   }
 
   _getModel(gl) {
-    const positions = [-1, -1, 0, -1, 1, 0, 1, 1, 0, 1, -1, 0];
+    const positions = [-1, -1, 1, -1, 1, 0, 1, 1, 0, 1, -1, 1, -1, -1, 1, -1, -1, 1];
```
Pessimistress: z values should be `0`.

ellie-farjoud: It 's fixed.

tsherif:
TRIANGLE_FAN works fine in IE. Can you share a bit about why you think otherwise? Are you getting errors with deck.gl?

ellie-farjoud:
Actually the TRIANGLE_FAN doesn't work fine in ie11, while TRIANGLES does work fine in IE11.

I do not get any error in deck.gl. 

Here is the issue I had: https://github.com/uber/deck.gl/issues/2486

and by changing the positions, found that the triangle-fan is not supported in IE11.

By converting The TRIANGLE_FAN to TRIANGLES the icons is shown completely as below:

fixed-iconlayer

ellie-farjoud:
I do not get any error in deck.gl.

Here is the issue I had: #2486

and by changing the positions from TRIANGLE_FAN to TRIANGLE the icons are shown completely as below:

fixed-iconlayer

If IE11 does support the TRIANGLE_FAN, still something happening that The TRIANGLE_FAN is converted to TRIANGLES.

tsherif:
Can you try loading this page in IE? https://tsherif.github.io/triangle-fan-test/gl.html

This is what it should look like:
Screen Shot 2019-09-19 at 5 42 42 PM

ellie-farjoud:
Yes, it looks same.

tsherif:
That scene is using TRIANGLE_FAN so that's not the root issue. We made some fixes recently in luma.gl to how drawing is handled in IE (https://github.com/uber/luma.gl/pull/1238). I'm curious if this is related. Could you try your code against a beta version of deck.gl? I believe the latest is 7.3.0-beta.2.

ellie-farjoud:
Yes, it's working, thanks so much!

tsherif:
Great!

etimberg:
@tsherif @ibgreen I tested the luma.gl change I made, but it doesn't resolve this issue. The icon layer still draws incorrectly. It almost looks like some kind of transform has been applied to the icon square turning it into a rotated rectangle.

bad icons

etimberg:
@tsherif @ibgreen I've done some debugging on this and I think I have a cause, but I'm not yet sure what the ultimate root cause is.

IE11 appears to be interpreting the list of positions as a set of `x,y` points rather than as a set of `x,y,z` points. If I override `IconLayer._getModel()` to change the positions to be 2d instead of 3d, IE11 works fine and Chrome breaks.

Any ideas on what might be causing this?

PR #3647, opened 2019-09-19
Remove attribute-transitions that have null settings

rolyatmax:
This makes sure we remove existing attribute transitions that now have `null` settings as a result of new transition settings being applied.

### DIFF
--- a/modules/core/src/lib/attribute-transition-manager.js
+++ b/modules/core/src/lib/attribute-transition-manager.js
@@ -55,7 +55,7 @@ export default class AttributeTransitionManager {
 
     for (const attributeName in this.transitions) {
       const attribute = attributes[attributeName];
-      if (!attribute || !attribute.supportsTransition()) {
+      if (!attribute || !attribute.getTransitionSetting(transitions)) {
         // Animated attribute has been removed
         this._removeTransition(attributeName);
       }


modules/core/src/lib/attribute-transition-manager.js
```diff
@@ -55,7 +55,8 @@ export default class AttributeTransitionManager {
 
     for (const attributeName in this.transitions) {
       const attribute = attributes[attributeName];
-      if (!attribute || !attribute.supportsTransition()) {
+      const settings = attribute && attribute.getTransitionSetting(transitions);
+      if (!attribute || !attribute.supportsTransition() || !settings) {
```
rolyatmax: I don't really like this - maybe there's a refactor that streamlines it a bit?

modules/core/src/lib/attribute-transition-utils.js
```diff
@@ -20,6 +20,9 @@ export function normalizeTransitionSettings(userSettings, layerSettings) {
     userSettings = {duration: userSettings};
   }
   userSettings.type = userSettings.type || 'interpolation';
+  if (userSettings.type === 'interpolation' && userSettings.duration < 1) {
+    return null; // return null because there's no transition to perform
+  }
```
rolyatmax: Feels a bit weird to return `null` from a `normalizeTransitionSettings()` function that has been supplied a valid transition setting, but I guess this is what we were doing before.

modules/core/src/lib/attribute-transition-manager.js
Pessimistress: I believe `attribute.getTransitionSetting` already returns `null` if transition is not supported, so you only need to check `settings`

modules/core/src/lib/attribute-transition-utils.js
Pessimistress: I'm handling this inside `GPUInterpolationTransition` in https://github.com/uber/deck.gl/pull/3649

PR #3649, opened 2019-09-19
Clean up transition classes, avoid crash with `duration:0`

Pessimistress:
For #3574 

#### Change List
- GPU transition classes: Rename `isTransitioning()` to `inProgress`, remove `getTransitioningAttribute()`
- Do not throw with `duration:0`
- Improve minification

### DIFF
--- a/modules/core/src/transitions/gpu-interpolation-transition.js
+++ b/modules/core/src/transitions/gpu-interpolation-transition.js
@@ -9,7 +9,6 @@ import {
   cycleBuffers
 } from '../lib/attribute-transition-utils';
 import Transition from './transition';
-import assert from '../utils/assert';
 
 export default class GPUInterpolationTransition {
   constructor({gl, attribute, timeline}) {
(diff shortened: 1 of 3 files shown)


modules/core/src/transitions/gpu-interpolation-transition.js
```diff
@@ -50,47 +47,49 @@ export default class GPUInterpolationTransition {
...
       // Hack: Float64Array is required for double-precision attributes
       // to generate correct shader attributes
-      value: this.attribute.value
+      value: attribute.value
     });
 
     this.transition.start(transitionSettings);
 
-    this.transform = this.transform || new Transform(this.gl, getShaders(this.attribute.size));
+    this.transform = this.transform || new Transform(gl, getShaders(attribute.size));
```
rolyatmax: Because we are dropping out of this function at the top in some cases, `this.transform` doesn't always get created (since we're lazily creating it here). Can we move the `this.transform` creation to the `constructor` with the other GL resources? So we can easily reason about the state of this instance at any given point in time?

rolyatmax: @Pessimistress - I moved `this.transform` creation to `constructor` in the commit I just pushed.

Issue #3591, opened 2019-09-13
7.3 Release Tracker


1chandu:
- [x] Bump all dependencies (math.gl, probe.gl, loaders.gl and luma.gl)  latest alpha/beta/ prod versions (@1chandu)
- [x] Publish alpha/beta  (@1chandu)
- [x] Bump all examples to latest alpha/beta (@1chandu)
- [x] Stage the website (@1chandu)
- [x] Verify whats-new (@tgorkin )
- [x] Verify upgrade-guide (@Pessimistress  )
- [x] Verify all issues in milestone are closed or move out (@tgorkin )
- [x] Verify all API changes are documented (@Pessimistress )
- [x] Website (docs and example) testing: [staging](http://pessimistress.github.io/deck.gl/#/)
    - [x] Testing on Chrome (@ajduberstein)
    - [x] Testing on Safari (@xintongxia)
    - [x] Testing on Firefox (@1chandu)
    - [x] Testing on Edge + IE11 (@tsherif)
    - [x] Testing on iOS (@Pessimistress )
    - [x] Testing on Android (@Pessimistress)
- [x] Cut 7.3-release branch after all fixes
- [x] Update website doc urls to point to `7-3-release` branch  (@1chandu )
- [x] Update 7.3 badge to point to 7.3 branch on Overview page. (@1chandu)
- [x] Update luma.gl dependency to 7.3 prod (@1chandu )
- [x] Publish 7.3 prod version.  (@1chandu )

Testing procedure
- access staged doc website
- go through all examples
- verify functionality in control panels
- look for console errors/warnings
- add results of testing as comments on this issue
- Test Playground (JSON Browser) (Instructions:  ?? TBD)

SymbolixAU:
I've been trying the [`BrushingExtension()`](https://github.com/uber/deck.gl/blob/master/docs/api-reference/extensions/brushing-extension.md) the but keep getting the error 

> deck.BrushingExtension is not a constructor

I'm using `https://unpkg.com/deck.gl@^7.3.0-alpha.7/dist.min.js`. 

The new `LightingEffects` seem to work fine

The snippet of code to generate the error is

```js

  const d = new deck.DirectionalLight({
	  color: [255, 255, 255],
	  intensity: 1.0,
	  direction: [-3, -9, -1]
	});
  console.log( d );
  const b = new deck.BrushingExtension();
  console.log( b );

```
And here's a screenshot

[...]

PR #3661, opened 2019-09-19
Add HeatmapLayer example to the website

Pessimistress:
For #3591 

#### Change List
- Add HeatmapLayer example to the website
- Rename the old `HeatmapDemo` to `HexagonDemo`
- Update thumbnails of the new demos


website/src/components/demos/heatmap-demo.js
```diff
@@ -1,29 +1,20 @@
...
-      coverage: {displayName: 'Coverage', type: 'range', value: 0.7, step: 0.1, min: 0, max: 1},
-      upperPercentile: {
-        displayName: 'Upper Percentile',
-        type: 'range',
-        value: 100,
-        step: 0.1,
-        min: 80,
-        max: 100
-      }
+      intensity: {displayName: 'Intensity', type: 'range', value: 1, step: 0.1, min: 0, max: 5},
```
1chandu: Include `radiusPixels` as well? If don't want more than two, I would replace `threshold` with `radiusPixels`.

PR #3668, opened 2019-09-20
Improve multi-picking performance

Pessimistress:
#### Change List
- Use `buffer.subData` in `clearPickingColor` instead of mutating the cached array. This also eliminates the need for `copyPickingColors`.
- Promote `attribute.getVertexOffset` to a public method
- Consolidate special-case handling in `PathLayer` and `SolidPolygonLayer`


Pessimistress:
> I am assuming this is just a change on an internal class

That is correct.

PR #3673, opened 2019-09-21
Split website bundle

Pessimistress:
**FOR DISCUSSION**

Requires publishing #3672 and #3674.

Pros:
- Website bundle 5.6M -> 3.2M
- We do not have to republish the website if a bug is fixed in the libraries.
- This helps verify the validity of the pre-bundled version.

Cons:
- Complicate the website setup and potentially harder to debug.

#### Change List
- Production build of the website includes deck, luma, mapboxgl, h3-js, and highlight.js from CDN.
- Avoid using `import DeckGL from 'deck.gl'` in examples.

### DIFF
--- a/website/webpack.config.js
+++ b/website/webpack.config.js
@@ -145,19 +145,6 @@ const addProdConfig = config => {
   });
 };
 
-// s2-geometry cannot be directly included in a script tag
-const S2_CONFIG = {
-  mode: 'production',
-
-  entry: './src/s2-geometry',
-
-  output: {
-    libraryTarget: 'umd',
-    path: resolve(__dirname, './dist'),
-    filename: 's2-geometry.js'
-  }
-};
-
 module.exports = env => {
   env = env || {};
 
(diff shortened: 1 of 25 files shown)


website/webpack.config.js
```diff
@@ -127,11 +125,39 @@ const addProdConfig = config => {
...
+    'mapbox-gl': 'mapboxgl'
+  };
+
   return Object.assign(config, {
     mode: 'production'
   });
 };
 
+// s2-geometry cannot be directly included in a script tag
+const S2_CONFIG = {
```
Pessimistress: I have come to find out that the script tag usage as documented in https://github.com/uber/deck.gl/blob/641f5b1521754dbc03917f9a8078badc5c7ef248/docs/layers/s2-layer.md#installation does not work. The s2-geometry lib contains a `require` so it is not umd compliant.

@ibgreen I really don't want to bundle it with the main module. Any suggestions? Shall we distribute a pre-bundled version of s2-geometry with geo-layers?

Pessimistress: Switching to use https://bundle.run

Pessimistress:
> I wonder if it works for draco and laz loaders as well?

The loader bundles need some work. I can look into it next.

PR #3681, opened 2019-09-24
Fix HeatmapLayer crash in iOS (Safari)

1chandu:
For #3591

#### Background
- Requires https://github.com/uber/luma.gl/pull/1257
- When rendering to float textures not supported, use unsigned byte texture and log a warning for lower precession. 

#### Change List
-

### DIFF
--- a/modules/aggregation-layers/src/heatmap-layer/heatmap-layer.js
+++ b/modules/aggregation-layers/src/heatmap-layer/heatmap-layer.js
@@ -239,22 +239,23 @@ export default class HeatmapLayer extends CompositeLayer {
       _targetTexture: weightsTexture,
       _targetTextureVarying: 'weightsTexture'
     });
+    const maxWeightTransform =  new Transform(gl, {
+      id: `${this.id}-max-weights-transform`,
+      _sourceTextures: {
+        inTexture: weightsTexture
+      },
+      _targetTexture: maxWeightsTexture,
+      _targetTextureVarying: 'outTexture',
+      vs: vs_max,
+      elementCount: textureSize * textureSize
+    });
 
     this.setState({
       weightsTexture,
       maxWeightsTexture,
       weightsTransform,
       model: weightsTransform.model,
-      maxWeightTransform: new Transform(gl, {
-        id: `${this.id}-max-weights-transform`,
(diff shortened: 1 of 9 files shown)


modules/aggregation-layers/src/heatmap-layer/heatmap-layer.js
```diff
@@ -423,6 +446,12 @@ export default class HeatmapLayer extends CompositeLayer {
     });
     this._updateMaxWeightValue();
 
+    // reset filtering parameters
+    weightsTexture.setParameters({
```
Pessimistress: Why is this needed? Are parameters changed elsewhere?

1chandu: Yes, `Transform` sets parameters to NEAREST to correctly access each individual pixel, we need a proper push pop technique that should happen within `Transform`, tracked by https://github.com/uber/luma.gl/issues/1193

modules/aggregation-layers/src/utils/resource-utils.js
```diff
@@ -29,6 +29,29 @@ export function getFloatTexture(gl, opts = {}) {
...
 }
 
+export function getUByteTexture(gl, opts = {}) {
+  const {
+    width = 1,
+    height = 1,
+    data = null,
+    unpackFlipY = true,
+    parameters = DEFAULT_PARAMETERS
+  } = opts;
```
Pessimistress: As an internal helper function, I'd like to keep these very concise and not too "smart", as in as few defaults/assumptions as possible. This function constructs a texture with a specific type. All other parameter defaults are either redundant because they are the same defaults as used by `Texture2D`, or error-prone because they are not implied by the name of the function. To keep this helper flexible, you can simply pass through the opts like

```js
return new Texture2D(gl, {type: GL.UNSIGNED_BYTE, ...opts});
```

Comparing to your implementation, it is easy to test (only one branch), and has minimal impact on bundle size.

1chandu: Updated, keeping `getFloatTexture` for now, as it is used in aggregator, will clean that up in a different PR.

PR #3686, opened 2019-09-24
Stress test app

tsherif:
Stress test app using in the performance report. Currently set up to render 100,000 points in 1000 ScatterplotLayers, but could be easily extended to be more customizable in the future.

Also update probe.gl to latest (there was a bug preventing the reading of frame times).


Pessimistress:
This is great!

Just a generic comment - shall we make this a pure JS app since React can pollute the profiler? You are already breaking the pattern by directly calling `deck.setProp`.

tsherif:
Yeah, I can definitely do that.

1chandu:
Nit: we have several random data generator utility methods in `test` with minor differences in data format, at some point we should consolidate all these :
https://github.com/uber/deck.gl/blob/0fec14d85975c5d74c96f698ca3e77d05c19de79/test/apps/data-filter/src/data-sample.js#L2,  https://github.com/uber/deck.gl/blob/0fec14d85975c5d74c96f698ca3e77d05c19de79/test/data/grid-aggregation-data.js#L147

test/apps/stress-tests/webpack.config.js
```diff
@@ -0,0 +1,26 @@
...
+const webpack = require('webpack');
+
+const CONFIG = {
+  mode: 'development',
+
+  entry: {
+    app: './app.js'
+  },
+
+  resolve: {
```
Pessimistress: I think this legacy config is no longer needed.

tsherif:
I also added the pico-mercator version of the stress test to make it easy to compare.

PR #3687, opened 2019-09-24
Make gallery options panel wider

ajduberstein:
For #3591

#### Change List
-


ajduberstein:
Old:
image


New:
image

Issue #3640, opened 2019-09-18
Tracker: pydeck Improvements


ajduberstein:
#### To Do List 
- [x] Add additional examples - p0
- [ ] Kepler.gl + pydeck integration - p2
- [ ] Control panel to allow users to configure layers with a GUI - p2
- [ ] Support blending and constants in Deck objects, e.g., [in this line](https://github.com/uber/deck.gl/blob/7.2-release/examples/website/line/app.js#L111) - p2
- [ ] Arrow support - p2
- [ ] Provide hook to track test coverage - p2
- [ ] Better error handling. Edge cases exist where layers do not render because of incorrect layer keywords
- [ ] Multiselect via rectangle/lasso select/clicking individual objects. Pass data back to pydeck via this method. Some support for this already via `deck_widget.selected_data`
- [ ] Visual feedback for region selection. Highlight the data selected.
- [ ] Include Python3 typing

Likely better pushed into `@deck.gl/json`:
- Incorporating legends.
- Incorporating scales.
- Incorporating time sliders.

#### Backlog
- [ ] Support for multiple MapViews (as in the [Playground minimap example](https://deck.gl/playground/))
- [ ] Support for time-based update on the client

## Release checklist
- [ ] **pydeck 0.5.0 release**

Tracked in #4823

- [x] **pydeck 0.4.0 release**

Tracked in #4323

- [x] **pydeck 0.3.0 release**

Tracked in #4103 
- [x] **pydeck 0.2.0 release**
    - [x] Publish standalone-html bundle from @deck.gl/jupyter-widget and change `deckgl_jupyter_widget_bundle` in `pydeck/io/html.py` to the URL for this bundle.
    - [x] Publish JupyterLab bundle and verify the JupyterLab installation
    - [x] Verify that the Jupyter Notebook examples all work locally
    - [x] Publish pydeck to PyPI
    - [x] Update the `binder` branch of deck.gl to install the latest pydeck and verify that the examples now work
   - [x] Add screen diff tests - p0, addressed in #3964
    - [x] Add Jupyterlab support - p1, addressed in #3638
    - [x] Simplify setup.py - p2, addressed in #3638
    - [x] Support 3D tiles - p3
    - [x] Add multiselect functionality–let a user select multiple data points - p1 - addressed in #3786 
    - [X] Use Sphinx to generate .rst files for hosting on readthedocs.org - p0, addressed in #3698 and #3710
[...]

PR #3700, opened 2019-09-26
Add more explicit warning and data selection to pydeck

ajduberstein:
For #3640

#### Change List
-


ajduberstein:
Will make this modification in a future version of pydeck

PR #3707, opened 2019-09-26
Ignore keys with undefined values in the jupyter-widget custom tooltip

ajduberstein:
For #

#### Change List
-

### DIFF
--- a/modules/core/src/lib/tooltip.js
+++ b/modules/core/src/lib/tooltip.js
@@ -52,10 +52,10 @@ export default class Tooltip {
       el.style.display = 'none';
       return;
     } else {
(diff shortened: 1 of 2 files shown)


modules/core/src/lib/tooltip.js
```diff
@@ -52,10 +52,10 @@ export default class Tooltip {
       el.style.display = 'none';
       return;
     } else {
-      if ('text' in displayInfo) {
+      if ('text' in displayInfo && displayInfo.text !== undefined) {
```
Pessimistress: `displayInfo.text !== undefined` sufficiently covers `'text' in displayInfo`.

PR #3708, opened 2019-09-26
Document layer.clone()

Pessimistress:


docs/api-reference/layer.md
```diff
@@ -476,6 +476,17 @@ The state object allows a layer to store persistent information cross rendering
 
 ### General Methods
 
+##### `clone`
+
+```js
+const updatedLayer = layer.clone(overrideProps);
+```
+
+Create a copy of this layer, optionally change some prop values.
```
xintongxia: what is the cost of this call? Is it expensive?

Pessimistress: Should be similar to calling `new Layer({...})`

PR #3709, opened 2019-09-26
Fix inlined version in published package

Pessimistress:
This was broken when lerna stopped bumping root package version.


xintongxia:
Should we do it for all our other libs?

Pessimistress:
> Should we do it for all our other libs?

Yes.

Issue #3554, opened 2019-09-10
heatmap layer shows only one color from colorRange


UttamRahane:
#### Repro Steps
const heatMap_layer = new HeatmapLayer({
                data: this.geoJson.features,
                id: 'heatmp-layer',
                opacity: 0.6,
                getPosition: d => d.geometry.coordinates,
                getWeight: d => d.properties['Max Capacity'],
               colorFormat: 'RGBA',
                colorRange: [[1, 152, 189, 255], [73, 227, 206, 255], [216, 254, 181, 255], [254, 237, 177, 255],[254, 173, 84, 255],[209, 55, 78, 255]],
                intensity: 1,
                radiusPixels: 80,
                threshold: 0.9
            });

 -->

#### Environment (please complete the following information):
 - Framework Version: [e.g. deck.gl 7.2.3]
 - Browser Version: [e.g. Chrome 71.0]
 - OS: [e.g. Mac OS X 10.4]

1chandu:
Can you share result image? Which of the colors in colorRange are used? Also try adjusting `radiusPixels` and using default `colorRange`.

UttamRahane:
![image](https://user-images.githubusercontent.com/21005702/64690401-ecdf5200-d4ad-11e9-842c-a9c2cae0e0fe.png)

UttamRahane:
expected result should be something like -
![image](https://user-images.githubusercontent.com/21005702/64690866-0f259f80-d4af-11e9-8ad5-dd46b50daa0e.png)

UttamRahane:
expected result should be something like -

![image](https://user-images.githubusercontent.com/21005702/64690929-37150300-d4af-11e9-9f8a-907b436d928e.png)

UttamRahane:
using default colorRange, it shows only yellow color. that means it shows only first color.

1chandu:
Based on your image it seems all the pixels are falling into lower end of heat map range. Locally I am not able to reproduce this problem, can you share your system details? Browser, GPU, OS (complete the environment section in description) ?  Also can you minimize the issue to few data samples and share that data here?

SterlingMcCall:
I am seeing this same issue with deck.gl v7.2.3 in Windows. It happens in both Chrome and Edge. It works as expected in Chrome on linux.

UttamRahane:
Browser : Chrome Version 76.0.3809.132
GPU: Intel(R) UHD Graphics 620
OS: windows 10
[...]

PR #3714, opened 2019-09-27
HeatmapLayer: Add 'colorDomain' prop

1chandu:
For #3591 & #3554 

#### Background
- Add `weightDomain` prop to specify custom domain range, this is a required as workaround due to ANGLE issues (#3554)
- Docs:
  - Remove `Experimental`
  - Remove restriction of `WebGL2`
  - Document new prop.
  - Update whats-new

heatmapWeightDomain2Resized

#### Change List
-

### DIFF
--- a/docs/layers/heatmap-layer.md
+++ b/docs/layers/heatmap-layer.md
@@ -9,8 +9,6 @@
 
 `HeatmapLayer` can be used to visualize spatial distribution of data. It internally implements [Gaussian Kernel Density Estimation](https://en.wikipedia.org/wiki/Kernel_(statistics%29#Kernel_functions_in_common_use) to render heatmaps.
 
-* NOTE: Current version of this layer is supported only for WebGL2 enabled browsers, support for WebGL1 browsers will be added in future releases.
-
 ```js
 import DeckGL from '@deck.gl/react';
 import {HeatmapLayer} from '@deck.gl/aggregation-layers';
--- a/docs/layers/heatmap-layer.md
+++ b/docs/layers/heatmap-layer.md
@@ -5,7 +5,7 @@
 </p>
 
 
-# HeatmapLayer **Experimental**
+# HeatmapLayer
 
 `HeatmapLayer` can be used to visualize spatial distribution of data. It internally implements [Gaussian Kernel Density Estimation](https://en.wikipedia.org/wiki/Kernel_(statistics%29#Kernel_functions_in_common_use) to render heatmaps.
 
@@ -97,6 +97,17 @@ Value that is multiplied with the total weight at a pixel to obtain the final we
 The `HeatmapLayer` reduces the opacity of the pixels with relatively low weight to create a fading effect at the edge. A larger `threshold` smoothens the boundaries of color blobs, while making pixels with low relative weight harder to spot (due to low alpha value).
 
 `threshold` is defined as the ratio of the fading weight to the max weight, between `0` and `1`. For example, `0.1` affects all pixels with weight under 10% of the max.
+
+##### `weightDomain` (Array, optional) ![transition-enabled](https://img.shields.io/badge/transition-enabled-green.svg?style=flat-square")
+
+* Default: `[0, 0]`
+
+Weight of each data object is distributed to to all the pixels in a circle centered at the object position, weight a pixel receives is inversely proportional to its distance from the center. Pixels that fall into multiple circles will have sum of all weights. And the weight of the pixel determines its color. Using `weightDomain`, any pixels that are out of the specified range are not rendered. When `weightDomain[1]` is `0`, `HeatmapLayer` automatically calculates the maximum weight, which is the default behavior.
+
+NOTES:
+- It is recommend to use default value for regular use cases.
+- Pixel weights change based on the zoom level, so to keep the rendering same, this range should be increased (when zoomed out) and decreased (when zoomed in) when zoom is changed.
+- On `Windows` browsers due to an ANGLE [issue](https://github.com/uber/deck.gl/issues/3554), auto calculation of maximum weight doesn't work, hence on `Windows`, `weightDomain` should be used with a non zero maximum value.
 
 ### Data Accessors
 
--- a/docs/layers/heatmap-layer.md
+++ b/docs/layers/heatmap-layer.md
@@ -102,7 +102,7 @@ The `HeatmapLayer` reduces the opacity of the pixels with relatively low weight 
 
 * Default: `null`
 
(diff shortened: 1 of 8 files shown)


1chandu:
@Pessimistress , I can remove changes to website/* and example/* , as this new prop is only used in special cases.
Also should I make this prop experimental, two reasons, 1. Once `ANGLE` bug is fixed we might not need it, 2. The `domain` specified is zoom specific, when zoom changed, range might not be effective.

Pessimistress:
- I think the same prop is called `colorDomain` in the grid/hexagon layers and default to `null`.
- theoretically you can adjust the domain by scale (already as uniform of the project module) in the shader.
- I wish we didn't have to expose this in the website example. Is there a way to detect the affected platforms from JavaScript?

1chandu:
> I wish we didn't have to expose this in the website example. Is there a way to detect the affected platforms from JavaScript?

So fare we know it is affecting Windows and iOS, we can detect OS like :
https://stackoverflow.com/questions/38241480/detect-macos-ios-windows-android-and-linux-os-with-js

1chandu:
Here is my proposal:
1. make `weightDomain` experimental and rename it to `_weightDomain`, we can remove it or re-purpose it once ANGLE issue fixed.
2. Remove all website demo / example changes to specify `weightMap`
3. I ok with `colorDomain`, but making it function like other layers `colorDomain` prop would clip off higher range of pixels, this may not be a desired behavior
heatmapWeightDomainClipResized2

examples/website/heatmap/app.js
```diff
@@ -17,10 +17,17 @@ const INITIAL_VIEW_STATE = {
...
 
 export class App extends PureComponent {
   _renderLayers() {
-    const {data = DATA_URL, intensity = 1, threshold = 0.03, radiusPixels = 30} = this.props;
+    const {
+      data = DATA_URL,
+      intensity = 1,
+      threshold = 0.03,
+      radiusPixels = 30,
+      colorDomain = COLOR_DOMAIN
```
Pessimistress: You shouldn't need any fallback here.

modules/aggregation-layers/src/heatmap-layer/heatmap-layer.js
```diff
@@ -57,14 +57,16 @@ const TEXTURE_OPTIONS = {
...
 
 const defaultProps = {
   getPosition: {type: 'accessor', value: x => x.position},
   getWeight: {type: 'accessor', value: 1},
   intensity: {type: 'number', min: 0, value: 1},
   radiusPixels: {type: 'number', min: 1, max: 100, value: 30},
   colorRange: defaultColorRange,
-  threshold: {type: 'number', min: 0, max: 1, value: 0.05}
+  threshold: {type: 'number', min: 0, max: 1, value: 0.05},
+  colorDomain: DEFAULT_COLOR_DOMAIN
```
Pessimistress: `null`

modules/aggregation-layers/src/heatmap-layer/heatmap-layer.js
```diff
@@ -118,8 +122,15 @@ export default class HeatmapLayer extends CompositeLayer {
       this._updateTextureRenderingBounds();
     }
 
-    this.setState({zoom: opts.context.viewport.zoom});
+    if (oldProps.colorDomain !== props.colorDomain || changeFlags.viewportChanged) {
+      const {viewport} = this.context;
+      const domainScale = viewport ? 1024 / viewport.scale : 1;
+      newState.colorDomain = props.colorDomain.map(x => x * domainScale);
```
Pessimistress: - Check `props.colorDomain` and fall back to `DEFAULT_COLOR_DOMAIN` here
- Just directly call `setState` here. Creating and merging the `newState` object is more expensive.

website/src/components/demos/heatmap-demo.js
```diff
@@ -15,7 +15,23 @@ export default class HeatmapDemo extends Component {
     return {
       radius: {displayName: 'Radius', type: 'range', value: 5, step: 1, min: 1, max: 50},
       intensity: {displayName: 'Intensity', type: 'range', value: 1, step: 0.1, min: 0, max: 5},
-      threshold: {displayName: 'Threshold', type: 'range', value: 0.03, step: 0.01, min: 0, max: 1}
+      threshold: {displayName: 'Threshold', type: 'range', value: 0.03, step: 0.01, min: 0, max: 1},
+      minWeight: {
```
Pessimistress: Let's make change to the website demo in a separate PR.

docs/layers/heatmap-layer.md
```diff
@@ -98,6 +96,20 @@ The `HeatmapLayer` reduces the opacity of the pixels with relatively low weight
 
 `threshold` is defined as the ratio of the fading weight to the max weight, between `0` and `1`. For example, `0.1` affects all pixels with weight under 10% of the max.
 
+`threshold` is ignored when `colorDomain` is specified.
+
+##### `colorDomain` (Array, optional) ![transition-enabled](https://img.shields.io/badge/transition-enabled-green.svg?style=flat-square")
+
+* Default: `null`
+
+Weight of each data object is distributed to to all the pixels in a circle centered at the object position, weight a pixel receives is inversely proportional to its distance from the center. Pixels that fall into multiple circles will have sum of all weights. And the weight of the pixel determines its color. When `colorDomain` is specified, all pixels weights are clamped to this range before mapping a color to each pixel.
```
Pessimistress: Explain that weights smaller than `colorDomain[0]` will be faded out.

docs/layers/heatmap-layer.md
```diff
@@ -98,6 +96,20 @@ The `HeatmapLayer` reduces the opacity of the pixels with relatively low weight
...
 
+`threshold` is ignored when `colorDomain` is specified.
+
+##### `colorDomain` (Array, optional) ![transition-enabled](https://img.shields.io/badge/transition-enabled-green.svg?style=flat-square")
+
+* Default: `null`
+
+Weight of each data object is distributed to to all the pixels in a circle centered at the object position, weight a pixel receives is inversely proportional to its distance from the center. Pixels that fall into multiple circles will have sum of all weights. And the weight of the pixel determines its color. When `colorDomain` is specified, all pixels with weight with in specified `colorDomain` will get mapped to `colorRange`, pixels with weight less than `colorDomain[0]` will fade out (reduced alpha) and pixels with weight more than `colorDomain[1]` will mapped highest color in `colorRange`.
+
+When not specified, `maximum weight` is auto calculated and domain will be set to [`0`, `maximum weight`].
```
Pessimistress: `maxWeight * threshold, maxWeight`

PR #3718, opened 2019-09-27
HeatmapLayer Demo: Conditionally add colorDomain controls

1chandu:
For #

#### Background
Custom color domain is needed due to an ANGLE [issue](https://bugs.chromium.org/p/angleproject/issues/detail?id=3941) for Windows and iOS platforms.

#### Change List
- HeatmapLayer Demo: Conditionally add colorDomain controls

### DIFF
--- a/website/src/components/demos/heatmap-demo.js
+++ b/website/src/components/demos/heatmap-demo.js
@@ -14,32 +14,11 @@ export default class HeatmapDemo extends Component {
   }
 
   static get parameters() {
-    const parameters = {
+    return {
       radius: {displayName: 'Radius', type: 'range', value: 5, step: 1, min: 1, max: 50},
       intensity: {displayName: 'Intensity', type: 'range', value: 1, step: 0.1, min: 0, max: 5},
       threshold: {displayName: 'Threshold', type: 'range', value: 0.03, step: 0.01, min: 0, max: 1}
     };
-    if (shouldEnableColorDomain()) {
-      Object.assign(parameters, {
-        minWeight: {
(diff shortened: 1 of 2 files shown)


website/src/components/demos/heatmap-demo.js
```diff
@@ -84,3 +94,11 @@ export default class HeatmapDemo extends Component {
     );
   }
 }
+
+// HELPER
+function shouldEnableColorDomain() {
+  const OS_NAMES_TO_DETECT = ['windows', 'ios'];
```
Pessimistress: I believe it's `Windows` and `iOS`

website/src/components/demos/heatmap-demo.js
```diff
@@ -84,3 +94,11 @@ export default class HeatmapDemo extends Component {
...
+  const browser = Bowser.getParser(window.navigator.userAgent);
```
Pessimistress: `import {Parser} from 'bowser'` to avoid bundling additional content

website/src/components/demos/heatmap-demo.js
```diff
@@ -84,3 +94,11 @@ export default class HeatmapDemo extends Component {
...
+  const currentOS = browser.getOSName(true);
+  return OS_NAMES_TO_DETECT.indexOf(currentOS) !== -1;
```
Pessimistress: [array.includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes)

PR #3730, opened 2019-10-02
TileLayer cache rendered sub layers

Pessimistress:
See discussion in https://github.com/uber/deck.gl/pull/3704

#### Change List
- Avoid excessive re-render during continuous viewport updates

### DIFF
--- a/modules/geo-layers/src/tile-layer/utils/tile-cache.js
+++ b/modules/geo-layers/src/tile-layer/utils/tile-cache.js
@@ -11,21 +11,22 @@ export default class TileCache {
    * Takes in a function that returns tile data, a cache size, and a max and a min zoom level.
    * Cache size defaults to 5 * number of tiles in the current viewport
    */
-  constructor({getTileData, maxSize, maxZoom, minZoom, onTileError}) {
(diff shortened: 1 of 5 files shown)


Pessimistress:
@jesusbotella

modules/geo-layers/src/tile-layer/utils/tile-cache.js
```diff
@@ -11,23 +11,29 @@ export default class TileCache {
...
+    this._tiles = [];
 
-    if (maxZoom && parseInt(maxZoom, 10) === maxZoom) {
-      this._maxZoom = maxZoom;
+    if (Number.isFinite(maxZoom)) {
+      this._maxZoom = Math.floor(maxZoom);
     }
-    if (minZoom && parseInt(minZoom, 10) === minZoom) {
-      this._minZoom = minZoom;
+    if (Number.isFinite(maxZoom)) {
```
xintongxia: maxZoom -> minZoom

jesusbotella:
I just wanted to say thank you! That was fast! 😀

jererobles:
Is the cache configurable and if so should we add document it somewhere?

PR #3732, opened 2019-10-03
HeatmapLyaer: set point size to address ANGLE bug

1chandu:
possible fix for #3554 
More details: https://bugs.chromium.org/p/angleproject/issues/detail?id=3941

#### Change List
-

### DIFF
--- a/modules/aggregation-layers/src/heatmap-layer/max-vs.glsl.js
+++ b/modules/aggregation-layers/src/heatmap-layer/max-vs.glsl.js
@@ -6,5 +6,7 @@ void main()
 {
 outTexture = inTexture;
 gl_Position = vec4(0, 0, 0, 1.);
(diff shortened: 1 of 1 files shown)


modules/aggregation-layers/src/heatmap-layer/max-vs.glsl.js
```diff
@@ -6,5 +6,7 @@ void main()
 {
 outTexture = inTexture;
 gl_Position = vec4(0, 0, 0, 1.);
+// Enforce default value for ANGLE issue
```
Pessimistress: Link to https://bugs.chromium.org/p/angleproject/issues/detail?id=3236

1chandu: added issue that I opened, which links to original issue too.

PR #3734, opened 2019-10-03
Use float texture for shadowmap

Pessimistress:


modules/core/src/passes/shadow-pass.js
```diff
@@ -1,52 +1,43 @@
...
-      id: 'shadowmap',
-      width: 1,
-      height: 1,
-      attachments: {
-        [gl.COLOR_ATTACHMENT0]: this.shadowMap,
-        // Depth attachment has to be specified for depth test to work
-        [gl.DEPTH_ATTACHMENT]: this.depthBuffer
-      }
-    });
+      this.fbo = new Framebuffer(gl, {id: 'shadowmap'});
```
1chandu: Nit: remove fbo.attach call by passing `attachments: {[GL.CO...} object to constructor.

modules/core/src/passes/shadow-pass.js
```diff
@@ -1,52 +1,43 @@
...
-        [gl.TEXTURE_WRAP_T]: gl.CLAMP_TO_EDGE
-      }
-    });
-
-    this.depthBuffer = new Renderbuffer(gl, {
-      format: gl.DEPTH_COMPONENT16,
-      width: 1,
-      height: 1
-    });
+    if (Framebuffer.isSupported(gl, {colorBufferFloat: true})) {
```
1chandu: This will fail on iOS, don't we need RGBA/UNSIGNED_BYTE as a backup?

modules/core/src/passes/shadow-pass.js
Pessimistress: If I supply `attachments` then no depth attachment will be automatically created.

PR #3739, opened 2019-10-03
Add `pass` parameter to layerFilter

Pessimistress:
#### Background

Currently, the `pickable` prop is tied to the following features:

- Callbacks such as `onHover` and `onClick`
- Auto highlighting
- Queries such as `pickObjects`

This makes it difficult for applications that need different behaviors in auto-picking on hover (e.g. tooltips, auto-highlight) vs. multi-depth picking on click (e.g. map editing).

This PR introduces an additional `renderPass` field to the `layerFilter` parameters so that applications can control whether a layer is rendered based on which render pass it's in.

Related: https://github.com/uber/streetscape.gl/pull/396

#### Change List
- Add picking mode to `pass` string
- Add `pass` to `layerFilter` parameters

### DIFF
--- a/docs/api-reference/deck.md
+++ b/docs/api-reference/deck.md
@@ -75,8 +75,12 @@ Arguments:
 
 - `layer` (Layer) - the layer to be drawn
 - `viewport` (Viewport) - the current viewport
-- `isPicking` (Boolean) - whether this is a picking pass (drawing to offscreen picking buffer)
-- `pass` (String) - the name of the current render pass
+- `isPicking` (Boolean) - whether this is a picking pass
+- `renderPass` (String) - the name of the current render pass. Some standard passes are:
+  + `'screen'` - drawing to screen
+  + `'picking:hover'` - drawing to offscreen picking buffer due to pointer move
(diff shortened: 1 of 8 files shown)


docs/api-reference/deck.md
```diff
@@ -49,7 +49,38 @@ The array of deck.gl layers to be rendered. This array is expected to be an arra
...
+
+- `layerFilter` does not override the visibility defined by the layer's `visible` and `pickable` props.
+- All the lifecycle methods are still triggered even a if a layer is filtered out using this prop.
+
+Arguments:
+
+- `layer` (Layer) - the layer to be drawn
+- `viewport` (Viewport) - the current viewport
+- `isPicking` (Boolean) - whether this is a picking pass (drawing to offscreen picking buffer)
+- `pass` (String) - the name of the current render pass
```
xintongxia: Is `pass` more like an enum? or call it `renderPass`?

Pessimistress: Sort of. We can list some of the standard pass names, but extensions/effects can add their own.

PR #3749, opened 2019-10-04
Fix 3d picking projection with camera offset

Pessimistress:
#### Change List
- When the camera is not pointing to the ground, common space z must compensate for the offset.


xintongxia:
move the comment to code.

PR #3769, opened 2019-10-11
Update wboit test app: add sample polygons

georgios-uber:


test/apps/wboit/app.js
```diff
@@ -1,34 +1,189 @@
...
+    };
+
+    const layers = [];
+    if (wboit) {
+      layers.push(new WBOITLayer({id: 'WBOITLayer', ...options}));
+    } else {
+      layers.push(new SolidPolygonLayer({id: 'SolidPolygonLayer', ...options}));
+    }
+
+    const mkButton = (label, name, value) => (
```
tsherif: This could be a react component as well.

georgios-uber: In that case, I will need to have a global state, making things more complex.

PR #3785, opened 2019-10-14
Optionally surface pydeck warnings in widget UI

ajduberstein:
For #3640

#### Change List
- Add warning banner to pydeck
- Wrap console.warn to catch displayed warnings

### DIFF
--- a/modules/jupyter-widget/src/utils.js
+++ b/modules/jupyter-widget/src/utils.js
@@ -66,6 +66,9 @@ export function initDeck({
         });
 
         const warn = deck.log.warn;
+        // TODO overrride console.warn instead
+        // Right now this isn't doable (in a Notebook at least)
+        // because the widget loads in deck.gl (and its logger) before @deck.gl/jupyter-widget
         deck.log.warn = injectFunction(warn, handleWarning);
         if (onComplete) {
           onComplete({jsonConverter, deckgl});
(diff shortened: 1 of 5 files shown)


modules/jupyter-widget/src/utils.js
```diff
@@ -55,6 +63,8 @@ export function initDeck({mapboxApiKey, container, jsonInput, tooltip, onComplet
           getTooltip,
           container
         });
+        const warn = deck.log.warn;
+        deck.log.warn = wrapWarn(warn, handleWarning);
```
Pessimistress: Wrap `console.warn` instead of `deck.log.warn`. `log.warn` also dedupes messages.

Pessimistress: No support for piping console error messages?

modules/jupyter-widget/src/utils.js
```diff
@@ -67,3 +79,10 @@ export function initDeck({mapboxApiKey, container, jsonInput, tooltip, onComplet
     });
   });
 }
+
+function wrapWarn(warnFunction, messageHandler) {
+  return (message, err) => {
```
Pessimistress: These argument names do not reflect what is passed to `log.warn`. Use `...args`.

modules/jupyter-widget/src/utils.js
```diff
@@ -67,3 +79,10 @@ export function initDeck({mapboxApiKey, container, jsonInput, tooltip, onComplet
     });
   });
 }
+
+function wrapWarn(warnFunction, messageHandler) {
```
Pessimistress: Call it `wrapFunction` or `injectFunction`. There is no logic specific to `warn`.

PR #3786, opened 2019-10-14
Add multiselect for data EDA to pydeck

ajduberstein:
For #3640

#### Change List
- Support single-click object selection
- Support clearing current selection
- Support meta/alt key + click to select multiple objects and push them
back to a data frame


ajduberstein:
TwoWayAPI

PR #3788, opened 2019-10-15
Invalidate cache when updateTriggers changed

xintongxia:
For https://github.com/uber/deck.gl/issues/3776

### DIFF
--- a/modules/geo-layers/src/tile-layer/tile-layer.js
+++ b/modules/geo-layers/src/tile-layer/tile-layer.js
@@ -29,9 +29,7 @@ export default class TileLayer extends CompositeLayer {
   updateState({props, oldProps, context, changeFlags}) {
     let {tileCache} = this.state;
     if (
-      !tileCache ||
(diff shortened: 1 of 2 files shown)


modules/geo-layers/src/tile-layer/tile-layer.js
```diff
@@ -28,11 +28,7 @@ export default class TileLayer extends CompositeLayer {
 
   updateState({props, oldProps, context, changeFlags}) {
     let {tileCache} = this.state;
-    if (
-      !tileCache ||
-      (changeFlags.updateTriggersChanged &&
-        (changeFlags.updateTriggersChanged.all || changeFlags.updateTriggersChanged.getTileData))
-    ) {
+    if (!tileCache || changeFlags.updateTriggersChanged) {
```
Pessimistress: @xintongxia This is unnecessarily re-downloading all the data when any update trigger changes. To invalid the cached layer instance, you should just set `tile.layer` to `null`.

PR #3791, opened 2019-10-16
[Refactor tooltip test] Add DOM testing utility

ajduberstein:
Supports increasing the test coverage on @deck.gl/jupyter-widget, where
many components of the module assume a global document and a canvas have
already been created.

See #3792 for usage

#### Change List
- Create a helper class for document testing


test/utils/document.js
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  _createDocument() {
+    // Creates a global document object via JSDOM or a deep clone of the browser one if available
+    if (typeof window === undefined || !window.document) {
+      const {JSDOM} = require('jsdom');
+      const dom = new JSDOM(`<!DOCTYPE html>`);
+      return dom.window.document;
+    }
+    const documentClone = window.document.cloneNode(true);
```
Pessimistress: Why do you need to clone the document? It may have heavy content during browser test.

test/utils/document.js
```diff
@@ -0,0 +1,41 @@
+/* global window */
+
+export default class DocumentTest {
```
Pessimistress: Add JSDoc for this class.

test/utils/document.js
```diff
@@ -0,0 +1,41 @@
...
+    canvasParent.className = 'canvas-parent';
+    canvasParent.appendChild(canvas);
+    this.canvasParent = canvasParent;
+
+    this.document.body.appendChild(canvasParent);
+    this.testObject = canvasCallback(canvas);
+  }
+
+  finalize() {
+    this.testObject.remove();
```
Pessimistress: This class has no control over the returned value of the user callback. Assuming `testObject` is a DOM element here is dangerous. There is no apparent reason why `testObject` has to be stored in this class. It will be a lot easier to read if you write your test like this

```js
const testCanvas = new TestCanvas();
const myElement = setup(testCanvas.canvas);
// ...
myElement.remove();
testCanvas.finalize();
```

test/utils/document.js
Pessimistress: The class name `DocumentTest` does not communicate what it does, e.g. the creation of a canvas. When you write tests using this class, it will be difficult for other maintainers to understand a) what the callback receives b) what the callback is supposed to return, from the code `new DocumentTest(...)`.

test/utils/document.js
ajduberstein: Just updated the PR, the class has been cut in favor of smaller functions and an API that looks more like the one here.

test/utils/canvas-test.js
```diff
@@ -0,0 +1,38 @@
+/* global window */
+
+/** Create a clean document object for testing */
+function createDocument() {
+  // Creates a global document object via JSDOM or a shallow clone of the browser's document
+  let documentClone;
+  if (typeof window === undefined || !window.document) {
+    const {JSDOM} = require('jsdom');
+    const dom = new JSDOM(`<!DOCTYPE html>`);
+    documentClone = dom.window.document.cloneNode(false);
```
Pessimistress: This node is brand new - why clone it?

test/utils/canvas-test.js
```diff
@@ -0,0 +1,38 @@
...
+  } else {
+    documentClone = window.document.cloneNode(false);
+  }
+  const html = window.document.createElement('html');
+  const body = window.document.createElement('body');
+  html.appendChild(body);
+  documentClone.appendChild(html);
+  documentClone.body = body;
```
Pessimistress: What is the purpose of this blob? Why don't you just use `document.body`?

test/utils/canvas-test.js
```diff
@@ -0,0 +1,38 @@
...
+/** Create a clean document object for testing */
+function createDocument() {
+  // Creates a global document object via JSDOM or a shallow clone of the browser's document
+  let documentClone;
+  if (typeof window === undefined || !window.document) {
+    const {JSDOM} = require('jsdom');
+    const dom = new JSDOM(`<!DOCTYPE html>`);
+    documentClone = dom.window.document.cloneNode(false);
+  } else {
+    documentClone = window.document.cloneNode(false);
```
Pessimistress: Is there a reason you cannot attach test elements to the current document?

test/utils/canvas-test.js
```diff
@@ -0,0 +1,38 @@
...
+  const html = window.document.createElement('html');
+  const body = window.document.createElement('body');
+  html.appendChild(body);
+  documentClone.appendChild(html);
+  documentClone.body = body;
+  return documentClone;
+}
+
+/* Create a clean canvas, its parent, and a document node for testing */
+export default function setup() {
```
Pessimistress: `setupTestCanvas` for clarity

ajduberstein:
Closed in favor of  #3792

PR #3792, opened 2019-10-16
[Refactor tooltip test] Refactor @deck.gl/core tooltip test to use DocumentTest utility

ajduberstein:
#### Background
Supports increasing the test coverage on @deck.gl/jupyter-widget, where
many components of the module assume a global document and a canvas have
already been created.

#### Change List
- Refactor tooltip test to use DocumentTest class (see #3791)


test/modules/jupyter-widget/utils.spec.js
```diff
@@ -59,7 +59,9 @@ export class DummyManager extends widgets.ManagerBase {
 
   loadClass(className, moduleName, moduleVersion) {
     if (moduleName === '@jupyter-widgets/base') {
+      // eslint-disable-next-line
```
Pessimistress: Why is eslint disabled here?

PR #3800, opened 2019-10-17
[Refactor tooltip test] Add widget-tooltip tests

ajduberstein:
Increase jupyter-widget module test coverage by providing tests for the
widget's tooltip

#### Change List
-


test/modules/jupyter-widget/widget-tooltip.spec.js
```diff
@@ -0,0 +1,137 @@
...
+  return module;
+}
+
+test('getDefaultTooltip', t => {
+  const wt = safelyGetModule(t);
+  if (!wt) {
+    return;
+  }
+
+  wt.getDiv = () => {
```
Pessimistress: Why is this needed? All users of `getDiv` are referencing the original function.

test/modules/jupyter-widget/widget-tooltip.spec.js
```diff
@@ -0,0 +1,130 @@
...
+  } catch (err) {
+    t.end();
+  }
+  return module;
+}
+
+test('getDefaultTooltip', t => {
+  const wt = safelyGetModule(t);
+  if (!wt) {
+    return;
```
Pessimistress: - Add a `t.comment()` on why the test is skipped.
- You must call `t.end()` before `return`

test/modules/jupyter-widget/widget-tooltip.spec.js
```diff
@@ -0,0 +1,143 @@
...
+  }
+};
+
+function safelyGetModule(t) {
+  // Skip tests if in headless browser environment
+  let module;
+  try {
+    module = getModule(TOOLTIP_LIB);
+  } catch (err) {
+    return null;
```
Pessimistress: You should never use try...catch without logging and/or documentation. If there is a bug in your code that causes crash, your tests here will be useless in catching it.

- What exactly in the tooltip module breaks in the Node environment?
- Why do you need nested try...catch in this util and `getModule`? What error could be generated?

test/modules/jupyter-widget/widget-tooltip.spec.js
```diff
@@ -0,0 +1,143 @@
...
+  }
+  return module;
+}
+
+test('getDefaultTooltip', t => {
+  const widgetTooltipModule = safelyGetModule(t);
+  if (!widgetTooltipModule) {
+    t.comment('Widget tooltip module failed to load');
```
Pessimistress: This sounds like a very bad reason to pass tests.

PR #3823, opened 2019-10-30
TileLayer: only invalidate tile.layer when updateTriggerChanged

xintongxia:
https://github.com/uber/deck.gl/pull/3788 followup

### DIFF
--- a/modules/geo-layers/src/tile-layer/tile-layer.js
+++ b/modules/geo-layers/src/tile-layer/tile-layer.js
@@ -47,11 +47,10 @@ export default class TileLayer extends CompositeLayer {
       });
       this.setState({tileCache});
     } else if (
-      changeFlags.updateTriggersChanged &&
-      Object.keys(changeFlags.updateTriggersChanged).length
+      changeFlags.updateTriggersChanged
     ) {
       // if any updateTriggersChanged (other than getTileData), delete the layer
-      this.state.tileCache.tiles.map(tile => {
+      this.state.tileCache.tiles.forEach(tile => {
         tile.layer = null;
       });
     }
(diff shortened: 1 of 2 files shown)


modules/geo-layers/src/tile-layer/tile-layer.js
```diff
@@ -28,7 +28,10 @@ export default class TileLayer extends CompositeLayer {
 
   updateState({props, oldProps, context, changeFlags}) {
     let {tileCache} = this.state;
-    if (!tileCache || changeFlags.updateTriggersChanged) {
+    if (
+      !tileCache ||
+      (changeFlags.updateTriggersChanged.all || changeFlags.updateTriggersChanged.getTileData)
```
Pessimistress: `changeFlags.updateTriggersChanged` can be null. Add a test?

modules/geo-layers/src/tile-layer/tile-layer.js
```diff
@@ -42,7 +46,16 @@ export default class TileLayer extends CompositeLayer {
         onTileError: this._onTileError.bind(this)
       });
       this.setState({tileCache});
+    } else if (
+      changeFlags.updateTriggersChanged &&
+      Object.keys(changeFlags.updateTriggersChanged).length
```
Pessimistress: `updateTriggersChanged` is never an empty object.

modules/geo-layers/src/tile-layer/tile-layer.js
```diff
@@ -42,7 +46,16 @@ export default class TileLayer extends CompositeLayer {
...
+    ) {
+      // if any updateTriggersChanged (other than getTileData), delete the layer
+      this.state.tileCache.tiles.map(tile => {
```
Pessimistress: `forEach`

PR #3833, opened 2019-10-31
[JupyterLab support] Remove RequireJS from JS jupyter-widget notebook module

ajduberstein:
RequireJS makes our build process fairly complex. This PR removes RequireJS from the @deck.gl/jupyter-widget module. An additional set of PRs is still needed to address pydeck's expectation that RequireJS is present.

#### Change List
-


ajduberstein:
Doing this in a different PR, not stacked against the JupyterLab changes

Issue #3583, opened 2019-09-12
Integrate GPU Attribute filtering with GPUGridAggregator


1chandu:
Right now GPU Attribute filtering is performed by Layer system setting up required extra `attributes` and injecting required `shader modules`.  Aggregation layers , GridLayer (CPUGridLayer and GPUGridLayer), ContourLayer and ScreenGridLayer, use `GPUGridAggregator` but it handles attributes directly and doesn't use AttributeManger.

- [x] ScreenGridLayer
     - [x] Simplify the layer by merging WebGL1 and WebGL2 versions.

- [X] All aggregation layers must be composite layers

     - [x] Make `ScreenGridLayer` a composite layer which produces attribute `Buffer`s and provides them to sub layer (cell) which actually renders the cell.
     - [x] Verify other aggregation layers to make sure, Buffer attributes are not directly updated instead provided to sub layers.

- [x] New composite layers `AggregationLayer` and `GridAggregationLayer`.
    - [x] Prepare GPUGridAggregator and Utility methods to accept attributes from `AttributeManager`.
    - [x] Update Unit tests.
    - [x] Start with one layer (`GPUGridLayer`) and Integrate AttributeManager.
    - [x] Extract AttributeManager integration code into new composite super layers. 
    - [x] Make all aggregation layers extend this new Composite layer.

- [x] GPU Data Filtering (extension)
    - [x] GPUGridLayer
    - [x] ContourLayer
    - [x] ScreenGridLayer

- [ ] CPU Aggregation
    - [x] Write RFC for CPU Aggregation refactor/integration
    - [x] Remove CPU Aggregation code from GPUGridAggregator.
    - [x] Update CPU Aggregation to consume data from Attribute Manager. 
    - [x] Update all layer's CPU Aggregation interface
       - [x] HexagonLayer
       - [x] CPUGridLayer
       - [x] ScreenGridLayer (CPU fallback case WebGL1)
       - [ ] ContourLayer (CPU fallback case WebGL1)

- [ ] Data filtering for CPU Aggregation cases
   - [ ] Investigate kepler.gl data filtering and map it to deck.gl
   - [ ] Investigate and develop any fallbacks for cases that are not supported by deck.gl GPU Data filtering.

- [x] Cleanup FP64 aggregation flag

- [ ] Misc
    - [ ] Remove `getFloatTexture` usage with inline Textuer2D creation (see: #3681 
[...]

PR #3845, opened 2019-11-05
RFC: CPU Aggregation Refactor

1chandu:
For #3583 

#### Background

RFC for CPU Aggregation refactor to integrate with `AttributeManager` and support data filtering.

#### Change List

- RFC: CPU Aggregation Refactor


1chandu:
This feature is implemented in 8.0 already, merging.

PR #3848, opened 2019-11-06
Fix texture crash in Safari

xintongxia:
For https://github.com/uber/deck.gl/issues/3830


modules/layers/src/icon-layer/icon-manager.js
```diff
@@ -55,26 +55,19 @@ function buildRowMapping(mapping, columns, yOffset) {
...
-    parameters: DEFAULT_TEXTURE_PARAMETERS
+    height: oldHeight
   });
 
-  texture.generateMipmap();
-
-  oldPixels.delete();
-  return texture;
+  texture.delete();
+  return newTexture;
```
Pessimistress: How are you testing this? The return value is not used anywhere.

modules/layers/src/icon-layer/icon-manager.js
```diff
@@ -55,26 +55,19 @@ function buildRowMapping(mapping, columns, yOffset) {
 }
 
 // resize texture without losing original data
-function resizeTexture(texture, width, height) {
+function resizeTexture(gl, texture, width, height) {
```
1chandu: Nit: update comment to `create a new texture and copy original data`.

test/render/test-cases.js
```diff
@@ -584,12 +592,13 @@ export const TEST_CASES = [
...
       bearing: 0
     },
-    // rendering times
-    renderingTimes: 2,
     layers: [
       new IconLayer({
         id: 'icon-lnglat-auto',
         data: dataSamples.points,
+        updateTriggers: {
+          getIcon: () => getIconAccessor('icon-lnglat-auto')
```
Pessimistress: Just use `1`

Issue #3844, opened 2019-11-05
Extensions: Add support for querying whether a redraw is required.


1chandu:
When props that belong to an extension are changed, it results in new set of uniforms and layer models need to be re-drawn. This works fine since all layers are re-drawn when any prop is changed (`changeFlags.propsChanged` is true). But for for aggregation layers, `aggregation` is performed as seperate step in a `CompositeLayer` and re-draw only happens for sub-layers when a props are changed. We can force aggregation layers to re-aggregate when `changeFlags.propsChanged` is true, but that is un-necessary and expensive, aggregation should be performed only when any state/parameter is changed that affects aggregation. `DataFilterExtension` is one such case, when props of a `DataFilterExtension` are changed, aggregation should be re-run. To get more fine grained control over what props are changed, each `LayerExtension` should implement :
```
needsRedraw({oldProps, props, propTypes})
```
which returns true, if any prop corresponding to that extension is changed. Aggregation layers can call this method for required extension to check if they need to run the aggregation step. Default implementation of this method can just return `false`.

Relevant discussions in #3777 

- [ ] Implement `needsRedraw`
- [ ] Update `AggregationLayer` to use this new method
- [ ] Docs and unit tests.

PR #3850, opened 2019-11-06
AggregationLayer: add support for 'ignoreProps' for checking aggregation

1chandu:
For #3844 

#### Background
More details in #3844 

#### Change List
- LayerExtension: add needsRedraw()

### DIFF
--- a/modules/aggregation-layers/src/aggregation-layer.js
+++ b/modules/aggregation-layers/src/aggregation-layer.js
@@ -22,22 +22,10 @@ import {CompositeLayer, AttributeManager, experimental} from '@deck.gl/core';
 import {cssToDeviceRatio, log} from '@luma.gl/core';
 const {compareProps} = experimental;
 
-// props when changed results in new uniforms that requires re-aggregation
-const UNIFORM_PROPS = [
-  // DATA-FILTER extension
-  'filterEnabled',
-  'filterRange',
-  'filterSoftRange',
-  'filterTransformSize',
-  'filterTransformColor'
-];
-
 export default class AggregationLayer extends CompositeLayer {
   initializeState(aggregationProps = []) {
     super.initializeState();
-    this.setState({
-      aggregationProps: aggregationProps.concat(UNIFORM_PROPS)
-    });
+    this.setState({aggregationProps});
   }
 
   updateState(opts) {
@@ -94,9 +82,13 @@ export default class AggregationLayer extends CompositeLayer {
       oldProps[propName] = opts.oldProps[propName];
       props[propName] = opts.props[propName];
     }
-    return Boolean(
-      compareProps({oldProps, newProps: props, propTypes: this.constructor._propTypes})
-    );
+    if (compareProps({oldProps, newProps: props, propTypes: this.constructor._propTypes})) {
+      return true;
+    }
+
+    const {extensions} = opts.props;
+    const redrawOpts = Object.assign({}, opts, {propTypes: this.constructor._propTypes});
+    return extensions.some(extension => extension.needsRedraw(redrawOpts));
   }
 
   _updateShaders(shaders) {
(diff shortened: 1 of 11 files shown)


modules/core/src/lib/layer-extension.js
```diff
@@ -62,6 +63,21 @@ export class LayerExtension {
 
   initializeState(context, extension) {}
 
+  // Called with old and current props to determine if layer needs to redraw
+  needsRedraw(opts) {
+    const {redrawProps} = this;
```
Pessimistress: Can we just use the default props of the current extension? What is the use case where we need a separate list?

1chandu:
Based on discussion with @Pessimistress , I will update this PR with following :
Define a list of props (when changed don't require re-aggregation) and during update cycle, compare all props excluding above, to determine if re-aggregation is needed, this way we can avoid changes to the `LayerExtensions`.

modules/aggregation-layers/src/aggregation-layer.js
```diff
@@ -18,26 +18,14 @@
...
-  'filterEnabled',
-  'filterRange',
-  'filterSoftRange',
-  'filterTransformSize',
-  'filterTransformColor'
-];
-
 export default class AggregationLayer extends CompositeLayer {
-  initializeState(aggregationProps = []) {
+  initializeState(ignoreProps = {}) {
```
Pessimistress: You are filtering `defaultProps` in each sub layer to generate this argument. There are several potential issues:
- `defaultProps` may contain `[key]: 0` or `[key]: null` which will defeat the check in `compareProps`
- we may have multiple layers of inheritance in the future, and this process only accounts for the leaf class and the base Layer

If you instead pass the aggregation props array here, you can perform the filtering with `filterProps(this.constructor._propTypes, aggregationProps)`. `constructor._propTypes` is always normalized and merged up the inheritance chain.

1chandu:
This PR might change, based on how my CPU Aggregation refactor work going, will wait couple more days before updating this.

modules/aggregation-layers/src/aggregation-layer.js
```diff
@@ -87,16 +75,17 @@ export default class AggregationLayer extends CompositeLayer {
...
       return true;
     }
-    const {aggregationProps} = this.state;
-    const oldProps = {};
-    const props = {};
-    for (const propName of aggregationProps) {
-      oldProps[propName] = opts.oldProps[propName];
-      props[propName] = opts.props[propName];
+    if (
+      compareProps({
```
Pessimistress: Nit: `return compareProps(...)`

modules/aggregation-layers/src/utils/prop-utils.js
```diff
@@ -0,0 +1,28 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+export function filterProps(props, filterKeys) {
+  return Object.keys(props).reduce((accu, cur) => {
+    if (!filterKeys.includes(cur)) {
+      accu[cur] = props[cur];
```
Pessimistress: If `props[cur]` is falsy the key will not be ignored in `compareProps`

1chandu:
Closing in favor of a better solution : https://github.com/uber/deck.gl/pull/4008, will address remaining comments in 4008.

PR #3777, opened 2019-10-12
[WIP] Integrate all aggregation layers with AttributeManager

1chandu:
For #3583

#### Change List
-

PR #3851, opened 2019-11-06
AggregationLayer : minor fixes and unit tests.

1chandu:
For #3777 

#### Background
Minor fixes for crashes discovered in unit tests.
Add unit tests for new `AggregationLayer`.

#### Change List
- AggregationLayer : minor fixes and unit tests.

### DIFF
--- a/modules/aggregation-layers/src/screen-grid-layer/screen-grid-layer.js
+++ b/modules/aggregation-layers/src/screen-grid-layer/screen-grid-layer.js
@@ -75,7 +75,7 @@ export default class ScreenGridLayer extends GridAggregationLayer {
     super.updateState({oldProps, props, changeFlags});
 
     const cellSizeChanged = props.cellSizePixels !== oldProps.cellSizePixels;
-    const dataChanged = this._isAggregationDirty({oldProps, props});
+    const dataChanged = this._isAggregationDirty({oldProps, props, changeFlags});
 
     if (cellSizeChanged || changeFlags.viewportChanged) {
       this._updateGridParams();
--- a/modules/aggregation-layers/src/screen-grid-layer/screen-grid-layer.js
+++ b/modules/aggregation-layers/src/screen-grid-layer/screen-grid-layer.js
@@ -71,17 +71,17 @@ export default class ScreenGridLayer extends GridAggregationLayer {
     return this.state.supported && changeFlags.somethingChanged;
   }
 
-  updateState({oldProps, props, changeFlags}) {
-    super.updateState({oldProps, props, changeFlags});
+  updateState(opts) {
+    super.updateState(opts);
 
-    const cellSizeChanged = props.cellSizePixels !== oldProps.cellSizePixels;
-    const dataChanged = this._isAggregationDirty({oldProps, props, changeFlags});
+    const cellSizeChanged = opts.props.cellSizePixels !== opts.oldProps.cellSizePixels;
+    const dataChanged = this._isAggregationDirty(opts);
+    const {viewportChanged} = opts.changeFlags;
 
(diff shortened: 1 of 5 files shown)


modules/core/src/lib/composite-layer.js
```diff
@@ -191,7 +191,10 @@ export default class CompositeLayer extends Layer {
     for (const extension of extensions) {
       const passThroughProps = extension.getSubLayerProps.call(this, extension);
       Object.assign(newProps, passThroughProps, {
-        updateTriggers: Object.assign(newProps.updateTriggers, passThroughProps.updateTriggers)
+        updateTriggers: Object.assign(
```
1chandu: Without this change, new unit tests added will crash, `newProps.updateTriggers` was undefined or null.

modules/aggregation-layers/src/screen-grid-layer/screen-grid-layer.js
```diff
@@ -75,7 +75,7 @@ export default class ScreenGridLayer extends GridAggregationLayer {
     super.updateState({oldProps, props, changeFlags});
 
     const cellSizeChanged = props.cellSizePixels !== oldProps.cellSizePixels;
-    const dataChanged = this._isAggregationDirty({oldProps, props});
+    const dataChanged = this._isAggregationDirty({oldProps, props, changeFlags});
```
Pessimistress: Can we avoid exploding and reassembling the argument (`{oldProps, props, changeFlags}`) multiple times

modules/core/src/lib/composite-layer.js
Pessimistress: How is that possible with line 179?

1chandu: In my testcase, argument to `getSubLayerProps()` function, `sublayerProps` is undefined. Also it will only happen, if `extensions` is not an empty array, so I guess thats not being tested with exsisting unit tests of `CompositeLayer`.  One fix is, set a default argument (`getSubLayerProps(sublayerProps = {}) {...`)

modules/aggregation-layers/src/screen-grid-layer/screen-grid-layer.js
1chandu: Updated.

modules/core/src/lib/composite-layer.js
```diff
@@ -153,40 +153,40 @@ export default class CompositeLayer extends Layer {
...
-      if (overridingSublayerProps) {
-        const propTypes = this.constructor._propTypes;
-        for (const key in overridingSublayerProps) {
-          const propType = propTypes[key];
-          // eslint-disable-next-line
-          if (propType && propType.type === 'accessor') {
-            overridingSublayerProps[key] = this.getSubLayerAccessor(overridingSublayerProps[key]);
-          }
+    const overridingSublayerProps =
+      overridingProps && sublayerProps.id && overridingProps[sublayerProps.id];
```
Pessimistress: Nit: what is the scenario where `sublayerProps.id` is not specified but `overridingProps[sublayerProps.id]` is?

1chandu: Before this change, `sublayerProps` had no default value, so if  they are not provided, we never access `id` in `sublayerProps` object, but now we have default value, `{}`, so this check basically replaces old one.

Pessimistress: That is not the question. If `sublayerProps.id` is `undefined` then `overridingProps[undefined]` is not going to fail, so `overridingProps && overridingProps[sublayerProps.id]` should be sufficient.

1chandu: Got it,  updated.

Issue #3758, opened 2019-10-07
Apache Arrow support


1chandu:
This issue tracks remaining work for columnar table support in deck.gl

# deck.gl v8.0

### loaders.gl v2.0
- [Tracker: Columnar Table Support](https://github.com/uber-web/loaders.gl/issues/303) in loaders.gl v2

### GPUTable (@ibgreen/@1chandu)

- [ ] GPUTable documentation (@ibgreen)
- [ ] Arrow column->GPUTable attribute mappings (Arrow metadata semantics) documentation (@ibgreen)
- [ ] GPUTable initial implementation (@ibgreen)
- [ ] GPUTable full implementation (@1chandu)

### GPUTable transformations (@1chandu)

- [ ] GPU Table Transformer: Transform (combine etc) raw Apache Arrow columns in GPU tables into Layer-compatible Attribute tables (positions = vec3, colors = vec4s, float64 handling, matrices, ...)

      - Inputs: 
             - Schema defining Apache Arrow table.
             - One Buffer object for each Apache table column
      - Outputs:
            - Set of Buffers to match deck.gl Layer attributes.

     - [x] Make GPUTransform work for a single attribute (position column generated by longitude and latitude columns), write unit tests #3796 
     - [ ] Integrate GPUTable/GPUTransform with deck.gl and make single example work
        - [ ] Send GPUTable as `data` prop , and have AttributeManager/Layer extract attributes and set.
        - [ ] Use loaders.gl to generate column data, and use it to build GPUTable
             - [ ] Verify loaders.gl arrow loader with simple arrow data
             - [ ] Use csv loader to generate arrow data
        - [ ] Use 3D heatmap example to generate arrow data from csv file, build GPU table and pass it as data prop.
        - [ ] Move GPUTable and new column generation into deck.gl ??

1chandu:
Over to @Pessimistress , close it if this is covered by other issues.

ibgreen:
v10

PR #3852, opened 2019-11-06
Attribute class refactor

Pessimistress:
For #3758 

This PR refactors the `Attribute` class. Users of `Attribute` including `AttributeManager`, `AttributeTransitionManager`, `GPU*Transition` are only affected by API renaming.

#### Background

Currently, `Attribute` inherits from the `BaseAttribute` class, both providing luma.gl-compatible "attribute" interface (i.e. `getValue()`). 

The `BaseAttribute` class handles the following logic:

- luma.gl accessor options
- Creation of managed WebGL buffer
- Update buffer with data from typed array
- Set and normalize constant value

The `Attribute` class handles the following logic:

- deck.gl attribute options
- deck.gl update paths (external buffer, constant, updater)
- allocation of typed array
- data validation
- partial update
- double precision
- shader attributes

After this PR, there are 3 classes:

1. `DataColumn`: the owner of data (typed array and buffer)
  - luma.gl accessor options
  - Creation of managed WebGL buffer
  - Allocation of typed array
  - Update buffer from loaders.gl-compatible attribute descriptor
  - Set and normalize constant value
  - double precision
2. `ShaderAttribute`: a very light wrapper that "views" a `DataColumn` instance with custom accessor.
3. `Attribute`: inherits `DataColumn` and handles deck.gl-specific logic
  - deck.gl attribute options
  - deck.gl update paths (external buffer, constant, updater)
  - data validation
  - partial update
  - shader attributes

It is possible to use `DataColumn` independently from deck.gl. This prepares the code base for https://github.com/uber/deck.gl/pull/3798.

#### Breaking Changes
- `Attribute.state` is split into `Attribute.settings` (initial options) and `Attribute.state` (transient states)
- Accessor props `type`, `offset`, `stride` etc. are no longer surfaced on the top level. Use `Attribute.getAccessor()`.


Pessimistress:
@tsherif 
> Would it be possible for `Attribute` to work without being aware of buffers, etc?

The `Attribute` class (almost) depends on nothing luma.gl. It does inherit `DataColumn` which is WebGL heavy.

> `ShaderAttribute` feels unnecessary. Could this be implemented by just passing overrides when using a data column?

There still are various places in the code base that expect an "attribute-like" object from `getShaderAttributes()` (e.g. aggregation layers). @1chandu

PR #3857, opened 2019-11-08
Fix: custom controller events

Pessimistress:
For https://github.com/uber/deck.gl/issues/3822

#### Change List
- Make sure custom events are always registered


xintongxia:
Add a test?

PR #3868, opened 2019-11-11
Bump to luma 8.0

tsherif:
Main changes: 
- Update imports
- Materials defined as data objects
  - Material classes were removed in luma 8
- Define light classes fully. 
  - The deck classes subclassed luma classes which were removed. Full definition is now in deck.gl
- Rename `gouraudlighting`/`phonglighting` to `gouraudLighting`/`phongLighting` (fix case)

Perf improvement on the stress test is not as large as I was hoping (it was ~60% on the luma stress test), but I still haven't hit the state tracking which still shows up as the biggest bottleneck.

|         |Before|After|Improvement|
|---------|------|-----|-----------|
|CPU Time |75ms  |70ms |5ms (7%)   |
|GPU Time |17ms  |17ms |0ms (0%)   |


examples/experimental/sun/src/app.js
```diff
@@ -4,7 +4,6 @@ import {StaticMap} from 'react-map-gl';
 import DeckGL from '@deck.gl/react';
 import {AmbientLight, DirectionalLight, LightingEffect} from '@deck.gl/core';
 import {SolidPolygonLayer} from '@deck.gl/layers';
-import {PhongMaterial} from '@luma.gl/core';
```
Pessimistress: `PhongMaterial` is mentioned in all the layer docs.

examples/layer-browser/src/app.js
```diff
@@ -24,7 +24,7 @@ import LayerControls from './components/layer-controls';
 import LAYER_CATEGORIES from './examples';
 import Map from './map';
 
-import {ink} from '@luma.gl/effects';
```
Pessimistress: Update docs

modules/core/src/effects/lighting/lights.js
```diff
@@ -0,0 +1,88 @@
+import {Vector3} from 'math.gl';
```
Pessimistress: Nit: we usually use one file for each class.

package.json
```diff
@@ -41,7 +41,7 @@
     "jsdom": false
   },
   "devDependencies": {
-    "@luma.gl/effects": "^7.4.0-alpha.2",
+    "@luma.gl/engine": "^8.0.0-alpha.3",
```
Pessimistress: This should be dropped since it's already a dependency of deck core.

modules/core/package.json
```diff
@@ -32,8 +32,9 @@
   "dependencies": {
     "@loaders.gl/core": "^1.3.3",
     "@loaders.gl/images": "^1.3.3",
-    "@luma.gl/addons": "^7.4.0-alpha.2",
-    "@luma.gl/core": "^7.4.0-alpha.2",
+    "@luma.gl/addons": "^8.0.0-alpha.3",
+    "@luma.gl/core": "^8.0.0-alpha.3",
+    "@luma.gl/engine": "^8.0.0-alpha.3",
```
Pessimistress: Looks like engine should be a dependency of core?

test/modules/core/controllers/test-controller.js
```diff
@@ -1,4 +1,4 @@
-import {Timeline} from '@luma.gl/addons';
+import {Timeline} from '@luma.gl/engine';
```
Pessimistress: Can we re-export `Timeline` from luma core? This is the only place `@luma.gl/engine` is explicitly mentioned in deck.

docs/layers/column-layer.md
```diff
@@ -180,10 +180,10 @@ The maximum outline width in pixels.
 
 ##### `material` (Object, optional)
 
-* Default: `new PhongMaterial()`
+* Default: `true`
+
+This is an object that contains material props for [lighting effect](/docs/effects/lighting-effect.md) applied on extruded polygons. Check [the lighting guide](/docs/developer-guide/using-lighting.md) for more details.
```
Pessimistress: - link directly to `using-lighting.md#constructing-a-material-instance`
- "for more details" -> "for configurable settings"

modules/core/package.json
```diff
@@ -32,8 +32,8 @@
   "dependencies": {
     "@loaders.gl/core": "^1.3.4",
     "@loaders.gl/images": "^1.3.4",
-    "@luma.gl/addons": "^7.4.0-alpha.2",
-    "@luma.gl/core": "^7.4.0-alpha.2",
+    "@luma.gl/addons": "^8.0.0-alpha.5",
```
Pessimistress: I think this was only needed because of `Timeline`.

PR #3872, opened 2019-11-11
Fix attribute constant comparison

Pessimistress:
For #3852 

Fixes render test failure.

#### Background
`Attribute._areValuesEqual` only compares `size` elements, thus failing to detect change when comparing `[1]` with `Float32Array([1, 1, 2, ...])`.

#### Change List
- When checking for constant update, first check if the existing value is a constant

### DIFF
--- a/modules/core/src/lib/attribute/data-column.js
+++ b/modules/core/src/lib/attribute/data-column.js
@@ -167,7 +167,7 @@ export default class DataColumn {
       if (this.settings.normalized) {
         value = this._normalizeConstant(value);
       }
-      const hasChanged = !this._areValuesEqual(value, this.value);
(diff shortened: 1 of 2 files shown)


modules/core/src/lib/attribute/data-column.js
```diff
@@ -167,7 +167,7 @@ export default class DataColumn {
       if (this.settings.normalized) {
         value = this._normalizeConstant(value);
       }
-      const hasChanged = !this._areValuesEqual(value, this.value);
+      const hasChanged = !state.constant || !this._areValuesEqual(value, this.value);
```
xintongxia: this is not captured by any test?

PR #3874, opened 2019-11-12
v8.0 TripsLayer API change

Pessimistress:
For https://github.com/uber/deck.gl/issues/3869

#### Change List
- Remove support for the legacy use case (`[x, y, timestamp]` format)
- Remove arbitrary zShift applied in the vertex shader
- Docs
- Remove tests for the legacy use case

### DIFF
--- a/modules/geo-layers/src/trips-layer/trips-layer.js
+++ b/modules/geo-layers/src/trips-layer/trips-layer.js
@@ -19,7 +19,7 @@
 // THE SOFTWARE.
 
 import {PathLayer} from '@deck.gl/layers';
-import {createIterable} from '@deck.gl/core';
+import {createIterable, log} from '@deck.gl/core';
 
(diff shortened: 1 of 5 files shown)


modules/geo-layers/src/trips-layer/trips-layer.js
```diff
@@ -34,27 +34,12 @@ export default class TripsLayer extends PathLayer {
...
-vec2 timestamps = instanceTimestamps;
-if (!isPath3D) {
-  prevPosition.z = 0.0;
-  currPosition.z = 0.0;
-  nextPosition.z = 0.0;
-  timestamps.x = instanceStartPositions.z;
-  timestamps.y = instanceEndPositions.z;
-}
 `,
       // Apply a small shift to battle z-fighting
```
xintongxia: Does this comment still apply?

PR #3876, opened 2019-11-12
[Feat] add filterData prop to cpu aggregation layer

heshan0131:


heshan0131:
https://github.com/uber/deck.gl/issues/3583

1chandu:
CC: @Pessimistress 
@heshan0131 , my understanding is, this is a temporary change to enable Kepler to upgrade to 7.3 right. Any chance these can be done inside kepler (subclassing these two layers) to avoid any risk of regressions in 7.3 ?

test/modules/aggregation-layers/cpu-grid-layer/cpu-grid-layer.spec.js
```diff
@@ -32,7 +32,8 @@ test('CPUGridLayer', t => {
     Layer: CPUGridLayer,
     sampleProps: {
       data: FIXTURES.points.slice(0, 3),
-      getPosition: d => d.COORDINATES
+      getPosition: d => d.COORDINATES,
+      getFilterValue: null
```
1chandu: Nit: why set getFilterValue, this should be the default value, right? Same for hexagon layer test.

heshan0131: `testLayer` will automatically test all accessors, and throw error if `getFilterValue` is not defined. Pass in `getFilterValue` to override such behavior

modules/aggregation-layers/src/utils/cpu-aggregator.js
```diff
@@ -41,6 +41,16 @@ const defaultDimensions = [
         },
         aggregation: {
           prop: 'colorAggregation'
+        },
```
1chandu: `getBins` dimension is handling both `Filtering` and `Sorting`, can we split into two dimensions, `getFilteredBins` and `getSortedBins`? This will be inline with 8.0 refactor of aggregation layers as-well. Not all layers going to use filtering.

1chandu: actually , we can make sorting optional, which is more simpler change, I can do that in master during the refactor, this is ok for 7.3.

1chandu:
This enables data filtering on cpu-grid-layer only, so we have to
- Update grid-layer.js to fallback to cpu , when filtering is on.
- Update fallback case list in the docs

I am fine with this PR going to 7.3 given:
- 8.0 upgrade might take more than a month
- changes are limited to `Hexagon` and `CPUGridLayer`.

heshan0131:
After Discussing it with @Pessimistress, instead of provide cpu filter support for gpu filter prop, we are going to add a `filterData` property to cpu aggregation layer. User can provide custom filter function to filter points inside each bin.

1chandu:
Mostly looks good , one remaining item is, how do we handle this in `GridLayer`, when this prop specified, fallback to `CPUGridLayer` ?

heshan0131:
> Mostly looks good , one remaining item is, how do we handle this in `GridLayer`, when this prop specified, fallback to `CPUGridLayer` ?

Yes, `filterData` should be a CPU only prop

1chandu:
Once landed, please cherry-pick the change and push to master branch (No separate PR needed.)

Issue #3869, opened 2019-11-11
v8.0 deprecation/breaking changes


Pessimistress:
- [x] Layer opacity default to `1`
- [x] Zoom-independent common space
- [x] PathLayer: remove `getDashArray`, move functionality into an extension
- [x] Layers use actual shader modules, remove dependency on `registerShaderModules`
- [x] `Layer` class:
  + remove `setLayerNeedsUpdate`
  + remove `setUniforms`
  + remove `use64bitProjection`
- [x] `project` shader module:
  + remove `project_scale`
  + remove `project_to_clipspace`
  + remove `project_pixel_to_clipspace`
  + remove `project_to_clipspace_fp64`
- [x] `H3HexagonLayer`
  + remove `getColor`
- [x] `TripsLayer`
  + `getPath` no longer supports `[x, y, timestamp]` format
  + remove tiny z offset from the shader
- [x] `ArcLayer`
  + remove `getStrokeWidth`
- [x] `LineLayer`
  + remove `getStrokeWidth`
- [x] `WebMercatorViewport` class
  + remove `getLocationAtPoint`
  + remove `lngLatDeltaToMeters`
  + remove `metersToLngLatDelta`
- [x] `ScenegraphLayer` and `SimpleMeshLayer` should propagate layer’s `modelMatrix` to local positions
- [x] Remove auto-controller default syncing the state of multiple views
- [x] Standalone: ensure API consistency with `Deck`
  + Do not overwrite `width`/`height`
  + Do not auto-infer `initialViewState`
- [x] `TileLayer`
  + `onViewportLoaded` -> `onViewportLoad`
- [x] `Tile3DLayer`
  + `onTileLoadFail` -> `onTileError`
- [x] Use `ImageLoader` instead of `HTMLImageLoader`
- [x] Remove default `flipY` in texture loading
- [x] Remove `PerspectiveView` and `ThirdPersonView`

PR #3882, opened 2019-11-12
Remove dependency on `registerShaderModule`

Pessimistress:
For #3869

#### Change List
- Remove shader initialization
- Layers
- Tests
- Docs
- Upgrade guide

#### TODO
We don't have documentations for all available shader modules.

### DIFF
--- a/modules/core/src/shaderlib/index.js
+++ b/modules/core/src/shaderlib/index.js
@@ -19,11 +19,7 @@
 // THE SOFTWARE.
 
 import {ProgramManager} from '@luma.gl/core';
-import {
-  fp32,
-  gouraudlighting as gouraudLighting,
-  phonglighting as phongLighting
-} from '@luma.gl/core';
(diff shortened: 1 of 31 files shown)


modules/core/src/shaderlib/index.js
```diff
@@ -18,8 +18,8 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-import {registerShaderModules, ProgramManager} from '@luma.gl/core';
-import {fp32, gouraudlighting, phonglighting} from '@luma.gl/core';
+import {ProgramManager} from '@luma.gl/core';
+import {gouraudlighting as gouraudLighting, phonglighting as phongLighting} from '@luma.gl/core';
```
Pessimistress: @tsherif Can we rename these at the source?

tsherif: Yes, definitely.

Issue #3831, opened 2019-10-31
Projection system refactor


Pessimistress:
- Consistent common space
  + remove mercator-specific Y flipping
  + remove dependency on zoom level
- Better near/far planes
  + remove fov dependency on canvas height

PR #3885, opened 2019-11-14
Use 3D 64-bit positions

Pessimistress:
For #3831 

#### Change List
- All `vec2 *64xyLow` attributes are now `vec3 *64Low`
- Documentation

### DIFF
--- a/docs/shader-modules/project64.md
+++ b/docs/shader-modules/project64.md
@@ -22,14 +22,14 @@ The uniforms of the `project64` shader module are prefixed with `project64_` in 
 64 bit implementation of the `project_position_to_clipspace` interface.
 
 ```glsl
-vec4 project_position_to_clipspace(vec3 position, vec2 position64xyLow, vec3 offset)
-vec4 project_position_to_clipspace(vec3 position, vec2 position64xyLow, vec3 offset, out vec4 commonPosition)
+vec4 project_position_to_clipspace(vec3 position, vec3 position64Low, vec3 offset)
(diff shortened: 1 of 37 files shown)


modules/aggregation-layers/src/utils/gpu-grid-aggregation/gpu-grid-aggregator.js
```diff
@@ -150,8 +150,6 @@ export default class GPUGridAggregator {
       // cache weights and position data to process when data is not changed
       weights: null,
       gridPositions: null,
-      positionsBuffer: null,
-      positions64xyLowBuffer: null,
```
Pessimistress: @1chandu I'm pretty sure these are no longer used. They are still mentioned in the GPUGridAggregator documentation though.

1chandu: Yes, ok to remove,  code related to building/using these buffers is already removed in #3777.

docs/shader-modules/project64.md
```diff
@@ -43,7 +43,7 @@ Projected position in the clipspace.
 
 ```glsl
 void project_position_fp64(vec4 position_fp64, out vec2 out_val[2])
-void project_position_fp64(vec2 position, vec2 position64xyLow, out vec2 out_val[2])
+void project_position_fp64(vec2 position, vec2 position64Low, out vec2 out_val[2])
```
1chandu: Nit: `vec2 position64Low` -> `vec3 position64Low`

Pessimistress: `project_position_fp64` only deals with x and y.

PR #3888, opened 2019-11-14
Support 3D in ArcLayer

Pessimistress:
For #3869 

arc-lnglat-3d

#### Change List
- Use 3D positions for source/target
- Calculate parabola between 3D points
- Render test
- What's new


test/modules/layers/core-layers.spec.js
```diff
@@ -102,48 +102,6 @@ test('ArcLayer', t => {
   t.end();
 });
 
-test('ArcLayer#non-iterable', t => {
```
Pessimistress: This test targets an edge case in `calculateInstancePositions`, which is removed in this PR.

PR #3894, opened 2019-11-16
Change bufferLayout to startIndices

Pessimistress:
For #3758

#### Background

Variable-width attributes (e.g. those of the `PathLayer` or the `PolygonLayer`) contain a `bufferLayout` member that stores the vertex count of each object as `[chunk0Length, chunk1Length, ...]`.

This PR replaces the member with `vertexStarts` which stores the staring indices of each object as `[chunk0Start, chunk1Start, ...]`. This allows us to access arbitrary positions in the data without having to loop from the start.

This is in preparation for RFC #3884 -- if an application wants to supply a flat positions buffer for paths/polygons, they need to also supply a `vertexStarts` array, making the previously private API public.

I'm not happy with the name `vertexStarts`. Any suggestions?

#### Change List
- `Attribute` class
- `Tesselator` classes
- Attribute transition utils
- Unit tests

Verified with the layer browser and the attribute transition test app.

### DIFF
--- a/modules/core/src/lib/attribute/attribute-manager.js
+++ b/modules/core/src/lib/attribute/attribute-manager.js
@@ -216,7 +216,7 @@ export default class AttributeManager {
   update({
     data,
     numInstances,
-    bufferLayout,
+    vertexStarts,
     transitions,
     props = {},
     buffers = {},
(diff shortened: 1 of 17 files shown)


Pessimistress:
> How about `vertexOffsets` for the name?

We already use `offset` in the same context to refer to element positions:

```js
getVertexOffset(row, vertexStarts = this.vertexStarts) {
  const vertexIndex = vertexStarts ? vertexStarts[row] : row;
  return this.settings.elementOffset + vertexIndex * this.size;
}
```

@twojtasz is suggesting `startIndices`.

PR #3897, opened 2019-11-18
Support variable-width values in auto update

Pessimistress:
For #3758 

#### Change List
- Attribute auto updater allows assigning each vertex in a tesselated object different values.
- Simplify `TripsLayer`
- Tests

### DIFF
--- a/test/modules/core/lib/attribute/attribute.spec.js
+++ b/test/modules/core/lib/attribute/attribute.spec.js
@@ -422,6 +422,55 @@ test('Attribute#updateBuffer', t => {
   t.end();
 });
 
+test('Attribute#standard accessor - variable width', t => {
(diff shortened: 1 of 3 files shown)


test/modules/core/lib/attribute/attribute.spec.js
```diff
@@ -422,6 +422,81 @@ test('Attribute#updateBuffer', t => {
...
+        255,
+        0,
+        255,
+        128,
+        128,
+        128,
+        255,
+        255,
+        255,
+        255,
```
xintongxia: `[255, 255, 255, 255]` is default value of a color attribute?

Pessimistress: defaultValue is assigned in line 451.

PR #3902, opened 2019-11-19
Handle offset in DataColumn

Pessimistress:
For #3758 

This change moves the handling of attribute offset from typed array generation to the buffer management. Layers and application no longer have to consider the buffer layout when dealing with attributes.

#### Change List
- `DataColumn` can upload value to WebGL buffer at an offset
- Remove `Attribute.elementOffset`
- `PathTesselator` no longer adds padding to the positions array

### DIFF
--- a/modules/core/src/lib/attribute/data-column.js
+++ b/modules/core/src/lib/attribute/data-column.js
@@ -10,7 +10,7 @@ import log from '../../utils/log';
 function addDoublePrecisionAttributes(id, baseAccessor, shaderAttributeOptions) {
   const doubleShaderAttributeDefs = {};
   const offset =
-    'offset' in shaderAttributeOptions ? shaderAttributeOptions.offset : baseAccessor.offset || 0;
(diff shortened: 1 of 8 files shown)


modules/core/src/lib/attribute/data-column.js
```diff
@@ -99,6 +100,11 @@ export default class DataColumn {
     return this._buffer;
   }
 
+  get byteOffset() {
+    const {offset} = this.settings;
+    return this.doublePrecision && this.value instanceof Float64Array ? offset * 2 : offset;
```
tsherif: Does this need to be dynamic? Seems like it could be cached on updates to offset or value.

Pessimistress: This getter is fairly cheap and infrequently accessed (only on attribute updates). Because there are multiple ways to update data, reducing the number of properties that need maintenance will help control the complexity of the class.

Issue #3178, opened 2019-06-03
Text layer with background


maps82:
Hi! Can you give me a short and rough recipe on how I can achieve to render texts with text layer, but with white rounded background boxes, that have the same width as the text? I think I have to create a custom layer.

Is there a best practices for this without copying too much internals and messing around with a custom FontAtlasManager?

Thank you very much!

maps82:
I solved this by writing a custom background layer, based on the original text layer source, together with a custom fontAtlasManager to calculate the background width based on the text, and modified the canvas texture created by the fontAtlasManager.

I think this is currently the only way to achieve such a background box. I leave this open in case someone has better ideas. If not feel free to close this issue.

Pessimistress:
@xintongxia I think we could easily expose a TextLayer prop to set the background color of the texture canvas?

maps82:
Would be great. But additionally I need (and have now) some rounded corners on the left and right of the background.

maps82:
<img width="329" alt="Bildschirmfoto 2019-06-04 um 23 35 29" src="https://user-images.githubusercontent.com/12223156/58915516-7b12a200-8721-11e9-90ce-7063f9610759.png">

Thats what it looks currently in my case, but there is more to come...

vnugent:
@maps82  would you please share the code snippet?

danmarshall:
@Pessimistress : Would a background color enhance picking? I am hoping to get less accurate picking of text - for example, I would like the inside of the `o` and space between characters to be pickable.

![Text Picking](https://user-images.githubusercontent.com/11507384/63617252-d3797300-c59d-11e9-8a32-63e06f833fc6.gif)

Pessimistress:
@xintongxia The picking issue should be fixable even without supporting a secondary color , if we make the cuttoff configurable: https://github.com/uber/deck.gl/blob/master/modules/layers/src/icon-layer/icon-layer-fragment.glsl.js#L47

miguel-viola:
@maps82 
> I solved this by writing a custom background layer, based on the original text layer source, together with a custom fontAtlasManager to calculate the background width based on the text, and modified the canvas texture created by the fontAtlasManager.
> 
> I think this is currently the only way to achieve such a background box. I leave this open in case someone has better ideas. If not feel free to close this issue.

How does the fontAtlasManager work? I'm trying to do the same thing as you but with no success..

xintongxia:
I'll enable `text-layer` to support customized background color this week.

[...]

PR #3903, opened 2019-11-20
TextLayer: support background color

xintongxia:
For #3178


docs/layers/text-layer.md
```diff
@@ -113,6 +113,12 @@ The maximum size in pixels.
 
 If on, the text always faces camera. Otherwise the text faces up (z).
 
+##### `transparentColor` (Array) ![transition-enabled](https://img.shields.io/badge/transition-enabled-green.svg?style=flat-square")
```
Pessimistress: Why is it not called `backgroundColor`?

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer-fragment.glsl.js
```diff
@@ -47,12 +48,13 @@ void main(void) {
 
   // Take the global opacity and the alpha from vColor into account for the alpha component
   float a = alpha * vColor.a;
-
+  
+  vec4 color = vec4(vColor.rgb, a);;
   if (a < alphaCutoff) {
-    discard;
+    color = transparentColor;
```
Pessimistress: Does not account for `opacity`

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer.js
```diff
@@ -85,14 +86,15 @@ export default class MultiIconLayer extends IconLayer {
...
+    const {sdf, transparentColor} = this.props;
     super.draw({
       uniforms: Object.assign({}, uniforms, {
         // Refer the following doc about gamma and buffer
         // https://blog.mapbox.com/drawing-text-with-signed-distance-fields-in-mapbox-gl-b0933af6f817
         buffer: DEFAULT_BUFFER,
         gamma: DEFAULT_GAMMA,
-        sdf: Boolean(sdf)
+        sdf: Boolean(sdf),
+        transparentColor: transparentColor.map(c => c / 255.0)
```
Pessimistress: Does not handle missing alpha

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer-fragment.glsl.js
xintongxia: Do you mean `transparentColor` need apply the layer prop `opacity`?

Pessimistress: It’s not “transparent“, it’s background, and everything the layer renders must apply the opacity.

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer.js
```diff
@@ -74,25 +75,36 @@ export default class MultiIconLayer extends IconLayer {
...
       this.getAttributeManager().invalidate('instanceOffsets');
     }
+
+    if (props.backgroundColor !== oldProps.backgroundColor) {
+      const backgroundColor = props.backgroundColor.map(c => c / 255.0);
+      if (backgroundColor.length === 3) {
+        // alpha
+        backgroundColor.push(1.0);
+      }
+      this.setState({backgroundColor});
```
Pessimistress: Call `setUniforms` directly

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer-fragment.glsl.js
```diff
@@ -47,12 +48,13 @@ void main(void) {
...
   }
 
-  gl_FragColor = vec4(vColor.rgb, a);
+  gl_FragColor = color;
```
Pessimistress: Does this changes the default (no background) behavior in auto highlight?

modules/layers/src/text-layer/text-layer.js
```diff
@@ -64,6 +64,7 @@ const defaultProps = {
   sizeUnits: 'pixels',
   sizeMinPixels: 0,
   sizeMaxPixels: Number.MAX_SAFE_INTEGER,
+  backgroundColor: [0, 0, 0, 0],
```
georgios-uber: @xintongxia 
shouldn't this be 
`
backgroundColor: {type: 'color', value: [0, 0, 0, 0]},`

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer-fragment.glsl.js
```diff
@@ -36,23 +37,30 @@ varying vec2 uv;
...
   float a = alpha * vColor.a;
-
+  
   if (a < alphaCutoff) {
-    discard;
+    if (backgroundColor.a < 0.001) {
+      discard;
+    } else {
+      gl_FragColor = vec4(backgroundColor.rgb, backgroundColor.a * vColor.a);
+      return;
```
georgios-uber: @xintongxia the return statement fixes the issue of highlight.

docs/layers/text-layer.md
georgios-uber: It is now.

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer-fragment.glsl.js
georgios-uber: This is fixed now.

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer-fragment.glsl.js
georgios-uber: This is fixed in the updated shader.

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer-fragment.glsl.js
Pessimistress: Depending on the value of your backgroundColor, it may mess up picking.

Pessimistress: You can check `picking_uActive` as a hack.

georgios-uber: Any way to know if we are in picking mode?

georgios-uber: OK, I'll do that.

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer-fragment.glsl.js
```diff
@@ -36,23 +37,30 @@ varying vec2 uv;
...
+    if ((backgroundColor.a < 0.001) || picking_uActive) {
+      discard;
+    } else {
+      gl_FragColor = vec4(backgroundColor.rgb, backgroundColor.a * vColor.a);
+      return;
+    }
   }
 
-  gl_FragColor = vec4(vColor.rgb, a);
+  if (backgroundColor.a < 0.001) {
```
Pessimistress: Make a variable `bool shouldDrawBackground`

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer-fragment.glsl.js
```diff
@@ -36,23 +37,30 @@ varying vec2 uv;
...
+    gl_FragColor = vec4(vColor.rgb, a);
+  } else {
+    gl_FragColor = vec4(mix(backgroundColor.rgb, vColor.rgb, alpha), backgroundColor.a * vColor.a);
```
Pessimistress: See https://github.com/uber/luma.gl/blob/master/modules/shadertools/src/modules/picking/picking.js#L101 for how to do proper color blending.

Note that we don't blend colors in any other layer, e.g. between fill color and stroke color. I think it's fair and more efficient to just use `gl_FragColor = vec4(backgroundColor.rgb, backgroundColor.a * opacity)` here. If the user wants the color blended they can easily return a different instance color with `getColor`.

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer-fragment.glsl.js
```diff
@@ -36,23 +37,30 @@ varying vec2 uv;
...
+    alpha = smoothstep(buffer - vGamma, buffer + vGamma, alpha);
   }
 
   // Take the global opacity and the alpha from vColor into account for the alpha component
   float a = alpha * vColor.a;
-
+  
   if (a < alphaCutoff) {
-    discard;
+    if ((backgroundColor.a < 0.001) || picking_uActive) {
```
Pessimistress: This breaks the use case where we need to pick the text by pointing at empty pixels (e.g. the hole in "o").

I'm ok with making a breaking change in v8 to no longer support *NOT* picking the text in this case. i.e. never discard if `picking_uActive: true`.

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer-fragment.glsl.js
georgios-uber: uniform, set from the layer code?

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer-fragment.glsl.js
georgios-uber: Can you explain?

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer-fragment.glsl.js
georgios-uber: Done as uniform

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer-fragment.glsl.js
georgios-uber: This is a special case because the text is smooth (unlike other layers).
This is how it looks without blending:
Screen Shot 2019-12-06 at 3 48 52 PM

Pessimistress: I see. In that case, the opacity should also be interpolated (`mix(backgroundColor.a, alpha, alpha) * opacity`). Multiplying the two alpha channels make the in-between fragments more translucent than either the text or the background which doesn't make sense.

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer-fragment.glsl.js
Pessimistress: If you discard the background fragment in picking mode then the background won't be pickable. That makes it impossible to pick the text if your cursor is over a "hole" in the letter.

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer-fragment.glsl.js
georgios-uber: After playing with settings I think it is best to output constant opacity for both the text and background which will be `vColor.a`. This is basically the `opacity` property of the layer with no smoothing/interpolation in case of background.

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer-fragment.glsl.js
georgios-uber: Currently that is how TextLayer behaves in picking. It is a behavior I don't like.

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer-fragment.glsl.js
Pessimistress: Okay can you update the documentation and the uniform generation then? You can make the prop default to `null` by adding `optional: true`

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer-fragment.glsl.js
georgios-uber: Let me try to provide a fix.

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer-fragment.glsl.js
georgios-uber: OK, I'll do that.

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer-fragment.glsl.js
```diff
@@ -36,23 +38,37 @@ varying vec2 uv;
...
   if (a < alphaCutoff) {
-    discard;
+    // We are now in the background, let's decide what to draw
+    if (shouldDrawBackground && !picking_uActive) {
+      // draw background color and return if not picking
+      gl_FragColor = vec4(backgroundColor.rgb, vColor.a);
+      return;
+    } else if (picking_uActive) {
+      // allow picking to work and pick the background
+      DECKGL_FILTER_COLOR(gl_FragColor, geometry);
```
Pessimistress: Just let this fall through - the perf impact is minimal.

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer.js
```diff
@@ -80,25 +81,37 @@ export default class MultiIconLayer extends IconLayer {
...
     if (
       changeFlags.updateTriggersChanged &&
       (changeFlags.updateTriggersChanged.getAnchorX || changeFlags.updateTriggersChanged.getAnchorY)
     ) {
       this.getAttributeManager().invalidate('instanceOffsets');
     }
+
+    if (props.backgroundColor !== oldProps.backgroundColor) {
+      const backgroundColor = props.backgroundColor.map(c => c / 255.0);
+      if (backgroundColor.length === 3) {
```
Pessimistress: Alpha is no longer used, make `backgroundColor` a vec3, and `props.backgroundColor` default to `null`.

docs/layers/text-layer.md
```diff
@@ -113,6 +113,12 @@ The maximum size in pixels.
 
 If on, the text always faces camera. Otherwise the text faces up (z).
 
+##### `backgroundColor` (Array) ![transition-enabled](https://img.shields.io/badge/transition-enabled-green.svg?style=flat-square")
+
+- Default `[0, 0, 0, 0]`
+
+The color to use for text background, in `[r, g, b]` or `[r, g, b, a]`. Each component is in the `[0, 255]` range.
```
Pessimistress: `a` is not supported. Explain that alpha is determined by the per-object color.

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer-fragment.glsl.js
georgios-uber: ok...

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer.js
georgios-uber: `Uncaught Error: Invalid prop backgroundColor: null`. Why? the property is optional.

docs/layers/text-layer.md
```diff
@@ -113,6 +113,13 @@ The maximum size in pixels.
 
 If on, the text always faces camera. Otherwise the text faces up (z).
 
+##### `backgroundColor` (Array, optional) ![transition-enabled](https://img.shields.io/badge/transition-enabled-green.svg?style=flat-square")
+
+- Default `null`
+
+The color to use for text background, in `[r, g, b]`. Each component is in the `[0, 255]` range.
+Alpha is determined by the per-object color.
```
Pessimistress: "The alpha of the background matches the opacity of each object, controlled by the props `getColor` and `opacity`."

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer.js
```diff
@@ -80,25 +81,37 @@ export default class MultiIconLayer extends IconLayer {
...
+
     super.draw({
       uniforms: Object.assign({}, uniforms, {
         // Refer the following doc about gamma and buffer
         // https://blog.mapbox.com/drawing-text-with-signed-distance-fields-in-mapbox-gl-b0933af6f817
         buffer: DEFAULT_BUFFER,
         gamma: DEFAULT_GAMMA,
-        sdf: Boolean(sdf)
+        sdf: Boolean(sdf),
+        backgroundColor: shouldDrawBackground ? backgroundColor : [0, 0, 0],
```
Pessimistress: Nit: `backgroundColor || [0, 0, 0]`

PR #3913, opened 2019-11-21
Simplify bundle luma exports

Pessimistress:
#### Change List
- Remove classes that are not needed by custom layers
- Remove utility functions dropped in luma.gl v8
- Remove shader modules that are re-exported by deck


tsherif:
@Pessimistress just gonna merge this so I can do the fixes for the new alpha without conflicts.

PR #3919, opened 2019-11-21
Various core deprecations

Pessimistress:
For #3869 

Mostly unused and undocumented methods.

See inline comments.

### DIFF
--- a/modules/core/src/lib/layer.js
+++ b/modules/core/src/lib/layer.js
@@ -58,7 +58,6 @@ const defaultProps = {
     compare: false
   },
   updateTriggers: {}, // Update triggers: a core change detection mechanism in deck.gl
-  numInstances: undefined,
 
   visible: true,
   pickable: false,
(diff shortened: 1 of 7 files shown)


modules/core/src/lib/layer.js
```diff
@@ -586,10 +550,6 @@ export default class Layer extends Component {
     }
     // End subclass lifecycle methods
 
-    // TODO deprecated, for backwards compatibility with older layers
-    // in case layer resets state
-    this.state.attributeManager = this.getAttributeManager();
```
Pessimistress: This use case is still supported. This line is a duplicate.

modules/core/src/lib/layer.js
```diff
@@ -713,13 +667,6 @@ export default class Layer extends Component {
     this.props = currentProps;
   }
 
-  // {uniforms = {}, ...opts}
-  pickLayer(opts) {
```
Pessimistress: Only `getPickingInfo` is documented.

modules/core/src/lib/layer.js
```diff
@@ -599,12 +559,6 @@ export default class Layer extends Component {
     });
 
     this._updateState();
-
-    const model = this.getSingleModel();
```
xintongxia: Why this could be removed?

Pessimistress: - Model id does not serve any purpose other than for debugging
- All models created by build-in layers already have ids

PR #3920, opened 2019-11-22
Supporting quantile and other scale type in aggregation layer

heshan0131:
- Calculate domain based on style type
- Add test to bin-sorter

### DIFF
--- a/modules/aggregation-layers/src/utils/bin-sorter.js
+++ b/modules/aggregation-layers/src/utils/bin-sorter.js
@@ -83,6 +83,16 @@ export default class BinSorter {
       .sort((a, b) => a.value - b.value);
   }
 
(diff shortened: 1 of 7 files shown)


modules/aggregation-layers/src/utils/bin-sorter.js
```diff
@@ -56,6 +57,19 @@ export default class BinSorter {
       .sort((a, b) => a.value - b.value);
   }
 
+  percentileToIndex([lowerPercentile, upperPercentile]) {
```
1chandu: Nit: individual array elements are not being used, so keep the array as is, `percentileToIndex(percentileRange)`

1chandu:
Once landed,  cherry-pick this to master branch.

PR #3924, opened 2019-11-23
Fix FirstPersonView and FirstPersonController

Pessimistress:
#### Background
Needed by streetscape.

@twojtasz 

#### Change List
- Support pitch in `FirstPersonView`
- Fix zoom and move in `FirstPersonViewState`
- Update docs, move `FirstPersonView` out of experimental
- Fix test app


modules/core/src/controllers/first-person-controller.js
```diff
@@ -1,16 +1,16 @@
...
-  return Number.isFinite(value) ? value : fallbackValue;
-}
+const MOVEMENT_SPEED = 20;
+const DEFAULT_STATE = {
+  position: [0, 0, 0],
+  pitch: 0,
+  bearing: 0,
+  maxPitch: 90,
+  minPitch: -90
+};
```
xintongxia: Does `FirstPersonController` have a documentation?

Pessimistress: No. `FirstPersonController` is not exposed; it is only enabled via `controller: true` on `FirstPersonView`.

modules/core/src/controllers/first-person-controller.js
```diff
@@ -75,54 +63,37 @@ class FirstPersonState extends ViewState {
...
    * @param {[Number, Number]} pos - position on screen where the pointer grabs
    */
-  panStart({pos}) {
-    const {translationX, translationY} = this._viewportProps;
-
-    return this._getUpdatedState({
-      startPanPosition: [translationX, translationY],
-      startPanEventPosition: pos
-    });
+  panStart() {
```
xintongxia: maybe mention `FirstPersonView` does not support panning?

PR #3934, opened 2019-11-26
Updates to luma 8.0 alpha 10

tsherif:
Biggest update was adding some architecture so extensions can add their module injections to a `ProgramManager` local to a given `LayerManager` as the global shader hook logic has been deprecated.

Smaller changes:
- Update `BitmapLayer` and `IconLayer` to explicitly set flip y unpacking behaviour as the flip y default is deprecated.
- `getParameter` -> `getParameters` as `getParameter` has been deprecated.
- Switch to new module injection API


modules/core/src/lib/layer-manager.js
```diff
@@ -372,6 +374,14 @@ export default class LayerManager {
     // Set back pointer (used in picking)
     layer.internalState.layer = layer;
 
+    layer.props.extensions.forEach(extension => {
+      const name = extension.constructor.name;
+      if (!this._registeredExtensions[name]) {
+        extension.addShaderHooks(this.context.programManager);
+        this._registeredExtensions[name] = true;
+      }
+    });
```
tsherif: @Pessimistress this is the piece I'd most like to get your opinion on. It's what allows an extension to add it's injections to a program manager.

Pessimistress: We used to support an `inject` field in the shader module definition and let luma's `assembleShaders` handle it. From deck's perspective that makes more sense - all injections are defined with modules, and modules should not be aware of the context being used.

tsherif: Done.

modules/core/src/shaderlib/picking/picking.js
```diff
@@ -1,24 +1,24 @@
...
-    picking_setPickingColor(geometry.pickingColor);
-    // for picking depth values
-    picking_setPickingAttribute(geometry.position.z);
-`
-});
+export default Object.assign(
+  {
+    inject: {
+      'vs:DECKGL_FILTER_COLOR': `
+      picking_setPickingColor(geometry.pickingColor);
```
Pessimistress: Total nit: I usually format the injection code with the indentation it will need in the final glsl...

tsherif: Oh sure, I thought matched it. Is it 4 spaces?

PR #3939, opened 2019-11-27
Remove texture flip

Pessimistress:
For #3869 

#### Change List
- Switch `HTMLImageLoader` to `ImageLoader` (supports loading to `ImageBitmap` when `Image` is not available)
- Flip y in texture packing: `BitmapLayer`, `IconLayer`, `TextLayer`
- Upgrade guide


Pessimistress:
> Will textures still work on those browsers or would they now be flipped?

@ibgreen After this breaking change, `Image` object and `ImageBitmap` behave consistently. We couldn't use the `ImageLoader` before because with `UNPACK_FLIP_Y_WEBGL` set, `Image` is flipped but `ImageBitmap` is not.

PR #3941, opened 2019-11-28
Improve AttributeManager logging perf

Pessimistress:
#### Change List
- Avoid unnecessary construction/destruction of small objects

### DIFF
--- a/modules/core/src/lib/attribute/attribute-manager.js
+++ b/modules/core/src/lib/attribute/attribute-manager.js
@@ -33,13 +33,13 @@ const logFunctions = {
   timeStart: null,
   onUpdateStart: () => {
     logFunctions.savedMessages.length = 0;
-    logFunctions.timeStart = new Date();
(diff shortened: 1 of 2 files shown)


modules/core/src/lib/attribute/attribute-manager.js
```diff
@@ -27,35 +27,26 @@ import AttributeTransitionManager from './attribute-transition-manager';
...
     logFunctions.timeStart = new Date();
   },
-  onUpdate: ({level, message}) => {
-    if (logFunctions.savedMessages) {
-      logFunctions.savedMessages.push(message);
-    }
+  onUpdate: message => {
+    logFunctions.savedMessages.push(message);
   },
-  onUpdateEnd: ({level, id, numInstances}) => {
```
tsherif: Is this all internal? Do any docs need to be updated?

Pessimistress: This class is internal.

PR #3912, opened 2019-11-21
Aggregation Module: CPU Aggregation consolidation and integration with AttributeManager

1chandu:
For #3583

#### Background
- Integrate CPUAggregation modules to consume `position` data from AttributeManager.
- Add new CompositeLayer, `AggregationLayer`
- Sets up and maintains all the aggregation state
- Calls either CPU or GPU aggregation methods.
- Consolidate all CPU Aggregation methods, we now use same set of utility methods for all CPU aggregations. 
- Removed all CPU Aggregation code from GPUGridAggregator
- `GPUGridLayer` no longer supports CPU Aggregation, one should use `CPUGridLayer`.

#### Change List
- CPUAggregation (Grid and Hexagon): use attributes for position
- CPUAggregation consolidation.

PR #3952, opened 2019-12-03
[PART-2] CPUAggregation Refactor : BinSorter : Make sorting optional

1chandu:
For #3583

NOTE: Splitting #3912 to multiple stacked PRs, will close #3912 (keeping it open for existing discussions to be resolved).

#### Change List
- In preparation to consolidate all CPU Aggregation modules to same set of utility methods : 
   -  Make sorting of bins optional (for layers that don't have calculate lower/higher percentiles, sorting is not needed).
   - Min/Max values:  Calculate both min and max values to match GPU Aggregation.


modules/aggregation-layers/src/utils/bin-sorter.js
```diff
@@ -74,16 +77,6 @@ export default class BinSorter {
     return [this.sortedBins[lowerIdx].value, this.sortedBins[upperIdx].value];
```
heshan0131: sorting is used to quickly return value range by percentile. If the bins are not sorted, here you will have to first sort them before return value range by lower and upper percentile range

1chandu: With this refactor, I am going to use same cpu grid aggregation methods for all the layers (earlier I was using cpu aggregation in `GPUGridAggregator`), what I am doing here is, make sorting optional (by default it is true). Layers other than `CPUGridLayer` and `HexagonLayer`, don't need sorting, they don't support percentile props.

modules/aggregation-layers/src/utils/bin-sorter.js
```diff
@@ -64,38 +69,51 @@ export default class BinSorter {
...
 
   /**
    * Get a mapping from cell/hexagon index to sorted bin
    * This is used to retrieve bin value for color calculation
-   * @return {Object} bin index to sortedBins
+   * @return {Object} bin index to aggregatedBins
    */
   getBinMap() {
-    return this.sortedBins.reduce(
+    return this.aggregatedBins.reduce(
```
Pessimistress: Use for loop and `mapper[curr.i] = curr`

modules/aggregation-layers/src/utils/bin-sorter.js
```diff
@@ -64,38 +69,51 @@ export default class BinSorter {
...
+    return [this.aggregatedBins[lowerIdx].value, this.aggregatedBins[upperIdx].value];
   }
 
   /**
    * Get a mapping from cell/hexagon index to sorted bin
    * This is used to retrieve bin value for color calculation
-   * @return {Object} bin index to sortedBins
+   * @return {Object} bin index to aggregatedBins
    */
   getBinMap() {
```
Pessimistress: Is this a public method?

modules/aggregation-layers/src/utils/bin-sorter.js
```diff
@@ -23,37 +23,42 @@
...
    * @param {Function} getValue
+   * @param {Bool} sort
    * @return {Array} array of values and index lookup
    */
-  getSortedBins(bins, getValue) {
-    return bins
-      .reduce((accu, h, i) => {
-        const value = getValue(h.points);
+  getAggregatedBins(bins, getValue, sort) {
+    const aggregatedBins = bins.reduce((accu, h, i) => {
```
Pessimistress: - Is this a public method? Can the user update an existing instance?
- Use for loop

modules/aggregation-layers/src/utils/bin-sorter.js
1chandu: @heshan0131 is going to mark private/public methods in this file.

modules/aggregation-layers/src/utils/bin-sorter.js
1chandu: as discussed offline, added code to compute sortedBins on demand and save it to the class, removed `sort` flag.

PR #3954, opened 2019-12-03
[PART-4] CPUAggregation Refactor : Add CPU Aggregation support to GridAggregationLayer

1chandu:
For #3583

NOTE: Splitting #3912 to multiple stacked PRs, will close #3912 (keeping it open for existing discussions to be resolved).

#### Change List
- GridAggregationLayer:
    - Add support for CPU Aggregation (using same utility methods used for `CPUGridLayer`).
    - Add required state management to handle various aggregation scenarios (CPU vs GPU, WorldSpace vs ScreenSpace, single vs multiple weight aggregation)
- Update all grid aggregation layers, `ContourLayer`, `ScreenGridLayer`,  and `GPUGridLayer` to extend form `GridAggregationLayer`.  [NOTE: `CPUGridLayer` doesn't use `GridAggregationLayer`, which requires more sophisticated dimension management, same as `HexagonLayer]
- Add docs and unit tests.


docs/developer-guide/aggregation-layers.md
```diff
@@ -0,0 +1,55 @@
+# Aggregation Layers
```
1chandu: move to  aggregation-layers module.

modules/aggregation-layers/src/contour-layer/contour-layer.js
```diff
@@ -232,32 +221,6 @@ export default class ContourLayer extends GridAggregationLayer {
 
     return thresholds.some((_, i) => !equals(thresholds[i], oldThresholds[i]));
   }
-
-  _updateSubLayerTriggers(oldProps, props) {
```
Pessimistress: `_shouldRebuildContours` should be removed

modules/aggregation-layers/src/cpu-grid-layer/cpu-grid-layer.js
```diff
@@ -76,7 +76,7 @@ export default class CPUGridLayer extends AggregationLayer {
     };
     const attributeManager = this.getAttributeManager();
     attributeManager.add({
-      positions: {size: 3, accessor: 'getPosition' /* , type: GL.DOUBLE, fp64: false*/}
+      positions: {size: 3, accessor: 'getPosition'}
```
Pessimistress: - `type: GL:DOUBLE`?
- Why does this layer need the `attributeManager` for positions? It will create a WebGL buffer that it never uses.

modules/aggregation-layers/src/utils/gpu-grid-aggregation/aggregate-to-grid-vs.glsl.js
```diff
@@ -47,8 +51,17 @@ void main(void) {
 
   vec2 pos = project_to_pixel(windowPos);
 
+  vec2 pixelXY64[2];
```
Pessimistress: Looks like you are not handling `positions64Low` for `ScreenGridLayer`?

modules/aggregation-layers/src/cpu-grid-layer/cpu-grid-layer.js
1chandu: as discussed offline, using attribute manager for just positions and reading it from attributes array. This allows any AttributeManager features/optimization for position attribute available for cpu aggregation, we can remove this later if no such items are available. (Also original intetion was data filtering, but that has been made exclusive for GPU aggregation only.)

modules/aggregation-layers/src/utils/gpu-grid-aggregation/aggregate-to-grid-vs.glsl.js
1chandu: `pixelXY64` defined here only to perform fp64 arithmetic for division operation, applies to both world space aggregation and screen space, but I am not actually using low fp64 positions attributes, doesn't seem to be needed, we can add if there is a need.

Pessimistress: The projection to clipspace is inaccurate if you don't use the low parts. This applies to both geospatial and non-geospatial.

Pessimistress: I'm only making the comment here because GitHub doesn't allow me to comment on a line you didn't change.

modules/aggregation-layers/src/contour-layer/contour-layer.js
```diff
@@ -37,160 +37,158 @@ const defaultProps = {
...
+    const LinesSubLayerClass = this.getSubLayerClass('lines', LineLayer);
+    const BandsSubLayerClass = this.getSubLayerClass('bands', SolidPolygonLayer);
+
+    // Contour lines layer
+    const lineLayer =
+      contourSegments &&
+      contourSegments.length > 0 &&
+      new LinesSubLayerClass(
+        this.getSubLayerProps({
+          id: 'contour-line-layer'
```
Pessimistress: - Just use sub layer ids "lines" and "bands". They will be concatenated with the parent layer's id so you don't have to call out that they are children of a contour layer.
- The sub layer ids should be documented in `contour-layer.md`

modules/aggregation-layers/src/contour-layer/contour-layer.js
```diff
@@ -37,160 +37,158 @@ const defaultProps = {
...
-      getPolygon: d => d.vertices,
-      getFillColor: this._onGetSublayerColor.bind(this),
-      updateTriggers: {
-        getFillColor: colorTrigger
-      }
-    });
-  }
+  // Private (Sublayers)
 
   _onGetSublayerColor(element) {
```
Pessimistress: This is extremely expensive (creating inline functions inside an iteration).

You should be able to do something like `contours[object.contourIndex].color`.

Same thing for `_onGetSublayerStrokeWidth`

modules/aggregation-layers/src/cpu-grid-layer/cpu-grid-layer.js
Pessimistress: `GL.DOUBLE` packs a Float64Array instead of Float32Array. You lose precision without it in LNGLAT mode.

modules/aggregation-layers/src/utils/gpu-grid-aggregation/aggregate-to-grid-vs.glsl.js
1chandu: Updated attributes to be 64bit and added attribute to the shader.

PR #3968, opened 2019-12-05
Clean up change flags handling in Layer class

Pessimistress:
For bundle size optimization

#### Change List
- Use iteration to check change flag updates

### DIFF
--- a/modules/core/src/lib/layer.js
+++ b/modules/core/src/lib/layer.js
@@ -697,9 +697,7 @@ export default class Layer extends Component {
       changeFlags.extensionsChanged;
     changeFlags.propsOrDataChanged = propsOrDataChanged;
     changeFlags.somethingChanged =
-      propsOrDataChanged ||
-      flags.viewportChanged ||
-      flags.stateChanged;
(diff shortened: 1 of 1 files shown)


modules/core/src/lib/layer.js
```diff
@@ -680,47 +680,23 @@ export default class Layer extends Component {
...
 
-    // Update primary flags
-    if (flags.dataChanged && !changeFlags.dataChanged) {
-      changeFlags.dataChanged = flags.dataChanged;
-      debug(TRACE_CHANGE_FLAG, this, 'dataChanged', flags);
-    }
-    if (flags.updateTriggersChanged && !changeFlags.updateTriggersChanged) {
-      changeFlags.updateTriggersChanged =
-        changeFlags.updateTriggersChanged && flags.updateTriggersChanged
-          ? Object.assign({}, flags.updateTriggersChanged, changeFlags.updateTriggersChanged)
```
Pessimistress: No breaking change here - this merge never happens due to the if statement above

PR #3970, opened 2019-12-05
[New @deck.gl/json syntax support - Part 5] @deck.gl/json: Convert JSON playground examples to new syntax

ajduberstein:
Converts @deck.gl/json playground to the new syntax.

#### Change List
-


Pessimistress:
Fix lint before merge.

PR #3972, opened 2019-12-06
Tile3DLayer: v8.0 audit

xintongxia:
## Background
- v8.0 API audit

#### Change List
- `onTileLoadFail` -> `onTileError`


modules/geo-layers/src/tile-3d-layer/tile-3d-layer.js
```diff
@@ -36,6 +37,9 @@ export default class Tile3DLayer extends CompositeLayer {
   }
 
   async updateState({props, oldProps}) {
+    if ('onTileLoadFail' in props) {
+      log.warn('`onTileLoadFail` is deprecated. use `onTileError` instead.');
```
Pessimistress: - Move to `initializeState` - we usually only check deprecated props once
- Use `log.removed`
- `log.warn` returns a function. You need to call it e.g. `log.warn(...)()`.

PR #3973, opened 2019-12-06
Upgrade probe.gl (part 2)

Pessimistress:
#### Change List
- Remove globals, use `probe.gl/env`
- Rename `log.priority` (deprecated) to `log.level`


xintongxia:
does this need to be mentioned in upgrade guide?

PR #3978, opened 2019-12-09
GPUGridAggregator: use fp64arithmetic module instead of fp64

1chandu:
For #

#### Background
- Bump luma.gl dep to beta.2
- GPUGridAggregator: use `fp64arithmetic` module instead of `fp64`

#### Change List
- GPUAggregator: use fp64arithmetic instead of fp64


test/modules/core/shaderlib/project/project-32-64-glsl.spec.js
```diff
@@ -287,14 +287,10 @@ test('project32&64#vs', t => {
...
+          const dependencies = appendDependencies(module, []);
           const vsSource =
-            []
-              .concat(
-                usefp64 ? project64.dependencies : project32.dependencies,
-                project.dependencies
-              )
-              .map(dep => dep.vs)
-              .join('') + (usefp64 ? project64.vs : project32.vs);
+            dependencies.map(dep => dep.vs).join('') + (usefp64 ? project64.vs : project32.vs);
```
Pessimistress: Just add `module` to `dependencies`

PR #3989, opened 2019-12-10
Prop rename: filterData to _filterData

1chandu:
For #

#### Background
`filterData` is not a public API, rename it to`_filterData`.

#### Change List
-


1chandu:
The build issues are related to pydeck setup, and related to #3817.

PR #3994, opened 2019-12-11
Attribute update perf

Pessimistress:
A fix we made in `attribute._normalizeValue` from 7.3 to 8.0 is showing up in profiling. This PR restores the old implementation for size<=4 use cases for performance.

Before:

├─ initialize layers: 34.3 iterations/s ±20.04%
├─ update layers: 13.5 iterations/s ±6.94%

After:

├─ initialize layers: 37.5 iterations/s ±20.41%
├─ update layers: 39.4 iterations/s ±6.43%

#### Change List
- Improve `attribute._normalizeValue` perf
- Bench tests

### DIFF
--- a/modules/core/src/lib/attribute/data-column.js
+++ b/modules/core/src/lib/attribute/data-column.js
@@ -197,7 +197,7 @@ export default class DataColumn {
     if (opts.constant) {
       // set constant
       let value = opts.value;
-      value = this._normalizeValue(value);
+      value = this._normalizeValue(value, [], 0);
       if (this.settings.normalized) {
         value = this._normalizeConstant(value);
       }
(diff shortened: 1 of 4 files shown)


modules/core/src/lib/attribute/data-column.js
```diff
@@ -346,25 +346,48 @@ export default class DataColumn {
...
+    if (Number.isFinite(value)) {
+      out[start] = value;
       return out;
     }
+    if (!value) {
+      out[start] = defaultValue[0];
+      return out;
+    }
+
+    // Important - switch cases are 5x more performant than a for loop!
```
tsherif: 👍

PR #3995, opened 2019-12-11
@deck.gl/json: Fix function lookups in classes

ajduberstein:
While syntax like `@@=[lng, lat]` was evaluating correctly as a function, the old syntax (`[lng, lat]`) was still returning an evaluated expression rather than a string. This PR corrects that

New browser tests and modified unit tests are added as well.

For #3640 

#### Background
Supports @deck.gl/json syntactic sugar

#### Change List
- New browser test to assure @@= function syntax works
- Move syntactic sugar constants to their own file
- Modify the @deck.gl/json class conversion path
- Modify convert-functions.js to convert functions without propType sniffing
- Update unit tests for convertFunctions


test/modules/json/json-render.spec.js
```diff
@@ -11,7 +11,7 @@ test('JSONConverter#render', t => {
 
   const deckProps = jsonConverter.convert(JSON_DATA);
   t.ok(deckProps, 'JSONConverter converted correctly');
-
+  console.log({deckProps}); // eslint-disable-line
```
Pessimistress: Use `t.comment`

PR #4003, opened 2019-12-12
pydeck: Add CDN-hosted bundle for standalone HTML rendering

ajduberstein:
For #3640

The standalone HTML renderer needs to download a built version of @deck.gl/jupyter-widget from a CDN to work. Blocking for pydeck 0.2.0.

### DIFF
--- a/bindings/python/pydeck/pydeck/io/html.py
+++ b/bindings/python/pydeck/pydeck/io/html.py
@@ -12,6 +12,7 @@ TEMPLATES_PATH = os.path.join(os.path.dirname(__file__), './templates/')
 j2_env = jinja2.Environment(loader=jinja2.FileSystemLoader(TEMPLATES_PATH),
                             trim_blocks=True)
 
+STANDALONE_LIB = 'https://cdn.jsdelivr.net/npm/@deck.gl/jupyter-widget@8.0.0-alpha.2/dist/standalone-html-bundle.js'
 
 def render_json_to_html(json_input, mapbox_key=None, tooltip=True):
     js = j2_env.get_template('index.j2')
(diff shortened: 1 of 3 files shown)


bindings/python/pydeck/pydeck/io/html.py
```diff
@@ -20,8 +20,7 @@ def render_json_to_html(json_input, mapbox_key=None, tooltip=True):
     html_str = js.render(
         mapbox_key=mapbox_key,
         json_input=json_input,
-        # TODO change before publication to the NPM-hosted module
-        deckgl_jupyter_widget_bundle='',
+        deckgl_jupyter_widget_bundle='https://cdn.jsdelivr.net/npm/@deck.gl/jupyter-widget@8.0.0-alpha.2/dist/index.js',
```
Pessimistress: Is there a reason to lock the version of the package?

ajduberstein: Guaranteeing consistent behavior from one pydeck release to the next

Pessimistress: - There should not be any breaking changes between patch releases of `@deck.gl/jupyter-widget`.
- The pydeck release should require a minimum version to ensure certain bug fix/new feature is included. We frequently publish fixes to the core module and pydeck is not automatically released with the js packages. There is no reason for you to release a new version of pydeck without any change on the python side every time deck.gl is published.

modules/jupyter-widget/package.json
```diff
@@ -18,6 +18,8 @@
   "files": [
     "dist/index.js",
     "dist/index.js.map",
+    "dist/standalone-html-bundle.js",
```
Pessimistress: Where is this file from?

bindings/python/pydeck/pydeck/io/html.py
ajduberstein: Locked to deck.gl ^8.0 in the latest commit

modules/jupyter-widget/package.json
ajduberstein: It no longer exists; I've cut it in the latest commit.

bindings/python/pydeck/pydeck/io/html.py
```diff
@@ -11,7 +11,7 @@
 TEMPLATES_PATH = os.path.join(os.path.dirname(__file__), './templates/')
 j2_env = jinja2.Environment(loader=jinja2.FileSystemLoader(TEMPLATES_PATH),
                             trim_blocks=True)
-
+CDN_URL = 'https://cdn.jsdelivr.net/npm/@deck.gl/jupyter-widget@^8.0.0/dist/index.min.js'
```
Pessimistress: What is `index.min.js`? From your package.json you should only have `index.js`?

PR #4007, opened 2019-12-12
Improve createProps perf

Pessimistress:
Before

├─ createProps#empty: 575K iterations/s ±1.46%
├─ createProps#realWorld: 417K iterations/s ±6.83%

After

├─ createProps#empty: 6.01M iterations/s ±1.98%
├─ createProps#realWorld: 1.94M iterations/s ±2.27%

#### Change List
- Use `Symbol` instead of `Object.defineProperties` for props' non-enumerable fields
- Add deprecation warning/fallback to setters on the props prototype, instead of checking with for loop
- Bench tests

### DIFF
--- a/modules/core/src/lifecycle/create-props.js
+++ b/modules/core/src/lifecycle/create-props.js
@@ -1,9 +1,9 @@
 import log from '../utils/log';
 import {isAsyncIterable} from '../utils/iterable-utils';
 import {parsePropTypes} from './prop-types';
-import {PROP_KEYS} from './constants';
+import {PROP_SYMBOLS} from './constants';
 
-const {COMPONENT, ASYNC_ORIGINAL, ASYNC_RESOLVED, ASYNC_DEFAULTS} = PROP_KEYS;
+const {COMPONENT, ASYNC_ORIGINAL, ASYNC_RESOLVED, ASYNC_DEFAULTS} = PROP_SYMBOLS;
 
 // Create a property object
 export function createProps() {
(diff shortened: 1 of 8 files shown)


modules/core/src/lifecycle/create-props.js
```diff
@@ -1,139 +1,99 @@
...
+  propsInstance[ASYNC_ORIGINAL] = {};
+  // Note: the actual (resolved) values for props that are NOT set to urls or Promises.
+  // in this case the values are served directly from this map
+  propsInstance[ASYNC_RESOLVED] = {};
 
   // "Copy" all sync props
   for (let i = 0; i < arguments.length; ++i) {
-    Object.assign(propsInstance, arguments[i]);
+    const props = arguments[i];
+    // Do not use Object.assign here to avoid copying Symbols
```
tsherif: Any concern that we might have to worry about this distinction elsewhere?

Pessimistress:
> I wasn't aware that symbols had such a perf advantage.

`Object.defineProperties` is apparently much slower than direct assignment, not mentioning the re-creation of property definition objects. 

> Do we feel that we have sufficient test coverage to catch regressions here?

`lifecycle/*.js` has ~97% test coverage, and its usage in all layers is tested by automated layer tests.

modules/core/src/lifecycle/create-props.js
Pessimistress: The real concern is that user arguments may overwrite the private fields. At the end of this function we call `Object.freeze` so they will be protected from there.

PR #4021, opened 2019-12-14
Smooth edges in scatterplot

Pessimistress:
Before (devicePixelRatio=1):

image

After:

image

#### Change List
- Add a `smoothedge` function to the `geometry` module
- ScatterplotLayer shaders


modules/core/src/shaderlib/misc/geometry.js
```diff
@@ -30,9 +30,15 @@ struct VertexGeometry {
 `;
 
 const fs = `
+#define SMOOTH_EDGE_RADIUS 0.5
```
tsherif: Any chance the user will want to control this (maybe to create a fade out effect)?

Pessimistress: I'd keep it under the wraps for the current release cycle (i.e. a visual artifacts fix vs. a new feature)

Issue #3983, opened 2019-12-09
8.0 Release Tracker


tgorkin:
- [x] Bump all dependencies (math.gl, probe.gl and loaders.gl) to latest release (@Pessimistress)
- [x] Publish alpha/beta  (@Pessimistress)
- [x] Bump all examples to latest alpha/beta (@Pessimistress)
- [x] Stage the website (@Pessimistress)
- [x] Verify whats-new (@Pessimistress)
- [x] Verify upgrade-guide (@Pessimistress)
- [x] Verify all issues in milestone are closed or move out (@tgorkin)
- [x] Verify all API changes are documented (@Pessimistress)
- [x] Website (docs and example) testing: [staging]()
    - [x] Testing on Chrome (@tsherif)
    - [x] Testing on Safari (@xintongxia)
    - [x] Testing on Firefox (@1chandu)
    - [x] Testing on Edge + IE11 (@tsherif)
    - [x] Testing on iOS (@Pessimistress)
    - [x] Testing on Android (@Pessimistress)
- [x] Cut 8.0-release branch after all fixes (@Pessimistress)
- [x] Update website doc urls to point to `8.0-release` branch (@Pessimistress)
- [x] Publish 8.0 prod version.

Testing procedure
- access [staged doc website](http://pessimistress.github.io/deck.gl/)
- go through all examples and verify functionality in control panels
- go through all layers in [JSON playground](http://pessimistress.github.io/deck.gl/playground/)
- check visible results for regressions and console for errors/warnings
- add results of testing as comments on this issue

Pessimistress:
iOS:

- [ ] 3D surface example: text labels are up side down
- [ ] ScreenGridLayer example: aggregation result is not aligned with the base map

![E775E69D-F924-4B1D-A485-AAE380A2D601](https://user-images.githubusercontent.com/2059298/70772018-0cde1600-1d28-11ea-85c5-82ab466f2d44.jpeg)

1chandu:
- [ ] TripsLayer Example: when rotating only base map is rotated, also seeing these WebGL errors in console
```
Error: WebGL warning: <SetDimensions>: Exceeded 16 live WebGL contexts for this principal, losing the least recently used one. dist.min.js:1:3763
Error: WebGL warning: generateMipmap: Tex image TEXTURE_2D level 0 is incurring lazy initialization. dist.min.js:1:98609
[...]

PR #4026, opened 2019-12-16
[Fix]: CPU Aggregation: filter out points outside of viewport.

1chandu:
For #3983

#### Change List
- [Fix] CPUAggregation: filter out points ouside of viewport bounds.

### DIFF
--- a/modules/aggregation-layers/src/cpu-grid-layer/grid-aggregator.js
+++ b/modules/aggregation-layers/src/cpu-grid-layer/grid-aggregator.js
@@ -96,7 +96,7 @@ function pointsToGridHashing(props, aggregationParams) {
       const xIndex = Math.floor((x + offsets[0]) / gridOffset.xOffset);
       if (
         !projectPoints ||
-        // when doing screen space agggregation, filter points outside of the viewport range.
+        // when doing screen space agggregation (projectPoints = true), filter points outside of the viewport range.
         (xIndex >= 0 && xIndex < numCol && yIndex >= 0 && yIndex < numRow)
       ) {
         const key = `${yIndex}-${xIndex}`;
--- a/modules/aggregation-layers/src/cpu-grid-layer/grid-aggregator.js
+++ b/modules/aggregation-layers/src/cpu-grid-layer/grid-aggregator.js
@@ -70,7 +70,7 @@ export function getGridOffset(boundingBox, cellSize) {
 /* eslint-disable max-statements, complexity */
 function pointsToGridHashing(props, aggregationParams) {
   const {data = [], cellSize} = props;
-  const {attributes, viewport, projectPoints, numInstances, width, height} = aggregationParams;
+  const {attributes, viewport, projectPoints, numInstances} = aggregationParams;
   const positions = attributes.positions.value;
   const {size} = attributes.positions.getAccessor();
   const boundingBox =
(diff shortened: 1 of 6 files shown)


test/modules/aggregation-layers/grid-aggregation-layer.spec.js
```diff
@@ -214,7 +215,7 @@ test('GridAggregationLayer#CPUvsGPUAggregation', t => {
         data: fixture.data,
         getWeight: x => x.weight1[0],
         aggregation: 'SUM',
-        cellSize: 5000,
+        cellSize: 50,
```
Pessimistress: Wasn't this set to `5000` to reduce test runtime?

modules/aggregation-layers/src/cpu-grid-layer/grid-aggregator.js
```diff
@@ -92,11 +94,17 @@ function pointsToGridHashing(props, aggregationParams) {
...
       const xIndex = Math.floor((x + offsets[0]) / gridOffset.xOffset);
-      const key = `${yIndex}-${xIndex}`;
-
-      gridHash[key] = gridHash[key] || {count: 0, points: [], lonIdx: xIndex, latIdx: yIndex};
-      gridHash[key].count += 1;
-      gridHash[key].points.push(pt);
+      if (
+        !projectPoints ||
+        // when doing screen space agggregation, filter points outside of the viewport range.
+        (xIndex >= 0 && xIndex < numCol && yIndex >= 0 && yIndex < numRow)
```
Pessimistress: The filter needs to be generalized. What if the aggregator is used for a non-screen grid use case?

1chandu: `projectPoints` argument differentiates, screen space vs non screen space. This aggregator currently being used for both cases.
Updated the header doc for the public function.

test/modules/aggregation-layers/grid-aggregation-layer.spec.js
1chandu: Needed one test case with smaller cell size,  otherwise, everything gets aggregated to the same cell.
Earlier running time was an issue on CI only (browser tests were fast), but now I re-tested, test finished quickly.

modules/aggregation-layers/src/cpu-grid-layer/grid-aggregator.js
Pessimistress: If there are only two cases, then you should be able to use `viewport.width` and `viewport.height` instead of sending them separately.

1chandu: `width` and `height` are grid dimensions, when doing world space aggregation they refer to bounding box, when doing screen space they refer to viewport.

Pessimistress: As far as I can tell you are not using `width` and `height` in the world space aggregation case. This goes back to the confusion about how generic this method should be. If you are targeting very specific use cases and not caring about future expansion, why add arguments that you don't need?

1chandu: Ok, bounds checking only needed in screen space (as world space, entire bounding box is mapped to the grid), updated the PR to deduce `width` and `height` from `viewport`.

PR #4032, opened 2019-12-17
React module fixes

Pessimistress:
For #3983 

#### Change List
- Fix a crash when used with older versions of react-map-gl (missing `transitionInterpolator.getDuration`)
- Behavioral change: (except the base map) all child elements default to render above the canvas. This removes the dependency on `child.props.style.zIndex` to position children.
- Behavioral change: (except the base map) no longer overwrites `child.props` with information from the current viewport (viewState, width, height, etc.). This removes React warning about unsupported div props. Components can still use `ContextProvider` or functional child to listen to viewport change.
- Only create one wrapper for each view. This improves performance when a lot of child elements are used.
- Unit tests
- Docs

### DIFF
--- a/docs/api-reference/react/deckgl.md
+++ b/docs/api-reference/react/deckgl.md
@@ -69,41 +69,6 @@ import {_MapContext as MapContext, NavigationControl} from 'react-map-gl';
 </DeckGL>
 ```
 
-### View State Properties
-
-For backwards compatibility, the `DeckGL` component can be used without the `viewState` prop. If a `viewState` prop is not supplied, `DeckGL` will attempt to autocreate a geospatial view state from the following props.
-
-##### `latitude` (Number, optional)
-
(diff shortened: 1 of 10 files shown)


xintongxia:
Is this breaking change? Should mention it in the upgrade doc?

> Behavioral change: (except the base map) no longer overwrites child.props with information from the current viewport (viewState, width, height, etc.). This removes React warning about unsupported div props. Components can still use ContextProvider or functional child to listen to viewport change.

Does this mean DeckGL's children can have different sizes from the canvas?

Pessimistress:
> Is this breaking change? Should mention it in the upgrade doc?

Already is.

> Does this mean DeckGL's children can have different sizes from the canvas?

They always could. Injecting `width` and `height` props to child elements don't change their dimensions most of the time anyways, e.g. `<div width={...} height={...} />`

PR #4035, opened 2019-12-17
Support preprojection in PolygonTesselator

Pessimistress:
For https://github.com/uber/deck.gl/issues/3892

#### Change List
- Preproject LNGLAT points to the mercator plane before polygon tesselation


test/render/index.js
```diff
@@ -38,10 +38,10 @@ test('Render Test', t => {
       imageDiffOptions: {
         threshold: 0.99
         // uncomment to save screenshot to disk
-        // saveOnFail: true,
+        // , saveOnFail: true
         // uncomment `saveAs` to overwrite current golden images
         // if left commented will be saved as `[name]-fail.png.` enabling comparison
-        // saveAs: '[name].png'
+        // , saveAs: '[name].png'
```
tsherif: Ha, clever :)

modules/layers/src/solid-polygon-layer/polygon.js
```diff
@@ -287,12 +291,28 @@ export function normalize(polygon, positionSize, vertexCount) {
...
     holeIndices = normalizedPolygon.holeIndices.map(positionIndex => positionIndex / positionSize);
   }
+  let positions = normalizedPolygon.positions || normalizedPolygon;
+
+  // TODO - handle other coordinate systems and projection modes
+  if (preproject) {
+    // When tesselating lnglat coordinates, project them to the Web Mercator plane for accuracy
+    const n = positions.length;
+    // Clone the array
+    positions = positions.slice();
```
tsherif: Could prevent some memory churn by pre-allocating this array and just setting the length here. (Also possible for holeIndices).

Pessimistress: `positions` is a typed array. Yes I can pre-allocate the holeIndices.

PR #4037, opened 2019-12-17
Fix WebGL BlendEquation warnings

1chandu:
For #3983 

#### Background
Fix invalid blend mode warnings.
ContourLayer: add missing prop documentation.

#### Change List
- Fix WebGL BlendEquation warnings

### DIFF
--- a/modules/aggregation-layers/src/utils/gpu-grid-aggregation/gpu-grid-aggregator.js
+++ b/modules/aggregation-layers/src/utils/gpu-grid-aggregation/gpu-grid-aggregator.js
@@ -515,7 +515,7 @@ export default class GPUGridAggregator {
         });
       }
       framebuffers[id].resize(framebufferSize);
-      equations[id] = EQUATION_MAP[operation];
+      equations[id] = EQUATION_MAP[operation] || EQUATION_MAP.SUM;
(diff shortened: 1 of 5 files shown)


modules/aggregation-layers/src/utils/gpu-grid-aggregation/gpu-grid-aggregator.js
```diff
@@ -515,7 +515,7 @@ export default class GPUGridAggregator {
         });
       }
       framebuffers[id].resize(framebufferSize);
-      equations[id] = EQUATION_MAP[operation];
```
tsherif: Should this not fail? Silently changing the the requested operation seems like it could lead to tricky bugs (though I'll admit I'm not totally clear on where `operation` is coming from and what's expected of it).

Pessimistress: +1

1chandu: This is documented in the layer's page, `aggregation` prop, when not specified on invalid, we will fallback to `SUM`.

PR #4047, opened 2019-12-18
Refactor Styletron/BaseUI providers out of data filter example app file.

tgorkin:


examples/website/data-filter/wrapper.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
```
Pessimistress: Following React naming convention, I'd call

- the file root.js
- `function Root({data})`
- `render(<Root data={data} />, container)`

PR #4050, opened 2019-12-19
[WIP] Fix frustum plane calculation

Pessimistress:
Broken due to #3841 

- `fov` no longer depends on `height`
- `viewMatrix` now scales x and y as well as z
- Mercator common space y is flipped

Tested with /test/apps/frustum-culling

### DIFF
--- a/modules/core/src/utils/math-utils.js
+++ b/modules/core/src/utils/math-utils.js
@@ -83,7 +83,7 @@ export function getFrustumPlanes({aspect, near, far, fovyRadians, position, dire
     .add(nearCenter)
     .subtract(cameraPosition)
     .normalize();
-  normal = new Vector3(cameraUp).cross(a).negate();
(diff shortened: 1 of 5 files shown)


modules/core/src/utils/math-utils.js
```diff
@@ -79,7 +83,7 @@ export function getFrustumPlanes({aspect, near, far, fovyRadians, position, dire
     .add(nearCenter)
     .subtract(cameraPosition)
     .normalize();
-  normal = new Vector3(cameraUp).cross(a);
+  normal = new Vector3(cameraUp).cross(a).negate();
```
tsherif: Take the `negate` out and reverse the order of the `cross` (i.e. `new Vector3(a).cross(cameraUp)`). Same for all the other cross products.

This changes makes sense because of the y axis no longer being flipped. It changes the handedness of the space.

PR #4061, opened 2019-12-20
Bump loaders and fix infinite updateState calls

xintongxia:
v8.0


modules/geo-layers/src/tile-3d-layer/tile-3d-layer.js
```diff
@@ -39,19 +39,21 @@ export default class Tile3DLayer extends CompositeLayer {
...
       (props._ionAccessToken !== oldProps._ionAccessToken ||
         props._ionAssetId !== oldProps._ionAssetId)
     ) {
-      await this._loadTilesetFromIon(props._ionAccessToken, props._ionAssetId);
+      this._loadTilesetFromIon(props._ionAccessToken, props._ionAssetId);
     }
 
-    const {tileset3d} = this.state;
-    await this._updateTileset(tileset3d);
+    if (changeFlags && changeFlags.viewportChanged) {
```
Pessimistress: `changeFlags` should never be empty

PR #4064, opened 2019-12-20
fix 3d tiles culling

Pessimistress:
For #3983 

#### Change List
- Bump loaders.gl to 2.0.0-beta.8 (bounding volume fix)
- Fix frustum plane conversion

### DIFF
--- a/modules/geo-layers/src/tile-3d-layer/get-frame-state.js
+++ b/modules/geo-layers/src/tile-3d-layer/get-frame-state.js
@@ -43,7 +43,7 @@ export function getFrameState(viewport, frameNumber) {
     enuToFixedTransform.transformAsVector(new Vector3(cameraUp).scale(metersPerUnit))
   ).normalize();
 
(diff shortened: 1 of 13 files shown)


modules/geo-layers/src/tile-3d-layer/get-frame-state.js
```diff
@@ -80,15 +80,10 @@ function commonSpacePlanesToWGS84(viewport) {
 
     const cartesianPos = Ellipsoid.WGS84.cartographicToCartesian(cartographicPos, new Vector3());
 
-    scratchPlane.normal
-      .copy(cartesianPos)
-      .subtract(viewportCenterCartesian)
-      .scale(-1) // Want the normal to point into the frustum since that's what culling expects
-      .normalize();
-    scratchPlane.distance = Math.abs(scratchPlane.normal.dot(cartesianPos));
```
Pessimistress: This was the issue - `distance` should be `-scratchPlane .normal.dot(cartesianPos)`

PR #4066, opened 2019-12-21
Bump dependencies to production versions

Pessimistress:
For #3983


test/modules/geo-layers/tile-3d-layer/get-frame-state.spec.js
```diff
@@ -1,76 +0,0 @@
-// Copyright (c) 2015 - 2017 Uber Technologies, Inc.
```
Pessimistress: @xintongxia This needs to be moved to loaders.gl

PR #4079, opened 2019-12-31
Fix randomly failed icon layer render test

xintongxia:
For #4039 

#### Change List
- `IconManager`: add a `loaded` field indicating if the icons are fully loaded.

### DIFF
--- a/modules/layers/src/icon-layer/icon-manager.js
+++ b/modules/layers/src/icon-layer/icon-manager.js
@@ -185,7 +185,7 @@ export default class IconManager {
     this._externalTexture = null;
     this._mapping = {};
     // a counter recorded complete loading icons (including failed ones)
-    this._loadedIcons = 0;
+    this._pendingCount = 0;
 
     this._autoPacking = false;
 
@@ -241,8 +241,7 @@ export default class IconManager {
   }
 
   get loaded() {
-    const requestedIcons = Object.keys(this._mapping).length;
-    return this._autoPacking && requestedIcons && this._loadedIcons === requestedIcons;
+    return this._pendingCount === 0;
   }
 
   _updateIconAtlas(iconAtlas) {
@@ -315,6 +314,7 @@ export default class IconManager {
     const ctx = this._canvas.getContext('2d');
 
     for (const icon of icons) {
+      this._pendingCount++;
       loadImage(icon.url)
         .then(imageData => {
           const id = getIconId(icon);
@@ -339,7 +339,7 @@ export default class IconManager {
           log.error(error)();
         })
         .finally(() => {
-          this._loadedIcons++;
+          this._pendingCount--;
         });
     }
   }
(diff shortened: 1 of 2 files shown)


modules/layers/src/icon-layer/icon-manager.js
```diff
@@ -304,29 +307,33 @@ export default class IconManager {
...
 
-    for (const icon of icons) {
-      loadImage(icon.url).then(imageData => {
-        const id = getIconId(icon);
-        const {x, y, width, height} = this._mapping[id];
+    for (let i = 0; i < icons.length; i++) {
+      const imageData = await loadImage(icon.url);
 
-        const data = resizeImage(ctx, imageData, width, height);
+      if (i === icons.length - 1) {
```
Pessimistress: - This marks loaded when the last image is loaded, not when all images are loaded.
- What happens when the icon set changes?

xintongxia: > * What happens when the icon set changes?

reset `loaded` to false when there are more icons required to be loaded.

modules/layers/src/icon-layer/icon-manager.js
```diff
@@ -304,29 +307,36 @@ export default class IconManager {
...
-        const data = resizeImage(ctx, imageData, width, height);
-
-        this._texture.setSubImageData({
-          data,
-          x,
-          y,
-          width,
-          height
-        });
+      const imageData = await loadImage(icon.url);
```
Pessimistress: You are essentially only loading one image at a time.

modules/layers/src/icon-layer/icon-manager.js
```diff
@@ -307,8 +310,15 @@ export default class IconManager {
   _loadIcons(icons) {
     const ctx = this._canvas.getContext('2d');
 
+    let loaded = 0;
     for (const icon of icons) {
       loadImage(icon.url).then(imageData => {
+        loaded++;
```
Pessimistress: More scenarios to consider:

- `loaded` will never be set if one of the images fails to load, e.g. url is not found. Is that intended?
- This logic breaks if data is changed and a new batch of icons start loading. How do you handle that?

modules/layers/src/icon-layer/icon-manager.js
```diff
@@ -238,6 +240,11 @@ export default class IconManager {
     }
   }
 
+  get loaded() {
+    const requestedIcons = Object.keys(this._mapping).length;
+    return this._autoPacking && requestedIcons && this._loadedIcons === requestedIcons;
```
Pessimistress: I'd recommend tracking the number of pending requests rather than the number of fulfilled requests:

```js
this._pendingCount++;
loadSomething(...).finally(() => this._pendingCount--);
```

That way you don't need to know how to calculate the total number of requests here with `_autoPacking` and `_mapping`. The auto packing logic may change in the future and maintainers may forget to update this function, given that there are no tests. If requests are tracked where they start, it will be much more robust.

PR #4080, opened 2019-12-31
Improve deck.setProps perf

Pessimistress:
`deck.setProps` runs ~30% faster, saves ~6ms/60 frames.

#### Change List
- Merge initialization checks
- Avoid creating new objects

### DIFF
--- a/modules/core/src/lib/deck.js
+++ b/modules/core/src/lib/deck.js
@@ -127,6 +127,7 @@ const defaultProps = {
 export default class Deck {
   constructor(props) {
     props = Object.assign({}, defaultProps, props);
+    this.props = {};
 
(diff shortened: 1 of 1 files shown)


xintongxia:
> `deck.setProps` runs ~30% faster, saves ~6ms/60 frames.
> 
> #### Change List
> * Merge initialization checks
> * Avoid creating new objects

So changing `Object.assign({}, this.props, props)` to `Object.assign(this.props, props)` will improve the performance?

Pessimistress:
> So changing Object.assign({}, this.props, props) to Object.assign(this.props, props) will improve the performance?

Correct, no new object creation and no cloning the existing `this.props`. `Object.assign` iterate through all the keys and the `props` object is quite large .

PR #4081, opened 2020-01-01
Clean up hover handling

Pessimistress:
Move calling layer callbacks outside of `DeckPicker`. This makes the handling of hover events consistent with click, drag etc.

### DIFF
--- a/modules/core/src/lib/deck.js
+++ b/modules/core/src/lib/deck.js
@@ -532,7 +532,6 @@ export default class Deck {
       this.layerManager.context.mousePosition = {x: _pickRequest.x, y: _pickRequest.y};
     }
 
-    _pickRequest.callback = this.props.onHover;
     _pickRequest.event = event;
     _pickRequest.mode = 'hover';
   }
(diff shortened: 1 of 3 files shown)


modules/core/src/lib/deck.js
```diff
@@ -541,19 +540,28 @@ export default class Deck {
...
       if (this.props.getTooltip) {
         const displayInfo = this.props.getTooltip(pickedInfo);
         this.tooltip.setTooltip(displayInfo, pickedInfo.x, pickedInfo.y);
       }
-      if (_pickRequest.callback && !pickedInfo.handled) {
-        _pickRequest.callback(pickedInfo, _pickRequest.event);
+
+      // Execute callbacks
+      let handled = false;
+      if (pickedInfo.layer) {
```
xintongxia: how does it decide whether should call `onHover` or `onClick` to handle picking here?

Pessimistress: `_pickRequest` is only used for throttling hover events. Click is handled immediately in `onPointerDown`.

PR #4088, opened 2020-01-03
Update bundle config

Pessimistress:
Exclude all Node polyfills

Saves ~5k off minified bundle


Pessimistress:
@ibgreen I think `Buffer` (which was already excluded explicitly) was the biggest one. I saw `process` and `global` in the bundle analyzer prior to this PR. We use `probe.gl/env` in the new code bases but there might be dependencies that unintentionally brought them in. As far as I can tell there's no way to be sure other than manually testing the output.

PR #4092, opened 2020-01-03
update evaluate-children to pass deck gl view ports

georgios-uber:
to children with deckGLViewProps flag


modules/react/src/utils/evaluate-children.js
```diff
@@ -19,6 +19,8 @@ export default function evaluateChildren(children, childProps) {
     // Place map under the canvas
     childProps.style = MAP_STYLE;
     return cloneElement(children, childProps);
+  } else if (needsDeckGLViewProps(children)) {
```
Pessimistress: Did this pass lint? You don't need an `else` after `return`.

PR #4094, opened 2020-01-04
pydeck: Specify semver range for @deck.gl/jupyter-widget

ajduberstein:
For #3640 

Use a semver range instead of updating the exact version manually

#### Change List
-


ajduberstein:
Rolled into #4096

PR #4095, opened 2020-01-04
pydeck: Move project up a directory

ajduberstein:
For #3640 

Move pydeck up a directory level
Rename paths where necessary

#### Change List
-


ajduberstein:
Rolled into #4096

Issue #4104, opened 2020-01-06
Tracker: Geospatial data handling improvements


Pessimistress:
- [x] Offer the option to split features at the 180 meridian
- [x] Repeat maps at low zoom levels
- [ ] Fixed frame coordinates
- [x] Globe view
- [ ] Alternative 2D projections

PR #4105, opened 2020-01-06
Repeat maps at low zoom levels

Pessimistress:
For #4104

#### Background

Both Mapbox and Google Maps render multiple copies of the world at low zoom levels.

This PR adds a `repeat` option to `MapView` that mimics this behavior.
`repeat` is default to `false` to avoid breaking changes.

map-repeat

Under the hood, it adds support for `subViewports` to an `Viewport` instance. Sub viewports are used during redraw to achieve multiple renders without triggering `viewportChanged` layer updates.

Note that this option does not fix ALL longitude wrapping issues. If a geometry contains positions outside of the natural range ([-180, 180]) we may not render all copies that are visible in the current viewport.

#### Change List
- Add `WebMercatorViewport.subViewports`
- Support `subViewports` in `LayerPass`
- Documentation
- Unit and render tests


tsherif:
Will this end up rerendering the scene multiple times?

If so, could we not do this by rendering to a texture and repeating that?

Pessimistress:
> could we not do this by rendering to a texture and repeating that?

It would only make sense for 2D content.

PR #4110, opened 2020-01-07
pydeck: Reduce JupyterLab bundle size

ajduberstein:
For #4103

#### Change List
-

### DIFF
--- a/bindings/pydeck/README.md
+++ b/bindings/pydeck/README.md
@@ -117,6 +117,8 @@ python3 setup.py install
 
 ## Development
 
+Before 
+
 ```bash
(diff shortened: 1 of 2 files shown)


bindings/pydeck/README.md
```diff
@@ -117,16 +117,18 @@ python3 setup.py install
 
 ## Development
 
+Build the deck.gl project at the root of this repository:
```
Pessimistress: Is the documentation change related to the bundle config change?

ajduberstein: Much of it is–some changes are just for clarity

Issue #3804, opened 2019-10-20
Blurry Images with TileLayer + BitmapLayer


chrisgervang:
#### Description
I implemented the basic TileLayer usage example of OSM maps and Aviation maps and in both cases found the map tiles loading were very blurry.

Related to #3432, #3372

Actual
<img width="671" alt="Screen Shot 2019-10-19 at 9 52 44 PM" src="https://user-images.githubusercontent.com/2461547/67155032-71966a80-f2bb-11e9-88cb-f28c0fd3b774.png">
<img width="673" alt="Screen Shot 2019-10-19 at 9 51 14 PM" src="https://user-images.githubusercontent.com/2461547/67155034-78bd7880-f2bb-11e9-89eb-88e79e406acd.png">

Expected (Visible from [link](http://wms.chartbundle.com/charts/tms.html))
<img width="540" alt="Screen Shot 2019-10-19 at 9 53 23 PM" src="https://user-images.githubusercontent.com/2461547/67155033-73602e00-f2bb-11e9-9868-d6ce620aed6d.png">

I tried changed devicePixelRatio to true and false, didn't seem to effect.

#### Repro Steps
Created project based on https://github.com/uber/deck.gl/edit/7.3-release/examples/website/map-tile/app.js

See Implementation:
https://codesandbox.io/s/tilelayer-bitmaplayer-sji9h

#### Environment (please complete the following information):
 - Framework Version: deck.gl 7.3.3
 - Browser Version: Chrome 77.0
 - OS: Mac OS X 10.13

Issue #3432, opened 2019-08-12
Blurry map tiles in TileLayer example


MiKom:
#### Description
Tile layer renders blurry image (probably wrong zoom level is selected from tile server)

![image](https://user-images.githubusercontent.com/104767/62875953-3a886380-bd24-11e9-9895-da060f36fb42.png)

#### Repro Steps
Open https://deck.gl/#/examples/core-layers/tile-layer in latest Chrome

#### Environment (please complete the following information):
 - Framework Version: whatever is installed on Deck.gl homepage
 - Browser Version: Chrome 76.0.3809.100
 - OS: Windows 10 1903

#### Logs
    /#/examples/core-layers/tile-layer:1 Unchecked runtime.lastError: The message port closed before a response was received.

Issue #2931, opened 2019-04-09
TileLayer does not support COORDINATE_SYSTEM.IDENTITY?


mccalluc:
It looks like `TileLayer` assumes x and y are lat and long, and if they are out of range, it fails.
I'm starting with this:
```javascript
    return new TileLayer({
      id: 'tile-layer',
      coordinateSystem: COORDINATE_SYSTEM.IDENTITY,
      getTileData: ({ x, y, z }) => {
        console.log(x, y, z);
      },
    });
```
and I hit this error:
```
deck: error while initializing TileLayer({id: 'tile-layer'})
 Error: invalid latitude
    at assert (assert.js:3)
    at lngLatToWorld (web-mercator-utils.js:26)
    at getTileIndices (viewport-util.js:47)
    at TileCache.update (tile-cache.js:72)
    at TileLayer.updateState (tile-layer.js:112)
    at TileLayer._updateState (layer.js:799)
    at TileLayer._initialize (layer.js:767)
    at LayerManager._initializeLayer (layer-manager.js:700)
    at LayerManager._updateSublayersRecursively (layer-manager.js:631)
    at LayerManager._updateLayers (layer-manager.js:581)
```
It looks like [`viewport-util.js`](https://github.com/uber/deck.gl/blob/76291a07239343e9fe1de4e57b08317706facf8c/modules/geo-layers/src/tile-layer/utils/viewport-util.js#L44) assumes the coordinates are geographic:
```javascript
const [minX, minY] = lngLatToWorld([bbox[0], bbox[3]], viewport.scale).map(pixelsToTileIndex);
```
- Is my understanding of the problem correct?

I'm planning to copy out just these files, and see if adding a check for `IDENTITY` fixes this.
- Does that seem like a reasonable approach, and would you like a PR if I can get it working?

1chandu:
> Is my understanding of the problem correct?

Yes, AFAICT yes, this layer works with geographic data only CC: @Pessimistress 
Can you explain your use case to use non geographic data?

Pessimistress:
The TileLayer is designed around the geospatial use case. The tile indices (`x, y, z`) are defined by the OSM tile system. What do you expect the indices to be in the identity coordinate system?

mccalluc:
Thanks - Good point.

[...]

PR #4117, opened 2020-01-08
Non-Geospatial Tiling Usage

ilan-gold:
#### Background
I had a use case for non-geospatial tiling of images.  Therefore, this pull request contains the separation of the geospatial logic from the broader tiling logic.  Additionally it solves what appears to be a bug involving zoom level that caused blurriness
#### Change List
- Abstraction of `TileLayer` into `BaseTileLayer` to allow for custom tiling logic, which is implemented for the geospatial case under the `TileLayer` of `geo-layers` (#2931 )

- Fix outstanding blurriness bug (#3804 and #3432) by changing `Math.floor` to `Math.ceil` to fetch correct zoom level without affecting current view [here](https://github.com/uber/deck.gl/compare/master...ilan-gold:non_geo_tiling?expand=1#diff-747b0ebce440a45c3ec1cd2821af2d7fR41) and [here](https://github.com/uber/deck.gl/compare/master...ilan-gold:non_geo_tiling?expand=1#diff-385f7a7724ea41ba636b804e8978133fR118) (I would have split this off into its own pull request but it requires editing `BaseTileLayer`'s `getLayerZoomLevel`)

It's easier to see the difference if you click on the images:

**Before**

Screen Shot 2020-01-08 at 10 54 10 AM

**After**

Screen Shot 2020-01-08 at 10 54 48 AM

- Use correct tile size for OSM tiles (256x256, not 512x512) by exposing the parameter in the function used from `math.gl` [here](https://github.com/uber/deck.gl/compare/master...ilan-gold:non_geo_tiling?expand=1#diff-747b0ebce440a45c3ec1cd2821af2d7fR1-R11).  I can make a PR into `math.gl` if you would like about this, but I defer to you.

### DIFF
--- a/docs/layers/base-tile-layer.md
+++ b/docs/layers/base-tile-layer.md
@@ -1,0 +1,122 @@
+<p class="badges">
+  <img src="https://img.shields.io/badge/@deck.gl/geo--layers-lightgrey.svg?style=flat-square" alt="@deck.gl/geo-layers" />
+  <img src="https://img.shields.io/badge/lighting-yes-blue.svg?style=flat-square" alt="lighting" />
+</p>
+
+# BaseTileLayer
+
+This `BaseTileLayer` takes in a function `getTileData` that fetches tiles,`tileToBoundingBox` to calculate the bounds of each tile in the global cooridantes,
+`getTileIndices` to calculate the tile index given the current viewport, and `renderSubLayers` to render the tile in a layer.  This layer thus allows for non-geospatial tiling use-cases i.e with `COORDINATE_SYSTEM.IDENTITY`
+
+
+## Installation
+
+To install the dependencies from NPM:
+
+```bash
+npm install deck.gl
+# or
+npm install @deck.gl/core @deck.gl/layers @deck.gl/geo-layers
+```
+
+```js
+import {BaseTileLayer} from '@deck.gl/layers';
+new BaseTileLayer({});
+```
+
+To use pre-bundled scripts:
+
+```html
+<script src="https://unpkg.com/deck.gl@^8.0.0/dist.min.js"></script>
+<!-- or -->
+<script src="https://unpkg.com/@deck.gl/core@^8.0.0/dist.min.js"></script>
+<script src="https://unpkg.com/@deck.gl/layers@^8.0.0/dist.min.js"></script>
+<script src="https://unpkg.com/@deck.gl/geo-layers@^8.0.0/dist.min.js"></script>
+```
+
+```js
+new deck.TileLayer({});
+```
+
+
+## Properties
+
+Inherits from all [BaseLayer](/docs/api-reference/layer.md) properties and the following:
+
+##### `tileToBoundingBox` (Function, necessary)
+This function converts from `x, y, z` tile indices to a box in the global
+coordinates.
+
+Receives arguments:
+
+- `x` (Number)
+- `y` (Number)
+- `z` (Number)
+
+##### `getTileIndices` (Function, necessary)
+This function converts the given viewport to the indices needed to fetch tiles
+contained in the viewport.
+
+Receives arguments:
+
+- `viewport` (Viewport)
+- `minZoom` (Number) The minimum zoom level
+- `maxZoom` (Number) The maximum zoom level
+
+##### `maxZoom` (Number|Null, optional)
+
+Use tiles from this level when over-zoomed.
+
+- Default: `null`
+
+##### `minZoom` (Number, optional)
+
+Hide tiles when under-zoomed.
+
+- Default: 0
+
+##### `maxCacheSize` (Number|Null, optional)
+
+The maximum cache size for a tile layer. If not defined, it is calculated using the number of tiles in the current viewport times constant 5 (5 is picked because it's a common zoom range).
+
+- Default: `null`
+
+### Render Options
+
+##### `onViewportLoad` (Function, optional)
+
+`onViewportLoad` is a function that is called when all tiles in the current viewport are loaded. Data in the viewport is passed in as an array to this callback function.
+
+- Default: `onViewportLoad: (data) => null`
+
+##### `getTileData` (Function,  optional)
+
+`getTileData` given x, y, z indices of the tile, returns the tile data or a Promise that resolves to the tile data. For the default `renderSubLayers`, `bbox` is an object of `{west, north, east, south}`.
+
+- Default: `getTileData: ({x, y, z}) => Promise.resolve(null)`
+
+Receives arguments:
+
+- `x` (Number) The X coordinate of the tile index
+- `y` (Number) The Y coordinate of the tile index
+- `z` (Number) The Z coordinate of the tile index
+- `bbox` (Object, optional) - bounding box of the tile
+
+##### `onTileError` (Function, optional)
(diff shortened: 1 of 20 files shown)


modules/geo-layers/src/tile-layer/tile-layer.js
```diff
@@ -1,135 +1,21 @@
...
-  onTileError: {type: 'function', value: err => console.error(err), compare: false},
-  maxZoom: null,
-  minZoom: 0,
-  maxCacheSize: null
-};
+import {BaseTileLayer} from '@deck.gl/layers';
+import {tile2boundingBox} from './utils/tile-util';
+import {getTileIndices} from './utils/viewport-util';
 
 export default class TileLayer extends CompositeLayer {
```
Pessimistress: It might be cheaper to extend the `BaseTileLayer` and overwrite the default props.

modules/geo-layers/src/tile-layer/utils/viewport-util.js
```diff
@@ -18,10 +25,10 @@ function getBoundingBox(viewport) {
...
 }
 
-function getTileIndex(lngLat, scale) {
-  let [x, y] = lngLatToWorld(lngLat);
-  x *= scale / TILE_SIZE;
-  y = (1 - y / TILE_SIZE) * scale;
+function getTileIndex(lngLat, scale, tileSize) {
+  let [x, y] = lngLatToWorld(lngLat[0], lngLat[1], tileSize);
+  x *= scale / tileSize;
+  y = (1 - y / tileSize) * scale;
```
Pessimistress: `tileSize` does not matter here because you removed it right away. The old `getTileIndex` implementation should return the same result.

modules/layers/src/base-tile-layer/base-tile-layer.js
```diff
@@ -0,0 +1,156 @@
+import {log} from '@deck.gl/core';
+import {CompositeLayer} from '@deck.gl/core';
+import GeoJsonLayer from '../geojson-layer/geojson-layer';
+import TileCache from './utils/tile-cache';
+
+const defaultProps = {
+  renderSubLayers: {type: 'function', value: props => new GeoJsonLayer(props), compare: false},
+  getTileData: {type: 'function', value: ({x, y, z}) => Promise.resolve(null), compare: false},
+  tile2boundingBox: {
```
Pessimistress: We usually spell the name out in public APIs (`tileToBoundingBox` or `getTileBoundingBox` to align with the others)

modules/layers/src/base-tile-layer/base-tile-layer.js
```diff
@@ -0,0 +1,156 @@
...
+    type: 'function',
+    value: ({x, y, z}) => {
+      return {east: 0, west: 0, north: 0, south: 0};
+    }
+  },
```
Pessimistress: - I'd leave the default value to `null` since the prop has to be supplied for the layer to function.
- Add `compare: false` to avoid rerender on shallow change (happens in React if the prop is defined inline).

ilan-gold: Sounds good

modules/layers/src/base-tile-layer/base-tile-layer.js
ilan-gold: I think the reason I did this was because of the former name: https://github.com/uber/deck.gl/pull/4117.  I can change the name of the props in the public API - should I also change the name in `modules/geo-layers/src/tile-layer/utils/tile-util.js` (which is passed in as the prop) or leave it to match the old name, which was `tile2boundingBox` but not exposed as a prop?

modules/geo-layers/src/tile-layer/utils/viewport-util.js
ilan-gold: Thanks, the math definitely is redundant.  Should we just hardcode the `tileSize` prop for `TileLayer` then to be 512 to match the `math.gl` [implementation](https://github.com/uber-web/math.gl/blob/master/modules/web-mercator/src/web-mercator-utils.js#L15) even though it could be incorrect for the actual tiles?  I don't think the parameter is used anywhere else in the `geo-layers` `TileLayer` implementation so being "incorrect" won't have any real impact. I can also leave a comment.

modules/geo-layers/src/tile-layer/tile-layer.js
ilan-gold: This was a point I went back and forth on.  I landed on trying to match the other layers in `geo-layers` (which all render sublayers) but of course this layer only renders one sublayer while those generally did more. I will change it, then.

modules/geo-layers/src/tile-layer/utils/viewport-util.js
Pessimistress: Yeah I think hard coding it would be fine since it's just for this layer. `512` is needed by the `WebMercatorViewport` to match Mapbox projection.

Ultimately, `tileSize` is not needed by any of the public APIs, because `getTileIndices` and `getTileBoundingBox` would effectively determine the size of the tiles.

modules/layers/src/base-tile-layer/base-tile-layer.js
Pessimistress: `tile-util.js` is private so you can rename anything as you please.

test/modules/geo-layers/tile-layer/viewport-util.spec.js
```diff
@@ -90,7 +105,7 @@ function getTileIds(tiles) {
 
 test('getTileIndices', t => {
   for (const testCase of TEST_CASES) {
-    const result = getTileIndices(testCase.viewport, testCase.maxZoom, testCase.minZoom);
+    const result = getTileIndices(testCase.viewport, testCase.maxZoom, testCase.minZoom, TILE_SIZE);
```
Pessimistress: Can drop this now?

ilan-gold: Yes

ilan-gold:
I can add some documentation, yes.

Pessimistress:
@ilan-gold Can you also sign the CLA? The link is in the first comment above.

ilan-gold:
Sorry, I forgot to mention that the link seems to be dead.  Maybe I'm missing something, though.

docs/layers/base-tile-layer.md
```diff
@@ -0,0 +1,122 @@
...
+Renders one or an array of Layer instances with all the `TileLayer` props and the following props:
+
+* `data`: Resolved from `getTileData`
+* `tile`: An object containing tile index `x`, `y`, `z`, and `bbox` of the tile. `bbox` contains the global coordinates of the boundary of the box the tile occupies on the screen to be passed down to the sublayer.
+
+- Default: `props => new GeoJsonLayer(props)`
+
+# Source
+
+[modules/geo-layers/src/tile-layer](https://github.com/uber/deck.gl/tree/master/modules/geo-layers/src/tile-layer)
```
Pessimistress:
```suggestion
[modules/layers/src/base-tile-layer](https://github.com/uber/deck.gl/tree/master/modules/layers/src/base-tile-layer)
```

Issue #4116, opened 2020-01-08
StaticMap child overrides Deck cursor in 8.0


dnass:
#### Description
When using a `StaticMap` child with deck.gl 7.x, the div that wraps the Mapbox layer is placed below the deck canvas, so the user's cursor is set by the deck's `cursor` property (`grab` by default). In 8.0.x, the Mapbox wrapper is above the deck canvas, leading to an unstyled cursor.

#### Repro Steps
CodeSandbox example [here](https://codesandbox.io/s/falling-dream-yohxw). Toggling the `deck.gl` dependency between 8.0.x and 7.x illustrates the issue.

#### Environment (please complete the following information):
 - Framework Version: deck.gl 8.0.x
 - Browser Version: Chrome 79.0, Firefox 71.0, Safari 13.0.4
 - OS: Mac OS 10.14.6

#### Logs
Not clear if this is related, but under 8.0.x the example logs the warning `Each child in a list should have a unique "key" prop.`, which does not appear under 7.x and goes away if a `key` prop is added to `StaticMap`. (Adding the key does not fix the cursor issue, however.)

Pessimistress:
Thanks for reporting. This seems to be an issue introduced by us reordering React components.

The `key` warning should be fixed by #4098. We will publish a patch containing both fixes later today.

dnass:
So speedy! Thanks @Pessimistress

Pessimistress:
Fixed in 8.0.4

PR #4118, opened 2020-01-08
fix cursor style in React

Pessimistress:
For #4116

`canvas` is behind view containers in v8.0.

#### Change List
- Assign cursor style to container element

### DIFF
--- a/modules/core/src/lib/deck.js
+++ b/modules/core/src/lib/deck.js
@@ -544,8 +544,9 @@ export default class Deck {
   }
 
   _updateCursor() {
-    if (this.canvas) {
-      this.canvas.style.cursor = this.props.getCursor(this.interactiveState);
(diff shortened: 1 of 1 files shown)


modules/core/src/lib/deck.js
```diff
@@ -544,8 +544,9 @@ export default class Deck {
   }
 
   _updateCursor() {
-    if (this.canvas) {
-      this.canvas.style.cursor = this.props.getCursor(this.interactiveState);
+    const container = this.props.parent || this.canvas;
+    if (container) {
+      container.style.cursor = this.props.getCursor(this.interactiveState);
```
xintongxia: Is prop `parent` documented?

Pessimistress: Not right now, it is used as an internal prop by the react component.

Issue #4114, opened 2020-01-08
Using luma.gl buffer as data attributes


Darune:
Seeing binary data support in v8 described here: 
https://deck.gl/#/documentation/developer-guide/performance-optimization?section=supply-attributes-directly

I'm trying to use luma.gl Transform API do to some preprocess on my features.
As I understand it, creating new Buffer is not optimal, so I am trying to use the Transform output buffer directly as described in the documentation :

```
new PointCloudLayer({
    data: {
      length : data.pointCount,
      attributes: {
        getPosition: {buffer, size: 3, offset: 0, stride: 24},
(...)
``` 
But supplying attributes this way gives me the following error: 
`Error: invalid getPosition`

This assert seems to be raised from 
`src/lib/attribute/attribute.js:249`
```
(...)
assert(ArrayBuffer.isView(buffer.value), `invalid ${settings.accessor}`);
(...)
```

in my case buffer.value is not defined, so i'm not really sure what i'm doing wrong here.

I simplified my use case to a minimal .html joined here.
[buffer.txt](https://github.com/uber/deck.gl/files/4033993/buffer.txt)

Am I using the feature correctly ? 

changing the `getPosition` attribute to a 
`{value: new Float32Array([2.3522, 48.8566, 10]), size: 3, offset: 0, stride: 12}`
seems to fix the issue, but in my case where I want to use the luma.gl Transform api, 
does this mean that the operation will copy the transform output twice `transform <-> my code <-> deck.gl's own luma buffer` ?

1chandu:
CC: @Pessimistress 

You should be able to set `Transform` output as external buffer to a Layer attribute.

[Heatmap layer](https://github.com/uber/deck.gl/blob/5c8a156671a780c11ed33831381ad5f9a1b22ca7/modules/aggregation-layers/src/heatmap-layer/heatmap-layer.js#L182) uses `attributes: {attributeName: buffer, ...}`  can you try this way.

Pessimistress:
@1chandu We are moving away from the `[attributeName]: buffer` convention. This seems to be a bug in the Attribute class.

Pessimistress:
Fixed in 8.0.5.

Darune:
Just tried it, thank you for the quick fix / release !

PR #4121, opened 2020-01-08
Fix support for luma.gl buffers as external attributes

Pessimistress:
For #4114 

#### Change List
- Do not throw (in most cases) when `value` is missing from external buffer
- Allow external buffer to override the `type` accessor

### DIFF
--- a/modules/core/src/lib/attribute/data-column.js
+++ b/modules/core/src/lib/attribute/data-column.js
@@ -216,7 +216,7 @@ export default class DataColumn {
       this.value = opts.value;
 
       // Copy the type of the buffer into the accessor
-      accessor.type = buffer.accessor.type;
+      accessor.type = opts.type || buffer.accessor.type;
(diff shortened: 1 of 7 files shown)


modules/core/src/lib/attribute/data-column.js
```diff
@@ -216,7 +216,7 @@ export default class DataColumn {
       this.value = opts.value;
 
       // Copy the type of the buffer into the accessor
-      accessor.type = buffer.accessor.type;
+      accessor.type = opts.type || buffer.accessor.type;
```
1chandu: Once `Buffer` is created, `type` is fixed isn't it? why do we have to overwrite this? Is this covering non external buffer case?

Pessimistress: `buffer.accessor.type` is inferred from the data type array that is used to populate it by luma, but the chunk of memory can be interpreted as any data type in WebGL. This is needed if you use interleaved buffer with mixed data types.

PR #4122, opened 2020-01-08
Fix Tesselator handling of luma.gl buffers

Pessimistress:
For #4114 

#### Change List
- Do not throw if `geometryBuffer.value` is missing
- Deduce `instanceCount` from luma.gl Buffer instance
- Unit tests

### DIFF
--- a/modules/core/src/utils/tesselator.js
+++ b/modules/core/src/utils/tesselator.js
@@ -103,7 +103,7 @@ export default class Tesselator {
     const value = geometryBuffer.value || geometryBuffer;
     assert(ArrayBuffer.isView(value), 'cannot read geometries');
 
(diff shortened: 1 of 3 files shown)


modules/core/src/utils/tesselator.js
```diff
@@ -101,7 +100,10 @@ export default class Tesselator {
   }
 
   getGeometryFromBuffer(geometryBuffer) {
```
1chandu: Nit: If `geometryBuffer` is not always a `Buffer` object, consider renaming this method?

Pessimistress: We call both ArrayBuffer and luma Buffer `Buffer` in this code base.

I will do some refactor in a follow up PR after the patch.

PR #4126, opened 2020-01-09
Add offset feature to PathStyleExtension

Pessimistress:
See updated documentation.

#### Change List
- Add `offset` option to `PathStyleExtension`.


modules/extensions/src/path-style/shaders.glsl.js
```diff
@@ -57,3 +57,32 @@ float round(float x) {
...
+`,
+    'fs:#main-start': `
+  float isInside;
+  float offsetScale = (abs(vOffset * 2.0) + 1.0);
+  if (vOffset >= 0.0) {
+    isInside = step(vOffset * 2.0 - 1.0, vPathPosition.x * offsetScale);
+  } else {
+    isInside = step(vPathPosition.x * offsetScale, vOffset * 2.0 + 1.0);
+  }
+  if (isInside == 0.0) {
```
tsherif: Nice sculpting :)

docs/api-reference/extensions/path-style-extension.md
```diff
@@ -79,6 +80,16 @@ The dash array to draw each path with: `[dashSize, gapSize]` relative to the wid
 Only effective if `getDashArray` is specified. If `true`, adjust gaps for the dashes to align at both ends.
 
 
+##### `getOffset` ([Function](/docs/developer-guide/using-layers.md#accessors)|Number)
+
+Must be specified if the `offset` option is enabled.
+
+The offset to draw each path with, relative to the width of the path. Negative offset is to the right hand side, and positive offset is to the left hand side. `0` extrudes the path so that it is centered at the specified coordinates.
```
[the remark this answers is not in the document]
Pessimistress: No it just the uv happens to be this way... we can make a breaking change.

modules/extensions/src/path-style/shaders.glsl.js
Pessimistress: Eventually I want to use `smoothedge` and apply the result to the alpha. Right now this is not possible because I can only safely modify the color via the `FILTER_COLOR` hook, in which I don't have access to `vPathPositions`.

Pessimistress:
@rolyatmax Rebased #4111 and fixed miter and rounded joints. There still are issues with super sharp angles and high width/length ratios, but most cases should be fine.

modules/extensions/src/path-style/shaders.glsl.js
```diff
@@ -57,3 +57,27 @@ float round(float x) {
...
+`,
+    'vCornerOffset = offsetVec;': `
+  float offsetWidth = abs(instanceOffsets * 2.0) + 1.0;
+  vec2 offsetCenter = -instanceOffsets * (isCap ? perp : miterVec * miterSize) * 2.0;
+  vCornerOffset = vCornerOffset * offsetWidth - offsetCenter;
+`,
+    'fs:#main-start': `
+  float isInside;
+  isInside = step(-1.0, vPathPosition.x) * step(vPathPosition.x, 1.0);
+  if (isInside == 0.0) {
```
rolyatmax: Would this be easier to understand?:
```glsl
if (vPathPosition.x < -1.0 || vPathPosition.x > 1.0) {
  discard;
}
```
or even this?:
```glsl
bool isInside = vPathPosition.x < -1.0 || vPathPosition.x > 1.0;
```

Pessimistress: I need to add manual anti-aliasing here later (waiting for a change in luma) so ` isInside` can be between 0 and 1, in which case it will be multiplied with the output alpha.

rolyatmax: Ahhh gotcha

PR #4135, opened 2020-01-10
Error handling

Pessimistress:
For #4134

#### Change List
- Do not throw during layer lifecycles: initialize, update, finalize, draw
- Switch error logging from `log.warn` to `log.error`
- Add `onError` callback to `Deck` class (see updated documentation)
- Test


Pessimistress:
@tsherif We do something similar with props validation, based on the `debug` flag. We could potentially default `debug` to `true` in non-production mode?

PR #4142, opened 2020-01-13
Reorg dev docs

Pessimistress:
The existing RFC catalog is out of date and many links are dead. Many RFCs do not belong in the version directory that they reside in. This PR proposes simplified rules for organizing RFCs:

- All new RFCs are added to the `/dev-docs/RFCs/proposals` directory
- Once implemented, an RFC gets moved to its corresponding `/dev-docs/RFCs/<version>` directory
- No more catalog (top-level README)


Pessimistress:
> is the idea that an RFC is considered approved once it's merged in?

We don't currently have a process to approve RFCs outside of PRs. An RFC is accepted as a proposal when it's originally added. It's "approved" when the implementation is merged in, in which case the RFC should be moved to a release directory.

Pessimistress:
Discussed with @ibgreen, I will investigate an automated catalog/summary generator.

Algirdyz:
Hey,

I wonder if this ever got anywhere?
It would be really useful feature for me.

PR #4151, opened 2020-01-14
IconLayer: fix copy texture data when resize

xintongxia:
#### Background
for https://github.com/uber/deck.gl/issues/4146 and related to https://github.com/uber/deck.gl/issues/4039

#### Change List
- copy texture data without flipping Y when resize
- increase icon size in render test
- fix icon edge artifact


test/render/test-cases.js
```diff
@@ -791,35 +791,36 @@ export const TEST_CASES = [
...
-        sizeScale: 12,
+        sizeScale: 24,
         getPosition: d => d.COORDINATES,
         getColor: d => [64, 64, 72],
+        getSize: d => (d.RACKS > 2 ? 2 : 1),
         getIcon: d => {
           if (d.PLACEMENT === 'SW') {
             return {
-              url: './test/render/golden-images/pointcloud-meter.png',
+              url: './test/render/golden-images/examples/website_screen-grid.png',
```
Pessimistress: Can we not use a golden image as icon? They may change in the future for reasons unrelated to this layer.

test/render/test-cases.js
```diff
@@ -791,35 +791,36 @@ export const TEST_CASES = [
...
           return {
-            url: './test/render/golden-images/h3-hexagon-flat.png',
+            url: './test/render/golden-images/examples/website_geojson.png',
             width: 1024,
             height: 1024,
             anchorY: 1024,
             mask: false
           };
-        },
-        onAfterRender: ({layers, done}) => {
```
Pessimistress: Is this why the test was failing? 😒

xintongxia: yes

PR #4157, opened 2020-01-15
Refactor render tests; use stricter pass criteria

Pessimistress:


1chandu:
> Can we just define these props (`*aggregation`, `get*Weight`) inline? It's really difficult to track down their values in a big file.

Updated, also added couple of new tests for GridLayer elevation.

PR #4158, opened 2020-01-16
Fix H3HexagonLayer update when viewport jumps

Pessimistress:
For https://github.com/keplergl/kepler.gl/issues/887

#### Change List
- Check if `h3Distance` fails
- Tests


heshan0131:
Can this be in 7.3?

Pessimistress:
> Can this be in 7.3?

Yes we can patch 7.3 and 8.0

PR #4167, opened 2020-01-17
[Binary support, part 1] pydeck: Binary transport

ajduberstein:
For #4103

#### Change List
- Support binary transit using numpy
- Modify `pydeck.bindings.Layer` to take an argument indicating binary transit
- Expose functions for snake-case/camel-case conversion
- Test binary data conversion


bindings/pydeck/pydeck/bindings/layer.py
```diff
@@ -30,10 +30,12 @@ def __init__(
 
         type : str
             Type of layer to render, e.g., `HexagonLayer`
-        id : str
+        id : str, default None
             Unique name for layer
         data : str or list of dict of {str: Any} or pandas.DataFrame
             Either a URL of data to load in or an array of data
+        binary_transport : bool, default None
```
twojtasz: what is the convention here.  I like **is_binary_transport**

bindings/pydeck/pydeck/bindings/layer.py
```diff
@@ -1,22 +1,22 @@
...
+
 class Layer(JSONMixin):
-    def __init__(
-        self,
-        type,
-        data,
-        id=None,
-        **kwargs
-    ):
+    def __init__(self, type, data, id=None, binary_transport=None, **kwargs):
```
twojtasz: Is the boolean **binary_transport** needed.  Does this just come down to a **BytesIO** object (or bytes like that we can deduce from the type at runtime?)

twojtasz:
What is the guiding principle between to use binary or not?
The input for **data** seems to be a panda data frame or something else (dict?)

In the data frame case, if we can convert a column to binary why would we *not* want to do that.

I guess my question is on the interface, why bother with the boolean option and just if possible convert to binary if we know we can.

bindings/pydeck/pydeck/bindings/layer.py
ajduberstein: Right now I'm using numpy + pandas to convert to guarantee that we get a specific TypedArray on the client, so a BytesIO object may not be enough.

bindings/pydeck/pydeck/bindings/layer.py
ajduberstein: This is a better name

ajduberstein:
At the moment, there are a few edge cases around use that I haven't yet thoroughly address–

- Using the @deck.gl/json expression parser with accessors
- Guaranteeing that data provided to accessors are appropriately formatted (e.g. `getColor`, `getFillColor`, `getLineColor` values are normalized by dividing by 255)
- Complexity around flattened vs nested data frame columns in the data provided.
- Two way data passing: Not yet configured

By specifying `is_binary_transport=True` the user acknowledges that they're exposing a different feature set within pydeck.

I'll document the requirements and have binary conversion be non-default behavior at least in 0.3.0.

bindings/pydeck/pydeck/bindings/layer.py
```diff
@@ -30,10 +30,12 @@ def __init__(
 
         type : str
             Type of layer to render, e.g., `HexagonLayer`
-        id : str
+        id : str, default None
             Unique name for layer
         data : str or list of dict of {str: Any} or pandas.DataFrame
             Either a URL of data to load in or an array of data
+        use_binary_transport : bool, default None
+            Boolean indicating binary data
```
twojtasz: This only works with the pandas.DataFrame correct? Maybe add that to the comment, along with any other limitations (strings etc).

bindings/pydeck/pydeck/bindings/layer.py
```diff
@@ -89,35 +90,79 @@ def __init__(
...
+    def get_binary_data(self):
+        if not self.use_binary_transport:
+            raise BinaryTransportException(
+                "Layer must be flagged with `use_binary_transport=True`"
+            )
+        return self._binary_data
+
+    def _prepare_binary_data(self, data_set):
+        # Binary format conversion gives a sizable speedup but requires
+        # slightly stricter standards for data input
```
twojtasz: are these standards documented.

bindings/pydeck/pydeck/bindings/layer.py
```diff
@@ -89,35 +90,79 @@ def __init__(
...
+        if not is_pandas_df(data_set):
+            raise BinaryTransportException(
+                "Layer data must be a `pandas.DataFrame` type"
+            )
+
+        layer_accessors = self._kwargs
+        inv_map = {v: k for k, v in layer_accessors.items()}
+
+        blobs = []
+        for column in data_set.columns:
```
twojtasz: What happens if the JSON references something in the binary data that doesn't exist?  I was looking for a test case for this path but I don't see it.

PR #4169, opened 2020-01-17
[Binary support, part 2] @deck.gl/jupyter-widget: Process binary data within DeckGLModel

ajduberstein:
Change `npm run watch` for faster development within JupyterLab, which can call `jlpm run watch` on the package.json here.

Incorporate binary data transfer within the DeckGLWidget widget.

For #4103

### DIFF
--- a/modules/jupyter-widget/src/binary-transport.js
+++ b/modules/jupyter-widget/src/binary-transport.js
@@ -1,3 +1,5 @@
+import {jsonConverter} from './create-deck';
+
 // eslint-disable-next-line complexity
 function dtypeToTypedArray(dtype, data) {
   // Supports converting a numpy-typed array to a JavaScript-typed array
@@ -91,24 +93,6 @@ function makeLayerAccessorPropFunction({width, accessor}) {
   };
 }
 
-function getLayer(deckLayers, layerId) {
-  // Given a layer id, returns a deck.gl layer from a list of layers
-  for (const layer of deckLayers) {
-    if (layer.id === layerId) {
-      return layer;
-    }
-  }
-  return null;
-}
-
-function getCurrentLayerProps(layer) {
-  // Gets a deck.gl layer's current constructor and props
-  // TODO Uses proto, is there a way around this?
-  const LayerConstructor = layer.__proto__.constructor; // eslint-disable-line
-  const layerProps = layer.props;
-  return {LayerConstructor, layerProps};
-}
-
 function getAccessorNamesFrom(dataBuffer, layerId) {
   // Given data transferred from pydeck's backend,
   // get the names of the accessors using binary data on a particular layer
@@ -143,22 +127,19 @@ function makeBinaryAccessorProps(accessors, layerId, dataBuffer) {
   return dataAccessorProps;
 }
 
-function processDataBuffer({currentLayers, dataBuffer}) {
-  const updatedLayers = [];
-  const updateLayerIds = Object.keys(dataBuffer);
-  for (const layerId of updateLayerIds) {
-    const currentLayer = getLayer(currentLayers, layerId);
-    const newLayerData = constructData(dataBuffer, layerId);
-    const {LayerConstructor, layerProps} = getCurrentLayerProps(currentLayer);
-    const accessors = getAccessorNamesFrom(dataBuffer, layerId);
-    const accessorProps = makeBinaryAccessorProps(accessors, layerId, dataBuffer);
-    const combinedProps = {...layerProps, ...accessorProps};
-    combinedProps.data = newLayerData;
-    const layer = new LayerConstructor({...combinedProps});
-    updatedLayers.push(layer);
+function processDataBuffer({dataBuffer, jsonProps}) {
+  const convertedJsonProps = jsonConverter(jsonProps);
+  for (let i = 0; i < convertedJsonProps.layers.length; i++) {
+    const layer = convertedJsonProps.layers[i];
+    layer.data = constructData(dataBuffer, layer.id);
+    // use https://deck.gl/#/documentation/deckgl-api-reference/layers/layer?section=state-object- instead
+    const accessors = getAccessorNamesFrom(dataBuffer, layer.id);
+    const accessorProps = makeBinaryAccessorProps(accessors, layer.id, dataBuffer);
+    for (const accessor of accessors) {
+      layer[accessor] = accessorProps[accessor];
+    }
   }
-
-  return updatedLayers;
+  return convertedJsonProps;
 }
 
 export {deserializeMatrix, processDataBuffer};
--- a/modules/jupyter-widget/src/binary-transport.js
+++ b/modules/jupyter-widget/src/binary-transport.js
@@ -1,5 +1,9 @@
 import jsonConverter from './create-deck';
 
+// Functions to wrangle data from pydeck's row major order matrix dataframe to the deck.gl binary data format.
+// The format used is described here:
+// https://deck.gl/#/documentation/developer-guide/performance-optimization?section=supply-attributes-directly
+
 // eslint-disable-next-line complexity
 function dtypeToTypedArray(dtype, data) {
   // Supports converting a numpy-typed array to a JavaScript-typed array
@@ -81,16 +85,14 @@ function deserializeMatrix(arr, manager) {
   return renderable;
 }
 
-function makeLayerAccessorPropFunction({width, accessor}) {
+function makeDataLayerAttribute({width, accessor, dataframe}) {
   // Given a specific accessor name as a string and width as an integer,
-  // returns a function that can be used for lookups on the data at data.src
-  // data.src is a layer-specific data dictionary keyed by accessor
-  return (object, {index, data, target}) => {
-    for (let j = 0; j < width; j++) {
-      target[j] = data.src[accessor][index * width + j];
-    }
-    return target;
-  };
+  // returns a function that can be used for lookups on the dataframe
+  // here dataframe.src is a layer-specific data dictionary keyed by accessor
+  const dataPropAttribute = {};
+  dataPropAttribute.size = width;
+  dataPropAttribute.value = dataframe[accessor];
+  return dataPropAttribute;
 }
 
 function getAccessorNamesFrom(dataBuffer, layerId) {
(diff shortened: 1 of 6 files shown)


modules/jupyter-widget/src/binary-transport.js
```diff
@@ -143,22 +127,19 @@ function makeBinaryAccessorProps(accessors, layerId, dataBuffer) {
...
-    combinedProps.data = newLayerData;
-    const layer = new LayerConstructor({...combinedProps});
-    updatedLayers.push(layer);
+function processDataBuffer({dataBuffer, jsonProps}) {
+  const convertedJsonProps = jsonConverter(jsonProps);
+  for (let i = 0; i < convertedJsonProps.layers.length; i++) {
+    const layer = convertedJsonProps.layers[i];
+    layer.data = constructData(dataBuffer, layer.id);
+    // use https://deck.gl/#/documentation/deckgl-api-reference/layers/layer?section=state-object- instead
+    const accessors = getAccessorNamesFrom(dataBuffer, layer.id);
```
Pessimistress: You do not need to specify accessors when using binary data. They will not be called.

ajduberstein: They turn the accessor values into functions like [in the docs here](https://deck.gl/#/documentation/developer-guide/performance-optimization?section=use-binary-data).

```
    getFillColor: (object, {index, data, target}) => {
      return data.src.subarray(index * 6 + 3, index * 6 + 6);
    }
```

Pessimistress: There are multiple examples in that doc. You should follow the “directly supply attributes” one.

Pessimistress: There are multiple examples in that doc. You should follow the “directly supply attributes” one.

modules/jupyter-widget/src/binary-transport.js
```diff
@@ -79,34 +85,14 @@ function deserializeMatrix(arr, manager) {
...
-  return {LayerConstructor, layerProps};
+  // returns a function that can be used for lookups on the dataframe
+  // here dataframe.src is a layer-specific data dictionary keyed by accessor
+  const dataPropAttribute = {};
+  dataPropAttribute.size = width;
+  dataPropAttribute.value = dataframe[accessor];
+  return dataPropAttribute;
 }
 
 function getAccessorNamesFrom(dataBuffer, layerId) {
```
Pessimistress: There's no need to create an array from all the keys. Just use `for...in`.

modules/jupyter-widget/src/binary-transport.js
```diff
@@ -0,0 +1,117 @@
...
+// The format used is described here:
+// https://deck.gl/#/documentation/developer-guide/performance-optimization?section=supply-attributes-directly
+
+// eslint-disable-next-line complexity
+function dtypeToTypedArray(dtype, data) {
+  // Supports converting a numpy-typed array to a JavaScript-typed array
+  // based on a string value dtype and a DataView `data`
+  switch (dtype) {
+    case 'int8':
+      return new Int8Array(data);
```
Pessimistress:
```js
const ArrayType = dtypeToArrayType(dtype);
data = new ArrayType(data);
```

modules/jupyter-widget/src/binary-transport.js
```diff
@@ -0,0 +1,117 @@
...
+  if (!arr) {
+    return null;
+  }
+  const renderable = {};
+  for (const datum of arr.payload) {
+    // Each entry here represents a single column in
+    // a pandas.DataFrame arriving from the pydeck backend
+    const layerId = datum.layer_id;
+    const accessor = datum.accessor;
+    if (renderable[layerId] === undefined) {
```
Pessimistress: `!renderable[layerId]`

modules/jupyter-widget/src/binary-transport.js
```diff
@@ -0,0 +1,117 @@
...
+   *
+   * Objects at the JSON path `<accessor name>.matrix.data`
+   * are vectors representing a 1 x n matrix,
+   * or they can represent a m x n matrix in row major order form;
+   * shape of the matrix is given at <accessor name>.matrix.data.shape
+   */
+  if (!arr) {
+    return null;
+  }
+  const renderable = {};
```
Pessimistress: Use consistent naming - are `dataBuffer` `renderable` the same thing?

modules/jupyter-widget/src/binary-transport.js
```diff
@@ -0,0 +1,117 @@
...
+  for (const datum of arr.payload) {
+    // Each entry here represents a single column in
+    // a pandas.DataFrame arriving from the pydeck backend
+    const layerId = datum.layer_id;
+    const accessor = datum.accessor;
```
Pessimistress: `accessorName`. `accessor` refers to an object of shape `{value, size, stride, offset, ...}`.

modules/jupyter-widget/src/binary-transport.js
```diff
@@ -0,0 +1,117 @@
...
+}
+
+function constructDataProp(dataBuffer, layerId) {
+  const src = {length: 0, attributes: {}};
+  // For each accessor, we have a row major ordered matrix of data,
+  // represented as a single vector under `[accessor]matrix.data`
+  for (const accessor in dataBuffer[layerId]) {
+    const currentData = dataBuffer[layerId][accessor];
+    src.length = Math.max(currentData.matrix.shape[0], src.length);
+    src.attributes[accessor] = {
```
Pessimistress: Why the double conversion? Just get the objects into the right shape in `deserializeMatrix`?

modules/jupyter-widget/src/binary-transport.js
```diff
@@ -0,0 +1,117 @@
...
+
+function processDataBuffer({dataBuffer, jsonProps}) {
+  // Takes JSON props and combines them with the binary data buffer
+  jsonConverter.convert(jsonProps);
+  const convertedJsonProps = jsonConverter.convertedJson;
+  for (let i = 0; i < convertedJsonProps.layers.length; i++) {
+    const layer = convertedJsonProps.layers[i];
+    // Replace data on every layer prop
+    const data = constructDataProp(dataBuffer, layer.id);
+    const clonedLayer = layer.clone({data});
```
Pessimistress: `layer` here seems to be a plain JSON object. How does `clone` work?

Add unit tests.

ajduberstein: `layer.constructor.name` at this point gives a layer name, so I don't believe it's plain JSON. Unit tests are available in the latest update.

modules/jupyter-widget/src/binary-transport.js
ajduberstein: Yes that's correct, just to reshape the object.

Pessimistress: Why don't you get them into the correct shape the first time?

modules/jupyter-widget/src/binary-transport.js
Pessimistress: I see. You should not be calling `jsonConvertor.convert` in this function. One function should just handle one concern, and this one is to populate the JSON with binary data.

ajduberstein: convertedJsonProps are now received upstream from this function

modules/jupyter-widget/src/binary-transport.js
ajduberstein: This is reasonable. Let me modify the Python.

modules/jupyter-widget/src/binary-transport.js
```diff
@@ -0,0 +1,61 @@
...
+  }
+}
+
+function deserializeMatrix(obj, manager) {
+  if (!obj) {
+    return null;
+  }
+  for (const layerId in obj) {
+    for (const accessorName in obj[layerId].attributes) {
+      const {dtype, value} = obj[layerId].attributes[accessorName];
```
Pessimistress: Nit: assign `obj[layerId].attributes[accessorName]` to a variable

modules/jupyter-widget/src/binary-transport.js
```diff
@@ -0,0 +1,61 @@
...
+      const ArrayType = dtypeToTypedArray(dtype);
+      obj[layerId].attributes[accessorName].value = new ArrayType(value.buffer);
+      delete obj[layerId].attributes[accessorName].dtype;
```
Pessimistress: Theres no need to delete `dtype`, unrecognized fields are ignored.

test/modules/jupyter-widget/binary-transport.spec.js
```diff
@@ -0,0 +1,92 @@
...
+  const dv = new DataView(t.buffer, t.byteOffset, t.byteLength);
+  return dv;
+}
+
+const EXAMPLE_TRANSFER = {
+  'layer-id': {
+    length: 2,
+    attributes: {
+      getPosition: {
+        value: getDataView(),
```
Pessimistress: No need to use a function to create a constant.

modules/jupyter-widget/src/widget.js
```diff
@@ -119,6 +124,20 @@ export class DeckGLView extends DOMWidgetView {
...
+
+  dataBufferChanged() {
+    if (this.model.get('data_buffer')) {
+      const propsWithBinary = processDataBuffer({
+        dataBuffer: this.model.get('data_buffer'),
+        jsonProps: this.model.get('json_input')
+      });
+
+      jsonConverter.convert(propsWithBinary);
+      this.deck.setProps(jsonConverter.convertedJson);
```
Pessimistress: The cached conversion result should be private. Use the return value of `jsonConverter.convert`.

modules/jupyter-widget/src/binary-transport.js
```diff
@@ -0,0 +1,61 @@
...
+function deserializeMatrix(obj, manager) {
+  if (!obj) {
+    return null;
+  }
+  for (const layerId in obj) {
+    const attributes = obj[layerId].attributes;
+    for (const accessorName in attributes) {
+      const {dtype, value} = attributes[accessorName];
+      const ArrayType = dtypeToTypedArray(dtype);
+      obj[layerId].attributes[accessorName].value = new ArrayType(value.buffer);
```
Pessimistress:
```suggestion
      attributes[accessorName].value = new ArrayType(value.buffer);
```

PR #4170, opened 2020-01-18
[TileLayer] fix tile indices generation in edge cases

Pessimistress:
When the viewport's longitude is close to the 180th meridian, the bounding box will look something like

|  | minX | maxX |
|--|--|--|
| boundingBox | -200 | -150 |
| tileIndex (z=4) | -0.88 | 1.33 |

The old calculation clamps tileX to `[0, scale]` which was incorrect. tileX should loop between `[0, scale]` as the map repeats.

In the above case, the old calculation returns `[0, 0, 4]` and `[1, 0, 4]`. The correct result should be `[0, 0, 4]`, `[1, 0, 4]` and `[15, 0, 4]`.

#### Change List
- Fix bug in `getTileIndices`
- Unit tests

### DIFF
--- a/test/modules/geo-layers/tile-layer/viewport-util.spec.js
+++ b/test/modules/geo-layers/tile-layer/viewport-util.spec.js
@@ -82,7 +82,7 @@ const TEST_CASES = [
     title: 'z0 repeat',
     viewport: new WebMercatorViewport({
       width: 800,
-      height: 400,
+      height: 200,
       longitude: -90,
       latitude: 0,
       zoom: 0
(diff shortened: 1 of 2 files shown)


test/modules/geo-layers/tile-layer/viewport-util.spec.js
```diff
@@ -82,23 +82,35 @@ const TEST_CASES = [
...
 
 function getTileIds(tiles) {
-  const set = new Set();
+  const ids = [];
   for (const tile of tiles) {
-    set.add(`${tile.x},${tile.y},${tile.z}`);
+    ids.push(`${tile.x},${tile.y},${tile.z}`);
   }
-  return Array.from(set).sort();
+  return Array.from(ids).sort();
```
xintongxia: Why do you need create a copy of ids array?

Issue #4177, opened 2020-01-21
GeoJsonLayer doesn't use stroked and other properties in playground example


isaacbrodsky:
#### Description
I was using the deck.gl playground and trying to disable some parts of the GeoJsonLayer, and found that they did not disable rendering those parts.

I expected that the example JSON below would disable rendering the lines (paths) around the polygons because `stroked` was set to `false`, and that it would disable rendering the polygons themselves because both `filled` was set to `false` and `opacity` was set to `0.0`. In fact, it renders both the stroke and the fill. `opacity` was respected when it was >0.

![deckgl playground stroked false](https://user-images.githubusercontent.com/9139378/72846925-63f3d680-3c56-11ea-9517-ea14be23f4f3.png)

#### Repro Steps
In the deck.gl playground, I tried to use the following example JSON:
```
{
  "description": "A deck.gl get-started GeoJsonLayer example in JSON format",
  "initialViewState": {
    "latitude": 40,
    "longitude": -100,
    "zoom": 3,
    "bearing": 30,
    "pitch": 30
  },
  "map": true,
  "views": [
    {
      "@@type": "MapView",
      "controller": true
    }
  ],
  "layers": [
    {
      "@@type": "GeoJsonLayer",
      "data": "https://d2ad6b4ur7yvpq.cloudfront.net/naturalearth-3.3.0/ne_110m_admin_1_states_provinces_shp.geojson",
      "stroked": false,
      "filled": false,
      "lineWidthMinPixels": 2,
      "opacity": 0.0,
      "getLineColor": [
        0,
        0,
        0
      ],
      "getFillColor": [
        200,
        160,
        0,
        180
      ]
    }
  ]
}
```

#### Environment (please complete the following information):
 - Framework Version: On the playground and confirmed locally with deck.gl@8.0.1
 - Browser Version: Firefox 72.0.1, also happens on a few versions of Chrome
 - OS: Linux (Ubuntu Bionic), and others

#### Logs
N/a

ajduberstein:
Thanks for reporting this, @isaacbrodsky–it should be fixed by #4185.

Pessimistress:
Fix published in 8.0.10

PR #4185, opened 2020-01-22
@deck.gl/json: Fix bug dropping props with falsy values

ajduberstein:
For  #4177

#### Background

Boolean props on all layers created via @deck.gl/json appear to have been returning to their default value if specified as `false`.

The bug only affected situations where a prop's default value is `true`, and the user wished to deactivate it, e.g. `stroked` and `filled` exhibit this behavior in the GeoJsonLayer, while `extruded` does not, because of their [defaults](https://github.com/uber/deck.gl/blob/master/docs/layers/geojson-layer.md#filled-boolean-optional).

#### Change List
- Fix bug
- Add test to prevent regression

### DIFF
--- a/modules/json/src/helpers/convert-functions.js
+++ b/modules/json/src/helpers/convert-functions.js
@@ -26,11 +26,7 @@ export default function convertFunctions(props, configuration) {
       propValue = trimFunctionIdentifier(propValue);
       propValue = parseExpressionString(propValue, configuration);
     }
-
-    // Invalid functions return null, show default value instead.
-    if (propValue !== null) {
-      replacedProps[propName] = propValue;
-    }
+    replacedProps[propName] = propValue;
   }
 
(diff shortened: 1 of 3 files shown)


ajduberstein:
bugfix

Playground example running this PR

isaacbrodsky:
Does this also resolve the issue where setting `opacity` to `0` caused it to not be used?

modules/json/src/helpers/convert-functions.js
```diff
@@ -28,7 +28,7 @@ export default function convertFunctions(props, configuration) {
     }
 
     // Invalid functions return null, show default value instead.
-    if (propValue) {
+    if (propValue !== null) {
```
Pessimistress: Non-accessor props are allowed to be null. You must treat the return values of functions as its own special case.

PR #4193, opened 2020-01-23
[React] fix missing key error

goszczynskip:
Pull request #4098 partially fixed issue with missing key error. Context
element is used in array children and also misses key. This PR fully fixes
issue #4116.

### DIFF
--- a/modules/react/src/utils/position-children-under-views.js
+++ b/modules/react/src/utils/position-children-under-views.js
@@ -79,7 +79,7 @@ export default function positionChildrenUnderViews({children, viewports, deck, C
           deck._onViewStateChange(params);
         }
       };
-      return createElement(ContextProvider, {value: contextValue}, viewElement);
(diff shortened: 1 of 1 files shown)


modules/react/src/utils/position-children-under-views.js
```diff
@@ -79,7 +79,7 @@ export default function positionChildrenUnderViews({children, viewports, deck, C
           deck._onViewStateChange(params);
         }
       };
-      return createElement(ContextProvider, {value: contextValue}, viewElement);
+      return createElement(ContextProvider, {key: 'context', value: contextValue}, viewElement);
```
Pessimistress: This key is not unique.

Pessimistress: - Run `yarn lint fix` to apply prettier.
- You can just use the `key` variable in this function.

PR #4205, opened 2020-01-28
TextLayer improvements

Pessimistress:
#### Change List
- Move `getPixelOffset` to `IconLayer`, drop duplicate `MultiIconLayer` vertex shader
- Avoid pre-calculating and storing text layout information. This saves runtime memory and allows us to pass user provided `data` directly to the sublayer without datum and accessor wrapping, also makes it possible to use binary data.
- Use `startIndices` to leverage auto-packing for variable-width attributes
- Simplify `transformParagraph` - its return value is now only used to get the per-character offset, so we can drop a handful of irrelevant fields.
- Tests


b2kdaman:
@xintongxia does this address hovering events issue with inability to set text hovering area as rectangle, not the precise letters outline itself?

Pessimistress:
@b2kdaman if you are using 8.0 that is already the case.

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer.js
```diff
@@ -116,53 +95,16 @@ export default class MultiIconLayer extends IconLayer {
...
-    const {value, size} = attribute;
-    let i = startRow * size;
-    const pickingColor = [];
-    const {iterable} = createIterable(data, startRow, endRow);
-
-    for (const point of iterable) {
-      const index = getPickingIndex(point);
-      this.encodePickingColor(index, pickingColor);
+  getInstanceColorMode(icons) {
+    return 1; // mask
```
xintongxia: define a constant.

Pessimistress: It is not needed elsewhere.

Pessimistress: Context: colorMode is used as a boolean flag (`true` for masked, `false` for not) but because it is an attribute we have to fill the array with an int (`1` and `0`).

PR #4206, opened 2020-01-28
Support binary data in TextLayer

Pessimistress:
For #3758 

cc @ajduberstein @twojtasz 

#### Change List
- Add binary support to TextLayer
- Documentation
- Unit and render tests


xintongxia:
Why is this `Port website examples to ocular gatsby website (#4033)` included here?

Other than that, looks good to me.

Pessimistress:
> Why is this Port website examples to ocular gatsby website (#4033) included here?

Ah we force pushed to master to correct a commit message. I'll rebase.

PR #4207, opened 2020-01-28
tile-3d-layer: expose picked tile when enable picking

xintongxia:
#### Background
Enable picking feature for Tile 3D Layer, could used for debugging and rendering tile info

#### Change List
- change getPickingInfo in PointcloudLayer and ScenegraphLayer
- change tile.fullUri -> tile.id


modules/geo-layers/src/tile-3d-layer/tile-3d-layer.js
```diff
@@ -168,18 +180,20 @@ export default class Tile3DLayer extends CompositeLayer {
...
 
     const SubLayerClass = this.getSubLayerClass('scenegraph', ScenegraphLayer);
 
     return new SubLayerClass(
       {
-        _lighting: 'pbr'
+        _lighting: 'pbr',
+        pickable
       },
       this.getSubLayerProps({
```
Pessimistress: `getSubLayerProps` includes `pickable` from the parent layer.

modules/geo-layers/src/tile-3d-layer/tile-3d-layer.js
```diff
@@ -54,6 +54,18 @@ export default class Tile3DLayer extends CompositeLayer {
     }
   }
 
+  getPickingInfo({info}) {
+    const {layerMap} = this.state;
+    const sourceLayer = info.sourceLayer;
```
Pessimistress: Why do you need to add `sourceLayer` to `info`? Any reason not to use the `sourceLayer` param passed to this function?

modules/geo-layers/src/tile-3d-layer/tile-3d-layer.js
```diff
@@ -54,6 +54,17 @@ export default class Tile3DLayer extends CompositeLayer {
     }
   }
 
+  getPickingInfo({info, sourceLayer}) {
+    const {layerMap} = this.state;
+    const layerId = sourceLayer && sourceLayer.id;
+    if (layerId) {
+      const tileId = layerId.substring(this.id.length + 1).split('-')[1];
```
Pessimistress: What if `tileId` from some new loader contains `-`?

docs/layers/tile-3d-layer.md
```diff
@@ -122,6 +122,12 @@ const layer = new Tile3DLayer({
 })
 ```
 
+##### `pickable` (Boolean, Optional)
+
+- Default: false
+
+When enable picking, the picking info will contain the selected `tile` accessed by`info.object`.
```
Pessimistress: Link to `Tile3DHeader` documentation

modules/geo-layers/src/tile-3d-layer/tile-3d-layer.js
xintongxia: this should be resolved inside deck.gl `tile-3d-layer`, a better naming could be suggested. 

https://github.com/uber/deck.gl/blob/master/modules/geo-layers/src/tile-3d-layer/tile-3d-layer.js#L182

Pessimistress: If `tile.id` contains `-` you will end up with a partial id that does not match any thing. You need to take the rest of the string from the first `-`.

docs/layers/tile-3d-layer.md
```diff
@@ -122,6 +122,12 @@ const layer = new Tile3DLayer({
 })
 ```
 
+##### `pickable` (Boolean, Optional)
+
+- Default: false
+
+When enable [`picking`](/docs/developer-guide/custom-layers/picking.md), the picking info will contain the selected [`tile`](https://github.com/uber-web/loaders.gl/blob/master/modules/3d-tiles/docs/api-reference/tileset-3d.md) accessed by`info.object`.
```
Pessimistress: - You are linking to the `Tileset3D` class which is not what `tile` is.
- Link to the website please, master can contain unpublished changes.
- Wording: `info.object` will be a `Tile3DHeader` (link) object.

modules/geo-layers/src/tile-3d-layer/tile-3d-layer.js
xintongxia: layer id has three parts, `${this.id}-scenegraph-${tileHeader.fullUri}`

`layerId.substring(this.id.length + 1)` will give `scenegraph-${tileHeader.fullUri}`

`scenegraph-${tileHeader.fullUri}`.split('-')[1]` will give me the rest of the string after the `${tileHeader.fullUri}`.

docs/layers/tile-3d-layer.md
xintongxia: only the incomplete category spec `https://loaders.gl/docs/specifications/category-3d-tiles#tileheader-fields` is available at the moment.

PR #4211, opened 2020-01-29
Fix disappearing circles in the data-filter example

Pessimistress:
See inline comments for context.

We changed the default `farZMultiplier` from `10` to `1.01` in v8, making the issue more prominent than before.

This change still does not solve the clipping at high zoom levels, but should be sufficient for the purpose of this demo.


tgorkin:
This change seems fine, but did you consider pre-processing the data to remove the depth so that the data points are clamped to the ground plane?

Pessimistress:
> did you consider pre-processing the data to remove the depth so that the data points are clamped to the ground plane?

The z is not from the source data, but added in this app. If you rotate the map at z=0 it looks really cool.

PR #4227, opened 2020-02-04
pydeck: Enforce pydeck notebook deck.gl semver on sdist

ajduberstein:
For #4103 

#### Change List
-

### DIFF
--- a/bindings/pydeck/pydeck/frontend_semver.py
+++ b/bindings/pydeck/pydeck/frontend_semver.py
@@ -1,2 +1,1 @@
-"""Current associated version of NPM modules deck.gl and @deck.gl/jupyter-widget"""
 DECKGL_SEMVER = "^8.0.0"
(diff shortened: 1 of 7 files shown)


Pessimistress:
If I understand this correctly, it will block the ability to publish beta versions of pydeck that depends on a beta version of deck.gl.

ajduberstein:
> If I understand this correctly, it will block the ability to publish beta versions of pydeck that depends on a beta version of deck.gl.

There's nothing currently enforcing that a particular prerelease of pydeck depends on a prerelease of deck.gl, thought I can add that check. This just simply makes sure that a maintainer is intentional about the deck.gl release paired with pydeck.

bindings/pydeck/pydeck/frontend_semver.py
```diff
@@ -0,0 +1,2 @@
+"""Current associated version of NPM modules deck.gl and @deck.gl/jupyter-widget"""
```
ajduberstein: todo: read from lerna.json instead

ajduberstein: also check that the correct tag matches the current branch (beta/alpha from master only)

bindings/pydeck/pydeck/frontend_semver.py
```diff
@@ -0,0 +1 @@
+DECKGL_SEMVER = "^8.0.0"
```
Pessimistress: Add this to .gitignore if it's automatically generated

PR #4233, opened 2020-02-04
pydeck and @deck.gl/jupyter-widget: Add support for dynamic deck.gl custom Layer registration

ajduberstein:
For #4103

#### Change List
- Load custom scripts using XMLHttpRequest (works in Notebook, JupyterLab, and pure HTML environments)
- Layer registration


ajduberstein:
@Pessimistress working on a custom layer example here: https://github.com/ajduberstein/tagmap_layer_demo What should the webpack config look like?

The goal is to pass the URL for a custom layer to some function within @deck.gl/jupyter-widget and  then register it with the JSONConverter

ajduberstein:
This PR is currently working with the code from [this repo](https://github.com/ajduberstein/tagmap_layer_demo), which produces the bundle that I'm hosting [here](https://github.com/ajduberstein/ajduberstein.github.io/blob/master/layer-demo.js).

modules/jupyter-widget/src/dynamic-registration.js
```diff
@@ -0,0 +1,34 @@
...
+    xhr.send();
+  });
+}
+
+export default function addClassToConverter({jsonConverter, className, resourceUri}) {
+  loadScript(resourceUri).then(res => {
+    // Opinionated choice, requires that the user load only one layer at a time
+    // and that layer must be the sole default export of the library
+    // TODO better choice here?
+    const classConstructor = window[className].default;
```
Pessimistress: How is the user supposed to bundle their code? Add an example/test.

modules/jupyter-widget/src/dynamic-registration.js
```diff
@@ -0,0 +1,34 @@
...
+    const newConfiguration = {
+      classes: {}
+    };
+    newConfiguration.classes[className] = classConstructor;
+    jsonConverter.merge(newConfiguration);
```
Pessimistress: I don't see a `merge` method on `JSONConverter`. Add tests.

modules/jupyter-widget/src/create-deck.js
```diff
@@ -41,8 +42,13 @@ const jsonConverter = new deck.JSONConverter({
   configuration: jsonConverterConfiguration
 });
 
-export function updateDeck(inputJSON, deckgl) {
-  const results = jsonConverter.convert(inputJSON);
+export function updateClasses({className, resourceUri, onComplete}) {
```
Pessimistress: Why do we need a function that directly calls another? Where is `onComplete` used?

bindings/pydeck/pydeck/widget/_frontend.py
```diff
@@ -9,4 +9,4 @@
 module_name = "@deck.gl/jupyter-widget"
 # module_version is the current version of the module of the JS portion of the widget
 # It appears to be important only for JupyterLab and ignored for Jupyter Notebooks
-module_version = "^8.0.0"
+module_version = "8.1.0-alpha.1"
```
Pessimistress: Can we not hardcode any version?

modules/jupyter-widget/src/create-deck.js
```diff
@@ -41,19 +41,40 @@ const jsonConverter = new deck.JSONConverter({
   configuration: jsonConverterConfiguration
 });
 
-export function updateDeck(inputJSON, deckgl) {
-  const results = jsonConverter.convert(inputJSON);
+/* global document, fetch, window */
+function loadScript(resourceUri) {
+  const tag = document.createElement('script');
+  return Promise.resolve(
```
Pessimistress: Use

```js
tag.onload = resolve();
tag.src = resourceUri;
```

modules/jupyter-widget/src/create-deck.js
```diff
@@ -41,19 +41,40 @@ const jsonConverter = new deck.JSONConverter({
...
+  );
+}
+
+function addResourceToConverter({libraryName, resourceUri, onComplete}) {
+  loadScript(resourceUri)
+    .then(res => {
+      const module = window[libraryName];
+      const classes = Object.keys(module);
+      const newConfiguration = {classes: {}};
+      for (const c of classes) {
```
Pessimistress: Use `extractClasses`

modules/jupyter-widget/src/create-deck.js
```diff
@@ -41,19 +41,40 @@ const jsonConverter = new deck.JSONConverter({
...
+        newConfiguration.classes[c] = module[c];
+      }
+      jsonConverter.mergeConfiguration(newConfiguration);
+    })
```
Pessimistress: There are no async operations here, just call `onComplete()`

modules/jupyter-widget/src/index.js
```diff
@@ -31,5 +31,6 @@ module.exports = {
   MODULE_VERSION,
   MODULE_NAME,
   createDeck,
-  updateDeck
+  updateDeck,
+  addResourceToConverter
```
Pessimistress: Naming: `loadExternalClasses`?

modules/jupyter-widget/src/create-deck.js
ajduberstein: onComplete is used in the test associated with this function

modules/jupyter-widget/src/widget.js
```diff
@@ -125,9 +125,22 @@ export class DeckGLView extends DOMWidgetView {
...
+    this.model.on('change:classes', this.addNewCustomLayers.bind(this), this);
+
+    this.addNewCustomLayers();
     this.dataBufferChanged();
   }
 
+  addNewCustomLayers() {
+    if (this.model.get('classes')) {
+      const customLibraries = this.model.get('classes');
+      for (const obj of customLibraries) {
```
Pessimistress: What shape is `obj`? Add inline documentation.

modules/jupyter-widget/webpack.config.js
```diff
@@ -48,6 +48,7 @@ const config = [
       libraryTarget: 'umd'
     },
     devtool: 'source-map',
+    mode: 'development',
```
Pessimistress: Remove

test/modules/jupyter-widget/create-deck.spec.js
```diff
@@ -0,0 +1,48 @@
...
+  renderLayers() {
+    return new ScatterplotLayer(this.props);
+  }
+}
+
+// Mock the loading of the script
+const TEST_LIBRARY_NAME = 'DemoLibrary';
+window[TEST_LIBRARY_NAME] = {DemoCompositeLayer};
+const oldFetch = window.fetch;
+window.fetch = url => {
```
Pessimistress: You need to apply the override right before you use it. This will break other tests.

bindings/pydeck/pydeck/widget/_frontend.py
ajduberstein: Handled in a separate PR, this is only for development purposes, I'll change it back

modules/jupyter-widget/src/widget.js
```diff
@@ -125,9 +125,23 @@ export class DeckGLView extends DOMWidgetView {
...
+        // obj is an object of {libraryName: libraryUrl} e.g. {CustomLayersLibrary: 'https://example.com/customLayers.js'}
```
Pessimistress: Use `{libraryName: <string>, url: <string>}` for clarity

modules/jupyter-widget/src/create-deck.js
```diff
@@ -41,19 +41,35 @@ const jsonConverter = new deck.JSONConverter({
...
+  script.type = 'text/javascript';
+  script.src = url;
+  const head = document.querySelector('head');
+  head.appendChild(script);
+  return new Promise(resolve => {
+    script.onload = resolve;
+  });
+}
+
+function loadExternalClasses({libraryName, resourceUri, onComplete}, loadResource = loadScript) {
```
Pessimistress: Add a comment: `loadResource` is for testing only

modules/jupyter-widget/src/widget.js
```diff
@@ -125,9 +125,23 @@ export class DeckGLView extends DOMWidgetView {
...
+    this.addNewCustomLayers();
     this.dataBufferChanged();
   }
 
+  addNewCustomLayers() {
+    if (this.model.get('classes')) {
+      const customLibraries = this.model.get('classes');
+      for (const obj of customLibraries) {
+        // obj is an object of {libraryName: <string>, url: <string>}
+        const [libraryName, resourceUri] = Object.entries(obj)[0];
```
Pessimistress: `obj.libraryName`,
`obj.url`

bindings/pydeck/pydeck/settings.py
```diff
@@ -0,0 +1,28 @@
...
+    """Global settings for pydeck
+
+    custom_classes : list of dict
+        List of dictionaries of the format {'ClassName': 'deck.gl class URL'}.
+        For example, if there was a custom deck.gl Layer classed `TagmapLayer`
+        bundled for distribution at the path `https://demourl.libpath/bundle.js`,
+        one could load it into pydeck by doing the following:
+
+        ```
+        pydeck.settings.custom_classes = [{'TagmapLayer': 'https://demourl.libpath/bundle.js'}]
```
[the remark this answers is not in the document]
ajduberstein: The latest implementation of this allows for multiple classes. I'll be updated the docstring here to say `"tagmapLibrary": ...url...` instead.

modules/jupyter-widget/src/widget.js
ajduberstein: They're arriving as a list of dictionaries from the server in something like this format:

`[{"libraryName1": "url1"}, {"libraryName2": "url2"}]`

Pessimistress: Don’t you have control of the format on the server side?

modules/jupyter-widget/src/dynamic-registration.js
ajduberstein: The test is addressed in #4267 and I'll add documentation/an example directory elsewhere

modules/jupyter-widget/src/widget.js
ajduberstein: I do but I'm personally partial to the more concise form of 

`[{"libraryName1": "url1"}, {"libraryName2": "url2"}]`

instead of

`[{"libraryName": "libraryName1", "resourceUri": "url1"}, {"libraryName": "libraryName2", "resourceUri": "url2"}]`

It's anyone's choice, however–happy to switch it over.

Pessimistress: Think of each item as an entry in a database. Predictable data structured makes your code a lot more readable. Using the key name as the library name is implicit, and require additional documentation in the source. Explicitly calling out `libraryName` leaves no ambiguity.

If you insist on using keys as library names, which is fine, then there's no reason to have an array. All entries can be merged into one dictionary. `Object.entries(obj)[0]` is a bad pattern because you are relying on the existence and uniqueness of the object's content. It will easily crash if the input is malformed.

modules/jupyter-widget/src/widget.js
```diff
@@ -135,9 +135,7 @@ export class DeckGLView extends DOMWidgetView {
     const customLibraries = this.model.get('custom_libraries');
     if (customLibraries) {
       for (const obj of customLibraries) {
-        // obj is an object of {libraryName: <string>, url: <string>}
-        const [libraryName, resourceUri] = Object.entries(obj)[0];
-        loadExternalLibrary({libraryName, resourceUri});
+        loadExternalLibrary({libraryName: obj.libraryName, resourceUri: obj.resourceUri});
```
Pessimistress: `loadExternalLibrary(obj)`

ajduberstein:
Will merge after #4267 is merged

PR #4239, opened 2020-02-05
pydeck: Add background color configuration for standalone HTML renderer and PointCloudLayer example

ajduberstein:
For #4103

#### Change List
- Add background color support for standalone HTML rendering
- Use this change in the PointCloudLayer example


ajduberstein:
Closed in favor of #4243, #4242, and #4241

PR #4241, opened 2020-02-06
@deck.gl/jupyter-widget: Add sunlight effect

ajduberstein:
For #4103

#### Change List
-


modules/jupyter-widget/src/create-deck.js
```diff
@@ -17,6 +17,7 @@ function extractClasses() {
   for (const cls of classes) {
```
ajduberstein: todo: Allow for underscore and uppercase

Issue #4144, opened 2020-01-14
Google Maps overlay shows nothing when zoomed out too far


pwambach:
#### Description
We have a Google Map and use the Deck.Gl GoogleMapOverlay to sync the deck.gl layers. The problem is when zooming out so far that more than 360 degree of longitude is visible all deck.gl layers disappear.
The expected behavior would be that the points stay visible on the map.
Theoretically they should be duplicated if the same location on earth appears twice on screen but I guess this is another issue.

The console logs this warning `deck: Pixel project matrix not invertible`.

#### Repro Steps
I made a simple example on codepen: https://codepen.io/pwambach/pen/WNbKNzm
Depending on the screen resolution the two red dots around LatLng 0/0 disappear when zooming out 2 or 3 levels.

#### Environment (please complete the following information):
 - Framework Version: deck.gl 8.0.5
 - Browser Version: Chrome 79
 - OS: Mac OS X 10.15

Pessimistress:
This will be fixed in 8.1.0. You can test the feature with deck.gl@^8.1.0-alpha.

Issue #3576, opened 2019-09-11
Deck gl layers appearing on 'wrong' copy of world with Google Maps


SterlingMcCall:
#### Description
When displaying deck.gl layers on a google map, the layer 'disappears' when the center of the visible map crosses the anti-meridian. I say 'disappears' because what is actually happening is the deck.gl layer is prematurely jumping over to the second copy of the world that gets created when you pan too far west.

How it should work: Go to [this pure Google maps marker demo](https://developers.google.com/maps/documentation/javascript/examples/marker-simpleObserve) and zoom out until you see most of the world, but not so far that you see multiple Australias. As you move the map over to view more to the west, you will see that the Marker stays on the first Australia until it is off screen and then appears on the second Australia only when that Australia moves into the map bounds. If you zoom out to see multiple Australias, the marker will appear on which Australia is closer to the center of the screen, which makes a lot of sense.

Deck.gl jumps to new copies of the globe as soon as that copy of the globe takes up most of the screen.
So, if I have a layer with an image on San Francisco and I move the map east to view the next world's copy of San Francisco, I won't see the image until the new globe takes up most of the screen, which does not look good at all.

I know this is hard to convey with text so I made a short video:

![DECK_GL_SINGLE_GOOGLE_MAP_BUG](https://user-images.githubusercontent.com/8882589/64736126-9115c880-d4af-11e9-8eab-f99fd9219c4b.gif)

#### Repro Steps
Place any content on a layer. The effect is most noticeable when zoomed out very far and with elements near the anti-meridian (far east or west). 

#### Environment (please complete the following information):
 - Framework Version: 7.2.3

Pessimistress:
This is a known problem. We are planning to address it in the next major release; meanwhile there are two ways you can work around this:

- If you are using a point-like layer (Scatterplot, Icon, Text, Scenegraph etc.) or a line-like layer (Line, Arc) you can try set the `wrapLongitude: true` prop. This will automatically project a point to the copy of the map so that it's closest to the current viewport center.
- Otherwise you can manually make a copy of your data by shifting all longitudes +/- 360 degrees. This gives you the most control over where something is rendered.

Pessimistress:
This will be fixed in 8.1.0. You can test the feature with deck.gl@^8.1.0-alpha.

PR #4247, opened 2020-02-07
[google-maps] support lower zoom levels

Pessimistress:
Fixes issues from https://github.com/uber/deck.gl/pull/3643
For #4144 #3576

google-maps-low-zoom

#### Change List
- Supports zoom 0-3

### DIFF
--- a/modules/google-maps/src/utils.js
+++ b/modules/google-maps/src/utils.js
@@ -1,6 +1,7 @@
 /* global document, google */
 import {Deck} from '@deck.gl/core';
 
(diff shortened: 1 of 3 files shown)


modules/google-maps/src/utils.js
```diff
@@ -1,6 +1,8 @@
 /* global document, google */
 import {Deck} from '@deck.gl/core';
 
+const MAX_LATITUDE = 85.05113;
```
donmccurdy: Consider adding a comment referencing https://en.wikipedia.org/wiki/Web_Mercator_projection#Formulas or similar, this may be confusing to future readers. :)

PR #4248, opened 2020-02-07
[mapbox] support repeating at low zoom levels

Pessimistress:
#### Change List
- Repeat map at low zoom levels


xintongxia:
Is it necessary to make it optional?

Pessimistress:
> Is it necessary to make it optional?

The base map always shows multiple copies.

Issue #4236, opened 2020-02-05
Next Steps: TerrainLayer


chrisgervang:
Continuation of #3984 

## To Do List 
  - [x] Fix Martini (transpile published code to es5.. with rollup [their build tool]?) See: https://github.com/mapbox/martini/issues/10
  - [x] Move TerrainLayer to geo-layers

## Vital Extra Credit
  - [ ] Fix tile border seam line
    - Skirting?: https://github.com/mapbox/martini/issues/7
    - Additional scale for any XY seam (wouldn't solve Z)
  - [x] Improve performance (lots of CPU bound math going on. Can it be offloaded to workers or moved to GPU?) Done: https://github.com/uber-web/loaders.gl/pull/652
  - [x] Create a website demo that doesn't utilize mapbox API (serves as a good 2nd example)
    - [x] Could use FAA Sectional surface texture.
    - Could use US grayscale height map from wind demo.
    - Could use AWS Open Data Terrain Tiles

## Feature

- [ ] Support alternative terrain loaders e.g. quantized mesh

kylebarron:
> Create a website demo that doesn't utilize mapbox API (serves as a good 2nd example)

The main non-Mapbox source of terrain tiles that I know of is [AWS Open Data Terrain Tiles](https://registry.opendata.aws/terrain-tiles/). Note that these use a different encoding than Mapbox's tiles ([source](https://github.com/tilezen/joerd/blob/master/docs/formats.md#terrarium))
```
elevation = (red * 256 + green + blue / 256) - 32768
```
You can just use this tile endpoint with no access key currently required:
https://s3.amazonaws.com/elevation-tiles-prod/terrarium/{z}/{x}/{y}.png

Pessimistress:
Looks like Martini 0.2.0 has fixed the tranpilation issue.

chrisgervang:
Thanks @kylebarron, I think I got the AWS source working too :)

kylebarron:
For now this still only works with a raster satellite image texture, right? It would be awesome for it to work eventually with other vector layers, but I get that that's probably a lot of work.

chrisgervang:
> For now this still only works with a raster satellite image texture, right? It would be awesome for it to work eventually with other vector layers, but I get that that's probably a lot of work.

Yeah, only raster surface paint right now. I've gathered its pretty involved to convert the vector data to a texture. I think #3935 is discussing approaches.

kylebarron:
@chrisgervang 
[...]

PR #4275, opened 2020-02-14
Refactor TerrainLayer for performance and ease of use.

chrisgervang:
#### Background

Progress on #4236 

New TerrainLayer interface internalizes all data preparation logic so the user doesn't need to write that boilerplate. Also supports being used as a tiled layer (internally rendering a TileLayer of SimpleMeshLayers), or as a single mesh (internally rendering a SimpleMeshLayer).

This layer is suitable for rendering a few use cases:
- Terrain tiles globally from a tile server (similar to TileLayer).
- A static set of tiles from a tile server.
- A custom tile of your own dataset (e.g. non-[Slippy map tiles](https://wiki.openstreetmap.org/wiki/Slippy_map_tilenames)).

Interactions while loading new tiles now feels way smoother since we're now offloading terrain mesh generation to web workers.

New prop interface:
```
const defaultProps = {
  ...TileLayer.defaultProps,
  // Image url that encodes height data
  terrainImage: {type: 'string', value: null},
  // Image url to use as texture
  surfaceImage: {type: 'string', value: null},
  // Martini error tolerance in meters, smaller number -> more detailed mesh
  meshMaxError: {type: 'number', value: 4.0},
  // Bounding box of the terrain image, [minX, minY, maxX, maxY] in world coordinates
  bounds: {type: 'array', value: [0, 0, 1, 1], compare: true},
  // Color to use if surfaceImage is unavailable
  color: {type: 'color', value: [255, 255, 255]},
  // Object to decode height data, from (r, g, b) to height in meters
  elevationDecoder: {
    type: 'object',
    value: {
      rScaler: 1,
      gScaler: 0,
      bScaler: 0,
      offset: 0
    }
  },
  // Supply url to local terrain worker bundle. Only required if running offline and cannot access CDN.
  workerUrl: {type: 'string', value: null},
  // Same as SimpleMeshLayer wireframe
  wireframe: false
};
```

#### Change List
- Changed TerrainLayer prop interface changed.
- Integrated @loaders.gl/terrain for web worker mesh generation.
- Updated example to reflect new usage.
- Adding workerUrl prop

### DIFF
--- a/examples/experimental/terrain/src/app.js
+++ b/examples/experimental/terrain/src/app.js
@@ -4,6 +4,7 @@ import {render} from 'react-dom';
 import DeckGL from '@deck.gl/react';
 import {WebMercatorViewport, COORDINATE_SYSTEM} from '@deck.gl/core';
 import {load} from '@loaders.gl/core';
+import {TerrainLoader} from '@loaders.gl/terrain';
 import {TileLayer} from '@deck.gl/geo-layers';
 
 import TerrainLayer from './terrain-layer/terrain-layer';
@@ -24,11 +25,36 @@ const INITIAL_VIEW_STATE = {
 // Constants
 const TERRAIN_RGB = 'https://api.mapbox.com/v4/mapbox.terrain-rgb';
 
-const getTerrainData = ({x, y, z}) => {
-  const terrainTile = `${TERRAIN_RGB}/${z}/${x}/${y}.pngraw?access_token=${MAPBOX_TOKEN}`;
-  // Some tiles over the ocean may not exist
-  // eslint-disable-next-line handle-callback-err
-  return load(terrainTile).catch(err => null);
+const getTerrainData = ({bbox, x, y, z}) => {
+  const terrainImage = `${TERRAIN_RGB}/${z}/${x}/${y}.png?access_token=${MAPBOX_TOKEN}`;
+
+  // https://docs.mapbox.com/help/troubleshooting/access-elevation-data/#mapbox-terrain-rgb
+  // Note - the elevation rendered by this example is greatly exagerated!
+  const elevationDecoder = {
+    rScaler: 6553.6,
+    gScaler: 25.6,
+    bScaler: 0.1,
+    offset: -10000
+  };
+
+  const viewport = new WebMercatorViewport({
+    longitude: (bbox.west + bbox.east) / 2,
+    latitude: (bbox.north + bbox.south) / 2,
+    zoom: z
+  });
+  const bottomLeft = viewport.projectFlat([bbox.west, bbox.south]);
+  const topRight = viewport.projectFlat([bbox.east, bbox.north]);
+  const bounds = [bottomLeft[0], bottomLeft[1], topRight[0], topRight[1]];
+
+  const options = {
+    worker: false,
+    terrain: {
+      bounds,
+      meshMaxError: 4,
+      elevationDecoder
+    }
+  };
+  return load(terrainImage, TerrainLoader, options);
 };
 
 const SurfaceDropdown = ({surface, setSurface}) => {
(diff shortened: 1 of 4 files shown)


examples/experimental/terrain/src/app.js
```diff
@@ -20,16 +17,16 @@ const INITIAL_VIEW_STATE = {
...
 // Constants
 const TERRAIN_RGB = 'https://api.mapbox.com/v4/mapbox.terrain-rgb';
 
-const getTerrainData = ({x, y, z}) => {
-  const terrainTile = `${TERRAIN_RGB}/${z}/${x}/${y}.pngraw?access_token=${MAPBOX_TOKEN}`;
-  // Some tiles over the ocean may not exist
-  // eslint-disable-next-line handle-callback-err
-  return load(terrainTile).catch(err => null);
-};
+function getTerrain(tile) {
```
Pessimistress: Let's move this to a test. As a public-facing example, I prefer to keep things simple.

examples/experimental/terrain/src/surface.js
```diff
@@ -5,17 +5,29 @@ const SATELLITE = 'https://api.mapbox.com/v4/mapbox.satellite';
 // Set your mapbox token here
 const MAPBOX_TOKEN = process.env.MapboxAccessToken; // eslint-disable-line
 
-export const getSurface = ({x, y, z}, surface) => {
+export const getSurface = (surface, tile) => {
```
Pessimistress: Same thing here

examples/experimental/terrain/src/terrain-layer/terrain-layer.js
```diff
@@ -1,124 +1,147 @@
...
+    const viewport = new WebMercatorViewport({
+      longitude: (bbox.west + bbox.east) / 2,
+      latitude: (bbox.north + bbox.south) / 2,
+      zoom: z
+    });
+    const bottomLeft = viewport.projectFlat([bbox.west, bbox.south]);
+    const topRight = viewport.projectFlat([bbox.east, bbox.north]);
+    const bounds = [bottomLeft[0], bottomLeft[1], topRight[0], topRight[1]];
+
+    const options = {
```
Pessimistress: Extract the construction of options and calling `load` out to a utility function, so that you can reuse in `updateState`.

examples/experimental/terrain/src/surface.js
```diff
@@ -5,17 +5,29 @@ const SATELLITE = 'https://api.mapbox.com/v4/mapbox.satellite';
...
     case 'street':
-      surfaceImage = `${STREET}/${z}/${x}/${y}.png`;
+      surfaceImage = `${STREET}/{z}/{x}/{y}.png`;
+      if (tile) {
+        const {x, y, z} = tile;
+        surfaceImage = `${STREET}/${z}/${x}/${y}.png`;
+      }
       break;
     case 'none':
     default:
```
Pessimistress: The reason `<string> -> null` does not work has to do with the `texture` prop is async and default to `null`. The layer does not think the prop has changed. The fix will not be trivial. As an work around, you can set `surfaceImage` to `0`.

examples/experimental/terrain/src/app.js
chrisgervang: Are you referring to the two ways this function can be used, and reducing it to the simple TileLayer usage?

That’s fair since this becomes the website example which will just show one way.

I was going to prepare an example of how I’d typically use it without TileLayer and put it somewhere too, maybe explore some fun advance usages. Any suggestions?

Pessimistress: You can put the simple usage example in  https://github.com/uber/deck.gl/blob/master/website/src/components/demos/layer-demos.js — these are the examples embedded in the layer docs.

chrisgervang: Ok cool! I’ll merge this and get into geo-layers PR which can have tests and layer-demos example.

examples/experimental/terrain/src/surface.js
chrisgervang: Ok - good enough for me :) thanks for looking. I’ll just add a comment to document this.

chrisgervang: Hmm, this didn't work. I'm pushing up what I tried.

chrisgervang:
I'm going to merge this and continue implementing feedback in a integration branch.

PR #4288, opened 2020-02-19
Fix attribute check greater than attribute size

twojtasz:
Through the TypedArrayManager, it is possible that a buffer from the pool can be used that contains invalid values for the type and the update of the attribute may only affect the attribute.size, rather than the buffer.length.

Limit fix limits the validation to Math.min(attribute.size, 4)

### DIFF
--- a/modules/core/src/lib/attribute/attribute.js
+++ b/modules/core/src/lib/attribute/attribute.js
@@ -354,17 +354,31 @@ export default class Attribute extends DataColumn {
     }
   }
 
+  // check that the first few elements of the attribute are reasonable
+  /* eslint-disable no-fallthrough */
   _checkAttributeArray() {
     const {value} = this;
-    if (value && value.length >= 4) {
-      const valid =
-        Number.isFinite(value[0]) &&
-        Number.isFinite(value[1]) &&
-        Number.isFinite(value[2]) &&
-        Number.isFinite(value[3]);
+    const limit = Math.min(4, this.size);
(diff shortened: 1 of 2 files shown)


twojtasz:
I tried to tickle this through the TypedArrayManager (similar to how I found it in the internal app) but that was not fruitful.  The manager also sets the first 4 values to 0 (unless the copy flag is set), but I couldn't figure out how this was getting into this state in the app.

modules/core/src/lib/attribute/attribute.js
```diff
@@ -354,17 +354,31 @@ export default class Attribute extends DataColumn {
...
   _checkAttributeArray() {
     const {value} = this;
-    if (value && value.length >= 4) {
-      const valid =
-        Number.isFinite(value[0]) &&
-        Number.isFinite(value[1]) &&
-        Number.isFinite(value[2]) &&
-        Number.isFinite(value[3]);
+    const limit = Math.min(4, this.size);
+    if (value && value.length && value.length >= limit) {
```
Pessimistress: `value.length >= limit` should cover the `!value.length` case

twojtasz: I recall getting errors like cannot access attribute length.  But i'll double check.

twojtasz: must have been something else, i'll change the predicate to what it was before with the new limit var

PR #4296, opened 2020-02-21
[Feat] add lineOpacity prop to scatterplot layer

heshan0131:
#### Feature
It would be nice to be able to adjust fill opacity and line opacity separately. In Geojson layer e.g. I can achieve this by passing _sublayerProps to override opacity of polygon stroke and line string layer. But for point layer, there is currrently no way to control them separately. Here this PR

#### Background

related kepler.gl issue: [Separately adjust opacity of stroke and fill](https://github.com/keplergl/kepler.gl/issues/959)

### DIFF
--- a/modules/layers/src/scatterplot-layer/scatterplot-layer-vertex.glsl.js
+++ b/modules/layers/src/scatterplot-layer/scatterplot-layer-vertex.glsl.js
@@ -35,6 +35,7 @@ uniform float opacity;
 uniform float radiusScale;
 uniform float radiusMinPixels;
 uniform float radiusMaxPixels;
+uniform float lineOpacity;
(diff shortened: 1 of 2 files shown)


Pessimistress:
Many core layers have both `getLineColor` and `getFillColor`. It's hard to justify why only ScatterplotLayer gets two `opacity` props.

heshan0131:
Agreed it's not fair for just point layer to have it, I would suggest all core layers that offer stroke and fill styling should have it. because: 

Screen Shot 2020-02-24 at 2 06 02 PM

👯 😄

Pessimistress:
Would you add it to all core layers then? :)

We also need to update the documentation.

PR #4302, opened 2020-02-21
Use luma.gl in arcgis module

tsherif:
Update arcgis module to use luma.gl instead of raw gl calls as much as possible.

### DIFF
--- a/modules/arcgis/src/arcgis-deck-external-renderer.js
+++ b/modules/arcgis/src/arcgis-deck-external-renderer.js
@@ -8,6 +8,8 @@ import {
   initializeDeckGL
 } from './commons';
 
+import {withParameters} from '@luma.gl/core';
+
 export default function loadArcGISDeckExternalRenderer(externalRenderers, Collection) {
   function ArcGISDeckExternalRenderer(view, conf) {
     this.view = view;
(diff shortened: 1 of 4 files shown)


modules/arcgis/src/arcgis-deck-external-renderer.js
```diff
@@ -38,52 +40,51 @@ export default function loadArcGISDeckExternalRenderer(externalRenderers, Collec
...
     this.deckgl.setProps({
-      layers: this.deckLayers.items.slice(),
       viewState: {
         latitude: this.view.center.latitude,
         longitude: this.view.center.longitude,
-        zoom: this.view.zoom,
-        bearing: this.view.camera.heading,
-        pitch: this.view.camera.tilt
+        zoom: this.view.featuresTilingScheme.scaleToLevel(state.scale),
+        bearing: -state.rotation,
```
Pessimistress: I don’t think these should be changed...

modules/arcgis/src/arcgis-deck-layer-view-2d.js
```diff
@@ -72,22 +74,18 @@ export default function loadArcGISDeckLayerView2D(BaseLayerViewGL2D) {
...
-        this.program = null;
+      if (this.buffer) {
+        this.buffer.delete();
       }
 
-      if (this.vertexBuffer) {
-        gl.deleteBuffer(this.vertexBuffer);
-        this.vertexBuffer = null;
+      if (this.deckFbo) {
+        this.destroyFramebuffer(this.deckFbo);
```
Pessimistress: Just call `this.deckFbo.delete()`

modules/arcgis/src/arcgis-deck-external-renderer.js
tsherif: Weird, I don't remember changing this. Maybe copy pasta.

tsherif:
@Pessimistress Fixed.

PR #4311, opened 2020-02-26
Remove s2-geometry dependency

Pessimistress:
#### Background

deck.gl's pre-bundled version excluded s2-geometry because of its size (the module is not tree shakable). However, because s2-geometry does not offer an umd entry point, it is not easy to supply this dependency in a scripting environment. We recommended `https://bundle.run` in the documentation in the past, but it's not always reliable.

#### Change List
- Cherry pick s2-geometry utilities into the geo-layers module

#### TODO

There is an outstanding discussion with @macrigiuseppe on exposing some of these functionalities.

### DIFF
--- a/docs/layers/s2-layer.md
+++ b/docs/layers/s2-layer.md
@@ -74,7 +74,6 @@ new S2Layer({});
 To use pre-bundled scripts:
 
 ```html
-<script src="https://bundle.run/s2-geometry@1.2.10"></script>
 <script src="https://unpkg.com/deck.gl@^8.0.0/dist.min.js"></script>
 <!-- or -->
 <script src="https://unpkg.com/@deck.gl/core@^8.0.0/dist.min.js"></script>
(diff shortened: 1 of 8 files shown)


macrigiuseppe:
@Pessimistress 
why is this a breaking change?

developers are not required to include s2-geometry as peer dependency.

Do we require any changes on the consumer side?

docs/layers/s2-layer.md
```diff
@@ -74,7 +74,6 @@ new S2Layer({});
 To use pre-bundled scripts:
 
 ```html
-<script src="https://bundle.run/s2-geometry@1.2.10"></script>
 <script src="https://unpkg.com/deck.gl@^8.0.0/dist.min.js"></script>
```
macrigiuseppe: shouldn't this be updated to 8.1.0?

8.0.0 still requires s2-geometry in the html

ajduberstein:
Thanks for the heads up, will modify pydeck appropriately

PR #4312, opened 2020-02-26
Improve TerrainLayer loading experience

Pessimistress:
For #4236 

#### Change List
- Only display a tile if both the mesh and the surface image are loaded
- Fix an update issue in non-tiled scenario when `terrainImage` is changed
- Restore no texture option in the demo


Pessimistress:
> Was there a technical reason to load this way by any chance?

`TileLayer` determines the visibility of tiles based on whether their data loaders are resolved.

PR #4335, opened 2020-03-01
Fix projectPosition in CARTESIAN coordinate system

Pessimistress:
The CPU `projectPosition` to GPU `project_position` parity was broken when using CARTESIAN coordinate system with Web Mercator projection (#4140).

#### Change List
- `projectPosition` and `getUniformsFromViewport` now share the same logic
- Additional unit tests

### DIFF
--- a/modules/core/src/shaderlib/project/project-functions.js
+++ b/modules/core/src/shaderlib/project/project-functions.js
@@ -3,7 +3,7 @@
  * TODO: move to Viewport class?
  */
 import {COORDINATE_SYSTEM} from '../../lib/constants';
-import {LNGLAT_AUTO_OFFSET_ZOOM_THRESHOLD} from './viewport-uniforms';
+import {getOffsetOrigin} from './viewport-uniforms';
 
 import * as vec4 from 'gl-matrix/vec4';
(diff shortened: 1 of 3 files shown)


modules/core/src/shaderlib/project/project-functions.js
```diff
@@ -116,31 +106,26 @@ export function projectPosition(position, params) {
...
+    viewport,
+    modelMatrix,
+    coordinateSystem: fromCoordinateSystem,
+    coordinateOrigin: fromCoordinateOrigin,
+    offsetMode
+  });
+
+  if (offsetMode) {
+    const positionCommonSpace = viewport.projectPosition(
+      geospatialOrigin || shaderCoordinateOrigin
```
tsherif: Might be missing a usage, but could this be simplified by having the value of `geospatialOrigin` default to `shaderCoordinateOrigin` instead of `null`? And then having `getOffsetOrigin` just return one `offsetOrigin` value.

Pessimistress: `geospatialOrigin` is always in lnglat and used later to calculate units per pixel. I'd leave it as is for now for readability.

PR #4336, opened 2020-03-01
Add clipping to MVTTileLayer

Pessimistress:
Followup of #3935 

Before:

image

After:

image

#### Change List
- Support `draw` method in layer extensions
- Add private `ClipExtension`

@jesusbotella


modules/geo-layers/src/mvt-tile-layer/clip-extension.js
```diff
@@ -0,0 +1,113 @@
...
+// THE SOFTWARE.
+
+import {LayerExtension} from '@deck.gl/core';
+
+const defaultProps = {
+  clipBounds: [0, 0, 1, 1]
+};
+
+const shaderFunction = `
+uniform vec4 clipBounds;
```
tsherif: `clip_bounds` or `clip_clipBounds` per shader module prefix convention.

PR #4337, opened 2020-03-01
pydeck: Support rendering in Google Collab

ajduberstein:
For #4103 

#### Background
Supports rendering of HTML pages in Collab

The 2-way interactive widget won't work in Collab yet. We seem to be awaiting this issue: https://github.com/googlecolab/colabtools/issues/60

#### Change List
- Add built-in support for rendering .to_html in Google Collab

### DIFF
--- a/bindings/pydeck/pydeck/bindings/deck.py
+++ b/bindings/pydeck/pydeck/bindings/deck.py
@@ -124,6 +124,7 @@ class Deck(JSONMixin):
         notebook_display=True,
         iframe_width=700,
         iframe_height=500,
+        as_string=False,
         **kwargs
     ):
         """Write a file and loads it to an iframe, if in a Jupyter environment;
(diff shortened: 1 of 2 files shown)


lyzhan:
Does pydeck still only support rendering of HTML pages in Colab? 
I tried pdk.Deck.show() in Colab. It doesn't show anything.

ajduberstein:
@lyzhan yes, that’s correct, in the current version (pydeck 0.6) 2-way interactivity is only available in Jupyter.

lyzhan:
@ajduberstein Thanks for your reply! It is OK for me to just use Jupyter now. It will be very conviniet to share the pydeck results with others using colab if the two-way interactivity is supported for colab in the future.
By the way, I recently found your answer in #5151 is very helpful for my project! Thank again!

PR #4344, opened 2020-03-02
Workaround for invalid attribute index access

twojtasz:
In an application with heavy path layer usage the typed array manager
can get to a point where the number of points exactly fits in the
allocated buffer.  This problem is this does not appear to be accounting
for the shaderAttributes and results the runtime error:

  'glDrawElementsInstancedANGLE: attempt to access out of range vertices in attribute'

This extra padding 3 * 3 is based on how the type array manager needs to allocate.

### DIFF
--- a/modules/layers/src/path-layer/path-tesselator.js
+++ b/modules/layers/src/path-layer/path-tesselator.js
@@ -30,7 +30,9 @@ export default class PathTesselator extends Tesselator {
     super({
       ...opts,
       attributes: {
-        positions: {size: 3, type: opts.fp64 ? Float64Array : Float32Array},
+        // Padding covers shaderAttributes for last segment in largest case fp64
+        // additional vertex + hi & low parts, 3 * 6
+        positions: {size: 3, padding: 18, type: opts.fp64 ? Float64Array : Float32Array},
         segmentTypes: {size: 1, type: Uint8ClampedArray}
       }
     });


1chandu:
Related :
When allocating the buffer we always allocate twice the size of value array [link](https://github.com/uber/deck.gl/blob/4270ac86c44b49dac02e1830a57b3d2cce5e2756/modules/core/src/lib/attribute/data-column.js#L242)

```
        // Over allocation is required because shader attributes may have bigger offsets
        buffer.reallocate((value.byteLength + byteOffset) * 2);
```

But I think instead of doubling the size, we should properly scan all `shaderAttributes`  and get the extra space based on offsets. I wonder this logic is failing, in some cases that you are seeing. If the tesselator needed padding, we should update these unit tests to cover this, also we need docs on path-tesselator to better understand this.

That said, above WA seems safe, good to go if it is unblocking any other work.

twojtasz:
@1chandu Actually, this hits the case above with the `toDoublePrecision()` [link](https://github.com/uber/deck.gl/blob/4270ac86c44b49dac02e1830a57b3d2cce5e2756/modules/core/src/lib/attribute/data-column.js#L237) which does not allocate double the size, well not really. It does but that i because it stores the hi & lo values for each vertex.

Maybe this is a place where the two paths are out of sync?

PR #4365, opened 2020-03-09
Highlight features spread across tiles

jesusbotella:
#### Background
Right now, only one feature is highlighted when `autoHighlight` property is enabled and the feature is split into several tiles. We would need to highlight the whole feature instead of only the hovered part.

I misunderstood the object index and returned the index within the provided data array instead of the picking index. **Is there a way to get the picking index from a given property?**

#### Change List
- Add `highlightedObjectIndex` to rendered sublayers in `TileLayer`.
- Implement `getHighlightedObjectIndex` in `MVTTileLayer`.
- Save `hoveredFeatureId` into `MVTTileLayer` state.


Pessimistress:
I think one fix that needs to happen is in GeoJsonLayer - `pickingInfo` returned by `GeoJsonLayer` should contain the index of the overall feature array, not the index in the sublayer. I will look into this.

VictorVelarde:
Hi @Pessimistress will that fix be available for the next 8.1.0 release?

Pessimistress:
> will that fix be available for the next 8.1.0 release?

We are releasing soon - the fix will land as a patch.

VictorVelarde:
Great, ping us if you need any help or review on that!

Pessimistress:
@jesusbotella Landed #4403, this PR should have been unblocked.

jesusbotella:
Thank you @Pessimistress! That was fast!

I have just tested my code after the merge and it works but I have some problems with particular polygons. Here's a GIF:

highlightacrosstiles

Particularly with the top big polygon, it works to highlight most of it but there's a part that remains unhighlighted.

I've uploaded the example that I'm using so that you can see what happens. Some features are successfully highlighted but others don't.
[highlightExample.zip](https://github.com/uber/deck.gl/files/4356244/highlightExample.zip)

I need to tweak the code a bit so that features stop being highlighted, but that'll come later.

Thank you!

modules/geo-layers/src/mvt-layer/mvt-layer.js
```diff
@@ -9,6 +9,11 @@ import ClipExtension from './clip-extension';
 
 const WORLD_SIZE = 512;
 
+const defaultProps = {
+  ...TileLayer.defaultProps,
```
Pessimistress: `defaultProps` of the base class is automatically merged.

jesusbotella:
I've pushed commits to make it work properly with @Pessimistress changes.

First of all, I've tweaked the code to unhighlight features when mouse is moved off the feature.

Another important thing is that I've changed MVT custom highlighting to only work when `autoHighlighting` is active, mimicking its functionality but selecting the whole feature instead of the split parts of the whole feature.

And I noticed that this change broke `highlightedObjectIndex` in GeoJSON layer and MVT layer, so I returned `highlightedObjectIndex` value in `getHighlightedObjectIndex` if present.

Let me know what you think, and if you like it, I'll add docs and tests.

AdriSolid:
Hi! Thanks for working on the MVTLayer, awesome!

I'm faced the splits parts once a feature is highlighted, on version 8.1.1

Screenshot at 2020-04-14 21-41-29

VictorVelarde:
@AdriSolid That "partial highlighting" should be fixed when this PR gets merged.

Regarding to that, @Pessimistress do you think that we should go on and create docs & tests for it with the current approach for unhighlighting?

jesusbotella:
Hey @Pessimistress! I have just set this PR as ready for review. It would be great if you (or any other contributor) could give it a look and tell us your thoughts.

Thank you very much!

docs/layers/mvt-layer.md
```diff
@@ -81,12 +81,23 @@ Inherits all properties from [`TileLayer`](/docs/layers/tile-layer.md) and [base
...
 
 The `getTileData` prop from the `TileLayer` class will not be called.
 
+##### `uniqueIdProperty` (String)
+
+Optional. Needed for highlighting a feature split across two or more tiles if no [feature id](https://github.com/mapbox/vector-tile-spec/tree/master/2.1#42-features) is provided.
+
+An string pointing to a tile attribute containing a unique identifier for features across tiles.
+
+##### `highlightedObjectIndex` (Number)
```
Pessimistress, line 96: Instead of overriding the base layer prop, I suggest just adding a new prop `highlightedFeatureId` which will describe this feature better.

modules/geo-layers/src/mvt-layer/mvt-layer.js
```diff
@@ -9,6 +9,10 @@ import ClipExtension from './clip-extension';
 
 const WORLD_SIZE = 512;
 
+const defaultProps = {
+  uniqueIdProperty: {type: 'string', value: '', compare: false}
```
Pessimistress, line 13: Strings are always compared.

modules/geo-layers/src/tile-layer/tile-layer.js
```diff
@@ -20,7 +20,8 @@ const defaultProps = {
   maxCacheSize: null,
   maxCacheByteSize: null,
   refinementStrategy: STRATEGY_DEFAULT,
-  zRange: null
+  zRange: null,
+  getHighlightedObjectIndex: {type: 'function', value: () => -1, compare: false}
```
Pessimistress, line 24: This is not documented.

If this is going to be a public API, we need to look into whether it's sufficient to just pass it the `tile` argument.

I think it's ok to not expose this until we find other use cases.

modules/geo-layers/src/tile-layer/tile-layer.js
```diff
@@ -138,6 +139,10 @@ export default class TileLayer extends CompositeLayer {
     return this.props.renderSubLayers(props);
   }
 
+  getHighlightedObjectIndex(tile) {
+    return this.props.getHighlightedObjectIndex(tile);
```
Pessimistress, line 143: Drop the prop and `return -1` by default.

modules/geo-layers/src/mvt-layer/mvt-layer.js
```diff
@@ -30,12 +34,68 @@ export default class MVTLayer extends TileLayer {
...
+
+    return super.onHover(info, pickingEvent);
+  }
+
+  getHighlightedObjectIndex(tile) {
+    const {highlightedFeatureId} = this.state;
+    const {uniquePropertyId, highlightedObjectIndex: highlightedFeatureProp} = this.props;
+    const {data} = tile;
+
+    if ((!highlightedFeatureId && !highlightedFeatureProp) || !Array.isArray(data)) {
```
Pessimistress, line 70: `highlightedFeatureId` could be `0`

modules/geo-layers/src/mvt-layer/mvt-layer.js
```diff
@@ -30,12 +34,68 @@ export default class MVTLayer extends TileLayer {
...
+          : feature.properties[uniquePropertyId] === featureIdToHighlight
+    );
+  }
+}
+
+function getFeatureUniqueId(feature, uniquePropertyId) {
+  const hasFeatureId = Boolean(feature.id);
+  const hasUniquePropertyId = Boolean(uniquePropertyId);
+
+  if (hasFeatureId) {
```
Pessimistress, line 89: I'd expect the user-specified `hasUniquePropertyId` to override the default

docs/layers/mvt-layer.md
jesusbotella: Sure, I didn't like it either. It was a silly thing to reuse an existing prop instead of adding a new one.

modules/geo-layers/src/tile-layer/tile-layer.js
jesusbotella: Yeah, makes sense. Let's remove it and wait until we find some other use cases.

modules/geo-layers/src/mvt-layer/mvt-layer.js
jesusbotella: Sure, makes more sense. I have just changed it.

docs/layers/mvt-layer.md
jesusbotella: Sure, I didn't like it either. It was a silly thing to reuse an existing prop instead of adding a new one.

jesusbotella:
I have just changed the suggestions mentioned in the review.

I have also had to remove the interaction test because it was fragile. I guess that I have not been able to guess a proper wait time for firing the hover events and the property checking. I don't know if you have some ideas to make it work (besides increasing timeout).

jesusbotella:
Hi @Pessimistress! Sorry for disturbing, I performed the review changes, and I would like some validation whenever you can. Thank you!

docs/layers/mvt-layer.md
```diff
@@ -81,12 +81,25 @@ Inherits all properties from [`TileLayer`](/docs/layers/tile-layer.md) and [base
...
 
 The `getTileData` prop from the `TileLayer` class will not be called.
 
+##### `uniqueIdProperty` (String)
+
+Optional. Needed for highlighting a feature split across two or more tiles if no [feature id](https://github.com/mapbox/vector-tile-spec/tree/master/2.1#42-features) is provided.
+
+An string pointing to a tile attribute containing a unique identifier for features across tiles.
+
+##### `highlightedFeatureId` (Number)
```
Pessimistress, line 96: Does the feature id have to be a number? Is string accepted? If so, the default should be `null` for clarity?

modules/geo-layers/src/mvt-layer/mvt-layer.js
```diff
@@ -30,12 +35,64 @@ export default class MVTLayer extends TileLayer {
...
+    );
+  }
+}
+
+function getFeatureUniqueId(feature, uniqueIdProperty) {
+  if (uniqueIdProperty) {
+    return feature.properties[uniqueIdProperty];
+  }
+
+  if (feature.id) {
```
Pessimistress, line 90:
```suggestion
  if ('id' in feature) {
```

docs/layers/mvt-layer.md
jesusbotella: As per MVT docs, feature ID can be [only a number](https://github.com/mapbox/vector-tile-spec/blob/master/2.1/vector_tile.proto#L32). But it's true that if anyone uses `uniqueIdProperty` to reference an ID within feature properties, it could be a string.

I have just changed it to support strings as well in the code. Let me know if you see something weird.

jesusbotella:
I see that coverage has decreased. I tried to make a module test for highlighting to improve it, but I don't know how to make it pass because fetch is not supported in Node and the only way for MVT Layer to load data is through a URL.

ibgreen:
> I see that coverage has decreased. I tried to make a module test for highlighting to improve it, but I don't know how to make it pass because fetch is not supported in Node and the only way for MVT Layer to load data is through a URL.

I am not up-to-date on the test setup in deck.gl, but the [`@loaders.gl/polyfills`](https://loaders.gl/modules/polyfills/docs/api-reference) module installs a polyfill of fetch under Node.js. This is used for all Node.js tests in loaders.gl, so it is pretty well tested.

jesusbotella:
> I am not up-to-date on the test setup in deck.gl, but the @loaders.gl/polyfills module installs a polyfill of fetch under Node.js. This is used for all Node.js tests in loaders.gl, so it is pretty well tested.

Thank you @ibgreen! I tried modifying this file (https://github.com/visgl/deck.gl/blob/master/test/modules/geo-layers/mvt-layer.spec.js) but I believe that the output already prints a `fetch not available in node` message as far as I see for those tests already.

jesusbotella:
Ok, I have been able to get more insight on those test issues.

First of all, I think I have found a bug because we're just passing a simple template string to the sample data in the MVT test, but we're not considering it as valid in `getURLFromTemplate` and the tests print a `Invalid URL message`.

And then, when I change it to an array, I think the helper is not passing the proper data value to `Layer should have sublayers` tests because these errors pop up due to the failing tiles as you can see in the image below:
Screenshot 2020-05-27 at 19 49 43

Below those errors, I still see some `fetch not available in node` messages, but I cannot see where those errors come from because it seems like they come from the same test in the image as there are no other test case name below.

Going to debug for a bit and see if I can fix them.

jesusbotella:
I kinda found a fix for those errors, that is checking if our template value type has a replace method. That way we avoid passing unwanted values like `{ length: 1 }` to MVTLoader and it doesn't throw anymore.

About fetch, there is this line in Node tests setup:
https://github.com/visgl/deck.gl/blob/583cde50f7b30e130173aa60342411e870f062de/test/node.js#L3-L5

It seems difficult for me to test highlighting without tile data, so I don't know if you have a clue for me to inject that data in the layer and if you have some scheduled work to see why using `@loaders.gl/polyfills` froze the tests or not. I could give it a try and see the cause when I have time.

VictorVelarde:
@ibgreen @Pessimistress we're a bit stuck on this... any advice on how we could move on?

Pessimistress:
@jesusbotella Since we are not testing the loader here, you can bypass the default loading routine with this:

```js
class TestMVTLayer extends MVTLayer {
  getTileData(tile) {
    // return parsed MVT data directly
  }
}
```

Pessimistress:
> We could certainly get fetch working in Node tests similar to how it works in loaders.gl

Loaders has a quite complex system to reroute URLs in both node and browser. I prefer not to add that extra layer of complexity if we don't have to.

jesusbotella:
Thank you both for your comments!

@ibgreen @Pessimistress There's no need to make fetch work with those tests. I think that it's better to override `getTileData` and return a plain data object as @Pessimistress said.

That said, I think I am hitting a rock here 😓 .

The thing is that the method that I added in this PR (`getHighlightedObjectIndex`) to return the actual index of the feature in the data array rather than feature ID is not working properly on the tests.

The method needs the actual data to find the feature within the tile data array, but the tile data that I get when the method is called in the tests is a promise. I understand that it gets wrapped in a promise in case `getTileData` returns a promise, so no problem. But I don't know how to test the  code that I added.

When using the layer in a normal environment, I'm relying on a layer re-render that triggers on every tile load to find the index, and here's not happening.

So, I don't know what to do now, to be honest :(

Here's the test that I did, and wanted to check highlightedObjectId within all sublayers, but I have not been able to do that:

```js
test('MVT Highlight', t => {
  class TestMVTLayer extends MVTLayer {
    getTileData() {
      return geoJSONData;
    }
  }

  const testCases = [
    {
      props: {
        data: ['https://a.basemaps.cartocdn.com/rastertiles/voyager_labels_under/{z}/{x}/{y}.png'],
        id: 'mvt-highlight-test',
        filled: true,
        pickable: true,
        autoHighlight: true,
        highlightedFeatureId: 1
      },
      onAfterUpdate: ({ subLayers }) => {
        for (const layer of subLayers) {
          t.comment(layer.props.pickable);
          t.comment(layer.props.autoHighlight);
          t.comment(layer.props.highlightedObjectIndex);
          t.comment(layer.props.highlightedFeatureId);
        }
      }
    }
  ];

  testLayer({ Layer: TestMVTLayer, testCases, onError: t.notOk });

  t.end();
});
```

jesusbotella:
> It's a long due backlog item to support asynchronous checks in testLayers. :( I suggest we land this PR with this test skipped (test.skip) and I will look into fixing it.

Oh, didn't know about that, but don't worry about fixing it for this. It sounds as a relief because I was kinda hitting my head against the wall 😁 

Ok, so I am gonna commit it and skip that test. I'm not worried about this particular test because there's another one with a render image that works perfect, so we are good with this.

Issue #4317, opened 2020-02-26
8.1 Release Tracker


1chandu:
- [x] Bump all dependencies (math.gl, probe.gl and loaders.gl) to latest release (@Pessimistress)
- [x] Publish alpha/beta  (@Pessimistress)
- [x] Bump all examples to latest alpha/beta (@Pessimistress)
- [x] Stage the website (@Pessimistress)
- [x] Verify whats-new (@Pessimistress)
- [x] Verify upgrade-guide (@Pessimistress)
- [x] Verify all issues in milestone are closed or move out (@tgorkin)
- [x] Verify all API changes are documented (@Pessimistress)
- [x] Website (docs and example) testing: [staging](http://pessimistress.github.io/deck.gl/#/)
    - [x] Testing on Chrome (@1chandu)
    - [x] Testing on Safari (@ 1chandu)
    - [x] Testing on Firefox (@1chandu)
    - [ ] Testing on Edge + IE11 (@tsherif)
    - [x] Testing on iOS (@Pessimistress)
    - [ ] Testing on Android (@Pessimistress)
- [x] Cut 8.1-release branch after all fixes (@Pessimistress)
- [x] Update website doc urls to point to `8.1-release` branch (@Pessimistress)
- [x] Publish 8.1 prod version.

Testing procedure
- access [staged doc website](http://pessimistress.github.io/deck.gl/)
- go through all examples and verify functionality in control panels
- go through all layers in [JSON playground](http://pessimistress.github.io/deck.gl/playground/)
- check visible results for regressions and console for errors/warnings
- add results of testing as comments on this issue

1chandu:
Chrome and Safari :

- [ ] `IconLayer`, `TextLayer` and `3DSurface Explorer` examples crash
```
bundle.js:27 TypeError: Class constructor CompositeLayer cannot be invoked without 'new'
    at new t (bundle.js:85)
```

1chandu:
Docs:
- [ ] 8.1 section in whats new.
- [ ] (Nit) Update mesh-layer/terrain-layers docs calrify picking/highlighting don't work as it is rendered as single instance.

Chrome:
- [ ] TextLayer example doesn't show all texts

<img width="1754" alt="Screen Shot 2020-03-14 at 11 58 33 AM" src="https://user-images.githubusercontent.com/9502731/76688944-e06bbb80-65ee-11ea-808c-10dbdc4875ce.png">

[...]

PR #4377, opened 2020-03-10
Terrain layer API audit

Pessimistress:
For #4317 #4346

#### Change List
- Rename `terrainImage` to `elevationData`
- Rename `surfaceImage` to `texture`
- Docs

### DIFF
--- a/docs/layers/terrain-layer.md
+++ b/docs/layers/terrain-layer.md
@@ -136,11 +136,14 @@ The default value of `elevationDecoder` decodes a grayscale image:
 
 ##### `bounds` (Array, optional)
 
-Bounds of the image to fit x,y coordinates into. In `[left, top, right, bottom]` world coordinates. Must be supplied when using non-tiled elevation data.
+Bounds of the image to fit x,y coordinates into. In `[left, bottom, right, top]`. 
+`left` and `right` refers to the world longitude/x at the corresponding side of the image.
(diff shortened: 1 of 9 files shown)


docs/layers/terrain-layer.md
```diff
@@ -140,25 +136,25 @@ The default value of `elevationDecoder` decodes a grayscale image:
 
 ##### `bounds` (Array, optional)
 
-Bounds of the image to fit x,y coordinates into. In [minX, minY, maxX, maxY] world coordinates. Must be supplied when using as a Non-Tiled terrain, otherwise optional.
+Bounds of the image to fit x,y coordinates into. In `[left, top, right, bottom]` world coordinates. Must be supplied when using non-tiled elevation data.
```
chrisgervang: The left, top, right, bottom notation is confusing to me. For example, I don’t remember what top means - I think indicating it’s the “min y” is helpful. This is especially confusing when different tools order lat/lng differently.

Pessimistress: minY/maxY refer to different things in different context. In a geospatial viewport, the minLat is actually at the bottom.

chrisgervang: That makes sense. Would you be able to clarify the relation between min/max lat/lng and top/bottom/left/right in the docs?

Pessimistress: Next line in doc

Issue #4329, opened 2020-02-27
Feature request: Allow mixed resolution hexagons in `H3HexagonLayer`


ajfriend:
#### Target Use case
I'd like to use this layer to visualize sets of hexagons. Those sets may contain hexagons with different resolutions. One example would be for visualizing the output of the `h3.compact()` function.

#### Proposed feature

Enable `H3HexagonLayer` to plot hexagons of multiple resolutions.

#### To Do List 
  - [ ] Add label and assign to milestone 
  - [ ] Coding
  - [ ] Doc update
  - [ ] What’s new update 
  - [ ] Test

ajfriend:
Awesome! Thanks @nrabinowitz and @Pessimistress!

PR #4381, opened 2020-03-12
Add support for multiple H3 resolutions to H3HexagonLayer

nrabinowitz:
For #4329

#### Change List
- Detects when the input hexagon set has multiple resolutions, setting rendering to "high precision" when true
- Fixes color accessor in H3HexagonLayer example
- Adds tests for high precision rendering


nrabinowitz:
> Can you also add this to the documentation: https://github.com/uber/deck.gl/blob/master/docs/layers/h3-hexagon-layer.md#highprecision-boolean-optional

Done.

PR #4396, opened 2020-03-15
Some minor changes for 8.1 ArcGIS docs

damix911:
This PR does two things:

- _Important_. Fixes two links in `docs/api-reference/arcgis/overview.md`
- _Less important but worth considering_. Adds the word _"experimental"_ in a couple of places wrt the novel `DeckRenderer` class.


damix911:
@Pessimistress I think the `arcgis/` path fix is needed to make the docs render and reference each other properly. For the _"experimental"_ wording fix it's your call.

Thank you!

Issue #4290, opened 2020-02-20
pydeck: Ability for rendered HTML documents to load without internet access


bandersen23:
#### Target Use case
Embedding the required javascript needed for deck.gl to work within pydeck would enable ability to dynamically create maps in a situation where you might not have access to the internet.

#### Proposed feature
This could be established fairly easily by creating an argument within `Deck` that would then change the CDN to something local to the package.
https://github.com/uber/deck.gl/blob/4211e5ed777cdaa4412de7f6e6f423cf2f5c08f8/bindings/pydeck/pydeck/io/html.py#L22

The same thing could be done with the CSS referenced as well:
https://github.com/uber/deck.gl/blob/4211e5ed777cdaa4412de7f6e6f423cf2f5c08f8/bindings/pydeck/pydeck/io/templates/index.j2#L5

https://github.com/uber/deck.gl/blob/4211e5ed777cdaa4412de7f6e6f423cf2f5c08f8/bindings/pydeck/pydeck/io/templates/index.j2#L6

#### To Do List 
  - [ ] Add label and assign to milestone 
  - [ ] Coding
  - [ ] Doc update
  - [ ] What’s new update 
  - [ ] Test

ajduberstein:
Thanks for the suggestion! This is a good idea and shouldn't be too hard to implement. Give me a few weeks.

bandersen23:
@ajduberstein Awesome, thank you!

ajduberstein:
Addressed by #4405

AhmedThahir:
Is there a way to add a custom URL for the assets? As you can see, despite the column chart appearing, the mapbox graph is not appearing offline.

![Screen Shot 2023-08-20 at 1 50 20 PM](https://github.com/visgl/deck.gl/assets/94914628/43d3b594-1058-445e-808e-f71e6c409704)

I really like Plotly's implementation of offline export

```
include_plotlyjs: bool or string (default True)
        Specifies how the plotly.js library is included/loaded in the output
        div string.
    
        If True, a script tag containing the plotly.js source code (~3MB)
        is included in the output.  HTML files generated with this option are
        fully self-contained and can be used offline.
    
        If 'cdn', a script tag that references the plotly.js CDN is included
        in the output. HTML files generated with this option are about 3MB
        smaller than those generated with include_plotlyjs=True, but they
        require an active internet connection in order to load the plotly.js
        library.
    
        If 'directory', a script tag is included that references an external
[...]

PR #4400, opened 2020-03-17
pydeck: add offline capability for pydeck for #4290

bandersen23:
For #4290 

#### Background
The pydeck install already has the necessary javascript come with it on install, so I set up an argument to allow for HTML documents to point to the absolute path of the javascript file. 

#### Change List
- Add an argument in pydeck to allow the option to have offline HTML documents rendered.
- Add a function that processes that by getting the absolute path of the javascript needed. 

@ajduberstein - If this isn't what you're looking for, just let me know and I'd be happy to edit it! I think this might still need a test written for it, which I'm happy to provide if desired.

### DIFF
--- a/bindings/pydeck/pydeck/io/html.py
+++ b/bindings/pydeck/pydeck/io/html.py
@@ -23,9 +23,15 @@ TEMPLATES_PATH = os.path.join(os.path.dirname(__file__), "./templates/")
 j2_env = jinja2.Environment(
     loader=jinja2.FileSystemLoader(TEMPLATES_PATH), trim_blocks=True
 )
-CDN_URL = "https://cdn.jsdelivr.net/npm/@deck.gl/jupyter-widget@{}/dist/index.js".format(
-    DECKGL_SEMVER
-)
+
+def cdn_picker(offline=False):
+    if offline:
+        CDN_URL = os.path.abspath('../nbextension/static/index.js')
+    else:
+        CDN_URL = "https://cdn.jsdelivr.net/npm/@deck.gl/jupyter-widget@{}/dist/index.js".format(
+            DECKGL_SEMVER
+        )
+    return CDN_URL
 
 
 def render_json_to_html(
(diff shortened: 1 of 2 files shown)


bindings/pydeck/pydeck/io/html.py
```diff
@@ -23,9 +23,15 @@ def convert_js_bool(py_bool):
 j2_env = jinja2.Environment(
     loader=jinja2.FileSystemLoader(TEMPLATES_PATH), trim_blocks=True
 )
-CDN_URL = "https://cdn.jsdelivr.net/npm/@deck.gl/jupyter-widget@{}/dist/index.js".format(
-    DECKGL_SEMVER
-)
+
+def cdn_picker(offline=False):
+    if offline:
```
ajduberstein: Small style preference–

```
if offline:
    return os.path.abspath('../nbextension/static/index.js')
return  "https://cdn.jsdelivr.net/npm/@deck.gl/jupyter-widget@{}/dist/index.js".format(DECKGL_SEMVER)
```

ajduberstein:
> If this isn't what you're looking for

This is great. The one thing that I was considering was putting the contents of that `index.js` in a `script` tag of the .html file's Jinja template–this would let you share the file to other consumers without them having pydeck at the path in the script tag in the PR here. If you have thoughts on that method vs this, let me know. I can add this later.

Otherwise, if you sign the CLA, we're good to go.

bindings/pydeck/pydeck/io/html.py
bandersen23: Thank you for the feedback!

bandersen23:
@ajduberstein That's a much better idea - it will then make the HTML's portable, which is a great idea. I'm closing this PR as that idea is better and I had my git config messed up for these commits. I'll probably take a stab at the embedding of the javascript in the template instead. Thank you for the feedback!

PR #4405, opened 2020-03-18
pydeck: embed javascript for offline HTML generation

bandersen23:
For #4290

#### Background

Expansion from #4400 - closed due to a better idea for the issue. 
This will allow for an argument of `offline` and if so, will embed the necessary javascript into a `script` tag in the HTML file to allow for portability. 
#### Change List
- add offline argument to bindings
- embed index.js from the CDN url into a static folder - will need to be manually updated each release.
- write function to grab binding and add it into the template.

@ajduberstein - here's an update from your idea in #4400. I manually added the javascript from the CDN, but if there's a way to automate that on build, that would make it easier when there are updates. I don't have any ideas how to do that, but wanted to flag that for you!


ajduberstein:
Thank you for your patience here @bandersen23. Verified it locally on my machine without internet access–everything works.

> if there's a way to automate that on build

This should be fine for now–I'll get it documented and released in a beta.

PR #4421, opened 2020-03-25
Fix PointCloud sizeUnits as "meters"

twojtasz:
When the prop **sizeUnits** was set to **meters** it would access a field that didn't exist.
This has been corrected and a test added.


twojtasz:
Also verified here: https://codepen.io/robotim/pen/xxGJZvG

PR #4428, opened 2020-03-27
fix docs

damix911:
Address an error in the docs for `FirstPersonView` where the pitch angle `90` was deemed _"up"_ and `-90` was deemed _"down"_, while it's actually the opposite, which for mapping makes sense.


damix911:
@Pessimistress FIY.

PR #4441, opened 2020-04-01
pydeck: Add comments to describe example visualizations

ajduberstein:
For #4103

#### Change List
-


bindings/pydeck/examples/custom_layer.py
```diff
@@ -1,6 +1,16 @@
-import pydeck
+"""
+CustomLayer
+===========
+
+A custom layer named LabeledGeoJsonLayer from copied from this Observable Notebook for use in pydeck:
```
twojtasz, line 5: one too many *from*s

bindings/pydeck/examples/path_layer.py
```diff
@@ -1,13 +1,20 @@
+"""
+PathLayer
+=========
+
+Locations of the Bay Area Rapid Transit lines.
+"""
+
```
twojtasz, line 7: consistency nit: Just noticing i see different levels of whitespace between these comments and the imports.

PR #4485, opened 2020-04-14
pydeck: Shrink documentation thumbnail images

ajduberstein:
#### Background

The thumbnail images for our docs page can take a few seconds to load. This PR shrinks the thumbnail assets.


kylebarron:
I deployed with this commit and it is noticeably faster: https://pydeck.gl

PR #4502, opened 2020-04-15
Update trips layer documentation links

akre54:


xintongxia:
I guess there are more broken links..

PR #4517, opened 2020-04-21
pydeck: Lock pydeck to deck.gl semver ~8.1.*

ajduberstein:
For #

#### Change List
-


kylebarron:
Is there supposed to be a change other than to the Changelog?

ajduberstein:
There's also the semver lock here: bindings/pydeck/pydeck/frontend_semver.py

PR #4526, opened 2020-04-24
Add maxBoundingBox prop

### LINKED ISSUE #4373: TileLayer minZoom Parameter
Hello, again.  I am here to ask about the `minZoom` parameter.  Before the recent changes, I think that I could essentially use it to control what tiles are created but it didn't control what was rendered.  Is there a strong opinion here on this?  Was I doing it wrong?  I preferred using it for controlling creation, but not rendering.  As an example, if you go to the [example](https://github.com/uber/deck.gl/blob/master/examples/website/image-tile/app.js) I committed recently, and change the `minZoom` parameter to `-8`, the picture disappears when you zoom out.  My use case is a tiled image pyramid that does not have all the levels created (i.e, in the data for the example I provided, it would be as if there was no images past the eighth level).  If this is desired behavior, is there a way around it?

### LINKED ISSUE #4448: How can I keep cached tiles of TileLayer visible when zoom < minZoom?
I have a TileLayer that fetches nodes from my server. I set `minZoom` equal to `maxZoom` otherwise I would be fetching the same nodes but multiple times. It behaves as I want it when my current `zoom` is `>= maxZoom` by keeping the tiles in view, but when `zoom < minZoom` they disappear abruptly.

Now I know that `TileLayer` is supposed to be used when you have cached tiles of some location with different resolutions etc.. but is there any way to keep the cached tiles visible even when `zoom < minZoom`? I tried zoom = 0 and zoom = null, but then it would try to fetch bigger and bigger areas.

Here is a GIF of the behavior I have right now:

ezgif com-optimize

Relevant part of the code:

```new TileLayer({
   minZoom: zoomValue,
   maxZoom: zoomValue
   ...otherProps
})
```

ilan-gold:
This fixes #4373 and fixes #4448 as discussed here: https://github.com/visgl/deck.gl/issues/4373#issuecomment-598811633.  Please let me know if I have the geospatial stuff wrong - It's not my expertise.  I think the change https://github.com/visgl/deck.gl/compare/master...ilan-gold:ilan-gold/tile_layer_bounding_box?expand=1#diff-053e9e13138536bce20e0f1e4b409d40L110 makes sense since there is now a box provided but perhaps we should not be providing the box mentioned by default?

CC: @Pessimistress 

#### Background
Since the cache changes in the `TileLayer`, underzooming didn't allow for the lowest zoom level to remain visible.  This allows users to provide a `maxBoundingBox` prop to control this behavior.

#### Change List
- Add `maxBoundingBox` props to `TileLayer`
- Update tests and docs


test/modules/geo-layers/tile-layer/tileset-2d.spec.js
```diff
@@ -141,16 +141,18 @@ test('Tileset2D#under-zoomed', t => {
     getTileData,
     minZoom: 11,
     maxZoom: 13,
+    maxBoundingBox: [-180, -85.05113, 180, 85.05113],
     onTileLoad: () => {}
   });
-  const zoomedOutViewport = new WebMercatorViewport(
+  const zoomedInViewport = new WebMercatorViewport(
```
Pessimistress, line 147: Don't modify existing test cases; add a new one.

test/modules/geo-layers/tile-layer/utils.spec.js
```diff
@@ -107,7 +107,7 @@ const TEST_CASES = [
       zoom: 3
     }),
     maxZoom: 2,
-    output: ['0,1,2', '0,2,2', '3,1,2', '3,2,2']
+    output: ['0,1,2', '0,2,2']
```
Pessimistress, line 110: This is broken by your change

ilan-gold:
Thanks for the review.  I was unsure of the geospatial wrapping/bounding box, so I'll fix that.

modules/geo-layers/src/tile-layer/utils.js
```diff
@@ -42,7 +42,7 @@ export function getURLFromTemplate(template, properties) {
 /**
  * gets the bounding box of a viewport
  */
-function getBoundingBox(viewport, zRange) {
+function getBoundingBox(viewport, zRange, maxBoundingBox) {
```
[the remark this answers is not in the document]
ilan-gold: Should we make this a default prop for the layer or leave it to the utility as the default (just not sure about best practices for `defaultProps` and want to follow what's best) ?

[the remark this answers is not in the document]
Pessimistress: The value in `defaultProps` should be `null` for backward compatibility. Then the easiest way to handle it is in this function with `extent = extent || DEFAULT_EXTENT`.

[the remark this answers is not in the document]
ilan-gold: Gotcha, that makes sense.

docs/layers/tile-layer.md
```diff
@@ -142,6 +142,12 @@ Hide tiles when under-zoomed.
 
 - Default: 0
 
+##### `extent` (Array, optional)
+
+If provided, the layer will load and render the tiles at the in this box when underzoomed.  The box is of the form [left, top, right, bottom]
```
Pessimistress, line 147: <del>at the</del>

modules/geo-layers/src/tile-layer/tile-layer.js
```diff
@@ -14,6 +14,7 @@ const defaultProps = {
   onTileLoad: {type: 'function', value: tile => {}, compare: false},
   // eslint-disable-next-line
   onTileError: {type: 'function', value: err => console.error(err), compare: false},
+  extent: {type: 'array', optional: true, value: null, compare: false},
```
Pessimistress, line 17: Should set `compare: true` to allow deep comparison

docs/layers/tile-layer.md
```diff
@@ -142,6 +142,12 @@ Hide tiles when under-zoomed.
 
 - Default: 0
 
+##### `extent` (Array, optional)
+
+If provided, the layer will load and render the tiles in this box when underzoomed.  The box is of the form [left, top, right, bottom]
```
Pessimistress, line 147: This does not apply to geospatial views - I'd suggest `[minX, minY, maxX, maxY]`

PR #4528, opened 2020-04-25
[bin-sortor] correctly sort all values

heshan0131:
- this improvement will account for all type of values including 'string'


heshan0131:
https://github.com/keplergl/kepler.gl/pull/1070

PR #4551, opened 2020-04-29
Add website deploy script

Pessimistress:


Pessimistress:
> We should document how to use this command to publish (npm run deploy) , may be some where in dev-docs?

It is already in dev-docs: https://github.com/visgl/tsc/blob/master/developer-process/publish-checklist.md#publish-the-website This repo is missing the script.

Issue #4549, opened 2020-04-29
Weird visualizations with points and lines with large size


manmorjim:
#### Description
There is a weird behavior with points with large radius:
![points](https://user-images.githubusercontent.com/1272750/80616950-ca507800-8a41-11ea-9bae-9e0151fe9b19.png)

And the same with lines with large width:
![lines](https://user-images.githubusercontent.com/1272750/80616966-cf152c00-8a41-11ea-8f88-56359dd88acc.png)

I think it is related to the issue https://github.com/visgl/deck.gl/issues/3824
They seem to be affected by [clipping extension](https://github.com/visgl/deck.gl/blob/eb6f16321ced8e19aef926cf83f59ef899382620/modules/geo-layers/src/mvt-layer/mvt-layer.js#L35) which was added for MVTLayers in https://github.com/visgl/deck.gl/pull/4336

An suggestion would be removing that extension for lines and points layers.

#### Repro Steps
I've created two examples which reproduce the behavior.

- [example with points](https://stackblitz.com/edit/deck-gl-mvt-viz-error).
- [example with lines](https://stackblitz.com/edit/deck-gl-mvt-viz-error-lines)

#### Environment (please complete the following information):
 - Framework Version: deck.gl 8.1.0
 - Browser Version: Chrome 81.0.4044.122 (Official Build) (64-bit)
 - OS: Ubuntu 19.04

Pessimistress:
The clipping extension indeed has a bug when applied to the ScatterplotLayer. I am unable to reproduce the PathLayer issue though, can you provide a data sample?

manmorjim:
> can you provide a data sample?

Yes, [here](https://stackblitz.com/edit/deck-gl-mvt-viz-error-lines) I created an example with lines. If you zoom in in the lines then it appears some blank parts like:

![image](https://user-images.githubusercontent.com/1272750/80684206-d88bac00-8ab4-11ea-89ce-5ed465e7c783.png)

Pessimistress:
This is because you didn't create any buffer around your tile. the PathLayer is rendering an end cap right at the edge:

![image](https://user-images.githubusercontent.com/2059298/80743797-7b9eee80-8ad2-11ea-8896-0c8f48a319f6.png)

It is technically impossible to handle the joins at tile edges correctly without any additional information. The tool you use to generate the vector tiles could be configured to add a buffer, [example]([buffer](http://mapnik.org/documentation/node-mapnik/3.6/#VectorTile)).

[...]

PR #4557, opened 2020-04-30
Fix MVT clipping for ScatterplotLayer

Pessimistress:
For #4549

#### Change List
- When a layer is anchored (geometry is populated around `instancePositions`), use world position instead of common position for filtering
- Update map tile test app to use `MVTLayer`
- Tests


modules/geo-layers/src/mvt-layer/clip-extension.js
```diff
@@ -97,15 +98,19 @@ export default class ClipExtension extends LayerExtension {
 
   draw({uniforms}) {
     const {clipBounds = defaultProps.clipBounds} = this.props;
-    const corner0 = this.projectPosition([clipBounds[0], clipBounds[1], 0]);
-    const corner1 = this.projectPosition([clipBounds[2], clipBounds[3], 0]);
+    if (this.state.clippingAnchor) {
+      uniforms.clip_bounds = defaultProps.clipBounds;
```
1chandu, line 102: `uniforms.clip_bounds = clipBounds` ?

modules/geo-layers/src/mvt-layer/clip-extension.js
```diff
@@ -81,8 +81,9 @@ varying vec2 clip_commonPosition;
 };
 
 export default class ClipExtension extends LayerExtension {
-  getShaders(opts) {
+  getShaders() {
     const hasAnchor = 'instancePositions' in this.getAttributeManager().attributes;
```
1chandu, line 85: Nit: This checking to decided VS for FS clipping seems error prone, may be add a prop to each layer , similar to `shadowEnabled` ??

Pessimistress: This extension is supposed to work automatically without the user marking any layers, and I believe this is a reasonable default. I will add a prop that allows users to override the default behavior.

Issue #4521, opened 2020-04-22
Incorrect vector tile extent when using StaticMap in Deck.gl's `gl` context


kylebarron:
#### Description

Vector tiles aren't clipped:
![image](https://user-images.githubusercontent.com/15164633/79928816-3ce9a400-8401-11ea-9f89-e573d495683a.png)

Edit: This is solved, see below
White canvas while panning:
![Screen Recording 2020-04-21 at 6 52 36 PM](https://user-images.githubusercontent.com/15164633/79928929-8508c680-8401-11ea-8144-f39c37c635f9.gif)

#### Repro Steps

A small reproducible `App.js`, and the custom `style.json` it references. (`mapbox-gl.css` is the standard Mapbox GL JS css).

<details>
<summary>App.js</summary>

```jsx
import React from "react"
import DeckGL from "@deck.gl/react"
import { StaticMap } from "react-map-gl"
import { pushContextState, popContextState } from "@luma.gl/gltools"

// You'll get obscure errors without including the Mapbox GL CSS
import "../css/mapbox-gl.css";

const mapStyle = require("./style.json")

const initialViewState = {
  longitude: -112.1861,
  latitude: 36.1284,
  zoom: 12.1,
  pitch: 0,
  bearing: 0,
}

export default class Map extends React.Component {
  state = {}

  // DeckGL and mapbox will both draw into this WebGL context
  _onWebGLInitialized = gl => {
    this.setState({ gl })
  }
  
  render() {
    const { gl } = this.state
    return (
      <DeckGL
        ref={ref => {
          // save a reference to the Deck instance
          this._deck = ref && ref.deck
        }}
        initialViewState={initialViewState}
        onBeforeRender={() => pushContextState(gl)}
        onAfterRender={() => popContextState(gl)}
        controller
        onWebGLInitialized={this._onWebGLInitialized}
      >
        {gl && (
          <StaticMap
            ref={ref => {
              // save a reference to the mapboxgl.Map instance
              this._map = ref && ref.getMap()
            }}
            gl={gl}
            mapStyle={mapStyle}
            mapOptions={{ hash: true }}
          />
        )}
      </DeckGL>
    )
  }
}
```

</details>

<details>
<summary>style.json</summary>

```json
{
    "version": 8,
    "name": "OSM Liberty",
    "metadata": {
        "maputnik:license": "https://github.com/maputnik/osm-liberty/blob/gh-pages/LICENSE.md",
[...]

PR #4559, opened 2020-05-02
Update MapboxLayer docs

kylebarron:
#### Change List

- Consistent capitalization of `Mapbox` in prose
- Update Mapbox, deck.gl version numbers
- Add React example of injecting deck.gl layers into Mapbox, including responses from https://github.com/visgl/deck.gl/issues/4521

### DIFF
--- a/docs/api-reference/mapbox/mapbox-layer.md
+++ b/docs/api-reference/mapbox/mapbox-layer.md
@@ -45,8 +45,6 @@ map.on('load', () => {
 
 This option allows one to take full advantage of the `Deck` API, e.g. top-level props such as `pickingRadius`, `onHover`, and adding/removing/updating layers in a reactive fashion by setting the `layers` array.
 
-#### Pure JS
-
 ```js
 import {MapboxLayer} from '@deck.gl/mapbox';
 import {Deck} from '@deck.gl/core';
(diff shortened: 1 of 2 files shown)


docs/api-reference/mapbox/mapbox-layer.md
```diff
@@ -89,6 +91,86 @@ map.on('load', () => {
...
+  _onMapLoad = () => {
+    const map = this._map;
+    const deck = this._deck;
+
+    // You must initialize an empty deck.gl layer to prevent flashing
+    map.addLayer(
+      // This id has to match the id of the deck.gl layer
+      new MapboxLayer({ id: "my-scatterplot", deck }),
+      // Optionally define id from Mapbox layer stack under which to add deck layer
+      'beforeId'
```
ajduberstein, line 128: Probably double quotes for consistency

kylebarron: There's already some quote inconsistency across examples. Earlier examples use single quotes. Maybe I should update all quotes for all examples to be either single quotes or double quotes.

docs/api-reference/mapbox/mapbox-layer.md
```diff
@@ -89,6 +91,86 @@ map.on('load', () => {
...
+    return (
+      <DeckGL
+        ref={(ref) => {
+          // save a reference to the Deck instance
+          this._deck = ref && ref.deck;
+        }}
+        layers={layers}
+        initialViewState={initialViewState}
+        controller
+        // To render vector tile polygons correctly
```
Pessimistress, line 153: This is not valid JSX syntax

kylebarron: You're right. Fixed.

docs/api-reference/mapbox/overview.md
```diff
@@ -135,6 +135,8 @@ export class App extends React.Component {
         initialViewState={INITIAL_VIEW_STATE}
         controller={true}
         onWebGLInitialized={this._onWebGLInitialized}
+        // To render vector tile polygons correctly
```
Pessimistress, line 138: Same issue here.

I would prefer to avoid duplicating examples in multiple documentations. It's very hard to keep track of which ones to update when the API is changed. You can link to another doc instead.

kylebarron: Ok, I removed the example in `mapbox-layer.md`

PR #4574, opened 2020-05-07
[ScenegraphLayer] Fix context.animationProps undefined

chrisgervang:
#### Background
After upgrading a project that used animations embedded in a gltf model from deck 7 to 8, a new error appeared in the console and the model with the animation wasn't rendering. Other models without animations still render correctly, and removing the animation prop from the broken model fixed the rendering (although it was static). The issue was isolated to an undefined object that was previously defined.

Screen Shot 2020-05-01 at 9 11 27 PM

I think the `animationProps` object in question was deprecated and now we should use `timeline` under `context.deck.animationLoop`, and perhaps @tsherif has some more context around this history.

#### Change List
- Reference `timeline` from `deck.animationLoop` instead of `context.animationProps.time`

### DIFF
--- a/modules/mesh-layers/src/scenegraph-layer/scenegraph-layer.js
+++ b/modules/mesh-layers/src/scenegraph-layer/scenegraph-layer.js
@@ -242,7 +242,7 @@ export default class ScenegraphLayer extends Layer {
     if (!this.state.scenegraph) return;
 
     if (this.props._animations && this.state.animator && context.deck) {
-      this.state.animator.animate(context.deck.animationLoop.animationProps.time);
+      this.state.animator.animate(context.deck.animationLoop.timeline.time);
     }
 
     const {viewport} = this.context;
--- a/modules/mesh-layers/src/scenegraph-layer/scenegraph-layer.js
+++ b/modules/mesh-layers/src/scenegraph-layer/scenegraph-layer.js
@@ -241,7 +241,7 @@ export default class ScenegraphLayer extends Layer {
   draw({moduleParameters = null, parameters = {}, context}) {
     if (!this.state.scenegraph) return;
 
-    if (this.props._animations && this.state.animator) {
+    if (this.props._animations && this.state.animator && context.deck) {
       this.state.animator.animate(context.deck.animationLoop.animationProps.time);
     }
 
--- a/modules/mesh-layers/src/scenegraph-layer/scenegraph-layer.js
+++ b/modules/mesh-layers/src/scenegraph-layer/scenegraph-layer.js
@@ -242,7 +242,7 @@ export default class ScenegraphLayer extends Layer {
     if (!this.state.scenegraph) return;
 
     if (this.props._animations && this.state.animator) {
-      this.state.animator.animate(context.animationProps.time);
+      this.state.animator.animate(context.deck.animationLoop.animationProps.time);
     }
 
     const {viewport} = this.context;
--- a/modules/mesh-layers/src/scenegraph-layer/scenegraph-layer.js
+++ b/modules/mesh-layers/src/scenegraph-layer/scenegraph-layer.js
@@ -241,8 +241,9 @@ export default class ScenegraphLayer extends Layer {
   draw({moduleParameters = null, parameters = {}, context}) {
     if (!this.state.scenegraph) return;
 
-    if (this.props._animations && this.state.animator && context.deck) {
-      this.state.animator.animate(context.deck.animationLoop.timeline.time);
+    if (this.props._animations && this.state.animator) {
+      this.state.animator.animate(context.timeline.getTime());
+      this.setNeedsRedraw();
     }
 
     const {viewport} = this.context;


modules/mesh-layers/src/scenegraph-layer/scenegraph-layer.js
```diff
@@ -241,8 +241,8 @@ export default class ScenegraphLayer extends Layer {
   draw({moduleParameters = null, parameters = {}, context}) {
     if (!this.state.scenegraph) return;
 
-    if (this.props._animations && this.state.animator) {
-      this.state.animator.animate(context.animationProps.time);
+    if (this.props._animations && this.state.animator && context.deck) {
```
ibgreen, line 244: This is confusing. I can see `context.deck.animationLoop` being undefined, but I would expect `context.deck` to always be defined?

modules/mesh-layers/src/scenegraph-layer/scenegraph-layer.js
```diff
@@ -241,8 +241,8 @@ export default class ScenegraphLayer extends Layer {
...
+      this.state.animator.animate(context.deck.animationLoop.timeline.time);
```
Pessimistress, line 245: `context.timeline.getTime()`

modules/mesh-layers/src/scenegraph-layer/scenegraph-layer.js
chrisgervang: I was confused by this too. The ScenegraphLayer generated layer tests fail indicating `context.deck` is undefined. Is there supposed to be a phase where it's undefined or is it just a test anomoly?

modules/mesh-layers/src/scenegraph-layer/scenegraph-layer.js
chrisgervang: My old code is fixed when using `deck.animationLoop.timeline`, and `context.timeline` didn't fix it.

Do you know the differences, and how they should be used? I don't know how to play `context.timeline` (time doesn't move forward when I used it in my app). But `deck.animationLoop.timeline` is playable through the deck object.

Pessimistress: They should be the same object : https://github.com/visgl/deck.gl/blob/master/modules/core/src/lib/deck.js#L586

modules/mesh-layers/src/scenegraph-layer/scenegraph-layer.js
Pessimistress: There is no deck instance in the layer testing context. It is driven by a standalone `LayerManager` instance.

modules/mesh-layers/src/scenegraph-layer/scenegraph-layer.js
chrisgervang: Interesting, they don't behave like they are. Where does the context object get set? This code looks like its setting `deck.animationLoop.timeline` and `context.timeline` is set up somewhere else.

Pessimistress: https://github.com/visgl/deck.gl/blob/7b3b615edb86c7a6adf9230bdcb139a58e9933d3/modules/core/src/lib/deck.js#L584-L586

I just tested in layer-browser and they are the same object.

chrisgervang: Created an isolated example at https://github.com/visgl/deck.gl/tree/chr-animated-gltf/examples/website/animated-scenegraph to test out the two timeline object references. It seems like the intention the code is for these two objects to be the same, but I wonder if somewhere they get memoized, redefined, or copied such that they don't reference the same object by the time scenegraph layer is executed. 

Where does a layer's `context` get defined?

Here's what version with `context.deck.animationLoop.timeline.getTime()` looks like.
dce142b9-384d-ec9c-73e4-86f7d5088eed

And this with `context.timeline.getTime()`
f5474eff-3e9e-273f-a918-cdaa580dde0c

I'm using hubble.gl to record these, which is using `deck.animationLoop.timeline.play()` on a reference of the deck object.

chrisgervang: Are the object passed to `LayerManager` a layer's `context` eventually?

If so I can see how we expect the timeline nested in deck and the timeline at the surface should be the same.

chrisgervang:
We were able to resolve the "two object" issue I was noticing in my projects. It's caused from me attaching a new timeline in my hubble project, which creates.. two objects. I remember doing that when timings in deck were in refactor and timeline's weren't always attached by default.

Since everything appears to be working now from here I'll
- Update my PR to use the context timeline.
- Add a browser render test.
- Add this embedded example to the scenegraph page (separate PR)
- Update hubble to assume there's always a timeline attached already. (separate PR)

Pessimistress (approved):

chrisgervang:
Yes, I'll be following up with a browser test for the animated glTF shortly. @Pessimistress suggested I do it separately since we may run into caveats.

Issue #4576, opened 2020-05-08
pydeck: offline javascript throws error on Windows as encoding isn't set


bandersen23:
#### Description

Windows throws an error on the `js.read()` part of the cdn_picker as the default windows encoding is different from linux. 

https://github.com/visgl/deck.gl/blob/32bf7848396b817c71e87f7e9fb387f64057507f/bindings/pydeck/pydeck/io/html.py#L27

#### Repro Steps

Used the example code to test: https://github.com/visgl/deck.gl/blob/master/bindings/pydeck/examples/scatterplot_layer.py

This line is changed to:
https://github.com/visgl/deck.gl/blob/32bf7848396b817c71e87f7e9fb387f64057507f/bindings/pydeck/examples/scatterplot_layer.py#L43

```python
r.to_html("scatterplot_layer.html", notebook_display=False, offline=False)
```
#### Environment (please complete the following information):
 - Framework Version: pydeck 0.3.1
 - Browser Version: N/A in this bug
 - OS: Windows 10

#### Logs

Error thrown: 
```python
UnicodeDecodeError: 'charmap' codec can't decode byte 0x90 in position 1063293: character maps to <undefined>
```
@ajduberstein - PR coming soon.

ajduberstein:
Fixed by #4577

PR #4577, opened 2020-05-08
pydeck: set offline javascript read to use utf-8 encoding

bandersen23:
pydeck throws an error on windows due to different environment encodings.

For #4576

#### Background
Previously added in #4405

#### Change List
- set `encoding="utf-8"` on javascript read when set offline is set to True.

@ajduberstein


ajduberstein:
Looks good, thanks @bandersen23!

PR #4592, opened 2020-05-15
Add viewport.getBounds

Pessimistress:
#### Background

This is in preparation for fixing tile fetching at extreme pitch angles.

We currently calculate visible bounds of a viewport by unprojecting the four screen corners. At pitch ~= 72, the horizon comes into view and the top two corners no longer unproject to reasonable coordinates (points fall behind the camera). This PR switches to use the `getBounds` utility in @math.gl/web-mercator which produces correct results.

#### Change List
- Add `viewport.getBounds()`. Each viewport type may need to implement its own bounding box calculation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,9 @@
   "browser": {
     "jsdom": false
   },
+  "resolutions": {
+    "math.gl": "3.2.0-alpha.3"
+  },
   "devDependencies": {
     "@loaders.gl/csv": "^2.1.5",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -40,6 +40,9 @@
   "browser": {
     "jsdom": false
   },
+  "resolutions": {
```
Pessimistress, line 43: This is temporary - luma.gl still depends on math.gl 3.1.

PR #4593, opened 2020-05-15
Use frustum culling in OSM tile traversal

Pessimistress:
For #4482 

#### Change List
- Use frustum culling in OSM tile traversal
- Show tile boundary in website demo

### DIFF
--- a/examples/website/map-tile/app.js
+++ b/examples/website/map-tile/app.js
@@ -4,13 +4,14 @@ import {render} from 'react-dom';
 import DeckGL from '@deck.gl/react';
 import {MapView} from '@deck.gl/core';
 import {TileLayer} from '@deck.gl/geo-layers';
-import {BitmapLayer} from '@deck.gl/layers';
+import {BitmapLayer, PathLayer} from '@deck.gl/layers';
 
 const INITIAL_VIEW_STATE = {
   latitude: 47.65,
   longitude: 7,
   zoom: 4.5,
   maxZoom: 20,
+  maxPitch: 89,
   bearing: 0
 };
 
@@ -40,7 +41,7 @@ export default class App extends PureComponent {
   }
 
   _renderLayers() {
-    const {autoHighlight = true, highlightColor = [60, 60, 60, 40]} = this.props;
+    const {showBorder = false} = this.props;
 
     return [
       new TileLayer({
@@ -49,8 +50,8 @@ export default class App extends PureComponent {
 
         pickable: true,
         onHover: this._onHover,
-        autoHighlight,
-        highlightColor,
+        autoHighlight: showBorder,
+        highlightColor: [60, 60, 60, 40],
         // https://wiki.openstreetmap.org/wiki/Zoom_levels
         minZoom: 0,
         maxZoom: 19,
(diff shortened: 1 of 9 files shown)


modules/geo-layers/src/tile-layer/tile-2d-traversal.js
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    // Avoid loading overlapping tiles - if a descendant is requested, do not request the ancester
+    if (!this.childVisible) {
+      let {z} = this;
+      if (z < maxZ && z >= minZ) {
+        // Adjust LOD by distance to camera
+        const distance =
+          (boundingVolume.distanceTo(viewport.cameraPosition) * viewport.scale) / viewport.height;
+        z += Math.floor(Math.log2(distance));
```
xintongxia, line 46: why do you need apply `* viewport.scale / viewport.height` to  `(boundingVolume.distanceTo(viewport.cameraPosition)`? Maybe add a comment about your calculation?

modules/geo-layers/src/tile-layer/tile-2d-traversal.js
```diff
@@ -0,0 +1,132 @@
...
+  const minZ = viewport.pitch <= 60 ? maxZ : 0;
+
+  const root = new OSMNode(0, 0, 0);
+  const traversalParams = {
+    viewport,
+    cullingVolume,
+    elevationBounds: [elevationMin, elevationMax],
+    minZ,
+    maxZ,
+    offset: 0
```
xintongxia, line 112: what is `offset`?

modules/geo-layers/src/tile-layer/tile-2d-traversal.js
```diff
@@ -0,0 +1,132 @@
...
+  );
+  const cullingVolume = new CullingVolume(planes);
+
+  // Project zRange from meters to common space
+  const unitsPerMeter = viewport.distanceScales.unitsPerMeter[2];
+  const elevationMin = (zRange && zRange[0] * unitsPerMeter) || 0;
+  const elevationMax = (zRange && zRange[1] * unitsPerMeter) || 0;
+
+  // Always load at the current zoom level if pitch is small
+  const minZ = viewport.pitch <= 60 ? maxZ : 0;
```
kylebarron, line 103: Is this necessary to keep in sync with basemaps, e.g. Mapbox GL? I could envision cases where a user might want to relax this to load fewer tiles at moderate pitch, e.g. 40-60 degrees.

modules/geo-layers/src/tile-layer/tile-2d-traversal.js
```diff
@@ -0,0 +1,132 @@
...
+  // Check worlds in repeated maps
+  traversalParams.offset = -1;
+  while (root.update(traversalParams)) {
+    if (--traversalParams.offset < -MAX_MAPS) {
+      break;
+    }
+  }
+  traversalParams.offset = 1;
+  while (root.update(traversalParams)) {
+    if (++traversalParams.offset > -MAX_MAPS) {
```
kylebarron, line 126: Should this be positive `MAX_MAPS`?

modules/geo-layers/src/tile-layer/tile-2d-traversal.js
```diff
@@ -0,0 +1,132 @@
+import {CullingVolume, Plane, AxisAlignedBoundingBox} from '@math.gl/culling';
+
+const TILE_SIZE = 512;
+// number of world copies to check
+const MAX_MAPS = 3;
```
kylebarron, line 5: Should checking world copies be configurable? If I turn off `repeat: true`

image

The `onViewportLoad` callback in this viewport still holds 75 tiles, while I only count 15 tiles in the viewport. Anecdotally, it seems that many more tiles are loaded when looking west than when looking north.

examples/website/map-tile/app.js
```diff
@@ -49,8 +50,8 @@ export default class App extends PureComponent {
 
         pickable: true,
         onHover: this._onHover,
-        autoHighlight,
-        highlightColor,
+        autoHighlight: showBorder,
```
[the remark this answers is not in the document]
Pessimistress: This is an example.

modules/geo-layers/src/tile-layer/tile-2d-traversal.js
Pessimistress: This is also what mapbox-gl is doing. Lowering the threshold to 40-50 saves 2-3 tile loads tops, and may have unexpected behavioral changes for existing users (60 is the default `maxPitch`)

Issue #4596, opened 2020-05-17
Hidden surfaces elimination in ArcGIS's `DeckRenderer` extension is not implemented


damix911:
This is being worked on in #4597

#### Description
`@deck.gl/arcgis`'s `DeckRenderer` (i.e. the deck.gl extension for ArcGIS's `SceneView` component) does not seem to cull hidden surfaces with certain 3D models. It was known (the 3D integration between ArcGIS and deck.gl is not even documented yet) but we were waiting to fix other issues before working on that (like camera alignment). It has been re-discovered by the devs working on I3S but we don't think that it is specific to I3S. I think I would like to fix this first, it would make working on camera alignment less frustrating.

It should look like on the right, but with 8.1.0 and even in `master` it looks like on the left.
![image](https://user-images.githubusercontent.com/12469086/82133045-b162f900-979b-11ea-93e3-819cd92db6d4.png)

#### Repro Steps
Go to this pen https://codepen.io/dawken/pen/KKdGRvx and see that it does not look right.

#### Environment
 - deck.gl 8.1.0, loaders.gl 2.1.6, ArcGIS JS API 4.15
 - Chrome 80.0.3987.132 and probably others
 - Windows 10 and Fedora 31 and probably others

#### Logs
Mainly these two errors in the console, but believed to be unrelated:

```
[esri.views.3d.support.cameraUtils] #scaleToZoom() Cannot compute zoom from scale without a tiling scheme
dist.min.js:11 deck: WebGL2 not supported by this browser. Transition animation is disabled.
```

damix911:
I did a very rough analysis of the problem.

I think that this _could_ be an issue with `Tile3DLayer` not enabling depth testing.

If you look at the code for [ScenegraphLayer](https://github.com/visgl/deck.gl/blob/master/modules/mesh-layers/src/scenegraph-layer/scenegraph-layer.js#L241) for instance it does not enable depth testing explicitly.

On the converse this [ElevationLayer](https://github.com/visgl/deck.gl/blob/master/showcases/wind/src/layers/elevation-layer/elevation-layer.js#L83) here, does enable it.

Now, if you look at the initialization code for the entire Deck, it seems like it is [enabled at the top level](https://github.com/visgl/deck.gl/blob/master/modules/core/src/lib/deck.js#L577). So maybe this is why also the layers which are not enabling it explicitly can get away with it.

[...]

PR #4597, opened 2020-05-17
Depth testing in @deck.gl/arcgis 3D mode

damix911:
For #4596

~This is a draft PR; we haven't tracked down the issue to the actual module that is responsible. I think it's the one I maintain (`@deck.gl/arcgis`) but the problem could also be in `luma.gl`, `loaders/i3s` or `deck.gl`.~

I think it is fairly established now that the issue was in `@deck.gl/arcgis`.


damix911:
I updated the fix to (hopefully) reflect a suggestion from @Pessimistress.

Issue #4407, opened 2020-03-18
TransitionInterpolator and FlyToInterpolator for Non-Geospatial


ilan-gold:
#### Target Use case

Hello again, I am back with a familiar sort of request.  

First, could you export `TransitionInterpolator` from the core?  It does not seem to be currently exported but if it is, could you tell me how to import it?  I am prone to mistakes so could be missing something.  

Second, much like the `TileLayer`, would you be interested in separating out the geo-spatial logic of `FlyToInterpolator`?  The current `flyToViewport` function that it uses from `math.gl` is hardcoded with the geospatial use case but the algorithm it uses is general.  I can make PR's there and here if you'd be interested in separating it out.   Thanks!
#### Proposed features

 - Export `TransitionInterpolator` from the `@deck.gl/core`
 - Make `FlyToInterpolator` more general than the geo-spatial use case (PR's here and in `math.gl`).

BertCh:
I also think this would be useful

Pessimistress:
Thank you! The class is exported starting 8.2: https://deck.gl/docs/api-reference/core/transition-interpolator

PR #4607, opened 2020-05-20
Export TransitionInterpolator

ilan-gold:
This allows people to write their own `TransitionInterpolator`s without copying that code into their own code base. 

#### Background

This begins to address #4407.   I think, judging by the docs, that this is something that might have been assumed to be happening, but isn't: https://github.com/visgl/deck.gl/blob/master/docs/api-reference/view-state-transitions.md#view-state-transitions: " ...But a user can provide any custom implementation for this object using TrasitionInterpolator base class."
#### Change List
- Export `TransitionInterpolator` from the core.


Pessimistress:
8.1.8

Issue #4323, opened 2020-02-27
pydeck: 0.4.0 release tracker


ajduberstein:
For #3640

Planned updates
- [x] JupyterLab >= 2.0 integration (#4488)
- [x] Notebook/Lab integration tests in CI

Tentative updates

- [ ] Update tooltip to use @deck.gl/json
- [x] Provide a standalone HTML file that doesn't need internet (#4290)
- [x] Google Maps base maps (#4632)
- [ ] Deeper widget binding with pydeck visualization objects (#4511)
- [ ] Support multiple map views, as in the second deck.gl/playground HexagonLayer example

ajduberstein:
#4818, released here: https://pypi.org/project/pydeck/0.4.0/

PR #4573, opened 2020-05-07
Pydeck: Update @jupyter-widgets/base to support JupyterLab v2

kylebarron:
Closes #4488

#### Change List
- As suggested [here](https://github.com/jupyter-widgets/ipywidgets/issues/2798#issuecomment-593716999), allows `@jupyter-widgets/base` to resolve to `3.x`.

Tested manually so far on JupyterLab 2.1.2 and Jupyter Notebook 6.0.3.

PR #4632, opened 2020-06-02
pydeck: Add Google Maps base maps to pydeck

ajduberstein:
For #4323

#### Background

Support the addition of the Google Maps base map to pydeck

Works exclusively via `.to_html` for now, since I can add the Google Maps JS library as a script tag. Where is the most convenient place to fetch the library for the Jupyter transport module--as a custom library?

#### Change List
- Add @deck.gl/google-maps to jupyter-widget's bundle
- Add Python tests
- Include a parameter to allow switching the base map within pydeck
- Read Google Maps token in the same way the Mapbox token is read


ajduberstein:
image

modules/jupyter-widget/src/playground/create-deck.js
```diff
@@ -116,15 +118,34 @@ function createDeck({
...
-    deckgl = new deck.DeckGL({
-      ...props,
-      map: mapboxgl,
-      mapboxApiAccessToken: mapboxApiKey,
-      onClick: handleClick,
-      getTooltip,
-      container
-    });
+    const {mapProvider} = props;
+    if (mapProvider === 'mapbox') {
```
ibgreen, line 122: Nit: prefer switch statement in this situation: 

```
switch (mapProvider) {
  case 'mapbox': ... break;
  case 'google': 
```

modules/jupyter-widget/src/playground/create-deck.js
```diff
@@ -116,15 +118,34 @@ function createDeck({
...
+      deckgl = new deck.DeckGL({
+        ...props,
+        map: mapboxgl,
+        mapboxApiAccessToken: mapboxApiKey,
+        onClick: handleClick,
+        getTooltip,
+        container
+      });
+    } else if (mapProvider === 'google_maps') {
```
ibgreen, line 131: Q: Provider = `google_maps` or just `google`?  

I suppose `google_maps` is more consistent with `googleMapsKey` etc

modules/jupyter-widget/src/playground/create-deck.js
ibgreen: I suppose we could even add arcgis here now if we wanted too...

modules/jupyter-widget/src/playground/create-deck.js
```diff
@@ -97,8 +99,46 @@ function missingLayers(oldLayers, newLayers) {
   return oldLayers.filter(ol => ol && ol.id && !newLayers.find(nl => nl.id === ol.id));
 }
 
+function createStandaloneFromProvider(
+  mapProvider,
+  {props, mapboxApiKey, googleMapsKey, handleClick, getTooltip, container}
+) {
+  switch (mapProvider) {
+    case 'mapbox':
+      console.debug('Using Mapbox base maps');
```
Pessimistress, line 108: Use `log` from `@deck.gl/core`

modules/jupyter-widget/src/playground/utils/google-maps-utils.js
```diff
@@ -0,0 +1,31 @@
...
+  props,
+  googleMapsKey,
+  onClick,
+  container,
+  getTooltip,
+  onComplete
+}) {
+  googleMapsKey = props.googleMapsKey || googleMapsKey;
+  if (!googleMapsKey) {
+    deck.log.warn('No Google Maps API key set');
```
Pessimistress, line 14: This returns a function. You need to call it: `log.warn(...)()`

Make sure that you test this use case.

yarn.lock
```diff
@@ -815,94 +815,63 @@
     unique-filename "^1.1.1"
     which "^1.3.1"
 
-"@jupyter-widgets/base@^2 || ^3":
-  version "3.0.0"
-  resolved "https://registry.yarnpkg.com/@jupyter-widgets/base/-/base-3.0.0.tgz#3996d566cddb742d275b007e0713de1e17f55a44"
-  integrity sha512-un1ZTHALCwE/SAYk2gEaonYM1JoaFyhosN8a3y2bhl4N26yCB3dP1PqGHLsAFur6ZB7fwuWwBEkIZx+nOwstAQ==
+"@jupyter-widgets/base@^2":
```
Pessimistress, line 818: Why is this changed without a corresponding change in package.json?

kylebarron: I don't know, but `yarn bootstrap` also resolves this to `2.0.2` for me now.

bindings/pydeck/pydeck/io/templates/index.j2
```diff
@@ -2,6 +2,9 @@
   <head>
     <meta http-equiv="content-type" content="text/html; charset=UTF-8" />
     <title>pydeck</title>
+{% if google_maps_key %}
```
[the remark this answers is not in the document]
ajduberstein: Noted here #4644

yarn.lock
kylebarron: I think this is my fault. Somehow in https://github.com/visgl/deck.gl/pull/4573 I moved `@jupyter-widgets/base` to `peerDependencies` but either forgot to remove it from `dependencies` or rebased incorrectly or something. 

So I think I had previously built and pushed the `yarn.lock` having removed `@jupyter-widgets/base` from `dependencies`, and it resolved to `3.0.0`.

PR #4655, opened 2020-06-05
BitmapLayer supports globe projection

Pessimistress:
For #4638 

Screen Shot 2020-06-05 at 12 09 39 PM

Note that this implementation assumes that the world coordinates can be linearly interpolated along the pixels. This is not true with e.g. OSM raster tiles, which have been preprojected to the Web Mercator space. Ultimately, the issue is that the layer does not ask the user for sufficient information regarding the image. We will need to introduce additional APIs for a more generalized solution.

#### Change List
- Introduce `resolution` option to `GlobeView`
- BitmapLayer creates a mesh if needed
- Doc
- Test


kylebarron:
> Note that this implementation assumes that the world coordinates can be linearly interpolated along the pixels. This is not true with e.g. OSM raster tiles, which have been preprojected to the Web Mercator space.

I still trying to wrap my mind around this.

In the example, you're projecting [this image](https://commons.wikimedia.org/wiki/File:PathfinderMap_hires_(4996917742).jpg) onto the globe. That image is 4096x2048px, so it maps linearly to degrees of latitude and longitude spanning the globe.

If you just stretched that image vertically to be a square, that would be nearly equivalent to an OSM web mercator tile at zoom 0 (except that in web mercator the latitude extremes are 85 degrees).

So it's not clear why it works to wrap the example image around the globe, but doesn't work to project an OSM web mercator image onto the globe, given that there's a linear combination relating the two.

Pessimistress:
Stretching a non-linear y axis still yields non-linear. See https://en.wikipedia.org/wiki/Mercator_projection#/media/File:Mercator_with_Tissot's_Indicatrices_of_Distortion.svg

kylebarron:
> Stretching a non-linear y axis still yields non-linear

I understand that there will be warping no matter what when you're projecting a rectangular image onto an ellipsoid, but why is this not an issue for the provided example?

Pessimistress:
image

kylebarron:
Tissot's indicatrix Mercator:
image

Tissot's indicatrix Equirectangular (wgs84):
image

PR #4658, opened 2020-06-06
Resolve promise before calling renderSubLayers in TileLayer

kylebarron:
#### Change List

- [x] Resolves the promise created by `getTileData` before calling `renderSubLayers`. This makes it easier to pass resolved data as props to sub layers, especially when you fetch multiple data objects, such as the mesh and the texture in the TerrainLayer.
- [ ] Update docs
- [x] Simplify TerrainLayer, removing promise handling in `renderSubLayers`

I don't think this is a backwards incompatible change, since the current implementation passes _either_ a promise _or_ the resolved object to `renderSubLayers`, so the new behavior is a strict subset of the old behavior.


Pessimistress:
To make the tests work you need to avoid returning promise in `Tile2DHeader.loadData`

kylebarron:
> avoid returning promise in `Tile2DHeader.loadData`

`Tile2DHeader.loadData` returns `undefined`, but sets a promise on `this._loader`. So do you mean `this._loader` shouldn't exist?

Pessimistress:
Yes - but you also need to check if `getTileData` returns a promise. If it's not a promise, then assign it to `this.content` directly and mark `isLoaded` as true.

kylebarron:
Is this blocked by #4659?

kylebarron:
> You need to check if `getTileData` returns a promise. If it's not a promise, then assign it to `this.content` directly and mark `isLoaded` as true.
> I would not recommend removing `this._loader`. If we allow cancelling tile loading in the future, you'll need this to tell if a loading request is in progress.

I updated to reflect these comments, but I don't understand why the tests are failing.

modules/geo-layers/src/tile-layer/tile-layer.js
```diff
@@ -34,9 +34,7 @@ export default class TileLayer extends CompositeLayer {
 
   get isLoaded() {
     const {tileset} = this.state;
-    return tileset.selectedTiles.every(
-      tile => tile.layers && tile.layers.every(layer => layer.isLoaded)
-    );
+    return tileset.isLoaded;
```
Pessimistress, line 37: Sublayers may have other async props that require loading?

kylebarron: I thought those would be covered by `tile.isLoaded`, but maybe not. I reverted this.

PR #4659, opened 2020-06-07
[test-utils] Add testLayerAsync

Pessimistress:
For https://github.com/visgl/deck.gl/pull/4365#issuecomment-639642708

Note that the skipped MVTLayer test still does not pass with this PR. It will rely on https://github.com/visgl/deck.gl/pull/4658

#### Change List
- Add `testLayerAsync` to test-utils
- Remove deprecated `testUpdateLayer` and `testDrawLayer` (not documented, not used)
- Docs


Pessimistress:
> Is it just waiting for all the async props for that layer to load

It's waiting for `layer.isLoaded` to be true. See #4091

Issue #3476, opened 2019-08-21
Using fill patterns for polygons 


jefffriesen:
I'm personally using the GeojsonLayer, but I imagine this questions applies to polygons or other layers in general. 

Is it possible to use fill patterns instead of solid fills, either on their own or on hover?

I realize that it is using canvas and not SVG, but I don't know if that means fill patterns are impractical. 

In SVG, you could define a pattern:
```
  <pattern id="pattern-circles" x="0" y="0" width="40" height="40" patternUnits="userSpaceOnUse">
    <circle mask="url(#fade)" cx="20" cy="20" r="20" />
  </pattern>
  <!-- The canvas with our applied pattern -->
  <rect x="0" y="0" width="100%" height="100%" fill="url(#pattern-circles)" />
```

Or using it as a fill style (http://iros.github.io/patternfills/sample_css.html)
```
<svg height="100" width="100" style="float:left" class="pattern-swatch">
  <rect style="fill: url(#circles-1) #fff;" x="0" y="0" height="100" width="100"></rect>
</svg>
```

Thanks

1chandu:
> Is it possible to use fill patterns instead of solid fills, either on their own or on hover?

deck.gl uses WebGL to render triangles with a color. Supporting such a feature requires some kind of post processing, our [PostProcessing Effects ](https://deck.gl/#/documentation/deckgl-api-reference/effects/post-process-effect) is currently experimental. You can try using `dotScreen` effect CC @jianhuang01

scoobyshi:
It doesn't appear this is yet available in core?  Is there an experimental release we can import to try it out?  Also, is this something that could be used on a per layer basis or the effect applies to all layers?

danielwhatmuff:
+1

Pessimistress:
The [FillStyleExtension](https://github.com/visgl/deck.gl/blob/master/docs/api-reference/extensions/fill-style-extension.md) will be available in 8.2.0. You can try it now with `8.2.0-beta.1`.

PR #4676, opened 2020-06-13
Add FillStyleExtension

Pessimistress:
For #3476 

deck gl-pattern-fill

cc @alee8 

#### Change List
- Add `FillStyleExtension`
- Tests
- Documentation


modules/extensions/src/fill-style/shaders.glsl.js
```diff
@@ -0,0 +1,109 @@
+import {project} from '@deck.gl/core';
+
+/*
+ * data filter shader module
```
xintongxia, line 4: fill style shader

examples/layer-browser/data/pattern.json
```diff
@@ -0,0 +1,30 @@
+{
+  "hatch-1x": {
+    "x": 4,
+    "y": 4,
+    "width": 120,
+    "height": 120,
+    "mask": true
```
xintongxia, line 7: does this extension support per object mask?

Pessimistress: Not at the moment. I am trying to reduce the number of attributes that are added by this extension so that it can be used with more layers.

PR #4681, opened 2020-06-15
Extract children from Fragment too

kevinresol:
#### Background

Support extracting Layers from Fragment. e.g.

```jsx
<DeckGL>
  {renderLayers()}
</DeckGL>

function renderLayers() {
  return (<><IconLayer/><IconLayer/></>);
}
```

#### Change List
- Added support for extracting layers from Fragment


Pessimistress:
Could you add a test case?

Pessimistress:
This is included in v8.1.10

PR #4686, opened 2020-06-17
Relax maxPitch in terrain example

kylebarron:
#### Change List

- Set `maxPitch` to 89
- Fix occasional errors in TerrainLayer when data prop is `null`. Specifically, some Mapbox Terrain RGB tiles don't exist (e.g. over oceans), so I was getting `TypeError: Invalid attempt to destructure non-iterable instance` errors.

image

cc @chrisgervang


modules/geo-layers/src/terrain-layer/terrain-layer.js
```diff
@@ -146,7 +146,7 @@ export default class TerrainLayer extends CompositeLayer {
   renderSubLayers(props) {
     const SubLayerClass = this.getSubLayerClass('mesh', SimpleMeshLayer);
     const {data, color} = props;
-    const [mesh, texture] = data;
+    const [mesh, texture] = data || [];
```
Pessimistress, line 149: Just return null if data is not available?

kylebarron: Oh duh that does make more sense

modules/geo-layers/src/terrain-layer/terrain-layer.js
```diff
@@ -165,7 +170,7 @@ export default class TerrainLayer extends CompositeLayer {
     }
 
     const {zRange} = this.state;
-    const ranges = data.map(arr => {
+    const ranges = data.filter(arr => arr).map(arr => {
```
kylebarron, line 173: I had to add this filter because there can now be `null`s, in the list, and `arr[0]` will fail.

Pessimistress: `filter(Boolean)`

kylebarron: So does that mean I can use just `filter(arr)` or `filter(Boolean(arr))`?

Pessimistress: `arr.filter(Boolean)`. `Boolean` is a function.

Issue #4695, opened 2020-06-19
No longer able to detect when hovering ends


mitchwilliamson:
#### Description
After updating to 8.1.10 our tooltip logic appears to no longer function. We use a similar method to the [tooltip example from the docs](https://deck.gl/#/documentation/developer-guide/adding-interactivity?section=the-picking-info-object). Previously we could detect the mouse leaving a point by checking if the picking info `.object` property is defined. After updating it appears that this is no longer the case as `onHover` isn't called with an empty object.

I've used the example code from the link above and it also appears to no longer function correctly.

Downgrading to 8.1.9 fixes the issue, so it may be related to https://github.com/visgl/deck.gl/pull/4685? 

#### Repro Steps
```js
    const layer = new ScatterplotLayer({
      data: [{ position: [-122.45, 37.78], message: 'Hover over me' }],
      getPosition: d => d.position,
      getRadius: 1000,
      getFillColor: [255, 255, 0],
      pickable: true,
      onHover: info => console.log(info.object ? 'hovering' : 'left')
    });
```
1. Hover over the circle.
2. Leave the circle.
3. `left` is never logged.

My output on v8.1.9
```
hovering
left
```

My output on v8.1.10
```
hovering
```

#### Environment (please complete the following information):
 - deck.gl 8.1.10
 - Browser Version: Electron 8.3.2 (Chrome 80)
 - OS: Windows 10

Pessimistress:
Try 8.1.11

mitchwilliamson:
It works fine again in 8.1.11, thanks!

PR #4697, opened 2020-06-19
Call onHover on all affected layers

Pessimistress:
For https://github.com/visgl/deck.gl/issues/4695

See inline comments

#### Change List
- Restore calling `onHover` for mouse out event


Pessimistress:
@xintongxia We don't have anything set up to test interactive picking. I'll look into it in a followup.

PR #4703, opened 2020-06-20
v8.2 what's new

Pessimistress:
cc @jesusbotella @ilan-gold

### DIFF
--- a/docs/whats-new.md
+++ b/docs/whats-new.md
@@ -2,6 +2,54 @@
 
 This page contains highlights of each deck.gl release. Also check our [vis.gl blog](https://medium.com/vis-gl) for news about new releases and features in deck.gl.
 
+## deck.gl v8.2
+
+Release Data: TBD, 2020
+
+(Need screenshots: globe view, pattern fill, MVT cross-tile highlight)
+
+### Tiled Layers
+
+Many new features are added to `TileLayer`, `MVTLayer` and `TerrainLayer` to improve correctness, performance and ease of use.
+
+- `MVTLayer` projects more accurately at high zoom levels.
+- `autoHighlight` in the `MVTLayer` now works on features that are split across multiple tiles, identified by the new `uniqueIdProperty` prop.
+- All geospatially tiled layers now handle high pitch angles better. Far away tiles from the camera are loaded at lower zoom levels, avoiding loading too many tiles.
+- Tiled layers now use a request scheduler to prioritize loading the most recently visible tiles during viewport navigation. See the new `throttleRequests` and `maxRequests` props.
+- `TileLayer` may continue to display tiles when underzoomed. See the new `extent` prop.
+- Fine-tune the tiles displayed by the `TileLayer` with the new `zRange` prop when dealing with 3D content. This is used by the `TerrainLayer` when viewing high-altitude regions.
+- `TileLayer`'s `tileSize` prop can be used to fine-tune the zoom level at which tiles are loaded.
+- `TileLayer`'s `renderSubLayers` is now always called after the tile layer is loaded, i.e. `props.data` is never a Promise.
+
+
+### GlobeView
+
+For geospatial data, a new projection mode -- globe projection is now available alongside Web Mercator projection. In this release, the feature is exposed via the experimental `GlobeView` class.
+
+Currently there is no base map provider under this view. You may use the `BitmapLayer` or `GeoJsonLayer` to render a backdrop for your data.
+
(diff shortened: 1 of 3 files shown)


docs/whats-new.md
```diff
@@ -2,6 +2,54 @@
...
+
+
+### Antimeridian Handling
+
+- `GreatCircleLayer` now renders correctly across the 180th meridian.
+- The `PathLayer`, `PolygonLayer` and `GeoJsonLayer` now supports the `wrapLongitude` prop.When enabled, the connection between any two neighboring vertices is drawn on the shorter side of the world, and split into two if it crosses the 180th meridian. Note that this introduces CPU overhead at runtime.
+
+### Pydeck
+
+Pydeck 0.4.0 introduces support for JupyterLab 2.0, support for a Google Maps base map, a new UI element for providing text descriptions of a map, and many of the new features of deck.gl 8.4, like the ability to render data to either a globe or a Mercator projection. See the new [website](https://pydeck.gl/) for documentation and examples.
```
ajduberstein, line 41: *deck.gl 8.2

ajduberstein, line 41: My fault here, deck.gl 8.2, not 8.4

docs/whats-new.md
```diff
@@ -2,6 +2,54 @@
...
+
+### Tiled Layers
+
+Many new features are added to `TileLayer`, `MVTLayer` and `TerrainLayer` to improve correctness, performance and ease of use.
+
+- `MVTLayer` projects more accurately at high zoom levels.
+- `autoHighlight` in the `MVTLayer` now works on features that are split across multiple tiles, identified by the new `uniqueIdProperty` prop.
+- All geospatially tiled layers now handle high pitch angles better. Far away tiles from the camera are loaded at lower zoom levels, avoiding loading too many tiles.
+- Tiled layers now use a request scheduler to prioritize loading the most recently visible tiles during viewport navigation. See the new `throttleRequests` and `maxRequests` props.
+- `TileLayer` may continue to display tiles when underzoomed. See the new `extent` prop.
```
kylebarron, line 19: From the API docs

> If provided, the layer will load and render the tiles in this box when underzoomed

I can't tell if that means _new_ tiles will also be loaded or only that tiles already loaded will continue to be displayed. Could be good to clarify, maybe in a separate PR.

ilan-gold:
Looks great.  Thanks for all the good work everybody!

jesusbotella:
This is awesome. Thank you very much for your work and improvements.

Pessimistress:
> Do you want me to take some MVT highlighting screenshots comparing previous and current behaviour?

@jesusbotella that would be great! Thank you!

jesusbotella:
I captured these two animated GIF to showcase the previous and the current auto highlight behaviour:

brokenautoHighlight
goodautoHighlight

Let me know if these two are OK, or if you will make some changes.

ibgreen:
@jesusbotella That looks great.

By the way, I was looking at the [MVTTileLayer documentation](https://deck.gl/#/documentation/deckgl-api-reference/layers/mvtlayer) in deck.gl and I don't seem to be able to turn on auto highlighting in that example. 

Not sure if it is the tile source or some missing setting, but I think the cross-tile auto-highlighting is one of the strongest features of this layer, would be nice if the example came with working auto highlight.

But perhaps it is hard to set up declaratively via just props?

Pessimistress:
@ibgreen autohighlight was not supported in 8.1.

ibgreen:
@Pessimistress Actually I was testing this on 8.2 staging where it still wasn't working. I just didn't want to add that link here.

jesusbotella:
@ibgreen @Pessimistress Going to take a look at it ;)

jesusbotella:
@ibgreen Highlighting across tiles doesn't work because there's no ID or property provided to identify a unique feature.

Mapbox's tile features have a numeric ID, but the IDs collide because tiles contain several layers and we are treating features from different layers as if they were all from the same layer. So the highlight is kinda misleading because selects features that do not intersect or that are not part of the same feature because of the ID collision. We can tackle that issue to provide a better approach to it.

Given that the initial viewport of the example showcases SF streets, there's a common property across all those lines called `name` that could work for highlighting that kind of geometries.

Do you want me to open a PR for that?

ibgreen:
> Do you want me to open a PR for that?

Yes, if not too much work, I personally think it would be really neat to show off the highlighting feature.

I guess the user would still need to click the autohighlight checkbox in the side panel to enable it, but at least it would now work.

PR #4705, opened 2020-06-20
Update TileLayer API docs

kylebarron:
#### Change List

- Add `zRange`, `throttleRequests`, `maxRequests` props to tile layer API docs


docs/layers/tile-layer.md
```diff
@@ -173,19 +174,37 @@ How the tile layer refines the visibility of tiles. One of the following:
...
 
+##### `throttleRequests` (Boolean, optional)
+
+Whether to throttle `getTileData` requests.
+
+If `true`, a maximum of `maxRequests` instances of `getTileData` will be called concurrently. Additionally, for requests that are throttled, if the tile is no longer visible in the viewport when a request slot opens up, that request will be cancelled. When requests are not throttled, `getTileData` will be called for all visible tiles, regardless of how long that tile is or was visible.
+
+- Default: `true`
+
+##### `maxRequests` (Number, optional)
```
Pessimistress, line 185: Can we merge these two props? I.e. if `maxRequest > 0` then enable `throttleRequests`.

docs/layers/tile-layer.md
```diff
@@ -173,19 +174,37 @@ How the tile layer refines the visibility of tiles. One of the following:
...
 * `id`: An unique id for this sublayer
-* `data`: Resolved from `getTileData`
+* `data`: Resolved from `getTileData`. As of deck.gl 8.2, this prop is always the data resolved from the Promise and is never a Promise itself.
 * `tile`: An object containing tile index `x`, `y`, `z`, and `bbox` of the tile.
 
 - Default: `props => new GeoJsonLayer(props)`
 
+##### `zRange` (Array, optional)
+
+An array representing the range of minimum and maximum heights in the tile. This is designed to support extruded layers, such as the TerrainLayer, ensuring the right tiles are loaded and rendered. This prop currently only has effect when used with a geospatial view.
```
Pessimistress, line 205: I believe it is respected in non-geospatial as well.

kylebarron: From [here](https://github.com/visgl/deck.gl/blob/d96e91f56f2c120d59c1c5ee2832848bed9be139/modules/geo-layers/src/tile-layer/utils.js#L140-L142) I assumed it was only geospatial

docs/layers/tile-layer.md
kylebarron: I figured that would come up during API review; I already set `throttleRequests` to `false` on the request scheduler when `maxRequest <=0`

https://github.com/visgl/deck.gl/blob/d96e91f56f2c120d59c1c5ee2832848bed9be139/modules/geo-layers/src/tile-layer/tileset-2d.js#L61

kylebarron: I removed `throttleRequests` from the docs, pending #4727.

PR #4710, opened 2020-06-22
Bump loaders.gl to 2.2.3

kylebarron:
#### Change List

- Bump loaders.gl to 2.2.3


modules/geo-layers/package.json
```diff
@@ -45,6 +45,6 @@
     "@deck.gl/core": "^8.0.0",
     "@deck.gl/layers": "^8.0.0",
     "@deck.gl/mesh-layers": "^8.0.0",
-    "@loaders.gl/core": "^2.2.0"
+    "@loaders.gl/core": "^2.2.3"
```
Pessimistress, line 48: Don’t bump `peerDependencies`. These do not affect package resolution and we intend to keep them relaxed.

modules/layers/package.json
```diff
@@ -29,13 +29,13 @@
...
   "dependencies": {
-    "@loaders.gl/images": "^2.2.2",
+    "@loaders.gl/images": "^2.2.3",
     "@mapbox/tiny-sdf": "^1.1.0",
     "@math.gl/polygon": "^3.2.1",
     "earcut": "^2.0.6"
   },
   "peerDependencies": {
     "@deck.gl/core": "^8.0.0",
-    "@loaders.gl/core": "^2.2.0"
```
Pessimistress, line 39: Same

modules/geo-layers/package.json
kylebarron: Ok, I think those were automatically incremented by `ocular-bump`..

PR #4727, opened 2020-06-25
Remove throttleRequests TileLayer prop

kylebarron:
#### Change List

- When maxRequests <= 0, set `throttleRequests` to `false`
- Docs to be updated in #4705


test/modules/geo-layers/tile-layer/tile-2d-header.spec.js
```diff
@@ -5,7 +5,7 @@ import {RequestScheduler} from '@loaders.gl/loader-utils';
 test('Tile2DHeader', async t => {
   let onTileErrorCalled = false;
 
-  const requestScheduler = new RequestScheduler({throttleRequests: false});
+  const requestScheduler = new RequestScheduler({maxRequests: -1});
```
kylebarron, line 8: Oh here I'm instantiating the RequestScheduler directly, so I still need `throttleRequests: false`

PR #4743, opened 2020-06-29
pydeck: Simplify Deck.to_html()

ajduberstein:
Simplifies `Deck.to_html`, which will now return a string or an iframe with a srcdoc depending on which arguments are passed.

The srcdoc is particularly useful because this means that pydeck doesn't have to write a separate document from a Notebook, which makes it easier to get work into nbviewer

#### Change List
- Automatically detect whether or not pydeck is running in a Jupyter environment (with manual override in case this fails)
- Write a srcdoc in an iframe instead of a separate document
- Eliminate temporary files in pydeck

### DIFF
--- a/bindings/pydeck/pydeck/bindings/deck.py
+++ b/bindings/pydeck/pydeck/bindings/deck.py
@@ -9,6 +9,7 @@ from ..settings import settings as pydeck_settings
 from ..widget import DeckGLWidget
 from .view import View
 from .providers import Providers
+from .map_styles import LIGHT
 
 
 class Deck(JSONMixin):
@@ -16,7 +17,7 @@ class Deck(JSONMixin):
         self,
         layers=[],
         views=[View(type="MapView", controller=True)],
-        map_style="mapbox://styles/mapbox/dark-v9",
+        map_style=LIGHT,
         mapbox_key=None,
         google_maps_key=None,
         initial_view_state=None,
@@ -51,7 +52,7 @@ class Deck(JSONMixin):
             to get an API token.
             If not using a basemap, you can set this value to `''`.
         google_maps_key : str, default None
-            Read on initialization from the ``GOOGLE_MAPS_API_KEY`` environment variable if not set.
+            Read on initialization from the ``PYDECK_GOOGLE_MAPS_API_KEY`` environment variable if not set.
             Defaults to None if the environment variable is also not set.
             Not used on all layers.
         map_provider : str, default 'mapbox'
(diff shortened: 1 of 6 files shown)


ajduberstein:
@ajfriend as mentioned / requested

bindings/pydeck/pydeck/io/html.py
```diff
@@ -64,37 +75,27 @@ def render_json_to_html(
...
-    return path
+    return HTML(iframe)
 
 
-def add_html_extension(fname):
-    SUFFIX = ".html"
-    if fname.endswith(SUFFIX):
-        return str(fname)
-    return str(fname + ".html")
+def render_for_collab(html_str, iframe_height):
```
ajfriend, line 95: `collab` to `colab`?

bindings/pydeck/pydeck/io/html.py
```diff
@@ -10,6 +11,16 @@
 from ..frontend_semver import DECKGL_SEMVER
 
 
+def in_jupyter():
+    try:
+        ip = get_ipython()  # noqa
+
+        if ip.has_trait("kernel"):
```
ajfriend, line 18: This function will return `None` if `ip.has_trait("kernel")` is false. Maybe just do `return ip.has_trait("kernel")`?

ajfriend:
Nice!

bindings/pydeck/pydeck/bindings/deck.py
```diff
@@ -150,17 +151,19 @@ def to_html(
...
-            Attempts to display the HTML output in an iframe if True. Only works in a Jupyter environment.
-        iframe_width : int, default 700
-            Height of Jupyter notebook iframe in pixels, if rendered in a Jupyter environment.
-        iframe_height : int, default 500
+            Whether a browser window will open or not after write.
+        notebook_display : bool, default None
+            Display the HTML output in an iframe if True. Set to True automatically if rendering in Jupyter.
+        iframe_width : str or int, default '100%'
             Width of Jupyter notebook iframe in pixels, if rendered in a Jupyter environment.
+        iframe_heigth : int, default 500
```
ajfriend, line 161: `heigth` to `height`

ajfriend:
For the `Deck` object, what about adding something like

```python
def _repr_html_(self):
    return self.to_html(notebook_display=True)
```

Or maybe we can play with the `_repr_`s in a different PR?

ajduberstein:
> Or maybe we can play with the _repr_s in a different PR?

This is great. Yes, let's do a separate PR.

bindings/pydeck/pydeck/bindings/deck.py
```diff
@@ -51,7 +52,7 @@ def __init__(
             to get an API token.
             If not using a basemap, you can set this value to `''`.
         google_maps_key : str, default None
-            Read on initialization from the ``GOOGLE_MAPS_API_KEY`` environment variable if not set.
+            Read on initialization from the ``PYDECK_GOOGLE_MAPS_API_KEY`` environment variable if not set.
```
kylebarron, line 55: Is this important to be prefixed with `PYDECK`? The Mapbox key isn't

bindings/pydeck/pydeck/bindings/deck.py
```diff
@@ -150,17 +151,19 @@ def to_html(
         Parameters
         ----------
         filename : str, default None
-            Name of the file. If no name is provided, a randomly named file will be written locally.
+            Name of the file.
```
kylebarron, line 154: What happens now when not provided?

PR #4750, opened 2020-07-03
Fix TextLayer fragment shader in Edge Legacy browser

danmarshall:
I had a bug that crashed the Edge Legacy browser. Determined the cause to be usage of the "return" statement in the fragment shader. Re-worked the logic flow with "else" clauses to achieve the same logic without using "return."


danmarshall:
In the Edge Legacy browser, this page will fail: https://deck.gl/docs/api-reference/layers/text-layer
But other layers succeed.

PR #4762, opened 2020-07-09
pydeck: Support globe view example

ajduberstein:
#### Background

Support the globe view within pydeck

image

#### Change List
- Add globe example
- Modify view to allow for height and width arguments
- Apply Black py34 formatter

### DIFF
--- a/bindings/pydeck/examples/globe_example.py
+++ b/bindings/pydeck/examples/globe_example.py
@@ -1,0 +1,67 @@
+import pydeck
+import pandas as pd
+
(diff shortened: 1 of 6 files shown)


bindings/pydeck/examples/globe_example.py
```diff
@@ -0,0 +1,59 @@
...
+]
+
+deck = pydeck.Deck(
+    views=[view],
+    initial_view_state=view_state,
+    tooltip={"text": "{name}, {primary_fuel} plant, {country}"},
+    layers=layers,
+)
+# Note that this must be set for the globe to be opaque
+deck.parameters = {"cull": True}
```
kylebarron, line 56: Can this be added in the initial `Deck()` call?

Issue #4767, opened 2020-07-10
Mapbox layer throws an error when leaving the page


jjballano-qatium:
#### Description
I'm working in a React app where we want to have deck and mapbox layers. 
The code we have is something like this, removing many other lines of code that I think are not needed here:
```
const onLoad = (): void => {
    map.current.addLayer(
      new MapboxLayer({ id: "mylayerid", deck: deck.current }),
      "settlement-label"
    );
  };
[....]
<DeckGL width="100%"
            height="100%"
            effects={[]}
            ref={(ref): void => {
              deck.current = ref?.deck;
            }}
            initialViewState={viewState}
            onViewStateChange={onViewStateChange}
            layers={layers()}
            onHover={onMouseMove}
            onDrag={onMouseMove}
            controller={true}
            onLoad={onLoadDeck}
            onWebGLInitialized={onWebGLInitialized}
          >
            {glRef.current && (
              <ReactMapGL
                onLoad={onLoad}
                ref={(ref): void => {
                  map.current = ref?.getMap();
                }}
                width="100%"
                height="100%"
                mapboxApiAccessToken={token}
                preventStyleDiffing={true}
                mapStyle={styleId}
                gl={glRef.current}
              >
              </ReactMapGL>
            )}
          </DeckGL>
```

This is working well, but the error is thrown when I navigate to another page using react-router-dom. 
This makes Deck to execute [finalize method](https://github.com/visgl/deck.gl/blob/master/modules/react/src/deckgl.js#L111) which, among other things, removes the layerManager.
After that happened, Mapbox fire the render event, which [mapbox-layer is subscribed](https://github.com/visgl/deck.gl/blob/master/modules/mapbox/src/deck-utils.js#L58) so it gets executed and uses the layerManager.
In some other tests I made, [drawLayer is also executed by Mapbox](https://github.com/visgl/deck.gl/blob/master/modules/mapbox/src/mapbox-layer.js#L40) with same result of an error being thrown.

I've made a couple of changes [[1]](https://github.com/qatium/deck.gl/blob/master/modules/mapbox/src/deck-utils.js#L59) and [[2]](https://github.com/qatium/deck.gl/blob/master/modules/mapbox/src/deck-utils.js#L87) and now it's working, but I'm not sure if it's the best solution. If you think it's good enough, i could create a PR to you.

#### Environment (please complete the following information):
[...]

PR #4774, opened 2020-07-11
Avoid error when mapbox fires a render event and layerManager is null because finalize has been executed already

jjballano-qatium:
Issue #4767


jjballano-qatium:
I'll create another PR

PR #4776, opened 2020-07-12
@deck.gl/jupyter-widget: Modify Mapbox warning suppression logic

ajduberstein:
Add pydeck attribution to maps and move Mapbox CSS warning suppression.

#### Change List
-


modules/jupyter-widget/src/playground/utils/mapbox-utils.js
```diff
@@ -19,11 +19,19 @@ export function loadMapboxCSS(url = MAPBOX_CSS_URL) {
...
+  for (const c of classes) {
+    const el = document.getElementsByClassName(c)[0];
+    if (el) {
+      el.style.display = 'none';
+    }
   }
+  const footer = document.getElementsByClassName('mapbox-ctrl-mapboxgl-ctrl-attrib');
+  const aTag = document.createElement('a');
+  aTag.href = 'https://pydeck.gl';
+  aTag.innerText = '© pydeck';
```
kylebarron, line 36: Should we have a copyright symbol? There's no pydeck licensing terms. It might be better to only have `pydeck`?

PR #4782, opened 2020-07-13
Fix MapboxLayer error after Deck is finalized

jjballano-qatium:
Issue #4767


jjballano-qatium:
I've made another push to my master branch and it is not updated here, I don't know if it is related with the problems Github is facing this morning. 
Anything else I could do to update this PR?

Issue #4798, opened 2020-07-17
Point cloud example PLY


cdigruttola:
Hi,

I'm trying to use the example in https://github.com/visgl/deck.gl/tree/master/examples/website/point-cloud and in having a problem with PLY file. I uncommented the PLY rows and commented the LAS rows
`//import {LASWorkerLoader} from '@loaders.gl/las';`
`import {PLYWorkerLoader} from '@loaders.gl/ply';`
`//registerLoaders(LASWorkerLoader);`
`registerLoaders(PLYWorkerLoader);`
`//const LAZ_SAMPLE =`
`//  'https://raw.githubusercontent.com/visgl/deck.gl-data/master/examples/point-cloud-laz/indoor.0.1.laz';`
`// Data source: The Stanford 3D Scanning Repository`
`const PLY_SAMPLE =`
`   'https://raw.githubusercontent.com/visgl/deck.gl-data/master/examples/point-cloud-ply/lucy800k.ply';`
and I change data from LAZ_SAMPLE to PLY_SAMPLE
`  const layers = [`
`    new PointCloudLayer({`
`     id: 'ply-point-cloud-layer',`
`     data: PLY_SAMPLE,`
`    onDataLoad,`
`   coordinateSystem: COORDINATE_SYSTEM.CARTESIAN,`
`     getNormal: [0, 1, 0],`
`     getColor: [255, 255, 255],`
`     opacity: 0.5,`
`     pointSize: 0.5`
`   })`
` ];`

then I run yarn start-local but on the web page I can't see the cloud point.

Are there problems with PLY?

Pessimistress:
The point cloud is loaded, it's the camera that is not zoomed correctly. I think the example is currently looking for LAZ-specific metadata. You can try

```js
const onDataLoad = ({header}) => {
    if (header.boundingBox) {
      const [mins, maxs] = header.boundingBox;
      updateViewState({
      ...
    }
    ...
};
```

cdigruttola:
Thank you :)

PR #4800, opened 2020-07-17
Use standard API in point cloud example

Pessimistress:
For https://github.com/visgl/deck.gl/issues/4798

#### Change List
- Use standard loaders.gl mesh header API


Pessimistress:
Yes https://loaders.gl/docs/specifications/category-mesh#header

PR #4805, opened 2020-07-20
Forward more props from TerrainLayer to TileLayer

kylebarron:
#### Background

The TerrainLayer documentation says 

> When in Tiled Mode, inherits from all TileLayer properties

but this isn't actually true, because while `defaultProps` includes `TileLayer` props, those props are never actually passed to the `TileLayer`. Currently it appears that _no `TileLayer` props_ are passed on:
https://github.com/visgl/deck.gl/blob/0afd4e99a6199aeec979989e0c361c97e6c17a16/modules/geo-layers/src/terrain-layer/terrain-layer.js#L197-L217

We have an application for which we want to lower `maxRequests` for the `TerrainLayer`, but including it on the `TerrainLayer` has no effect.

Is there an ideal way to pass all props down to the `TileLayer`? I've heard we wish to avoid `...props`?

#### Change List

- Forward more props from TerrainLayer to TileLayer


grahambates-code:
If I wanted to set a property on bitmap layer, so even deeper than tile layer, is there any way to do this dynamically, without requesting a change like was made here. My use case is rendering the terrain Tiles in greyscale / desaturated.

PR #4807, opened 2020-07-21
Do not create unnecessary picking colors buffer

Pessimistress:
#### Background

After this PR, if a layer is never pickable or uses auto highlight, its picking colors attribute is not populated. This makes it so that:

- A non-pickable layer using large external binary will not throw the 'index out of picking color range' error
- Saves attribute packing time and memory for non-pickable layers

Note that a layer that was once pickable and then turned off will still update its picking colors buffer.

#### Change List
- Disable `instancePickingColors`/`pickingColors` attribute if `pickable` is off and `highlightedObjectIndex` is not specified

### DIFF
--- a/modules/core/src/lib/attribute/attribute.js
+++ b/modules/core/src/lib/attribute/attribute.js
@@ -294,6 +294,9 @@ export default class Attribute extends DataColumn {
 
   /* eslint-disable max-depth, max-statements */
   _autoUpdater(attribute, {data, startRow, endRow, props, numInstances}) {
+    if (attribute.constant) {
+      return;
+    }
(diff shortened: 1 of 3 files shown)


Pessimistress:
> Do instanced layers share picking buffers these days?

They share the same TypedArray on CPU, but not the WebGL buffer (modified per-layer during multi-depth picking)

contra:
Not sure if anyone noticed yet but this PR caused nebula.gl to start blowing up - opened a PR to address it there, not sure if it was intentional to make oldProps required in updateState or not.

https://github.com/uber/nebula.gl/pull/486

PR #4817, opened 2020-07-23
Add missing @loaders.gl/core dependency

walkerburgin:
#### Background
Add a missing `@loaders.gl/core` dependency to `@deck.gl/geo-layers`. I came across this while moving a repo from Yarn to PNPM (which is stricter about requiring dependencies to be fully declared). 

#### Change List
-Add missing @loaders.gl/core dependency


walkerburgin:
I just realized it's already declared as a peer dependency so this isn't quite right. Going to close for now.

PR #4829, opened 2020-07-27
pydeck: Add comment to GlobeView example

ajduberstein:
#### Background

Add missing data source comment.

#### Change List
- Add comment
- Remove unused keyword `get_line_color`
- Add water geojson

### DIFF
--- a/bindings/pydeck/examples/globe_view.py
+++ b/bindings/pydeck/examples/globe_view.py
@@ -1,8 +1,16 @@
+"""
+GlobeView
+=========
+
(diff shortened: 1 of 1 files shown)


ajduberstein:
Adding the oceans appears to create some kind of artifact: https://codepen.io/ajduberstein/pen/oNbOagb

Pessimistress:
You should avoid using overly complicated GeoJSONs. See https://deck.gl/docs/api-reference/core/globe-view#remarks for your options

Issue #4809, opened 2020-07-21
How to change mapStyle?


jhubley:
Hello!
I have a deckgl visualization of some arcs. I want to be able to switch between two mapbox styles (light and satellite). I'm trying to use setProps to do this, but the mapStyle isn't changing. Below is the full code showing an example of what I'm trying to do. What am I doing wrong? Thanks!

```
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="UTF-8">
  <meta name="viewport" content="width=device-width, initial-scale=1.0">
  <meta http-equiv="X-UA-Compatible" content="ie=edge">
  <title>Switch map style</title>
  <link rel='stylesheet' href='https://api.tiles.mapbox.com/mapbox-gl-js/v0.44.1/mapbox-gl.css'>
  <script src="https://d3js.org/d3.v5.min.js"></script>
  <script src="//code.jquery.com/jquery-1.10.2.js"></script>
  <style>
  html{
    margin,padding:0;
  }
  body {
    width: 100vw;
    height: 100vh;
    margin: 0;
    padding: 0;
    font-family: 'Khula', sans-serif;
  }
  #switch-satellite{
    position: absolute;
    top:20px;
    right:90px;
    z-index: 1;
    background: #fff000;
    padding:8px 15px;
    margin-top:10px;
  }
  #switch-light{
    position: absolute;
    top:20px;
    right:20px;
    z-index: 1;
    background: #fff000;
    padding:8px 15px;
    margin-top:10px;
  }
  </style>
</head>
<body>
  <div id="switch-light">light</div>
  <div id="switch-satellite">satellite</div>

  <script src='https://api.tiles.mapbox.com/mapbox-gl-js/v0.53.1/mapbox-gl.js'></script>
  <script src='https://unpkg.com/deck.gl@~7.0.0/dist.min.js'></script>
  <script>
[...]

PR #4886, opened 2020-08-23
Support changing mapStyle in standalone bundle

Pessimistress:
For #4809

#### Change List
- Support changing mapStyle in standalone bundle

### DIFF
--- a/modules/core/bundle/deckgl.js
+++ b/modules/core/bundle/deckgl.js
@@ -114,6 +114,9 @@ export default class DeckGL extends Deck {
       this.onBeforeRender = props.onBeforeRender;
       props.onBeforeRender = this._onBeforeRender;
     }
+    if ('mapStyle' in props && this._map) {
+      this._map._map.setStyle(props.mapStyle);
+    }
 
     super.setProps(props);
   }


kylebarron:
> This feature has nothing to do with react

I understand, but wondered if it had similar limitations.

> Not true:
> http://visgl.github.io/react-map-gl/docs/api-reference/static-map#mapstyle

I see that written in the docs, but I've never gotten it to work without using `Immutable.Map`. For example, [here](https://github.com/kylebarron/naip-cogeo-mosaic/blob/a203b2a751b409cccb0d69cbae9297afffe997ac/site/src/App.js#L38-L60) is a simple example. Based on user input it modifies one of the `sources` values. When I use a JS object, I can see in the debugger the value of `mapStyle` that is passed to React Map GL changes, but the map never updates.

Pessimistress:
The React component cannot detect nested change in an object. Style is only considered changed if it is provided with a different object. The advantage of using Immutable is that the container shallowly changes if you change any nested property.

This conversation belongs in the react-map-gl repo.

Pessimistress:
> wondered if it had similar limitations.

No.

The React component can rerender every animation frame. It has to compare old and new prop values so that it doesn't invoke this expensive operation unless the style has truly changed.

This standalone API calls mapbox's `setStyle` immediately. It is the user's responsibility to only call it when needed.

PR #4894, opened 2020-08-26
Pass layer object to getTileData function

### LINKED ISSUE #4893: Ability to pass additional context to `getTileData`
#### Target Use case

Sometimes the implementation of `getTileData` changes, but in a way that only affects getting additional tiles. The current way to inform `TilLayer` that your `getTileData` changed is to pass:

```
new TileLayer({
  getTileData(tile) {
    // do something with tile and somethingYourGetTileDataFuncIsDependentOn
  },
  updateTriggers: {
    getTileData: somethingYourGetTileDataFuncIsDependentOn
  }
})
```

However, when using this approach, `TileLayer` will re-request all the tiles.

#### Proposed feature

Passing the current layer instance to `getTileData` provides a way to give different props to `TileLayer` and have those props passed to `getTileData` for subsequent tile requests without needing to re-request the already-loaded tiles.

With this approach you can accomplish the same like this:

```
new TileLayer({
  getTileData(tile, layer) {
    // do something with tile and layer.props.somethingYourGetTileDataFuncIsDependentOn
  },
  somethingYourGetTileDataFuncIsDependentOn,
})
```

#### To Do List 
  - [ ] Coding
  - [ ] Doc update
  - [ ] What’s new update 
  - [ ] Test

supersonicclay:
For #4893

#### Change List
-

Fixes #4893

### DIFF
--- a/docs/api-reference/geo-layers/tile-layer.md
+++ b/docs/api-reference/geo-layers/tile-layer.md
@@ -107,14 +107,18 @@ If the value is an array: multiple URL templates. Each endpoint must return the 
 
 - Default: `tile => load(tile.url)`
 
-If supplied, `getTileData` is called to retrieve the data of each tile. It receives one argument `tile` which contains the following fields:
+If supplied, `getTileData` is called to retrieve the data of each tile. It receives two arguments
 
-- `x` (Number) - x index of the tile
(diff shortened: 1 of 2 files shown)


supersonicclay:
Going with #4901 instead

Issue #4895, opened 2020-08-26
getCursor support for ArcGIS (2D/DeckLayer and 3D/DeckRenderer)


damix911:
#### Description
**I am still debating if this is a valid issue or we should pursue another way of obtaining the same result**.

When using ArcGIS maps, it should be possible to specify a cursor by doing (for 2D):

    const layer = new DeckLayer({
      "deck.getCursor": () => "ne-resize"
    });

Or, for 3D:

    const renderer = new DeckRenderer(sceneView, {
      "getCursor": () => "nw-resize"
    });

But, currently, these have no effect.

#### Repro Steps
[This pen](https://codepen.io/dawken/pen/QWNvxPy?editors=1010) unsuccessfully tries to override the cursor.

#### Environment (please complete the following information):
 - Framework Version: can repro with the latest `master` branch.

#### Logs
N/A.

damix911:
A fix was installed with PR #4895.

Pessimistress:
8.2.7

PR #4896, opened 2020-08-26
Possible solution to the ignored getCursor

damix911:
Tentative fix for #4895 (I am still evaluating other strategies).

#### Change List
- Add a `getCursor` descriptor in `deck-props.js` so that a `deck.getCursor` property is recognized and copied when specified for a `DeckLayer` (2D, `MapView`) or a `DeckRenderer` (3D, `SceneView`).


Pessimistress:
Should this be mentioned in the documentation?

damix911:
@Pessimistress Yes, good point. I added it for the [DeckLayer constructor](https://deck.gl/docs/api-reference/arcgis/deck-layer#constructor-1) and the [DeckRenderer constructor](https://deck.gl/docs/api-reference/arcgis/deck-renderer#constructor-1).

PR #4901, opened 2020-08-27
Get current layer instance for TileLayer

### LINKED ISSUE #4893: Ability to pass additional context to `getTileData`

supersonicclay:
For #4893

#### Change List
-

Fixes #4893

### DIFF
--- a/modules/geo-layers/src/tile-layer/tile-layer.js
+++ b/modules/geo-layers/src/tile-layer/tile-layer.js
@@ -143,8 +143,8 @@ export default class TileLayer extends CompositeLayer {
   // Methods for subclass to override
 
   getTileData(tile) {
-    const {props} = this.getCurrentLayer();
-    const {getTileData, fetch, data} = props;
+    const {fetch, data} = this.props;
+    const {getTileData} = this.getCurrentLayer().props;
     const {signal} = tile;
 
     tile.url = getURLFromTemplate(data, tile);
--- a/modules/geo-layers/src/tile-layer/tile-layer.js
+++ b/modules/geo-layers/src/tile-layer/tile-layer.js
@@ -143,7 +143,7 @@ export default class TileLayer extends CompositeLayer {
   // Methods for subclass to override
 
   getTileData(tile) {
-    const {getTileData, fetch, data} = this.props;
+    const {getTileData, fetch, data} = this.getCurrentLayer().props;
     const {signal} = tile;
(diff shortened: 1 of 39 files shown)


Pessimistress:
Theoretically, when props like `data` and `elevationDecoder` change, all tiles should be invalidated and reload.

Let's limit this PR to using the current `getTileData`.

supersonicclay:
Done @Pessimistress

kylebarron:
So instead of passing the `layer` instance, this just uses the "current" value of `getTileData`? Would be good to add a line in What's New

supersonicclay:
Done @kylebarron

supersonicclay:
@kylebarron @Pessimistress is there anything else you need with this or can we get it merged? Thanks!

modules/geo-layers/src/tile-layer/tile-layer.js
```diff
@@ -143,7 +143,8 @@ export default class TileLayer extends CompositeLayer {
   // Methods for subclass to override
 
   getTileData(tile) {
-    const {getTileData, fetch, data} = this.props;
+    const {fetch, data} = this.props;
```
Pessimistress, line 146: `fetch` should be taken from the current props too

supersonicclay:
Done @Pessimistress

PR #4905, opened 2020-08-30
pydeck: Additional formatting updates

ajduberstein:
For #3640

This is from running `make pre-commit` within the pydeck folder. I think I'd prefer a shorter line length.

#### Change List
-


kylebarron:
> I think I'd prefer a shorter line length.

In `pyproject.toml` you have
```
line-length = 120
```

Are you suggesting decreasing it? (I personally always use line length 80)

Issue #4948, opened 2020-09-16
deck.gl should not resize the canvas if it does not own it


damix911:
#### Description

This bug is related to the experimental property `deck.props._framebuffer`.

The `deck.props._framebuffer` property is an experimental extension point to integrate with 3rd party APIs.

When specified, deck.gl is supposed to draw in that framebuffer instead of using it's own context.

This mechanism is currently used by the `@deck.gl/arcgis` module to integrate deck.gl and Esri's ArcGIS API for JavaScript.

**The problem is that deck.gl will try, under certain circumstances, to resize the canvas even when it is intended to be managed by the 3rd party API and this can lead to all sorts of bugs.**

#### Repro Steps
The repro app is far from minimal, sorry; it comes from a bug report originally submitted to Esri. It showcases a weird use of CSS to pad the map.

- Open [pen GRZBbEb](https://codepen.io/dawken/pen/GRZBbEb?editors=1000) and notice how the deck.gl layer and the Esri basemap are not aligned.

#### Explanation
- This happens because that app does something (IMHO) weird; setting a CSS `padding` on the Esri map element;
- Although the app is (I think) misusing the ArcGIS API for JavaScript, it helped uncover the issue;
- Integration between deck.gl and ArcGIS is implemented through WebGL canvas (and context) sharing; the canvas (and context) is owned by the ArcGIS map;
- The ArcGIS app creates an offscreen surface and passes it as the experimental property `props._framebuffer` to deck.gl; deck.gl should render in that framebuffer;
- However, under certain circumstances deck.gl will try to resize the canvas and mess up things.

#### Environment (please complete the following information):
 - Framework Version: deck.gl 8.2.8 and probably many releases before
 - Browser Version: all browsers
 - OS: all OSes

#### Logs
Nothing to report.

PR #4949, opened 2020-09-16
Avoid canvas resizing when not owned by deck.gl (i.e. when `props._framebuffer` is set)

damix911:
This fixes an issue that can arise when deck.gl is sharing the WebGL context and the `deck.props._framebuffer` property is set.

deck.gl will try, under certain circumstances, to resize the canvas even when it is intended to be managed externally and this can lead to all sorts of bugs.

For #4948.

#### Change List
- Don't resize the canvas if an external framebuffer was specified


damix911:
Feel free to advise; I think more work is needed here.

Pessimistress:
Resizing can be disabled with `width: null, height: null`.

damix911:
@Pessimistress awesome, thanks! I made a change. Does it make sense?

Pessimistress (approved):

Issue #4952, opened 2020-09-17
RFC: Viewport features


padawannn:
#### Target Use case
Allow an easy way to get viewport features from MVTLayer

#### Proposed feature
Add an onViewportFeatures function callback in MVTLayer that returns the viewport features

#### To Do List 
  - [ ] Add label and assign to milestone 
  - [ ] Coding
  - [ ] Doc update
  - [ ] What’s new update 
  - [ ] Test

padawannn:
And this is a bit more elaborated RFC

# RFC: Viewport features

# RFC: onViewportFeatures function callback in MVTLayer

- **Author**: Javier Aragón (CARTO)
- **Date**: September, 2020
- **Status**: **Draft**

References

- Filter features within map view: [https://docs.mapbox.com/mapbox-gl-js/example/filter-features-within-map-view/](https://docs.mapbox.com/mapbox-gl-js/example/filter-features-within-map-view/)
- Carto VL Viewport aggregations: [https://carto.com/developers/carto-vl/v0.10.0/guides/aggregations/#viewport-aggregations](https://carto.com/developers/carto-vl/v0.10.0/guides/aggregations/#viewport-aggregations)
- Arcgis: Query features from a FeatureLayerView: [https://developers.arcgis.com/javascript/latest/sample-code/featurelayerview-query/index.html](https://developers.arcgis.com/javascript/latest/sample-code/featurelayerview-query/index.html)

### Summary

This RFC proposes the creation of a new callback function in the MVTLayer that returns the features that are displayed on the map.

### Motivation

Sometimes you want to extract relevant information from the current view of the map (the viewport), to better understand *What am I seeing in the current map view?* Particularly in the MVTLayer where you don't know the data unless you make a call to the backend.

Currently, the MVTLayer has the onViewportLoad callback that returns all visible tiles features when all tiles in the current viewport are loaded, but this function is not enough to filter viewport features for the following reasons:

- You can drag or zoom the map and new tiles are not loaded so the onViewportLoad would not be called.
- Get viewport features outside the MVTLayer code is not trivial because you need to wait for the user interaction has finished (dragend, zoomend, etc) and be sure that all tiles have been loaded. Also, the code to filter the tiles is not easy.

The function that calculates the viewport features would be called inside the updateState function (like _updateTileset in tile-layer) but it would wait for user interaction has finished through a debounce function (similar to the one used in the heatmap-layer) and it would also wait to all tiles have been loaded.

[...]

PR #4953, opened 2020-09-17
onViewportFeatures function callback in MVTLayer

padawannn:
Related with https://github.com/visgl/deck.gl/issues/4952

I still need to add the tests, but I’d really appreciate some feedback from your side.

We do a double transformation of coordinates when this feature is enabled: one for rendering and we add another one to compare against the frustum panels. We’ll need some guidance to improve that.

Any feedback is more than welcome!

### DIFF
--- a/docs/api-reference/geo-layers/mvt-layer.md
+++ b/docs/api-reference/geo-layers/mvt-layer.md
@@ -111,13 +111,6 @@ Optional. When provided, a feature with ID corresponding to the supplied value w
 If `uniqueIdProperty` is provided, value within that feature property will be used for ID comparison. If not, [feature id](https://github.com/mapbox/vector-tile-spec/tree/master/2.1#42-features) will be used.
 
 
-
-##### `maxFeatures` (Number )
-
-* Default: 10000
-
-Optional. Max number of features to retrieve when getRenderedFeatures is called.
-
 ### Callbacks
 
 ##### `onDataLoad` (Function, optional)
@@ -128,7 +121,6 @@ Receives arguments:
 
 tileJSON (Object) - the tileJSON fetched
 
-
 ##### `onViewportChange` (Function, optional)
 
 `onViewportChange` called when the viewport changes or all tiles in the current viewport have been loaded. A function (`getRenderedFeatures`) to calculate the features rendered in the current viewport is passed as a JSON to this callback function.
@@ -136,22 +128,22 @@ tileJSON (Object) - the tileJSON fetched
 
 Receives arguments:
 
-`getRenderedFeatures` (Function)
+* `getRenderedFeatures` (Function)
 
-Parameters:
+  + format: Optional. Posible values are `json` or `geojson` (geometries are returned in [GeoJSON](https://tools.ietf.org/html/rfc7946)). Default to `json`. 
 
-* format: Optional. Posible values are `json` or `geojson` (geometries are returned in [GeoJSON](https://tools.ietf.org/html/rfc7946)). Default to `json`. 
+  + maxFeatures: Optional. Max number of features to retrieve when getRenderedFeatures is called. Default to `null`.
 
-Requires `pickable` to be true.
+  Requires `pickable` to be true.
 
-It is not recommended to call `getRenderedFeatures` every time `onViewportChange` is executed, instead, use a debounce function.
+  It is not recommended to call `getRenderedFeatures` every time `onViewportChange` is executed, instead, use a debounce function.
 
-If a `uniqueIdProperty` is provided only unique properties are returned.
+  If a `uniqueIdProperty` is provided only unique properties are returned.
 
 ```javascript
 const onViewportChange =  e => {
   const features = e.getRenderedFeatures();
-  // ...
+  // ... or  e.getRenderedFeatures({format: 'geojson', maxFeatures: 1000});
 };
 
 new MVTLayer({
--- a/docs/api-reference/geo-layers/mvt-layer.md
+++ b/docs/api-reference/geo-layers/mvt-layer.md
@@ -110,6 +110,16 @@ Optional. When provided, a feature with ID corresponding to the supplied value w
 
 If `uniqueIdProperty` is provided, value within that feature property will be used for ID comparison. If not, [feature id](https://github.com/mapbox/vector-tile-spec/tree/master/2.1#42-features) will be used.
 
+
+
+##### `maxFeatures` (Number )
+
+* Default: 10000
+
+Optional. Max number of features to retrieve when getRenderedFeatures is called.
+
+### Callbacks
+
 ##### `onDataLoad` (Function, optional)
 
 `onDataLoad` called when a tileJSON is successfully fetched
@@ -117,6 +127,41 @@ If `uniqueIdProperty` is provided, value within that feature property will be us
 Receives arguments:
 
 tileJSON (Object) - the tileJSON fetched
+
+
+##### `onViewportChange` (Function, optional)
+
+`onViewportChange` called when the viewport changes or all tiles in the current viewport have been loaded. A function (`getRenderedFeatures`) to calculate the features rendered in the current viewport is passed as a JSON to this callback function.
+
+
+Receives arguments:
+
+`getRenderedFeatures` (Function)
+
+Parameters:
+
+* format: Optional. Posible values are `json` or `geojson` (geometries are returned in [GeoJSON](https://tools.ietf.org/html/rfc7946)). Default to `json`. 
+
+Requires `pickable` to be true.
+
+It is not recommended to call `getRenderedFeatures` every time `onViewportChange` is executed, instead, use a debounce function.
+
+If a `uniqueIdProperty` is provided only unique properties are returned.
+
+```javascript
+const onViewportChange =  e => {
+  const features = e.getRenderedFeatures();
+  // ...
+};
+
+new MVTLayer({
+  id: "..."
+  data: "..."
+  pickable: true
+  onViewportChange: debounce(onViewportChange, 500)
+  uniqueIdProperty: 'geoid'
+})
+```
 
 ## Source
 
--- a/docs/api-reference/geo-layers/mvt-layer.md
+++ b/docs/api-reference/geo-layers/mvt-layer.md
@@ -130,8 +130,6 @@ Receives arguments:
 
 * `getRenderedFeatures` (Function)
 
-  + format: Optional. Posible values are `json` or `geojson` (geometries are returned in [GeoJSON](https://tools.ietf.org/html/rfc7946)). Default to `json`. 
-
   + maxFeatures: Optional. Max number of features to retrieve when getRenderedFeatures is called. Default to `null`.
 
   Requires `pickable` to be true.
@@ -140,10 +138,11 @@ Receives arguments:
 
   If a `uniqueIdProperty` is provided only unique properties are returned.
 
+* `viewport` (Object). A instance of the current [`Viewport`](/docs/api-reference/core/viewport.md).
+
 ```javascript
 const onViewportChange =  e => {
   const features = e.getRenderedFeatures();
-  // ... or  e.getRenderedFeatures({format: 'geojson', maxFeatures: 1000});
 };
 
 new MVTLayer({
--- a/docs/api-reference/geo-layers/tile-layer.md
+++ b/docs/api-reference/geo-layers/tile-layer.md
@@ -267,34 +267,6 @@ Receives arguments:
 
 - `tile` (Object) - the tile that has been cleared from cache.
 
-
-##### `onViewportChange` (Function, optional)
-
-`onViewportChange` is a function that is called when the viewport changes or all tiles in the current viewport have been loaded. A function (`getRenderedFeatures`) to calculate the features rendered in the current viewport is passed as a JSON to this callback function.
-
-`getRenderedFeatures` only works if pickable property is true.
-
-It is not recommended to call `getRenderedFeatures` every time `onViewportChange` is executed, instead, better use `getRenderedFeatures` inside a debounce function.
-
-```
-let timeout;
-function yourDebouncedFunction(func) {
-  clearTimeout(timeout)
-  timeout = setTimeout(() => {
-    const features = func.apply();
-    ...
-  }, 500);
-};
-
-new TileLayer({
-  ...
-  pickable: true,
-  onViewportLoad: (evt) => yourDebouncedFunction(evt.getRenderedFeatures)
-})
-```
-
-- Default: `data => null`
-
 ## Source
 
 [modules/geo-layers/src/tile-layer](https://github.com/visgl/deck.gl/tree/master/modules/geo-layers/src/tile-layer)
--- a/docs/api-reference/geo-layers/tile-layer.md
+++ b/docs/api-reference/geo-layers/tile-layer.md
@@ -267,6 +267,34 @@ Receives arguments:
 
 - `tile` (Object) - the tile that has been cleared from cache.
 
+
+##### `onViewportChange` (Function, optional)
+
+`onViewportChange` is a function that is called when the viewport changes or all tiles in the current viewport have been loaded. A function (`getRenderedFeatures`) to calculate the features rendered in the current viewport is passed as a JSON to this callback function.
+
+`getRenderedFeatures` only works if pickable property is true.
+
+It is not recommended to call `getRenderedFeatures` every time `onViewportChange` is executed, instead, better use `getRenderedFeatures` inside a debounce function.
+
+```
+let timeout;
+function yourDebouncedFunction(func) {
+  clearTimeout(timeout)
+  timeout = setTimeout(() => {
+    const features = func.apply();
+    ...
+  }, 500);
+};
+
+new TileLayer({
+  ...
+  pickable: true,
+  onViewportLoad: (evt) => yourDebouncedFunction(evt.getRenderedFeatures)
+})
+```
+
+- Default: `data => null`
+
 ## Source
 
 [modules/geo-layers/src/tile-layer](https://github.com/visgl/deck.gl/tree/master/modules/geo-layers/src/tile-layer)
--- a/modules/core/src/lib/deck-picker.js
+++ b/modules/core/src/lib/deck-picker.js
@@ -281,7 +281,7 @@ export default class DeckPicker {
     width = 1,
     height = 1,
     mode = 'query',
-    depth = Number.MAX_SAFE_INTEGER,
+    depth = null,
     onViewportActive
   }) {
     layers = this._getPickable(layers);
@@ -328,7 +328,13 @@ export default class DeckPicker {
     // Only return unique infos, identified by info.object
     const uniqueInfos = new Map();
 
+    const isDepth = Number.isFinite(depth);
+
     for (let i = 0; i < pickInfos.length; i++) {
+      if (isDepth && uniqueInfos.size >= depth) {
+        break;
+      }
+
       const pickInfo = pickInfos[i];
       let info = {
         color: pickInfo.pickedColor,
@@ -345,10 +351,6 @@ export default class DeckPicker {
       info = getLayerPickingInfo({layer: pickInfo.pickedLayer, info, mode});
       if (!uniqueInfos.has(info.object)) {
         uniqueInfos.set(info.object, info);
-      }
-
-      if (uniqueInfos.size >= depth) {
-        break;
       }
     }
 
--- a/modules/core/src/lib/deck-picker.js
+++ b/modules/core/src/lib/deck-picker.js
@@ -281,7 +281,7 @@ export default class DeckPicker {
     width = 1,
     height = 1,
     mode = 'query',
-    depth = null,
+    maxFeatures = null,
     onViewportActive
   }) {
     layers = this._getPickable(layers);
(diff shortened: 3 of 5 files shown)


modules/geo-layers/src/mvt-layer/mvt-layer.js
```diff
@@ -84,6 +80,110 @@ export default class MVTLayer extends TileLayer {
...
+    const currentFrustumPlanes = this.context.viewport.getFrustumPlanes();
+    let viewportFeatures = [];
+
+    await tileset.selectedTiles.forEach(async tile => {
+      const features = (await tile.data) || [];
+      const transformationMatrix = new Matrix4()
+        .translate(this._getCoordinateOrigin(tile))
+        .scale(this._getModelMatrixScale(tile));
+
+      viewportFeatures = viewportFeatures.concat(
```
ibgreen, line 138: Nits:
- concat tends to be fairly inefficient, compares to array.push - it copies the array again and again. Lots of memory / GC pressure and usually lower perf.
- I'd just do `for (const tile of ...) { for (const feature of ... ) if (...) features.push() }}`
- I'd allow the app to provide a `props.maxFeatures`, and stop concatenating after that number (depends on use case, but what is the point of showing 100K item lists in the UI?).
- Alternatively, it could be interesting to add a benchmark test.

ibgreen (commented):
@padawannn Thanks for this PR, I think this is a very valuable feature!

I started off by adding some comments inline. 

Also as an alternative to doing picking using math, have you considered [`Deck.pickObjects`](https://deck.gl/docs/api-reference/core/deck#pickobjects) - it returns a list of indices to all visible objects. It is performed on the GPU. 

Not sure if there are any issues using it with tiled layers, but that approach could be extended to work with any layer.

modules/geo-layers/src/mvt-layer/mvt-layer.js
```diff
@@ -84,6 +80,110 @@ export default class MVTLayer extends TileLayer {
...
+
+  async _getViewportFeatures() {
+    const {tileset} = this.state;
+    const {onViewportFeatures} = this.props;
+    const featureCache = new Set();
+    const currentFrustumPlanes = this.context.viewport.getFrustumPlanes();
+    let viewportFeatures = [];
+
+    await tileset.selectedTiles.forEach(async tile => {
+      const features = (await tile.data) || [];
```
Pessimistress, line 133: You can use `tile.content` here to avoid using async

modules/geo-layers/src/mvt-layer/mvt-layer.js
```diff
@@ -84,6 +80,110 @@ export default class MVTLayer extends TileLayer {
...
+
+    onViewportFeatures(viewportFeatures);
+  }
+
+  _checkIfCoordinatesIsInsideFrustum(matrix, frustumPlanes, coordinates) {
+    if (Array.isArray(coordinates) && coordinates.length && typeof coordinates[0] === 'number') {
+      return this._coordinateInPlanes(frustumPlanes, matrix.transform(coordinates).concat(0));
+    }
+
+    return coordinates.some(c => {
```
Pessimistress, line 165: This is technically incorrect: a path can both start and end outside the viewport but still visible.

modules/geo-layers/src/mvt-layer/mvt-layer.js
padawannn: Thank you @ibgreen,

I didn't know about Deck.pickObjects. I've done a quick test and apparently it works fine with tiled layers so of course is a better solution than using math

> I'd allow the app to provide a props.maxFeatures, and stop concatenating after that number (depends on use case, but what is the point of showing 100K item lists in the UI?).

When a user wants to know the viewport features it's not always to show them in the UI. For example, you might want to display a histogram based on a property, calculate an average, max or min value, etc. I think it might be a good idea to return all features and have the user decide what to do with them.

Before modifying this PR I think it would be good to decide the design of the API, @Pessimistress talk about this in the RFC

modules/geo-layers/src/mvt-layer/mvt-layer.js
padawannn: Yes, you are right :(
As @ibgreen says, we can remove this code and use Deck.pickObjects instead, right?

modules/geo-layers/src/mvt-layer/mvt-layer.js
ibgreen: You could still add add a maxFeatures prop and support a (default) value of null or 0 meaning no limit

ibgreen: You could still add add a maxFeatures prop and support a (default) value of null or 0 meaning no limit

modules/geo-layers/src/mvt-layer/mvt-layer.js
ibgreen: Pickobjects isn’t perfect either - for something to be picked it needs to cover at least 1 pixel, so there is a chance some object won’t be picked if overlapped etc. but this is very rare and in practice it works really well and performance makes it well worth it.

But if you need 100% guarantee the number is correct you may need to count.

padawannn:
I have updated the PR taking into account your comments here and in the [RFC](https://github.com/visgl/deck.gl/issues/4952)

- Use Deck.pickObjects instead of math to get features.
- New API design proposed by @Pessimistress https://github.com/visgl/deck.gl/issues/4952#issuecomment-694419157 including onViewportChange callback
- I've added maxFeatures properties.
- Rename methods (getRenderedFeatures instead of getViewportFeatures).

Some considerations:
- getRenderedFeatures only works if pickable property is true. What do you think about setting this property true by default. Honestly, I don't know the impact of this.
- With the new API design we've moved the control of the debounce to the application side. I think this is not very intuitive and it would be good to provide an example. if you want to include it in any of the example folders let me know.

Pessimistress:
> What do you think about setting this property true by default in tile layer

We can't do that. There is a `layer.isPickable` method that you can override to check an internal flag.

> example

The example code can live in the documentation for the new callback.

modules/geo-layers/src/tile-layer/tile-layer.js
```diff
@@ -214,6 +225,22 @@ export default class TileLayer extends CompositeLayer {
       return tile.layers;
     });
   }
+
+  getPickObjects() {
+    const {deck} = this.context;
+    const width = deck.canvas.width;
```
Pessimistress, line 231: Use viewport.x, viewport.width etc.

modules/geo-layers/src/tile-layer/tile-layer.js
```diff
@@ -214,6 +225,22 @@ export default class TileLayer extends CompositeLayer {
...
+    const layerIds = [this.id];
+    const x = 0;
+    const y = 0;
+    return deck.pickObjects({x, y, width, height, layerIds});
+  }
+
+  getRenderedFeatures() {
+    const {maxFeatures} = this.props;
+    const features = this.getPickObjects();
+    return maxFeatures ? features.splice(0, maxFeatures) : features;
```
Pessimistress, line 242: The maxFeatures prop here is not working as a stop gap for large feature sets. You are still spending all the CPU cycles building the array before truncating it.

modules/geo-layers/src/tile-layer/tile-layer.js
```diff
@@ -214,6 +225,22 @@ export default class TileLayer extends CompositeLayer {
       return tile.layers;
     });
   }
+
+  getPickObjects() {
```
Pessimistress, line 229: Nit: Add underscore before private method names

Call it `pickObjects`?

padawannn: Done

modules/geo-layers/src/tile-layer/tile-layer.js
padawannn: Done

modules/geo-layers/src/tile-layer/tile-layer.js
padawannn: Yes... I thought about removing this property from the tile-layer and keeping it in the mvt-layer but I found it strange from the user's point of view. What do you think? remove it from tile-layer and keep it on in the mvt-layer?

modules/geo-layers/src/tile-layer/tile-layer.js
```diff
@@ -96,6 +99,12 @@ export default class TileLayer extends CompositeLayer {
       this.state.tileset.tiles.forEach(tile => {
         tile.layers = null;
       });
+    } else if (changeFlags.viewportChanged) {
+      if (tileset.isLoaded && onViewportChange) {
+        onViewportChange(tileset.selectedTiles.map(tile => tile.data), {
```
Pessimistress, line 104: - Add to documentation
- Drop the first argument for perf & usability. This is a new api and it does not have to be aligned with anything else.

modules/geo-layers/src/tile-layer/tile-layer.js
```diff
@@ -111,7 +120,9 @@ export default class TileLayer extends CompositeLayer {
     const tilesetChanged = this.state.frameNumber !== frameNumber;
 
     if (isLoaded && onViewportLoad && (loadingStateChanged || tilesetChanged)) {
-      onViewportLoad(tileset.selectedTiles.map(tile => tile.data));
+      onViewportLoad(tileset.selectedTiles.map(tile => tile.data), {
```
Pessimistress, line 123: Don't change this callback if it's unnecessary

padawannn: Ok, maybe I didn't understand your last comment in the RFC. The conclusion is:
- `onViewportLoad` is not going to change.
- The new `onViewportChange` call back is the only callback that is going to return the `getRenderedFeatures` function.

If that is correct there are two situations in which onViewportChange should be called:
- When the viewport changes and all tiles have been loaded, for example when we drag the map a little and it doesn't loads new tiles. We can control it in `updateState` function checking `if (changeFlags.viewportChanged && tileset.isLoaded)`
- Every time a new set of tiles have finished loading for the current viewport, for example, when zoom ends `changeFlags.viewportChanged` is true but tiles are still loading so `onViewportChange` should be called when tiles have finished loading. It's being controlled in `_updateTileset` with the condition `if (isLoaded && (loadingStateChanged || tilesetChanged))`

In conclusion, in the new commit I'm calling the `onViewportChange` callback in these two situations.

Sorry for the long comment.

modules/geo-layers/src/tile-layer/tile-layer.js
padawannn: Done

modules/geo-layers/src/tile-layer/tile-layer.js
ibgreen: Should it be called when the view changes (without triggering tile reloads)?

Pessimistress: > there are two situations in which onViewportChange should be called

That sounds right to me. If the viewport changes but not all tiles have been loaded, the callback is expected to trigger, and it's just delayed.

> Should it be called when the view changes?

I'd say yes, given that it affects the result of `getRenderedFeatures`

padawannn: Yes because if for example, you drag the map a little or your current zoom is 8 and you change it to zoom 8.4 it will not trigger a tile reloads but the features that you are seeing in the new viewport have probably changed. 

I'm trying to upload a video to show this but Github is returning an error. You can download it from [here](https://drive.google.com/file/d/1Vfts4PJyWqb0RfMDaW4weXzkveGvDaHu/view?usp=sharing) 

As you can see, in the browser tab new tiles are not downloaded and this condition `isLoaded && (loadingStateChanged || tilesetChanged)` is not satisfied.

docs/api-reference/geo-layers/tile-layer.md
```diff
@@ -261,6 +261,34 @@ Receives arguments:
...
 - `tile` (Object) - the tile that has been cleared from cache.
 
+
+##### `onViewportChange` (Function, optional)
+
+`onViewportChange` is a function that is called when the viewport changes or all tiles in the current viewport have been loaded. A function (`getRenderedFeatures`) to calculate the features rendered in the current viewport is passed as a JSON to this callback function.
+
+`getRenderedFeatures` only works if pickable property is true.
+
+It is not recommended to call `getRenderedFeatures` every time `onViewportChange` is executed, instead, better use `getRenderedFeatures` inside a debounce function.
```
Pessimistress, line 271: Based on your implementation, I do not think `getRenderedFeatures` works if called immediately. `onViewportChange` is triggered before `renderLayers`, therefore will not render the correct sub layers to the picking buffer.

modules/geo-layers/src/tile-layer/tile-layer.js
```diff
@@ -96,6 +98,8 @@ export default class TileLayer extends CompositeLayer {
       this.state.tileset.tiles.forEach(tile => {
         tile.layers = null;
       });
+    } else if (changeFlags.viewportChanged && tileset.isLoaded) {
```
Pessimistress, line 101: This does not look right. `tileset.isLoaded` is updated by `this._updateTileset()`. Before then it's only reflecting the last viewport state.

padawannn: For example, while you are zooming in on the map `_updateTileset` set `tileset.isLoaded` to false because new tiles have started to download. This condition prevent that  onViewportChange be called if you drag or zoom map while tiles are downloading

Pessimistress: Correct, except `_updateTileset` has not been called yet when you check this flag.

padawannn: mmm, we could change the order.

```
this._updateTileset();
if (changeFlags.viewportChanged && tileset.isLoaded) {
 this._onViewportChange();
}
```

but when you start to dag the map `tileset.isLoaded` is still true it will continue to be true until you've daged enough to ask for new tiles. Honestly, I can't think of how to do it without a debounce function

padawannn: In any case, I'll think about it a little more tomorrow

alasarr:
Implemented all the pending issues (except tests): 
- Improve limit on Max Features adding depth to _pickVisibleObjects in `deck.picker`
- Move all functionality to MVTLayer. 
- Complete documentation
- Call to onViewportChange when tileset is ready.

modules/geo-layers/src/tile-layer/tile-layer.js
```diff
@@ -109,8 +110,11 @@ export default class TileLayer extends CompositeLayer {
     const loadingStateChanged = this.state.isLoaded !== isLoaded;
     const tilesetChanged = this.state.frameNumber !== frameNumber;
 
-    if (isLoaded && onViewportLoad && (loadingStateChanged || tilesetChanged)) {
-      onViewportLoad(tileset.selectedTiles.map(tile => tile.data));
+    if (isLoaded && (loadingStateChanged || tilesetChanged)) {
+      if (onViewportLoad) {
+        onViewportLoad(tileset.selectedTiles.map(tile => tile.data));
+      }
+      this._onViewportChange();
```
Pessimistress, line 117: This semantically doesn't make sense. You can call `this._onViewportLoad` here and overwrite it in the subclass to invoke `onViewportChange`. `_onViewportChange` as a base class interface will lead to future confusions because no actual viewport change is happening here.

docs/api-reference/geo-layers/mvt-layer.md
```diff
@@ -118,6 +128,41 @@ Receives arguments:
...
 
+
+##### `onViewportChange` (Function, optional)
+
+`onViewportChange` called when the viewport changes or all tiles in the current viewport have been loaded. A function (`getRenderedFeatures`) to calculate the features rendered in the current viewport is passed as a JSON to this callback function.
+
+
+Receives arguments:
+
+`getRenderedFeatures` (Function)
```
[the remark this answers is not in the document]
alasarr: Nope. It returns the rendered features at the current viewport

```javascript
const onViewportChange = e => {
  const features = e.getRenderedFeatures();
};

<MVTLayer
        id="test"
       ...
        onViewportChange={debounce(onViewportChange, 400)}
        maxFeatures={1000}
        uniqueIdProperty={'cartodb_id'}
      />
    </DeckGL>
```

modules/core/src/lib/deck-picker.js
```diff
@@ -281,6 +281,7 @@ export default class DeckPicker {
     width = 1,
     height = 1,
     mode = 'query',
+    depth = Number.MAX_SAFE_INTEGER,
```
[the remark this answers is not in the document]
alasarr: Yes, for [_pickClosestObject](https://github.com/visgl/deck.gl/blob/76482e6b05b851897fb45490fad0aefb88bba71c/modules/core/src/lib/deck-picker.js#L146), so I tried to do the same for [_pickVisibleObjects](https://github.com/visgl/deck.gl/blob/76482e6b05b851897fb45490fad0aefb88bba71c/modules/core/src/lib/deck-picker.js#L275).

Those methods are called from [pickObject and pickObjects](https://github.com/visgl/deck.gl/blob/76482e6b05b851897fb45490fad0aefb88bba71c/modules/core/src/lib/deck-picker.js#L76).

Do you see some inconsistency in that?

docs/api-reference/geo-layers/mvt-layer.md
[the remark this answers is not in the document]
Pessimistress: I'm talking about `onViewportChange` - if I subscribe to an event with this name, I expect to be able to access `viewport` in my callback.

This comes back to the way the documentation for this prop is written. Without proper hierarchy we can't talk about "arguments" without confusion.

modules/core/src/lib/deck-picker.js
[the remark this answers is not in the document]
Pessimistress: `depth` in `pickObjects` refer to how deep the picking looks into overlapped features. `pickMultipleObjects` does not handle overlapping at all.

docs/api-reference/geo-layers/mvt-layer.md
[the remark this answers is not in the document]
alasarr: I see now your point, it's a good idea...Let's do it

alasarr:
All requested changes implemented

modules/core/src/lib/deck-picker.js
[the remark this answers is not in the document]
alasarr: Ok, I'll change then to maxFeatures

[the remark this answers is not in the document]
alasarr: Done

modules/core/src/lib/deck-picker.js
```diff
@@ -281,7 +281,7 @@ export default class DeckPicker {
     width = 1,
     height = 1,
     mode = 'query',
-    depth = null,
+    maxFeatures = null,
```
Pessimistress, line 284: Can you name it `maxObjects`? `Feature` is a concept unique to GeoJSON.

modules/geo-layers/src/mvt-layer/mvt-layer.js
```diff
@@ -165,6 +171,64 @@ export default class MVTLayer extends TileLayer {
...
+    for (const f of features) {
+      const featureId = getFeatureUniqueId(f.object, this.props.uniqueIdProperty);
+
+      if (featureId === -1) {
+        // we have no id for the feature, we just add to the list
+        renderedFeatures.push(getFeatureByFormat(f.object, format));
+      } else if (!featureCache.has(featureId) && featureId !== -1) {
+        // Add removing duplicates
+        featureCache.add(featureId);
+        renderedFeatures.push(getFeatureByFormat(f.object, format));
```
Pessimistress, line 199: What is the use case for the `format` argument? It's no longer "features" that are returned if you take out just the properties. There's also no perf savings since all features are already stored in memory and will be passed by reference.

modules/geo-layers/src/mvt-layer/mvt-layer.js
```diff
@@ -165,6 +171,64 @@ export default class MVTLayer extends TileLayer {
...
+    const featureCache = new Set();
+    const renderedFeatures = [];
+
+    for (const f of features) {
+      const featureId = getFeatureUniqueId(f.object, this.props.uniqueIdProperty);
+
+      if (featureId === -1) {
+        // we have no id for the feature, we just add to the list
+        renderedFeatures.push(getFeatureByFormat(f.object, format));
+      } else if (!featureCache.has(featureId) && featureId !== -1) {
```
Pessimistress, line 196: ` && featureId !== -1` is redundant

docs/api-reference/geo-layers/mvt-layer.md
```diff
@@ -118,6 +121,40 @@ Receives arguments:
...
 tileJSON (Object) - the tileJSON fetched
 
+##### `onViewportChange` (Function, optional)
+
+`onViewportChange` called when the viewport changes or all tiles in the current viewport have been loaded. A function (`getRenderedFeatures`) to calculate the features rendered in the current viewport is passed as a JSON to this callback function.
+
+
+Receives arguments:
+
+* `getRenderedFeatures` (Function)
```
Pessimistress, line 131: Missing `viewport`

modules/geo-layers/src/mvt-layer/mvt-layer.js
alasarr: After a second thought, I've just removed it to make the interface simpler. We always return the geometry + attributes and the users can filter out based on their use case.

Pessimistress (approved):

PR #4958, opened 2020-09-21
Support TMS (flipped-y) indexing in TileLayer

kylebarron:
For #4757

#### Change List

- This PR supports _two_ ways to declare that a TileLayer should be indexed as TMS instead of XYZ, either by passing the `indexing: 'tms'` prop or by using `{-y}` in the data prop. If desired I can remove one of these.


modules/geo-layers/src/tile-layer/tile-layer.js
```diff
@@ -24,6 +24,7 @@ const defaultProps = {
   maxCacheByteSize: null,
   refinementStrategy: STRATEGY_DEFAULT,
   zRange: null,
+  indexing: 'xyz',
```
Pessimistress, line 27: Is `xyz` the actual term? Isn't the TMS index also expressed in x,y,z?
Call it `OSM` (geospatial) or `default` (non-geo)?

kylebarron: From my reading, XYZ vs TMS [appear to be the standard terms](https://gist.github.com/tmcw/4954720).

Another example, the [TileJSON spec uses `xyz` and `tms`](https://github.com/mapbox/tilejson-spec/tree/master/2.2.0) (lower case) 

>     // OPTIONAL. Default: "xyz". Either "xyz" or "tms". Influences the y
>     // direction of the tile coordinates.
>     // The global-mercator (aka Spherical Mercator) profile is assumed.
>     "scheme": "xyz",

modules/geo-layers/src/tile-layer/utils.js
```diff
@@ -29,17 +30,25 @@ export const urlType = {
...
+  let {x, y, z} = properties;
+  if (indexing === 'tms') {
+    y = Math.pow(2, z) - y - 1;
+  }
+
+  return template
+    .replace('{x}', x)
+    .replace('{y}', y)
+    .replace('{z}', z)
+    .replace('{-y}', Math.pow(2, z) - y - 1);
```
Pessimistress, line 51: > either by passing the indexing: 'tms' prop or by using {-y} in the data prop

What should `-y` be, the OSM y or TMS y?

According to your implementation, the effect of `indexing: 'tms'` and `{-y}` negate each other. Can they be both used? Avoid overwriting `y` in this function for better clarity.

modules/geo-layers/src/tile-layer/tile-layer.js
Pessimistress: - We don't necessarily have to align with TileJSON, but looks like most of our enums are in lowercase, so this is ok.
- I think `scheme` or `indexingScheme` is a better name than `indexing`.

kylebarron: I renamed to `indexingScheme`

modules/geo-layers/src/tile-layer/utils.js
kylebarron: I struggled with this while first drawing up the PR last week. Elsewhere I've seen support for both 1) setting a TMS: true prop and 2) having `{-y}` refer to a TMS coordinate.

Pessimistress: Is this correct though? If an user specified `indexingScheme` then they would expect `tile.y` to be in the TMS system, which it is not. Since we do not overwrite `tile.y`, `-y` and no `indexingScheme` may be the less confusing solution?

kylebarron: We discussed this in https://github.com/visgl/deck.gl/issues/4757#issuecomment-655721368

> True that the user will need to understand that `y` is not from the TMS indexing system if they want to handle this information programmatically. However, as far as I can tell, the original issue is about the automatic URL resolution

I think part of my original intention with overwriting `y` was so that both the `data` url and `x, y, z` would be in `indexingScheme`. I think `indexingScheme` is a little more flexible than `{-y}`, so if I had to choose one or the other I'd choose the former.

Pessimistress: That is also fine, though you'll have to move the `indexingScheme` handling to `Tileset2D` so that all tile indices are correctly populated.

Pessimistress: My comment that you quoted were referring to using `{-y}` instead of `indexingScheme`

Pessimistress (approved):

PR #4972, opened 2020-09-27
Bump loaders.gl and math.gl dependencies

Pessimistress:
#### Change List
- loaders -> 2.3.0-beta
- math -> 3.3.0-beta

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,8 +41,8 @@
     "jsdom": false
   },
   "devDependencies": {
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -40,9 +40,12 @@
   "browser": {
     "jsdom": false
   },
+  "resolutions": {
```
Pessimistress, line 43: This will be dropped when prod version of math is published

kylebarron (approved):

PR #5010, opened 2020-10-09
add onDataLoad and onDataError callbacks to carto-layer.js

juandjara:
#### Background
This PR covers a need that is present in every CARTO-PS project. You need to know if the request to the tiler service failed and when it finished loading, so you can catch the error or display a loading spinner when necessary.

#### Change List
* Change documentation in carto-bqtiler-layer.md to include the new callbacks
* Change documentation in carto-sql-layer.md to include the new callbacks
* Add callbacks to carto-layer.js

### DIFF
--- a/docs/api-reference/carto/carto-sql-layer.md
+++ b/docs/api-reference/carto/carto-sql-layer.md
@@ -110,9 +110,9 @@ Optional. Tile extent in tile coordinate space as defined by MVT specification.
 
 ### Callbacks
 
-#### `onLoad` (Function, optional)
+#### `onDataLoad` (Function, optional)
 
-`onLoad` is called when the request to the CARTO tiler was completed successfully.
+`onDataLoad` is called when the request to the CARTO tiler was completed successfully.
 
 - Default: `tilejson => {}`
 
@@ -120,9 +120,9 @@ Receives arguments:
 
 - `tilejson` (Object) - the response from the tiler service
 
-##### `onError` (Function, optional)
+##### `onDataError` (Function, optional)
 
-`onError` is called when the request to the CARTO tiler failed.
+`onDataError` is called when the request to the CARTO tiler failed.
 
 - Default: `console.error`
 
--- a/docs/api-reference/carto/carto-sql-layer.md
+++ b/docs/api-reference/carto/carto-sql-layer.md
@@ -114,7 +114,7 @@ Optional. Tile extent in tile coordinate space as defined by MVT specification.
 
 `onLoad` is called when the request to the CARTO tiler was completed successfully.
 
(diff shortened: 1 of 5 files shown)


Pessimistress:
I'm ok with the carto layers having their own callbacks, though we should consider not adding confusion to the existing callbacks:

- `onDataLoad` (Layer) - called with `data` when async loading is finished. Should you just overload this?
- `onTileLoad` (TileLayer)
- `onTileError` (TileLayer)
- `onError` (Deck) - called with any uncaught error

juandjara:
All right, thanks for the tip @Pessimistress. I didn't know about `onDataLoad`. I think I can just override this callback and add another callback named `onDataLoadError`

docs/api-reference/carto/carto-sql-layer.md
```diff
@@ -108,6 +108,28 @@ Optional. Tile extent in tile coordinate space as defined by MVT specification.
 
 * Default: `4096`
 
+### Callbacks
+
+#### `onLoad` (Function, optional)
```
Pessimistress, line 113: `onDataLoad`

modules/carto/src/layers/carto-layer.js
```diff
@@ -16,7 +18,16 @@ export default class CartoLayer extends CompositeLayer {
...
     if (changeFlags.dataChanged && data) {
-      this._updateTileJSON();
+      this._updateData();
+    }
+  }
+
+  async _updateData () {
+    try {
+      const tilejson = await this._updateTileJSON();
+      this.props.onDataLoad(tilejson);
```
Pessimistress, line 28: `this.setState({tilejson});` should be moved here, given that other methods in this base class expect the state to exist. This will remove the burden of individual subclasses having to remember setting it.

modules/carto/src/layers/carto-layer.js
```diff
@@ -16,7 +18,16 @@ export default class CartoLayer extends CompositeLayer {
...
+    } catch (err) {
+      this.props.onDataLoadError(err);
```
Pessimistress, line 30: Call it `onDataError` to mirror the `onTileLoad`/`onTileError` pair?

Pessimistress (approved):

juandjara:
@Pessimistress thanks for the follow up. I adressed the last comments in the last commit

Pessimistress:
Lint is still failing. You need to run `yarn lint fix` at root.

juandjara:
Done. Added `eslint-disable-next-line` before `console.error` for the default value of `onDataError`

alasarr (approved):

Issue #4998, opened 2020-10-07
8.3 Release Tracker


Pessimistress:
- [x] Bump all dependencies (math.gl, probe.gl and loaders.gl) to latest release
- [x] Publish alpha/beta
- [x] Bump all examples to latest alpha/beta
- [x] Stage the website
- [x] Verify whats-new
- [x] Verify upgrade-guide
- [ ] Verify all issues in milestone are closed or move out
- [x] Verify all API changes are documented
- [x] Website (docs and example) testing: [staging](https://pessimistress.github.io/deck.gl/)
    - [x] Testing on Chrome
    - [x] Testing on Safari
    - [x] Testing on Firefox
    - [ ] Testing on Edge + IE11
    - [x] Testing on iOS
    - [ ] Testing on Android
- [x] Cut 8.3-release branch after all fixes
- [x] Update website doc urls to point to `8.3-release` branch
- [x] Publish 8.3 prod version

Pessimistress:
TODOs re: carto module @alasarr 

- [ ] `/examples/get-started/pure-js/carto` does not work standalone (missing `@deck.gl/carto` dependency)
- [ ] Nice example for the website

Pessimistress:
iOS:

- [ ] HeatmapLayer example all back
- [ ] ScreenGridLayer example not rendered

PR #5013, opened 2020-10-09
Fix HeatmapLayer on iOS 14

Pessimistress:
For #4998 

Sampling float texture in fragment shader worked in iOS 13 but not iOS 14

#### Change List
- Use RGBA texture for color scale


ibgreen (approved):
Assume it is not just related to WebGL2 now being available and our feature detection not working here? 

The litmus test would be to confirm this no longer works with WebGL1 context...

Pessimistress:
It worked in desktop Safari, which was still WebGL1.

PR #5020, opened 2020-10-11
Add integration category to examples page

Pessimistress:
For #4998

See [staged website](https://pessimistress.github.io/deck.gl/examples)

cc @Tamrat-B @damix911


kylebarron (approved):

alasarr (approved):

ibgreen (approved):
Looks amazing, and very nice to see the integrations listed together, having such great  integrations with the top geospatial services are big part of the value of deck.gl.

One observation is that while the info boxes talk about the data/visualization being shown, not much is said about the details of the integrations being offered.

The arcgis doesn't explain that this is an ArcGIS basemap, the Google one does not call out that this is now on top of Google Maps etc.

It may or may not be obvious to the users what is happening, but it would be nice to impress the reader with the depth of these integrations.

Maybe for V2.

Tamrat-B:
@Pessimistress the 3D integration with SceneView is experimental and needs to address some major performance issue -@damix911 is aware and working on it. So might be too early to have it as an integration example ? However, the 2d integration example for arcgis should be mature and could be featured...

The [i3s layer support](https://loaders.gl/examples/i3s), via loaders.gl in deck.gl is mature and performs good - could that be featured as part of the Tile3D layers integration example ?

PR #5029, opened 2020-10-13
Playground compatible with CARTO layers

alasarr:
CARTO layers don't work in the playground, this PR tries to fix that.

### DIFF
--- a/examples/playground/json-examples/carto-bq-tiler.json
+++ b/examples/playground/json-examples/carto-bq-tiler.json
@@ -1,0 +1,26 @@
+{
+  "description": "CartoBQTilerLayer declarative example",
+  "initialViewState": {
+    "latitude": 40.7368521,
+    "longitude": -73.8536065,
+    "zoom": 9,
+    "pitch": 0,
+    "bearing": 0
+  },
+  "views": [
+    {
+      "@@type": "MapView",
+      "controller": true,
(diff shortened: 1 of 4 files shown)


examples/playground/json-examples/carto-bq-tiler.json
```diff
@@ -0,0 +1,26 @@
...
+      "@@type": "MapView",
+      "controller": true,
+      "mapStyle": "https://basemaps.cartocdn.com/gl/positron-gl-style/style.json"
+    }
+  ],
+  "layers": [
+    {
+      "@@type": "CartoBQTilerLayer",
+      "data": "cartobq.maps.nyc_taxi_points_demo_id",
+      "getFillColor": [232, 133, 113],
```
alasarr, line 21: I'd like to have a function to style by color using a property. Is this something feasible?

Something like this:
```javascript
getFillColor: (object) => {
  // Apply color based on an attribute
  if (object.properties.aggregated_total > 1000000) {
    return [255, 0,0];

  } else {
    return [0, 0, 255] ;
  }
}
```

Any ideas?

kylebarron: See https://deck.gl/docs/api-reference/json/conversion-reference#functions-and-using-the--prefix

Pessimistress (approved):

ajduberstein (approved):

Issue #4966, opened 2020-09-22
How can we remove image smoothing from BitMap layer (deck.gl)?


Sgajjar08:
I am using the deck.gl (v7.3.3) library to load the layers on the map and currently, I am using one of the layers, which is the Bitmap layer to load the raster image on the map with the help of some dependant libraries. After seeing the result, the image smoothing property is already been applied to the image on the map. Is there any way that I can remove image smoothing from layer or map?

For the help to understand the issue more, I am sharing the code sandbox link => https://codesandbox.io/s/sleepy-borg-94zcg?file=/index.html. In this example, we have two canvases, one simple canvas which is on the top (required result and in the bottom-center, we have an image which is integrated from the Bitmap layer by deck.gl. [screenShot of result](https://i.stack.imgur.com/yjyBJ.png)

Pessimistress:
The default texture parameters are defined here:
https://github.com/visgl/deck.gl/blob/041600f4a92ad35b597a26ee154b2d39bd5326b5/modules/layers/src/bitmap-layer/bitmap-layer.js#L30-L35

I think it's a fair feature request to be able to overwrite them.

With the current releases, you can supply your own [Texture2D](https://luma.gl/docs/api-reference/webgl/texture-2d) object to the `image` prop:

```js
    const deckgl = new DeckGL({
        ...
        onWebGLInitialized: (gl) => renderBitmap(gl),
        layers: []
      });

      function renderBitmap(gl) {
        const texture = new luma.Texture2D(gl, {
          data: canvas,
          parameters: {
            [gl.TEXTURE_MIN_FILTER]: gl.NEAREST,
            [gl.TEXTURE_MAG_FILTER]: gl.NEAREST
          }
        });

        const bitmapLayer = new BitmapLayer({
          id: "bitmap-layer",
          bounds: bbox,
          opacity: 1,
          image: texture
        });

        deckgl.setProps({ layers: [bitmapLayer] });
      }
```

Sgajjar08:
Hello @Pessimistress ,

Thanks for your response and it worked perfectly fine. I am getting expected results. Once again, Thanks a lot.

PR #5031, opened 2020-10-14
Add textureParameters prop to BitmapLayer

Pessimistress:
For #4966

#### Change List
- Add `textureParameters` prop to `BitmapLayer`
- Documentation


kylebarron (approved):

chrisgervang:
Oh this is great! I can replace my custom layer in my gameboy visualization project with this now. I do the same customization to the IconLayer too, could we consider adding this prop to IconLayer and then send it as a parameter to IconManager too?

Issue #5041, opened 2020-10-17
TileLayer Locking


ilan-gold:
#### Description
It seems that if you zoom/pan/interact with a `TileLayer` application for long enough, it locks.  I haven't been able to fully determine why this is happening - I think I notice that the `requestScheduler` eventually stops giving "`priority`" for new tiles to fetch, but maybe I was mis-reading that.

#### Repro Steps
Zoom/pan somewhat haphazardly/quickly for a minute or two on a `TileLayer` (which I recognize may be an abuse of the server, but I also have seen this happen much quicker than a minute or two).  Maybe this is just an extreme edge-case, but I wanted to bring it up to see if there is a condition or something in the code that might be a known weak-point.  I'm going to keep digging in the interim as well.  I also recognize these are not great "Repro steps" but if people don't think this is really an issue, I can also close this.

#### Environment (please complete the following information):
 - Framework Version: 8.3.0
 - Browser Version: 81.0.2 (64-bit)
 - OS: 10.14.6

#### Screen-gif
![ezgif com-video-to-gif(8)](https://user-images.githubusercontent.com/43999641/96352421-fe236680-1090-11eb-8706-71c62955d1a9.gif)

Pessimistress:
I believe this was introduced by the abort controller PR:

https://github.com/visgl/deck.gl/blob/master/modules/geo-layers/src/tile-layer/tile-2d-header.js#L58-L61

If the tile is aborted, `requestToken` is never marked `done`. @kylebarron @supersonicclay

ilan-gold:
Yes I was mucking around the `esm` build @Pessimistress and I noticed that newly created layers would not go past this point.  That certainly sounds correct.

kylebarron:
I'm guessing the issue is from this abort https://github.com/visgl/deck.gl/pull/4838/files#diff-ebfc47135e3d8fe2bee8b3d33fae4ebbbc295d01887c751d4cb77f71fc85a87aR192, since the abort from the AbortController happens only after the fetch has already been initiated.

I guess you want:

```diff
- if (!requestToken || this._isCancelled) {
-   this._isCancelled = true;
-   return;
- }
+ if (this._isCancelled) {
+   requestToken && requestToken.done();
+   return;
+ }
+ if (!requestToken) {
+   this._isCancelled = true;
+   return;
+ }
```
?

Pessimistress:
Published 8.3.2

ilan-gold:
Thanks!

PR #5044, opened 2020-10-19
Fix Quickly Aborted Tiles

ilan-gold:
I tried my hand at fixing #5041 (using Kyle's recommendation) and added a test case.
#### Background
In #5041 Kyle/Xiaoji noted that the individual tiles were not calling `done` on their `requestToken` if the `AbortSignal` was raised before the actual call to `getTileData`.  This PR should fix this.

#### Change List
- Update `TileHeader2D` to handle quickly-aborted tiles, before calling `getTileData`.
- Add a test that calls `abort` before awaiting the response of `_loadData`.

### DIFF
--- a/modules/geo-layers/src/tile-layer/tile-2d-header.js
+++ b/modules/geo-layers/src/tile-layer/tile-2d-header.js
@@ -54,17 +54,16 @@ export default class Tile2DHeader {
       return tile.isSelected ? 1 : -1;
     });
 
-    // A tile can be cancelled while being scheduled
-    if (this._isCancelled) {
-      requestToken && requestToken.done(); // eslint-disable-line no-unused-expressions
-      return;
-    }
     if (!requestToken) {
       this._isCancelled = true;
       return;
     }
+    // A tile can be cancelled while being scheduled
+    if (this._isCancelled) {
+      requestToken.done();
+      return;
+    }
 
-    this._isCancelled = false;
     let tileData;
     let error;
     try {
--- a/modules/geo-layers/src/tile-layer/tile-2d-header.js
+++ b/modules/geo-layers/src/tile-layer/tile-2d-header.js
@@ -55,7 +55,11 @@ export default class Tile2DHeader {
     });
 
     // A tile can be cancelled while being scheduled
-    if (!requestToken || this._isCancelled) {
+    if (this._isCancelled) {
+      requestToken && requestToken.done(); // eslint-disable-line no-unused-expressions
+      return;
+    }
+    if (!requestToken) {
       this._isCancelled = true;
       return;
     }
--- a/test/modules/geo-layers/tile-layer/tile-2d-header.spec.js
+++ b/test/modules/geo-layers/tile-layer/tile-2d-header.spec.js
@@ -46,3 +46,21 @@ test('Tile2DHeader#Cancel request if not selected', async t => {
   t.ok(tile2.isCancelled, 'Second request was cancelled');
   t.end();
 });
+
+test('Tile2DHeader#Abort quickly', async t => {
+  const requestScheduler = new RequestScheduler({throttleRequests: true, maxRequests: 1});
+
+  const getTileData = () => null;
+  const onTileLoad = () => null;
+  const onTileError = () => null;
+
+  const tile = new Tile2DHeader({onTileLoad, onTileError});
+
+  // Await later so that request scheduler the abort could go off before the getTileData call.
+  const loader = tile._loadData(getTileData, requestScheduler);
+  tile.abort();
+  await loader;
+
+  t.notOk(requestScheduler.requestMap.has(tile), 'Scheduler deletes tile on abort');
+  t.end();
+});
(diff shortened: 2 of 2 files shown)


modules/geo-layers/src/tile-layer/tile-2d-header.js
```diff
@@ -55,7 +55,11 @@ export default class Tile2DHeader {
     });
 
     // A tile can be cancelled while being scheduled
-    if (!requestToken || this._isCancelled) {
+    if (this._isCancelled) {
+      requestToken && requestToken.done(); // eslint-disable-line no-unused-expressions
```
Pessimistress, line 59: Move this if statement after `if(!requestToken)` so you won't have to check this

test/modules/geo-layers/tile-layer/tile-2d-header.spec.js
```diff
@@ -46,3 +46,21 @@ test('Tile2DHeader#Cancel request if not selected', async t => {
   t.ok(tile2.isCancelled, 'Second request was cancelled');
   t.end();
 });
+
+test('Tile2DHeader#Abort quickly', async t => {
```
Pessimistress, line 50: This test passes regardless of your change

test/modules/geo-layers/tile-layer/tile-2d-header.spec.js
```diff
@@ -46,3 +46,21 @@ test('Tile2DHeader#Cancel request if not selected', async t => {
...
+  const requestScheduler = new RequestScheduler({throttleRequests: true, maxRequests: 1});
+
+  const getTileData = () => null;
+  const onTileLoad = () => null;
+  const onTileError = () => null;
+
+  const tile = new Tile2DHeader({onTileLoad, onTileError});
+
```
Pessimistress, line 58: Add `tile.isSelected = true;` so that a request is actually sceduled

test/modules/geo-layers/tile-layer/tile-2d-header.spec.js
```diff
@@ -46,3 +46,21 @@ test('Tile2DHeader#Cancel request if not selected', async t => {
...
+  // Await later so that the abort could go off before the getTileData call.
+  const loader = tile._loadData(getTileData, requestScheduler);
+  tile.abort();
+  await loader;
```
Pessimistress, line 62: Add `await new Promise((resolve) => setTimeout(resolve, 0));`, requestScheduler resolves asynchronously after the loader

Pessimistress:
This line is redundant: 
https://github.com/ilan-gold/deck.gl/blob/fix-5041/modules/geo-layers/src/tile-layer/tile-2d-header.js#L67

test/modules/geo-layers/tile-layer/tile-2d-header.spec.js
ilan-gold: https://github.com/visgl/deck.gl/pull/5044/files#diff-5dd332fb839a9666cc7d7489bfa32eda9f101a0f17b44034ea90a5d132d61b74R58

test/modules/geo-layers/tile-layer/tile-2d-header.spec.js
ilan-gold: https://github.com/visgl/deck.gl/pull/5044/files#diff-5dd332fb839a9666cc7d7489bfa32eda9f101a0f17b44034ea90a5d132d61b74R53

Pessimistress: That's not the same thing. After loader is resolved, requestScheduler's `_issueRequest` is still pending.

ilan-gold: Ok, I can add this - I don't fully understand why it's needed.  Can you explain a bit?

Pessimistress: I suggest you run this test with and without your change to `tile-2d-header.js` - If the test is constructed correctly, it should fail before the change and succeed after.

Pessimistress: Sorry ignore my previous suggestion - I thought you were calling `loadData`

test/modules/geo-layers/tile-layer/tile-2d-header.spec.js
```diff
@@ -46,3 +46,22 @@ test('Tile2DHeader#Cancel request if not selected', async t => {
   t.ok(tile2.isCancelled, 'Second request was cancelled');
   t.end();
 });
+
+test('Tile2DHeader#Abort quickly', async t => {
+  const requestScheduler = new RequestScheduler({throttleRequests: true, maxRequests: 1});
+
+  const getTileData = await new Promise(resolve => setTimeout(resolve, 0)); // eslint-disable-line no-undef
```
Pessimistress, line 53: Revert it to `() => null` because this `await` resolves to `undefined`

test/modules/geo-layers/tile-layer/tile-2d-header.spec.js
ilan-gold: Ok, yes, I was just checking this and was confused.  No worries. I will revert the loader change.

Pessimistress (approved):

kylebarron:
Thanks so much for the PR @ilan-gold!

PR #5049, opened 2020-10-21
[ScenegraphLayer] Animated Render Test

chrisgervang:
Regression test for #4574

#### Change List
- adds `onBeforeRender: ({deck, layers}) => {}` to [snapshot test case](https://deck.gl/docs/api-reference/test-utils/snapshot-test-runner#addtestcase-arrayobject).

I saw these logs from loaders, I think. The tests work but want to surface.
Screen Shot 2020-10-20 at 8 57 09 PM


chrisgervang:
@ibgreen there are a couple loaders yellow text outputs from the test. Wanna flag in case it's something to worry about.

test/render/test-cases/scenegraph-layer.js
```diff
@@ -0,0 +1,105 @@
...
+      pitch: 30,
+      bearing: 0
+    },
+    onBeforeRender: ({deck, layers}) => {
+      deck.animationLoop.timeline.pause();
+      deck.animationLoop.timeline.setTime(400);
+    },
+    layers: [
+      new ScenegraphLayer({
+        id: 'scenegraph-layer-frame-2',
```
Pessimistress, line 58: You shouldn't have to change the layer id in between these tests

Pessimistress (commented):
Nice!

Pessimistress:
Regarding the loaders.gl warning, I think this object: https://github.com/visgl/deck.gl/blob/master/modules/mesh-layers/src/scenegraph-layer/scenegraph-layer.js#L216 should be nested under `{gltf: {...}}`

test/render/test-cases/scenegraph-layer.js
chrisgervang: oh ok, I'll make them the same since they should be the same.

chrisgervang:
Huh, following that code though it looks like the options get merged with these defaults, which don't nest. https://github.com/visgl/luma.gl/blob/d8cad75b9f8ca3e578cf078ed9d19e619c2ddbc9/modules/experimental/src/gltf/gltf-instantiator.js#L19-L33

chrisgervang:
Interesting, so when they are the same `id`, deck's normal layer memoization kick in, right? That results in two observations.

1. The loaders warnings only print once, since its only loading the object once I presume.
2. The 3rd frame is only an 88.4% match.

Screen Shot 2020-10-21 at 10 42 45 AM

Pessimistress (approved):

ibgreen:
I suspect those warning are from options that are (or used to be) handled by luma.gl, I don't think these can be defined by loaders.gl. 

If we had good Scenegraph tests in luma.gl I would have expected that these warnings were found there first.

https://github.com/visgl/luma.gl/issues/1411

Issue #5051, opened 2020-10-21
ContourLayer aggregation: 'MEAN' does not work on Windows


Pessimistress:
#### Description
`ContourLayer` with `aggregation: 'MEAN'` does not render on Windows Chrome.

Other aggregation modes work as expected.

#### Repro Steps
- Open https://pydeck.gl/gallery/contour_layer.html
- Layer does not render

#### Environment (please complete the following information):
 - Framework Version: 8.3.2
 - Browser Version: Chrome, Edge
 - OS: Windows

#### Logs
No error in the console. `aggregationData` read from the `GPUGridAggregator` are all zeros: https://github.com/visgl/deck.gl/blob/8.3-release/modules/aggregation-layers/src/utils/gpu-grid-aggregation/gpu-grid-aggregator.js#L215

cc @ajduberstein

Pessimistress:
Fixed in 8.3.3

PR #4113, opened 2020-01-08
GPUGridAggregator: Add WA for ANGLE specific bug.

1chandu:
For #4065 

#### Background

Adding a workaround specific to ANGLE. With this fix, GPU Aggregation works on windows:

Screenshot (4)
Screenshot (5)
Screenshot (6)

#### Change List
-

PR #5052, opened 2020-10-21
Fix MEAN aggregation on Windows

Pessimistress:
For #5051

This bug affects ContourLayer, GridLayer and ScreenGridLayer

#### Background

Same work around for ANGLE bug in https://github.com/visgl/deck.gl/pull/4113, applied to MEAN aggregation

#### Change List

- Explicitly set `gl_pointSize`


ajduberstein (approved):

ibgreen (approved):
Classic

PR #5053, opened 2020-10-21
Integration Maps API v2

alasarr:
Integration with Maps API v2.

* Refactor maps-api-client to receive a mapConfig.
* Move mapConfig generation to Layers. It improves the generation of new layers, it's a standard process where each layer needs to define its mapConfig via the method `buildMapConfig`
* Rename auth.js to config.js
* Add SQL API endpoint, for future integration with small datasets

V2 is not the default, we'll make the default in a future release after the corresponding QA.


alasarr:
@VictorVelarde please take a look at this since you've the internals from CARTO.

modules/carto/src/api/maps-api-client.js
```diff
@@ -1,76 +1,67 @@
-import {getDefaultCredentials} from '../auth';
+import {getDefaultCredentials, getMapsVersion} from '../config';
 
 const DEFAULT_USER_COMPONENT_IN_URL = '{user}';
-const REQUEST_GET_MAX_URL_LENGTH = 2048;
+const DEFAULT_REGION_COMPONENT_IN_URL = '{region}';
```
VictorVelarde, line 4: Ok, I see the new param. 
One question: isn't the module going to use the MAX_URL threshold to use a GET vs a POST request?

docs/api-reference/carto/overview.md
```diff
@@ -54,3 +54,25 @@ You can see real examples for the following:
 
 * [Pure JS](https://github.com/CartoDB/viz-doc/tree/master/deck.gl/examples/pure-js): integrate in a pure js application, using webpack.
 
+
+### CARTO credentials
+
+CARTO credentials is an object to define the connection credentials:
```
VictorVelarde, line 60: Maybe "This is an object to define the connection to CARTO, including the credentials (and optionally the parameters to point to specific api endpoints)"

modules/carto/src/api/maps-api-client.js
```diff
@@ -84,83 +75,50 @@ function dealWithWindshaftError({response, layergroup, credentials}) {
...
     default:
-      throw new Error(`${JSON.stringify(layergroup.errors)}`);
+      const e = getMapsVersion() === 'v1' ? JSON.stringify(json.errors) : json.error;
+      throw new Error(e);
   }
 }
 
 /**
- * Create a GET or POST request, with all required parameters
+ * Build a URL all required parameters
```
VictorVelarde, line 86: missing `with`: "Build a URL with all required parameters"

modules/carto/src/api/maps-api-client.js
alasarr: Good point! We're deprecating the POST in vector because they're not cached in the CDN and the config is far smaller since we're removing all the stuff of raster (like cartocss, dataviews, etc...)

modules/carto/src/api/maps-api-client.js
```diff
@@ -84,83 +75,50 @@ function dealWithWindshaftError({response, layergroup, credentials}) {
...
  */
-function createMapsApiRequest({config, credentials}) {
+function buildURL({mapConfig, credentials}) {
+  const cfg = JSON.stringify(mapConfig);
   const encodedApiKey = encodeParameter('api_key', credentials.apiKey);
   const encodedClient = encodeParameter('client', `deck-gl-carto`);
```
VictorVelarde, line 91: "deck-gl-carto", could be also a parameter in the configuration, that could be injected from outside or have this default value

modules/carto/src/api/maps-api-client.js
```diff
@@ -84,83 +75,50 @@ function dealWithWindshaftError({response, layergroup, credentials}) {
...
 
 /**
- * Simple POST request
+ *  This function will be deprecated really soon when we'll make v2 the default
  */
-function postRequest(url, payload) {
-  return new Request(url, {
-    method: 'POST',
+async function getTileJSONBigQueryCloudFunction(mapConfig) {
+  const BQ_TILEJSON_ENDPOINT = 'https://us-central1-cartobq.cloudfunctions.net/tilejson';
```
VictorVelarde, line 116: Does it make sense to make this url also configurable?

modules/carto/src/config.js
```diff
@@ -0,0 +1,33 @@
...
+    ...credentials,
+    ...opts
+  };
+}
+
+export function getDefaultCredentials() {
+  return credentials;
+}
+
+export function getMapsVersion(creds) {
```
VictorVelarde, line 26: I see clearly the point for the getDefaultCredentials / setDefaultCredentials.... what's the use case for `getMapsVersion`?

modules/carto/src/index.js
```diff
@@ -1,3 +1,3 @@
-export {getDefaultCredentials, setDefaultCredentials} from './auth.js';
+export {getDefaultCredentials, setDefaultCredentials} from './config.js';
```
VictorVelarde, line 1: Yeah, I think a more generic name makes sense

modules/carto/src/api/maps-api-client.js
alasarr: Nope, because it won't be deprecated. The new endpoint for bqtiler layers is the Maps API, so it'll be configurable via config.mapsUrl

This code will be removed as soon as we'll do the switch to v2.

test/modules/carto/config.spec.js
```diff
@@ -1,35 +1,36 @@
...
   // partial (keeping other params)
   setDefaultCredentials({username: 'a-new-username'});
   let credentials = getDefaultCredentials();
   t.ok(credentials.username === 'a-new-username', 'user update');
   t.ok(credentials.apiKey === 'default_public', 'keep default apiKey');
-  t.ok(
-    credentials.serverUrlTemplate === 'https://{user}.carto.com',
-    'keep default serverUrlTemplate'
-  );
+  t.ok(credentials.mapsUrl === 'https://{user}.carto.com/api/v1/map', 'keep default mapsUrl');
```
VictorVelarde, line 17: How about adding also the 'region' to the specs?

modules/carto/src/config.js
alasarr: It's used from  CartoSQLLayer to get the mapsVersion to generate the mapConfig of v1 or v2.

Also used at maps-api-client to parse the response of mapsAPI v1 or v2

test/modules/carto/config.spec.js
alasarr: As we're just inserting and fetching an object it just make the test bigger without adding too much value. 

In fact, we should only keep 1, otherwise we're just testing object aassignation in JS 😄

modules/carto/src/config.js
VictorVelarde: Ok, I see

test/modules/carto/config.spec.js
VictorVelarde: Just in case we change this in the future, but yes... it doesn't matter much

VictorVelarde (approved):
👍

Issue #5048, opened 2020-10-20
Deformation of small polygons on mobile


doninpr:
I have a problem with small polygons (PolygonLayer) on mobile iOS 14 (any browser), it looks different than on desktop or android. And I don't have any idea how to fix this.
Can someone help me with this?

Sandbox: https://codesandbox.io/s/friendly-montalcini-esgdb?fontsize=14&hidenavigation=1&theme=dark&file=/src/App.js

This is how it looks on desktop:
![desktop](https://user-images.githubusercontent.com/23110750/96646777-55e1fc00-1335-11eb-8bfd-972b1f40a676.png)

And this is on mobile:
![IMG_5138](https://user-images.githubusercontent.com/23110750/96647010-ae18fe00-1335-11eb-876a-b731132286e7.jpg)

Pessimistress:
If you need help debugging, please report the OS, browser and version of deck.gl you are using, and create a Codepen/Codesandbox that reproduces your issue.

doninpr:
> If you need help debugging, please report the OS, browser and version of deck.gl you are using, and create a Codepen/Codesandbox that reproduces your issue.

Thank you, I updated message.

Pessimistress:
This seems to be a regression on iOS 14. I put together a test:
https://observablehq.com/@pessimistress/deck-gl-precision-test

Expected:

![image](https://user-images.githubusercontent.com/2059298/96775832-f4557800-139c-11eb-8740-e1511ee9d143.png)

Safari for iOS14:

<img width="304" alt="PNG image" src="https://user-images.githubusercontent.com/2059298/96775995-37175000-139d-11eb-9f8d-6fdb197befc7.png">

Pessimistress:
@doninpr while we investigate, you can consider using the `LNGLAT_OFFSETS` or `METER_OFFSETS` [coordinate system](https://deck.gl/docs/developer-guide/coordinate-systems#supported-coordinate-systems) to work around this issue.

Pessimistress:
Fixed in 8.3.3

PR #5056, opened 2020-10-21
Fix project precision on iOS 14

Pessimistress:
For https://github.com/visgl/deck.gl/issues/5048#issuecomment-713839826

The shader compiler in Safari iOS really wants to reorder these two:

```
position_world += project_uModelMatrix * vec4(position64Low, 0.0);
```

Safari is the new IE.

#### Change List
- Shift things around

Tested on iPhone.


ibgreen (approved):
I would probably add a comment that order matters here due to Safari issues.

Issue #4531, opened 2020-04-25
TerrainLayer binding in pydeck


LjubeBoskovski:
#### Target Use case
 I want to be able to use the TerrainLayer in pydeck. It is not listed under the pydeck examples, so I guess it is not implemented yet?

#### Proposed feature
Sample code snippet of how it probably would be implemented:
```
import pydeck as pdk

ELEVATION_LAYER_DATA = 'https://raw.githubusercontent.com/uber-common/deck.gl-data/master/website/terrain.png'
TEXTURE = 'https://raw.githubusercontent.com/uber-common/deck.gl-data/master/website/terrain-mask.png'

layer = pdk.Layer(
        "TerrainLayer",
        data=ELEVATION_LAYER_DATA,
        elevationDecoder={
                'rScaler': 2,
                'gScaler': 0,
                'bScaler': 0,
                'offset': 0
                },
        texture=TEXTURE,
        bounds=[-122.5233, 37.6493, -122.3566, 37.8159]
        )

view_state = pdk.ViewState(latitude=37.7749295, longitude=-122.4194155, zoom=11, bearing=0, pitch=45)

# Render
r = pdk.Deck(layers=[layer], initial_view_state=view_state)
r.to_html("terrain_layer.html", notebook_display=False)
```
This sample was created following the style of the other pydeck examples and the deck.gl TerrainLayer example.

kylebarron:
I made an example yesterday, but haven't put it on the website yet:

![image](https://user-images.githubusercontent.com/15164633/80288013-35bce180-86f2-11ea-9073-33ceeb9a56e9.png)

You'll need to add your own surface image satellite tile endpoint:

```py
import pydeck as pdk
view_state = pdk.ViewState(latitude=46.24, longitude=-122.18, zoom=11.5, bearing=140, pitch=60)
TERRAIN_IMAGE = '"https://s3.amazonaws.com/elevation-tiles-prod/terrarium/{z}/{x}/{y}.png"'
SURFACE_IMAGE = '"..."'
ELEVATION_DECODER = {
  'rScaler': 256,
  'gScaler': 1,
  'bScaler': 1 / 256,
  'offset': -32768
};
layer = pdk.Layer('TerrainLayer', None, min_zoom=0, max_zoom=23, strategy='no-overlap', elevation_decoder=ELEVATION_DECODER, texture=SURFACE_IMAGE, elevation_data=TERRAIN_IMAGE)
r = pdk.Deck(
    layers=[layer],
[...]

PR #5061, opened 2020-10-25
WIP add quantized mesh loader to terrain layer

kylebarron:
#### Background

The Quantized Mesh loader was added in loaders.gl 2.2, which parses the [Quantized Mesh format](https://github.com/CesiumGS/quantized-mesh).

#### Change List

- Add QuantizedMeshLoader to the loaders list when loading terrain

##### Web Mercator Projection only

Quantized Mesh is often tiled in the EPSG: 4326 projection, which is incompatible with the current TileLayer, which only accepts Web Mercator/OSM indexing.

##### Mesh orientation

This is an issue I've struggled with before. Without changing the existing code when I use quantized mesh I see the images flipped:

image

I might need to check that I'm serving the mesh according to the spec

cc @davidbrochart, #4531

### DIFF
--- a/modules/geo-layers/src/terrain-layer/terrain-layer.js
+++ b/modules/geo-layers/src/terrain-layer/terrain-layer.js
@@ -107,7 +107,7 @@ export default class TerrainLayer extends CompositeLayer {
         elevationDecoder
       },
       'quantized-mesh': {
-        bounds
+        bounds: [bounds[0], bounds[3], bounds[2], bounds[1]]
       }
     };
     if (workerUrl !== null) {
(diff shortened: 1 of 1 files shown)


Pessimistress:
To avoid overbundling dependencies and be more future proof, this layer could adopt the Tile3DLayer's [loader](https://deck.gl/docs/api-reference/geo-layers/tile-3d-layer#loader) approach. The default loader would be the terrain loader, but users can choose to import another.

Similarly, rather than hardcoding format-specific logic, loaders should come up with a schema that describes the mesh coordinate system/transform in the returned `header`, for this layer to consume.

Pessimistress:
The use of CARTESIAN coordinates in this layer should not be mandatory. It is an optimization for rendering a specific format (color-coded image) to a specific projection (Web Mercator). If we want this layer to support e.g. globe view, the loader will have to provide the option to output wgs84 coordinates.

ibgreen-openai:
Close due to inactivity

Issue #5065, opened 2020-10-26
@deck.gl/arcgis SceneView camera misalignment


damix911:
#### Description

When using the `@deck.gl/arcgis` with `SceneView`, the Esri basemap and the deck.gl layers don't align at certain tilt angles.

![misalignment](https://user-images.githubusercontent.com/12469086/97196214-74972700-1769-11eb-9833-640d5004ea8b.png)

#### Repro Steps
- Run `test/apps/arcgis`; the bottom half of the screen is a `SceneView`. Tilt the camera by dragging with RMB and see that basemap and layer are not aligned.

#### Environment
 - Framework Version: 8.3.3 and earlier
 - Browser Version: all
 - OS: all

#### Logs
Nothing to report.

Pessimistress:
8.3.4

llwinner:
Still have problems

llwinner:
> 8.3.4

Still have problems

PR #5066, opened 2020-10-26
ArcGIS 3D altitude fix

damix911:
Seems like it should fix the wrong alignment between an ArcGIS basemap with `SceneView` and deck.gl.

@Pessimistress gets full credit for this; I was on the wrong track by trying to match the FOV of a `FirstPersonView`. Thanks!

For #5065

### DIFF
--- a/modules/arcgis/src/deck-renderer.js
+++ b/modules/arcgis/src/deck-renderer.js
@@ -2,6 +2,12 @@
 
 import {initializeResources, render, finalizeResources} from './commons';
 
+function arcgisFOVToDeckAltitude(fov, aspectRatio) {
+  const D = Math.sqrt(1 + aspectRatio ** 2);
+  const halfFOV = fov / 2 / 180 * Math.PI;
+  return D / 2 / Math.tan(halfFOV);
+}
+
 export default function createDeckRenderer(DeckProps, externalRenderers) {
   class DeckRenderer {
     constructor(view, props) {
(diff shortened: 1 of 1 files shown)


modules/arcgis/src/deck-renderer.js
```diff
@@ -2,6 +2,12 @@
 
 import {initializeResources, render, finalizeResources} from './commons';
 
+function arcgisFOVToDeckAltitude(fov, aspectRatio) {
```
Pessimistress, line 5: Add a comment that ArcGIS fov is corner-to-corner?

Pessimistress (approved):

llwinner:
Still have problems

damix911:
@llwinner I'll look into this. Do you have a URL you can share? If not I will use the same test app `test/apps/arcgis`. Thanks!

llwinner:
> @llwinner I'll look into this. Do you have a URL you can share? If not I will use the same test app `test/apps/arcgis`. Thanks!

I found that there are problems in two ways.
First, in sceneView, modify viewingMode: "global".
Second, the ArcGIS basemap must be a vector tile layer. If you change it to basemap: "streets", problems will occur.

damix911:
> First, in sceneView, modify viewingMode: "global".

Yes, this is a current limitation; globes are not supported yet in the `@arcgis/deck.gl` plugin.

> Second, the ArcGIS basemap must be a vector tile layer. If you change it to basemap: "streets", problems will occur.

You are right, this is probably related to tile size; MapView and SceneView behave differently in this respect and we got it right for `MapView` but quite clearly not for `SceneView`.

I'll open an issue and hopefully fix it soon; I think we are basically off by a factor of 2.

Thanks!

damix911:
Opened this issue: #5271

sgavathe:
@damix911 

> Second, the ArcGIS basemap must be a vector tile layer. If you change it to basemap: "streets", problems will occur.

the above is still an issue with the latest deck.gl/arcgis. When you use the basemap for scene view anything other than **dark-gray-vector** or **light-gray-vector** , the issue occurs. The layers don't project correct. 

image

PR #5075, opened 2020-10-28
Fix: Maps APIv2 has changed the endpoints

alasarr:
During the beta of Maps API v2 we've changed the URLs and these changes are in production and they're not backward compatible.


VictorVelarde (approved):
👌

PR #5080, opened 2020-10-28
Model Matrix for TileLayer (non-geospatial)

### LINKED ISSUE #5047: Scale/Translate TileLayer
#### Target Use case
Much like you can scale/translate a `BitmapLayer` by affecting its bounds, it would be cool to do the same to the `TileLayer`.  The hairy part would be the interplay between the `Viewport` and tile fetching but I think it could be possible.  Is this something we'd be interested in supporting?  @kylebarron I would be happy to work on this PR (and make sure it's even possible).

#### Proposed feature
Add `scale` and `translate` parameters to the `TileLayer` to allow for uniform transformations to the layer and and underlying tiles.

#### To Do List 
  - [ ] Coding
  - [ ] Doc update
  - [ ] What’s new update 
  - [ ] Test

ilan-gold:
This fixes #5047 (I think).  It worked under `OrbitView` and `OrthographicView` for scaling, rotation, and translation (and mixtures of them).  

I'm curious on thoughts about documentation since nothing really changed - we only added support for a coordinate system.

The only catch is that under rotation, you may end up fetching tiles outside of the `extent` prop since transforming the unprojected screen coordinates prop by the `modelMatrix` can yield a rotated box, in which case we can't index tiles (since we currently just use two for loops from min to max on each side of the box).  Thus I do another transformation (see [this line](https://github.com/visgl/deck.gl/compare/master...ilan-gold:modelMatrix_tileLayer_nogeo?expand=1#diff-308305ea22abe6c8fbb34f14d60ba7c54451aa71026ab0d4057a75e7928f7635R106)) to get the circumscription box of that rotated box in order to fetch tiles, which in turn can yield tiles outside of the `extent` box.  I am certainly open to other solutions given my inexperience with this sort of thing.  Other than this hiccup, this solution should work under rotation and there should be no other 404s from missing tiles.

I also tried to sequester my solution off from the geospatial use-case, as it sounded like `modelMatrix` works there without this change.


modules/geo-layers/src/tile-layer/utils.js
```diff
@@ -61,19 +86,26 @@ function getBoundingBox(viewport, zRange, extent) {
...
+      Math.min(Math.max(bounds[3], extent[1]), extent[3])
+    ];
+  }
+  return bounds;
 }
 
-function getTileIndex([x, y], scale) {
-  return [(x * scale) / TILE_SIZE, (y * scale) / TILE_SIZE];
+function getIndexingCoords(bbox, scale, modelMatrix) {
+  modelMatrix.invert();
```
ilan-gold, line 105: Is this expensive to do? Should we store the inverse somewhere?

modules/geo-layers/src/tile-layer/utils.js
```diff
@@ -141,9 +180,18 @@ export function getTileIndices({viewport, maxZoom, minZoom, zRange, extent, tile
   if (Number.isFinite(maxZoom) && z > maxZoom) {
     z = maxZoom;
   }
+  let transformedExtent = extent;
+  if (modelMatrix && extent && !viewport.isGeospatial) {
+    transformedExtent = transformBox(extent, modelMatrix);
```
ilan-gold, line 185: Should this be stored outside of this since it doesn't need to be run every time?

modules/geo-layers/src/tile-layer/tile-layer.js
```diff
@@ -87,7 +88,8 @@ export default class TileLayer extends CompositeLayer {
         onTileLoad: this._onTileLoad.bind(this),
         onTileError: this._onTileError.bind(this),
         onTileUnload: this._onTileUnload.bind(this),
-        maxRequests
+        maxRequests,
+        modelMatrix
```
Pessimistress, line 92: This doesn't account for changes in the `modelMatrix` prop, which could potentially happen on every render. It should be passed into `getTileIndices` instead of the constructor.

modules/geo-layers/src/tile-layer/utils.js
```diff
@@ -141,9 +186,18 @@ export function getTileIndices({viewport, maxZoom, minZoom, zRange, extent, tile
...
+  }
   return viewport.isGeospatial
     ? getOSMTileIndices(viewport, z, zRange, extent || DEFAULT_EXTENT)
-    : getIdentityTileIndices(viewport, z, extent || DEFAULT_EXTENT);
+    : getIdentityTileIndices(
+        viewport,
+        z,
+        transformedExtent || DEFAULT_EXTENT,
+        modelMatrix || new Matrix4()
```
Pessimistress, line 199: I suggest you avoid assigning a default value here and check for null everywhere before you use the matrix. That way we can be confident that for existing use cases there will be no behavioral change/perf hit.

modules/geo-layers/src/tile-layer/utils.js
Pessimistress: Probably. Also `invert` twice does not revert to the original matrix due to precision loss. You should use `matrix.clone().invert()`.

modules/geo-layers/src/tile-layer/tileset-2d.js
```diff
@@ -104,14 +104,17 @@ export default class Tileset2D {
    * @param {*} viewport
    * @param {*} onUpdate
    */
-  update(viewport, {zRange} = {}) {
-    if (!viewport.equals(this._viewport)) {
+  update(viewport, {zRange} = {}, modelMatrix, modelMatrixInverse) {
+    if (!viewport.equals(this._viewport) || modelMatrixInverse !== this._modelMatrixInverse) {
```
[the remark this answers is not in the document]
ilan-gold: @Pessimistress Can you reconcile this with the above where did not do deep equality?  Is there a reason to do it here?  Am I misunderstanding?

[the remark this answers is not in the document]
Pessimistress: So thinking about it again. The most reasonable place to cache the matrices may be inside `Tileset2D`, for both testability and API clarity (we plan to expose this class in the near future).

- the layer passes `modelMatrix` prop to `getTileIndices`
- the tile cache casts the model matrix to `Matrix4`, and compares it to the cached version using `matrix.equals`
- if changed, compute the inverse matrix and save both, similar to how the viewport is cached

Pessimistress:
Documentation wise, you can add an entry for `modelMatrix` to the layer doc and explain that it affects both tile fetching and rendering. The doc contains a sweeping statement "this layer inherits all base Layer props" but does list the props that have special behaviors, such as `data`.

modules/geo-layers/src/tile-layer/tileset-2d.js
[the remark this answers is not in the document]
ilan-gold: Ok, this makes sense - thanks!

modules/geo-layers/src/tile-layer/tile-layer.js
```diff
@@ -97,14 +97,13 @@ export default class TileLayer extends CompositeLayer {
...
-
     this._updateTileset();
   }
 
   _updateTileset() {
     const {tileset} = this.state;
-    const {onViewportLoad, zRange} = this.props;
-    const frameNumber = tileset.update(this.context.viewport, {zRange});
+    const {onViewportLoad, zRange, modelMatrix} = this.props;
+    const frameNumber = tileset.update(this.context.viewport, {zRange}, modelMatrix);
```
Pessimistress, line 106: Place `modelMatrix` inside named options (`{zRange, modelMatrix}`)

modules/geo-layers/src/tile-layer/utils.js
```diff
@@ -64,6 +88,16 @@ function getBoundingBox(viewport, zRange, extent) {
   } else {
     bounds = viewport.getBounds();
   }
+  if (!viewport.isGeospatial) {
+    return [
```
Pessimistress, line 92: Should this apply to all viewports?

ilan-gold: I didn't know enough about the geospatial coordinates so I sequestered it off in an attempt not to change anything from the previous way things worked.  My instinct to check `!viewport.isGeospatial` was that geospatial coordinates could "wrap around" where you have a "left" latitude greater than the "right" latitude of the unprojected viewport.  But I guess if you're using the `extent` prop this is not necessarily the case since `extent` defines a box to begin with.  I can play around a bit to get a feel.

ilan-gold: Sorry if I have that wrong too about how geospatial coordinates can work.

Pessimistress: Let's keep it as is for now. I can dig into it.

Pessimistress (approved):

modules/geo-layers/src/tile-layer/utils.js
ilan-gold: Thanks!

jeffpeck10x:
I know this is an old PR, but in case anyone is running into the issue of occasional 404s as documented above when rotating, a workaround is to pad the extent to mitigate the imprecision coming from `Matrix4.transformPoint`.

Something like this:
```
function padExtent(
  bbox: [number, number, number, number],
  padding = 0.1,
) {
  return [
    bbox[0] + padding,
    bbox[1] + padding,
    bbox[2] - padding,
    bbox[3] - padding,
  ];
}
```

And then in your TileLayer props:
```
  extent: padExtent([0, 0, width, height]),
```

I am not sure if there is any drawback, but of course you should choose a reasonable `padding` value that works for your application.

Thank you @ilan-gold for implementing this, hiccups and all. It works really nicely despite this minor issue.

That said, I was initially worried when I saw those 404s and so I am leaving this comment here in case anyone else runs into this and searches for information about it. It _does_ work, but those rounding errors may occasionally cause an attempt to fetch a tile that does not exist.

ilan-gold:
Ah, this is a great suggestion @jeffpeck10x, thanks! If you have any suggestions/ideas on how to improve the feature, feel free to make a PR!  I often "get things to work" in a way that is good but not ideal, so there might be an ideal solution out there I did not think of!

jeffpeck10x:
I poked around a little, but as far as I can tell the issue _seems_ to come from the result of `Matrix4.transformPoint` not always being what might be expected. For example, rotating 90 degrees occasionally returns a transform that contains long decimal, when you know that the actual value is just `0`. To be fair, I am not 100% sure that this is what the actual problem is, but it seems likely.

The only sort of general purpose solution that I can potentially think of would be to include an optional `marginOfErrorPadding` prop (which, there is probably a more succinct name for) to the TileLayer to take into account when calculating tiles. I imagine that such a prop could cause confusion for basic usage though. Still, if it is documented well, it could work.

Given that this was implemented over two years ago and there is no other discussion of it on the issue tracker, maybe there is not a lot of interest in rotating the TileLayer anyway 🤷 ?

I think that it suffices for now to have this workaround, but maybe @Pessimistress can comment if this ought to be formalized in some way.

Pessimistress:
Let's move the conversation to Discussions so it's more visible. Are you essentially asking for a method on `Matrix4` that performs rounding to a given precision?

jeffpeck10x:
@Pessimistress I am not sure if that is the best route, but ultimately would be a nice-to-have if the TileLayer could handle rotations without occasionally trying to load tiles that are beyond the given extent.

I _think_ that the problem stems from precision of the result from `Matrix4`, but I am not sure if that this the best place to try to fix this.

jeffpeck10x:
I can make a new discussion and reference this PR. Will do in a few minutes.

jeffpeck10x:
Discussion: https://github.com/visgl/deck.gl/discussions/7573

PR #5082, opened 2020-10-30
Fix a bug polygon fill layer updates when line color changes

heshan0131:
Polygon layer fill should not update when line color update triggers change

Screen Shot 2020-10-29 at 10 22 24 PM

Signed-off-by: Shan He <heshan0131@gmail.com>

### DIFF
--- a/modules/layers/src/polygon-layer/polygon-layer.js
+++ b/modules/layers/src/polygon-layer/polygon-layer.js
@@ -195,7 +195,7 @@ export default class PolygonLayer extends CompositeLayer {
             getFillColor: updateTriggers.getFillColor,
             // using a legacy API to invalid lineColor attributes
             // if (extruded && wireframe) has changed
-            lineColor: extruded && wireframe,
+            lineColors: extruded && wireframe,
             getLineColor: updateTriggers.getLineColor
           }
         }),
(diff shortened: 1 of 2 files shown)


Pessimistress:
Line color applies to the SolidPolygonLayer when `wireframe: true`.

heshan0131:
> Line color applies to the SolidPolygonLayer when `wireframe: true`.

When I use `PolygonLayer` and provide getLineColor, and its updateTriggers, polygon layer renders PathLayer, passing getLineColor -> getColor. Change of lineColor triggers should only updates the PathLayer, it should not cause the solid polygon layer to update...

Pessimistress:
That is not caused by updateTriggers, but the accessor itself. `getLineColor` is only used when `extruded: true, wireframe: true`, and it shouldn't be passed to `SolidPolygonLayer` otherwise. This bug affects both `GeoJsonLayer` and `PolygonLayer`.

Can you open an issue first?

heshan0131:
https://github.com/visgl/deck.gl/issues/5085

modules/layers/src/polygon-layer/polygon-layer.js
```diff
@@ -193,6 +193,9 @@ export default class PolygonLayer extends CompositeLayer {
             getPolygon: updateTriggers.getPolygon,
             getElevation: updateTriggers.getElevation,
             getFillColor: updateTriggers.getFillColor,
+            // using a legacy API to invalid lineColor attributes
+            // if (extruded && wireframe) has changed
+            lineColor: extruded && wireframe,
```
Pessimistress, line 198: `Boolean(extruded && wireframe)` just to be safe

Pessimistress (approved):

PR #5096, opened 2020-11-04
Add 'data' argument in HeatmapLayer documentation

Eo300:
Example did not use 'data' in layer declaration, which could cause confusion for newer users since no heatmap is rendered without the data.

For #

#### Change List
-


Pessimistress (approved):
Thank you!

PR #5099, opened 2020-11-05
Update LICENSE

Pessimistress:


ibgreen (approved):
I believe there can be multiple copyrights, if we wanted to keep Uber's name.

PR #5118, opened 2020-11-12
Remove double buffer overallocation

Pessimistress:
- When using the default attribute packing, `value` is already overallocated, and we should not overallocate again when creating the WebGL buffer.
- When using custom attribute packing (binary attribute, `instancePickingColors`), this will be less efficient if the data size increases frequently over time. It is arguably a small use case comparing to the memory savings.

#### Change List
- Be conservative when allocating WebGL buffer size


ibgreen (approved):

igorDykhta (approved):
Nice

PR #5126, opened 2020-11-13
Add a third party deckgl library for vue js components to readme

wsmoyer:


kylebarron:
In my opinion we should have a section of the docs for "Community-developed deck.gl packages" so that it's clear there's extra functionality but that it's clear it's not maintained by the core deck.gl team.

PR #5131, opened 2020-11-15
Better support multiple map tile providers in pydeck

ajduberstein:
For #5063

#### Background

This PR refactors the way users specify base maps in pydeck. It also moves pydeck from Mapbox to Carto as our default provider of map tiles.

image

#### Change List
- Refactor pydeck and @deck.gl/jupyter-widget to allow for multiple map providers.
- Provide cross-provider map style names
- Update tests to assume pydeck uses Carto
- Update Google Maps support to better support multiple providers
- Begin use of Python 3 typing; use Python 3.7 in CI tests


bindings/pydeck/docs/scripts/embed_examples.py
```diff
@@ -8,13 +8,7 @@
...
 import sys
 
-from const import (
-    DECKGL_URL_BASE,
-    EXAMPLE_GLOB,
-    GALLERY_DIR,
-    HTML_DIR,
-    HOSTED_STATIC_PATH,
-)
+from const import DECKGL_URL_BASE, EXAMPLE_GLOB, GALLERY_DIR, HTML_DIR, HOSTED_STATIC_PATH
```
kylebarron, line 11: Seems like there are some lint changes? Did the config change, e.g. max line length? If so might be cleaner as a separate PR.

bindings/pydeck/pydeck/bindings/deck.py
```diff
@@ -18,15 +17,14 @@ def __init__(
         layers=[],
         views=[View(type="MapView", controller=True)],
         map_style=DARK,
-        mapbox_key=None,
-        google_maps_key=None,
+        api_keys={},
```
kylebarron, line 20: Nit: Avoid mutable default arguments ([1](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments), [2](https://stackoverflow.com/questions/1132941/least-astonishment-and-the-mutable-default-argument)). This has bitten me recently. (applies to the `list` arguments above as well`)

bindings/pydeck/pydeck/bindings/providers.py
```diff
@@ -1,4 +1,20 @@
...
+    @classmethod
+    def as_list(cls):
+        return [getattr(cls, a) for a in dir(cls) if not a.startswith("__") and not callable(getattr(cls, a))]
+
+    @classmethod
+    def in_list_or_raise(cls, value):
+        if value not in cls.as_list():
+            raise PydeckException(
+                f"{value} is not a recognized API key provider. Choose one of the following: {cls.as_list()}"
+            )
```
kylebarron, line 20: I'd suggest subclassing an `Enum` instead, then you don't have to define these class methods...

```py
from enum import Enum
class Providers(str, Enum):
    """Basemap providers available in pydeck"""
    MAPBOX = "mapbox"
    GOOGLE_MAPS = "google_maps"
    CARTO = "carto"
```

Then to get the list of values you can just do 
```py
[x.value for x in Providers]
# ['mapbox', 'google_maps', 'carto']
[x.name for x in Providers]
# ['MAPBOX', 'GOOGLE_MAPS', 'CARTO']
```

bindings/pydeck/pydeck/bindings/view.py
```diff
@@ -15,7 +15,7 @@ class View(JSONMixin):
         If enabled, camera becomes interactive.
     """
 
-    def __init__(self, type=None, controller=None, width=None, height=None):
+    def __init__(self, type=None, controller=None, width=None, height=None, **kwargs):
```
kylebarron, line 18: Why allow `kwargs` if never used?

modules/jupyter-widget/src/playground/create-deck.js
```diff
@@ -141,11 +141,17 @@ function createStandaloneFromProvider({
         mapboxApiAccessToken: mapboxApiKey,
         onLoad: modifyMapboxElements
       });
+    case 'carto':
+      deck.log.info('Using Carto base maps')();
+      return new deck.DeckGL({
+        ...sharedProps,
+        ...props
+      });
```
kylebarron, line 149: Does this mean the carto API key is never used? Or is it just not explicit like `googleMapsKey` below?

kylebarron (approved):

bindings/pydeck/pydeck/bindings/view.py
ajduberstein: See update

bindings/pydeck/docs/scripts/embed_examples.py
ajduberstein: Yes, appears to have been done by mistake and seems like I mixed it in with another commit.

bindings/pydeck/pydeck/bindings/deck.py
ajduberstein: Interesting point--in over a decade of writing Python I've never thought about this.

modules/jupyter-widget/src/playground/create-deck.js
ajduberstein: It's just not explicit, at least at the moment. I'll focus on getting the carto layers integrated in a separate PR.

bindings/pydeck/pydeck/bindings/map_styles.py
```diff
@@ -1,4 +1,60 @@
...
+DARK = "dark"
+LIGHT = "light"
+SATELLITE = "satellite"
+ROAD = "road"
+DARK_NO_LABELS = "dark_no_labels"
+LIGHT_NO_LABELS = "light_no_labels"
+
+MAPBOX_LIGHT = "mapbox://styles/mapbox/light-v9"
+MAPBOX_DARK = "mapbox://styles/mapbox/dark-v9"
+MAPBOX_DARK_NO_LABELS = "https://rivulet-zhang.github.io/dataRepo/mapbox/style/map-style-dark-v9-no-labels.json"
```
Pessimistress, line 13: Why are you depending on a style hosted on someone else's site? What if they delete it?

bindings/pydeck/pydeck/bindings/map_styles.py
```diff
@@ -1,4 +1,67 @@
...
+DARK = "dark"
+LIGHT = "light"
+SATELLITE = "satellite"
+ROAD = "road"
+DARK_NO_LABELS = "dark_no_labels"
+LIGHT_NO_LABELS = "light_no_labels"
+
+MAPBOX_LIGHT = "mapbox://styles/mapbox/light-v9"
+MAPBOX_DARK = "mapbox://styles/mapbox/dark-v9"
+MAPBOX_DARK_NO_LABELS = (
```
Pessimistress, line 13: Can you move this? We should not be maintaining non-official map styles.

PR #5148, opened 2020-11-18
Fix using-layers.md broken catalog link

rkhayat:
#### Change List
- `https://deck.gl/docs/api-reference/layers/README.md` is broken, replacing with `https://deck.gl/docs/api-reference/layers`


kylebarron (approved):

Pessimistress:
This seems to be a bug in our Gatsby template. The link works fine on GitHub.

Pessimistress:
@rkhayat Thanks for bringing this to our attention. A proper fix is landed in the Gatsby template and applied by #5204.

Issue #5121, opened 2020-11-13
Passing local url of texture/height to terrain layer doesn't work


danielXr:
#### Description
I'm trying to load the TerrainLayer with locally VTs (taken from the example in deck.gl site, png, file:///the_location_in_my_pc.png)
When providing the original url of the same file, it works like a charm.
#### Repro Steps
This works:
```
import pydeck as pdk
import os

# Define how to parse elevation tiles
ELEVATION_DECODER = {"rScaler": 2, "gScaler": 0, "bScaler": 0, "offset": 0}
elevation_data = '"https://raw.githubusercontent.com/visgl/deck.gl-data/master/website/terrain.png"'
texture_data = '"https://raw.githubusercontent.com/visgl/deck.gl-data/master/website/terrain-mask.png"'
# elevation_data = '"file:///Users/daniel/PycharmProjects/mwi/tmp_files/terrain_sample_img/terrain.png"'
# texture_data = '"file:///Users/daniel/PycharmProjects/mwi/tmp_files/terrain_sample_img/terrain-mask.png"'
bounds = [-122.5233, 37.6493, -122.3566, 37.8159]
    


terrain_layer = pdk.Layer(
    "TerrainLayer", data=None, 
    elevation_decoder=ELEVATION_DECODER, 
    texture=texture_data, 
    elevation_data=elevation_data, 
#     elevation_decoder=el_decoder, 
#     texture=texture_img, 
#     elevation_data=heights_img, 
    bounds=bounds
)

view_state = pdk.ViewState(latitude=37.6493, longitude=-122.5233, zoom=11.5, bearing=140, pitch=60)

r = pdk.Deck(terrain_layer, initial_view_state=view_state, map_style='')


r.show()
```
Download the images and give it a try - doesn't work:
```
import pydeck as pdk
import os

# Define how to parse elevation tiles
ELEVATION_DECODER = {"rScaler": 2, "gScaler": 0, "bScaler": 0, "offset": 0}
#elevation_data = '"https://raw.githubusercontent.com/visgl/deck.gl-data/master/website/terrain.png"'
#texture_data = '"https://raw.githubusercontent.com/visgl/deck.gl-data/master/website/terrain-mask.png"'
elevation_data = '"file:///Users/daniel/PycharmProjects/mwi/tmp_files/terrain_sample_img/terrain.png"'
texture_data = '"file:///Users/daniel/PycharmProjects/mwi/tmp_files/terrain_sample_img/terrain-mask.png"'
[...]

PR #5162, opened 2020-11-22
pydeck: Handle quoting of image URLs or base64-conversion automatically

ajduberstein:
For #5063 

#### Background

We've had a number of issues regarding images from local disk vs from a URL e.g. https://github.com/visgl/deck.gl/issues/5121. This PR should solve that and also get rid of the "'quote-in-quote'" syntax prevalent in pydeck.

#### Change List
- Add a helper class within pydeck for strings and images
- Infer on layer creation whether a particular layer argument is a local image, a remote image, or not an image at all.
- Remove the requirement that data is an argument
- Update examples
- Add tests


bindings/pydeck/pydeck/types/image.py
```diff
@@ -0,0 +1,45 @@
...
+    r"\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})"
+    r"(?::\d+)?"
+    r"(?:/?|[/?]\S+)$",
+    re.IGNORECASE,
+)
+
+valid_image_regex = re.compile(r".(gif|jpe?g|tiff?|png|webp|bmp)$", re.IGNORECASE,)
+
+
+ENCODING_PREFIX = "data:image/png;base64,"
```
Pessimistress, line 20: MIME type should be based on file extension

bindings/pydeck/pydeck/types/image.py
```diff
@@ -14,10 +15,15 @@
...
-valid_image_regex = re.compile(r".(gif|jpe?g|tiff?|png|webp|bmp)$", re.IGNORECASE,)
+valid_image_regex = re.compile(
+    r".(gif|jpe?g|tiff?|png|webp|bmp)$",
+    re.IGNORECASE,
+)
 
 
-ENCODING_PREFIX = "data:image/png;base64,"
+def get_encoding(path: str) -> str:
+    extension = pathlib.Path(path).suffix
```
kylebarron: So glad we're on Python 3 now 😉

kylebarron (approved):

PR #5209, opened 2020-12-02
Simplify FontAtlasManager

Pessimistress:
FontAtlasManager was not initializing its texture correctly; instead it was relying on the IconLayer to override the texture parameters:

https://github.com/visgl/deck.gl/blob/8.3-release/modules/layers/src/icon-layer/icon-manager.js#L262

This was broken by #5197, resulting in aliased fonts:

image

After this PR, the TextLayer no longer manages its own texture. It supplies the font atlas canvas directly to its sublayer, which is then converted to a texture via the new prop transform API.

#### Change List

- Remove texture creation from FontAtlasManager

### DIFF
--- a/modules/layers/src/text-layer/font-atlas-manager.js
+++ b/modules/layers/src/text-layer/font-atlas-manager.js
@@ -1,6 +1,4 @@
 /* global document */
-
-import {Texture2D} from '@luma.gl/core';
 import TinySDF from '@mapbox/tiny-sdf';
 
 import {buildMapping} from './utils';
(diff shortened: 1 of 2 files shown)


ibgreen (approved):
Nice decoupling! 

Side note: Whether it is dependent on textures or not, I think the font atlas is very generic and would make an awesome `@luma.gl/text` module. Should take a look at some point and see if there are any tricky dependencies...

kylebarron (approved):

PR #5210, opened 2020-12-02
Fix MVT autohighlight feature when no featureId or uniqueProp present

AdriSolid:
#### Background

Currently `MVTLayer` highlights incorrect features if `pickable` `autoHighlight` are set to true, and data comes without `feature.id` and `uniqueIdProperty` is not set.

In this specific case seems better not to highlight features.

#### Expected Behavior
featureNull

#### Actual Behavior
featureMinus1

#### Change List

- Return `null` in => [getFeatureUniqueId fn](https://github.com/visgl/deck.gl/blob/0407f01ebfe69544d874674d0d043ad083d5a766/modules/geo-layers/src/mvt-layer/mvt-layer.js#L223)

Cheers!


modules/geo-layers/src/mvt-layer/mvt-layer.js
```diff
@@ -229,7 +229,7 @@ function getFeatureUniqueId(feature, uniqueIdProperty) {
     return feature.id;
   }
 
-  return -1;
+  return null;
```
Pessimistress, line 232: What is the difference between -1 and null?

AdriSolid: At the end `highlightedObjectIndex` comes [here](https://github.com/visgl/deck.gl/blob/0407f01ebfe69544d874674d0d043ad083d5a766/modules/core/src/lib/layer.js#L865) as:

0 if `featureId` is set to -1
-1 if `featureId` is set to null

right?

Pessimistress: `highlightedObjectId` is calculated by `getHighlightedObjectIndex`. The return value of this function is irrelevant.

AdriSolid: Must be deeper then.. Setting it to null seems that solve the issue with that casuistry :confused:

AdriSolid: With the current approach each tile get a `highlightedObjectIndex` = 0, [here in TileLayer](https://github.com/visgl/deck.gl/blob/177bf02088f62e1809710b20a6fbf113192c01a4/modules/geo-layers/src/tile-layer/tile-layer.js#L198)

Same example with tile boundaries:
gifFlick

Returning null on `getHighlightedObjectIndex` method sets `highlightedObjectIndex` to -1, avoiding to highlight features if no `feaure.id` or `uniqueIdProperty` is set, how you see this? Thanks!

alasarr: I prefer to keep getFeatureUniqueId returns -1, because it's more consistent with the rest of the architecture. highlightedObjectIndex also returns -1 when not found.

What about this at: https://github.com/visgl/deck.gl/blob/master/modules/geo-layers/src/mvt-layer/mvt-layer.js#L146
```javascript
if (hoveredFeatureId !== newHoveredFeatureId && newHoveredFeatureId!=-1 ) {
        this.setState({hoveredFeatureId: newHoveredFeatureId});
      }
```

alasarr (changes requested):

modules/geo-layers/src/mvt-layer/mvt-layer.js
alasarr: I've not tested, but I think it could work

AdriSolid: LGTM, changes applied @alasarr

alasarr (approved):

PR #5212, opened 2020-12-03
Support plain object as image source

Pessimistress:
`image` prop type supports an additional format: plain object (see documentation).

- Fixes an issue in TextLayer where the sub layer does not update if the font atlas canvas is reused
- Provides a way to further customize texture generation, e.g. format, mipmaps
- This allows pydeck users to directly pass pixel data without having to encode it into an image. @ajduberstein 

#### Change List
- image prop transform supports plain object as input
- FontAtlasManager texture format
- Docs

### DIFF
--- a/docs/api-reference/layers/bitmap-layer.md
+++ b/docs/api-reference/layers/bitmap-layer.md
@@ -55,11 +55,18 @@ new deck.BitmapLayer({});
 
 ### Data
 
-##### `image` (String|Texture2D|Image|ImageData|HTMLCanvasElement|HTMLVideoElement|ImageBitmap)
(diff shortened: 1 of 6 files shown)


kylebarron (commented):
👍 I've been using such objects defining a `Texture2D` instance for a while with my raster work (https://github.com/kylebarron/deck.gl-raster/blob/dcd32a29432355ab0422eaf5ae8cdf32c978e33c/src/deckgl/images.js#L81-L84).

It can be quite handy when you want to fully customize texture parameters but keep the `gl` context handle isolated.

kylebarron (approved):

PR #5223, opened 2020-12-04
Improve Solid Polygon layer - basic winding order check and winding order enforcement.

igorDykhta:
#### Background
Extruded `SolidPolygonLayer` can produce a lot of overdraw. Optional back face culling looks like a good idea. Geojson has some requirements but they aren't followed well (A linear ring MUST follow the right-hand rule with respect to the area it bounds, etc.).

The following change checks for winding order of each polygon in solid-polygon-layer/polygon, and changes winding order to the desired one, with which back face culling properly hides back faces of extruded polygons. I don't see any side effects as I understand that culling wasn't used for this layer before. Also, I don't think that it breaks rendering with transparency, as right now it is not consistent, some triangles are rendered, some not - based on order of triangles in geometry.

Draw call speed improvements: 5%, 28%, 42%, +++, gives a lot of boost when there is more of overdraw.

The following check for winding order (which uses @math.gl/polygon) adds around ~190ms of overhead to 225ms used for geometry "normalization". Slightly changed implementation which operates directly on the array of positions adds only ~50ms of overhead to 225ms of execution time (added in the comment, but not sure where it can be added).

#### Change List
-  `ensureWindingOrder` checks for winding order of a closed ring, reverses the ring to proper winding order


igorDykhta:
Implementation without math.gl/polygon, which is much faster:

`import { equals } from '@math.gl/core';

function areaCallback(p1x, p1y, p2x, p2y){
  return (p1x + p2x) * (p1y - p2y);
}

function getWindingDirection(points, start, end, size){
  const numPoints = (end - start) / size;

  const isClosed = equals(points[start], points[start + (numPoints - 1) * size]) && 
    equals(points[start+1], points[start + (numPoints - 1) * size + 1]);

  let area = 0;
  for(let i = 0; i < numPoints - 1; i++){
    area += areaCallback(
      points[start + i * size], 
      points[start + i * size + 1], 
      points[start + (i+1)*size], 
      points[start + (i+1)*size + 1]);
  }

  if(!isClosed){
    area += areaCallback(
      points[start + (numPoints-1)*size], 
      points[start + (numPoints-1)*size + 1],
      points[start], 
      points[start + 1]);
  }

  return Math.sign(area/2);
}

function ensureWindingOrder(target, start, end, size) {

  const newSignedArea = getWindingDirection(target, start, end, size);
  if(newSignedArea < 0){
    const numPoints = (end - start) / size;
    const numSwaps = Math.floor(numPoints / 2);
    for(var i=0; i<numSwaps; i++){
      const b1 = start + i*size;
      const b2 = start + (numPoints - 1 - i)*size;
      for(var j=0; j<size; j++){
        const tmp = target[b1+j];
        target[b1+j] = target[b2+j];
        target[b2+j] = tmp;
      }
    }
  }
}`

Pessimistress:
I suggest we generalize the utilities provided by `@math.gl/polygon`:

- should not require constructing a class instance to use
- should work with flat position arrays 

Since it's called after ensuring the polygon is closed, it can skip the closing check.

kylebarron:
> Implementation without math.gl/polygon, which is x4 faster, but duplicates existing algorithm from math.gl/polygon

Why is this 4x faster? Just because it uses a `TypedArray` without creating many small array objects to make the coordinates geojson-like? It seems ideal to make `@math.gl/polygon` support `TypedArray` input?

Along the same lines, why not add `ensureWindingOrder` to `@math.gl/polygon`?

igorDykhta:
> > Implementation without math.gl/polygon, which is x4 faster, but duplicates existing algorithm from math.gl/polygon
> 
> Why is this 4x faster? Just because it uses a `TypedArray` without creating many small array objects to make the coordinates geojson-like? It seems ideal to make `@math.gl/polygon` support `TypedArray` input?
> 
> Along the same lines, why not add `ensureWindingOrder` to `@math.gl/polygon`?

Yes, the first solution with @math.gl/polygon is only slower because of additional polygon allocations & construction of input.

modules/layers/src/solid-polygon-layer/polygon.js
```diff
@@ -49,6 +50,30 @@ function isSimple(polygon) {
...
+function ensureWindingOrder(positions, start, end, size) {
+  const numPositions = (end - start) / size;
+  const points = [];
+  for (let i = 0; i < numPositions; i++) {
+    points[i] = [positions[i * size], positions[i * size + 1]];
+  }
+
+  const polygon = new Polygon(points);
+
+  if (polygon.getWindingDirection() < 0) {
```
Pessimistress, line 69: This is incorrect. The holes need to wind in the opposite direction from the outer bounds.

igorDykhta: But is it required for earcut or rendering?

Pessimistress: Not sure about earcut, but definitely rendering. You can test your solution locally with a polygon that contains holes.

Pessimistress: Btw that is a requirement from the GeoJSON spec and followed by all conforming formats e.g. MVT

igorDykhta: Correct, missed this

ibgreen:
Nice findings. Agree it would be awesome to improve in math.gl. We could also add [benchmark tests[(https://math.gl/examples/benchmarks) in math.gl for this function, happy to show how that works if not obvious.

modules/layers/src/solid-polygon-layer/polygon.js
```diff
@@ -165,7 +196,8 @@ export function normalize(polygon, positionSize) {
           srcPositions,
           positionSize,
           srcHoleIndices[i - 1],
-          srcHoleIndices[i]
+          srcHoleIndices[i],
+          i === 0 ? WINDING.CLOCKWISE : WINDING.COUNTER_CLOCKWISE
```
Pessimistress, line 200: > A linear ring MUST follow the right-hand rule with respect to the area it bounds, i.e., exterior rings are counterclockwise, and holes are clockwise.

https://tools.ietf.org/html/rfc7946#section-3.1.6

ibgreen: I suppose that GeoJSON input data must be organized that way, but here are no rules for how we organize data internally for optimal rendering?

@igorDykhta You indicated that following the GeoJSON standard leads to non-optimal results?

igorDykhta: I just selected WINDING.CLOCKWISE as front as it meant fewer changes on our side. There is no difference between cw & ccw. CCW is the default front in OpenGL and geojson, CW is considered the default by ESRI & Shapefile. 
I'll swap CW <> CCW, and add a note to the upgrade guide.

Pessimistress: It matters because our calculation of normals (and inherently lighting) calculation is dependent on this order. You need to verify that lighting is correct with this change.

igorDykhta: Looks like the lighting is consistent when CW is used for outer polygon, CCW for hole polygons.
outer-as-CCW
outer-as-CW

Pessimistress (approved):

Pessimistress:
Add an entry to update guide about potentially breaking visual change. (Technically a bug fix)

Pessimistress:
Some follow up thoughts:

- I wonder if there is noticeable perf difference in enforcing one winding direction vs another. The majority of the use case is GeoJSON (and we document the layer as accepting data that conforms to the GeoJSON spec), so we should optimize for this use case. Should we decide to do so, the normal calculation can be inserted in the vertex shader.
- There is also the use case when using binary data and `_normalize: false` the winding order check will be completely bypassed. I suppose in its current state this PR does not change the visual outcome for this case and that's a good thing.
- We recently discovered a catch with the MVTLayer, which uses a modelMatrix to flip Y. It affects the face orientation in SolidPolygonLayer. Polygons retrieved from MVT are CCW. Which means using binary and non-binary loader options will yield different lighting results (I need to verify this).
- It might worth adding a layer prop to control the winding order as a fallback.

ibgreen:
Just a thought: Would it help to add winding order normalization to loaders.gl loaders?

Pessimistress:
> Would it help to add winding order normalization to loaders.gl loaders?

There are three places where drawing order matters:

1. The top: front/back enforced by `earcut`
2. The sides: front/back dependent on winding direction
3. Lighting of the sides: normal calculation done in vertex shader, dependent on winding direction

Changing the default winding direction affects 2 & 3 if using GeoJSON input, and affects only 3 if using binary input. The MVTLayer use case (flipped Y space) affects all three. We should be able to eliminate the MVTLayer problem in the MVTLoader by doing tesselation during parsing, but the issue remains when no loaders are used.

PR #5251, opened 2020-12-10
Add style functions - @deck.gl/carto 

AdriSolid:
For #5222

- The goal is have a way to style layers based on functions @deck.gl/carto.

#### Change List
- Add style functions:
 - `colorBins`
 - `colorCategories`
 - `colorContinuous`

#### TODO
- [x] Update documentation
- [x] Add tests

cc @alasarr


examples/playground/json-examples/carto-bq-tiler.json
```diff
@@ -18,7 +18,12 @@
     {
       "@@type": "CartoBQTilerLayer",
       "data": "cartobq.maps.nyc_taxi_points_demo_id",
-      "getFillColor": [232, 133, 113],
+      "getFillColor": {
+        "@@function": "colorBins",
+        "prop": "properties.avg_fare_amount",
+        "breaks": [0, 10, 25, 50, 100],
```
alasarr, line 24: Let's rename breaks by `domain`

modules/carto/src/style/utils.js
```diff
@@ -0,0 +1,80 @@
+import * as cartoColors from 'cartocolor';
```
alasarr, line 1: Let's rename this file to palette

AdriSolid: Great, and, much better if we create a new folders called `utils`? So => `style/utils/palette.js`

modules/carto/src/style/color-bins-style.js
```diff
@@ -0,0 +1,18 @@
...
+
+export default function colorBins({breaks, colors, nullColor = NULL_COLOR}) {
+  if (Array.isArray(breaks)) {
+    const palette = typeof colors === 'string' ? getPalette(colors, breaks.length + 1) : colors;
+
+    const color = scaleThreshold()
+      .domain(breaks)
+      .range(palette);
+
+    return d => {
```
alasarr, line 12: How about this: 
```javascript
return d => Number.isFinite(d) ? color(d) : nullColor
```

modules/carto/src/style/color-bins-style.js
```diff
@@ -0,0 +1,18 @@
+import {scaleThreshold} from 'd3-scale';
+import {getPalette, NULL_COLOR} from './utils';
+
+export default function colorBins({breaks, colors, nullColor = NULL_COLOR}) {
```
alasarr, line 4: Rename breaks to domain

modules/carto/src/style/color-bins-style.js
```diff
@@ -0,0 +1,18 @@
...
+  if (Array.isArray(breaks)) {
```
alasarr, line 5: I don't understand this if. In which case breaks it's not an array?  Cannot we just remove this if?

alasarr (changes requested):

modules/carto/src/style/color-bins-style.js
AdriSolid: Let's say the user incorrectly passes `breaks` as other type different of `Array`, we are not actually handling errors in that case, just returning `NULL_COLOR`

modules/carto/src/style/color-bins-style.js
AdriSolid: LGTM

modules/carto/src/style/color-bins-style.js
AdriSolid: I would also propose `if (Array.isArray(breaks) && breaks.every(Number.isFinite))`

alasarr (changes requested):
Some changes requested, we also need:

- A new doc section for style helpers at carto module. You can get some info from here: https://carto.com/developers/cartoframes/reference/#heading-Viz

- Test

modules/carto/src/style/utils.js
```diff
@@ -0,0 +1,7 @@
+export function getAttrValue(attr, d) {
+  if (typeof attr === 'function') {
+    return attr(d);
+  }
+
+  return d.properties[attr];
```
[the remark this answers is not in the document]
AdriSolid: Type checking added

test/modules/carto/style/carto-color-bins.spec.js
```diff
@@ -0,0 +1,47 @@
+import test from 'tape-catch';
```
alasarr, line 1: WOuld be great to add a test with carto colors

alasarr (approved):
Looks good to me. Let's wait to merge https://github.com/visgl/deck.gl/pull/5250

docs/api-reference/carto/styles.md
```diff
@@ -0,0 +1,129 @@
...
+
+const colorBinsManual = colorBins({
+  attr: 'gn_pop',
+  domain: [1e5, 2e5, 3e5],
+  colors: 'Teal'
+});
+
+const binsLayer = new CartoSQLLayer({
+  data: 'SELECT the_geom_webmercator, gn_pop FROM populated_places',
+  getFillColor: row => colorBinsManual(row)
```
alasarr, line 26: I don't see why we add another function.

Cannot be something like this? 

```javascript
const binsLayer = new CartoSQLLayer({
  data: 'SELECT the_geom_webmercator, gn_pop FROM populated_places',
  getFillColor: colorBins({})
})
```

docs/api-reference/carto/styles.md
```diff
@@ -0,0 +1,129 @@
+# Styles
+
+`CARTO styles` are a set of scale functions to easily assign a color to a feature.
```
alasarr, line 3:
```suggestion
Helper functions to create data-driven map visualizations.
```

docs/api-reference/carto/styles.md
```diff
@@ -0,0 +1,129 @@
...
+
+Transform data values into visual variables and take the advantage of [CARTOColors](https://carto.com/carto-colors/) custom color schemes.
```
alasarr, line 5:
```suggestion
Take the advantage of [CARTOColors](https://carto.com/carto-colors/), custom color schemes built on top of well-known standards for color use on maps, with next generation enhancements for the web and CARTO basemaps.

```

docs/api-reference/carto/styles.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Color Bins
+
+`colorBins` maps continuous numeric input to discrete values defined by the range (`colors`).
+
+* Quantitative scales have a continuous domain.
+
```
alasarr, line 12: Need to explain each parameter, here.  The structure should be:

- Description
- Parameters
- Usage example

alasarr (changes requested):

Pessimistress:
The test coverage of this new code is only 67%. It is below the current coverage of this repo. Since it's pure JavaScript functions you should be able to get close to 100%.

alasarr (approved):

modules/carto/src/style/palette.js
```diff
@@ -0,0 +1,83 @@
...
+  const longestPaletteIndex = Math.max(...palettesColorVariants);
+  const smallestPaletteIndex = Math.min(...palettesColorVariants);
+
+  if (!Number.isInteger(numCategories) || numCategories > longestPaletteIndex) {
+    paletteIndex = longestPaletteIndex;
+  } else if (numCategories < smallestPaletteIndex) {
+    paletteIndex = smallestPaletteIndex;
+  }
+
+  let colors = palette[paletteIndex];
```
Jesus89, line 27: This could cause a bug because the palette is being updated (color.slice). Consider cloning the object `[ ...palette[paletteIndex] ]`.

AdriSolid: :wave:
`getPalette` is called each time we need an update, so there is no problem as long as `colors` is being created again

PR #5255, opened 2020-12-14
pydeck: Add Carto layer examples and update example base maps

ajduberstein:
For #5063

#### Change List
- Update base map in layer examples
- Add Carto examples
- Add instructions for adding new layers to the pydeck docs


kylebarron (approved):

alasarr (approved):
🚀

PR #5256, opened 2020-12-14
pydeck: Fix local image bug; add abstract base class for pydeck types

ajduberstein:
Add new types to examples and add abstract base class supporting the new types

#### Background

`String` forces pydeck to treat input as a string literal rather than a JS expression
`Image` takes input that resembles an image's file path. If the image is found locally, pydeck converts the image to a binary representation before passing it to @deck.gl/json.

This PR fixes a bug which made neither of those types were being honored. pydeck works by converting Python objects to JSON, and then passing that JSON to @deck.gl/json. This PR interrupts the JSON conversion--if an object is detected that's a subclass of `base.PydeckType`, that item will be converted to its string representation.

#### Change List
- Fix local image bug
- Add PydeckType abstract base class
- Add local image JSON conversion test


kylebarron (approved):
For examples it might be nice to have `from pydeck.types import String`

bindings/pydeck/pydeck/io/templates/index.j2
```diff
@@ -5,7 +5,9 @@
 {% if google_maps_key %}
     <script src="https://maps.googleapis.com/maps/api/js?key={{google_maps_key}}&libraries=places"></script>
 {% endif %}
+{% if mapbox_key %}
```
Pessimistress, line 8: - latest Mapbox version is 1.13.0
- Does this mean that you can't use the Carto map styles (no key required)?

ajduberstein: Defaults to Carto if no key is set.
I'll update the version.

kylebarron: I think the point was: are you still including the `mapbox-gl.js` script when a mapbox key is not set?

PR #5261, opened 2020-12-17
MVTLayer coordinates transformation to WGS84

AdriSolid:
`MVTLayer` coordinates transformation to WGS84.

#### Background

Users usually think that something is wrong when trying to extract `object.coordinates` from `onClick/onHover` events in `MVTLayer`. Returned values are relative (0-1) coordinates in the tile, for better performance.

This PR includes a coordinates transformation to WGS84 latitude/longitude for `MVTLayer`.

#### Change List
- Add coordinates transformation function in `MVTLayer` for `onClick/onHover` events
- Link `MVTLayer` `onClick` method in `carto-layer`
- Unit tests

### DIFF
--- a/modules/geo-layers/src/mvt-layer/coordinate-transform.js
+++ b/modules/geo-layers/src/mvt-layer/coordinate-transform.js
@@ -9,11 +9,9 @@ const availableTransformations = {
   MultiPolygon
 };
 
-function Point([pointX, pointY], bbox, viewport) {
-  const [minX, minY] = viewport.projectFlat([bbox.west, bbox.north]);
-  const [maxX, maxY] = viewport.projectFlat([bbox.east, bbox.south]);
-  const x = lerp(minX, maxX, pointX);
-  const y = lerp(minY, maxY, pointY);
+function Point([pointX, pointY], [nw, se], viewport) {
+  const x = lerp(nw[0], se[0], pointX);
+  const y = lerp(nw[1], se[1], pointY);
 
   return viewport.unprojectFlat([x, y]);
 }
@@ -43,8 +41,16 @@ function MultiPolygon(multiPolygon, bbox, viewport) {
 }
 
 export function transform(geometry, bbox, viewport) {
+  const nw = viewport.projectFlat([bbox.west, bbox.north]);
+  const se = viewport.projectFlat([bbox.east, bbox.south]);
+  const projectedBbox = [nw, se];
+
   return {
     ...geometry,
-    coordinates: availableTransformations[geometry.type](geometry.coordinates, bbox, viewport)
+    coordinates: availableTransformations[geometry.type](
+      geometry.coordinates,
+      projectedBbox,
+      viewport
+    )
   };
 }
--- a/modules/geo-layers/src/mvt-layer/coordinate-transform.js
+++ b/modules/geo-layers/src/mvt-layer/coordinate-transform.js
@@ -1,3 +1,5 @@
+import {lerp} from 'math.gl';
+
 const availableTransformations = {
   Point,
   MultiPoint,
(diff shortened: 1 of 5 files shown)


modules/geo-layers/src/mvt-layer/mvt-layer.js
```diff
@@ -148,9 +150,23 @@ export default class MVTLayer extends TileLayer {
       }
     }
 
+    if (object) {
+      info = transformTileCoordsToWGS84(info, object, tile);
```
Pessimistress, line 154: This is not the right place to handle this. Use `getPickInfo`.

modules/geo-layers/src/mvt-layer/coordinate-transform.js
```diff
@@ -0,0 +1,52 @@
...
+  const y2 = 180 - ((pointY + originY) * 360) / zoomLevelSize;
+
+  return [
+    ((pointX + originX) * 360) / zoomLevelSize - 180,
+    (360 / Math.PI) * Math.atan(Math.exp((y2 * Math.PI) / 180)) - 90
+  ];
+}
+
+function getPoints(geometry, tile) {
+  return geometry.map(g => availableTransformations.Point(g, tile));
```
Pessimistress, line 24: Any reason to reference the wrapper object instead of calling the function directly?

modules/geo-layers/src/mvt-layer/coordinate-transform.js
```diff
@@ -0,0 +1,52 @@
...
+};
+
+function Point([pointX, pointY], {x, y, z}) {
+  const originX = x;
+  const originY = y;
+  const zoomLevelSize = Math.pow(2, z);
+
+  const y2 = 180 - ((pointY + originY) * 360) / zoomLevelSize;
+
+  return [
```
Pessimistress, line 17: Use `worldToLngLat` from `@math.gl/web-mercator`

modules/geo-layers/src/mvt-layer/mvt-layer.js
```diff
@@ -236,5 +252,15 @@ function isFeatureIdDefined(value) {
   return value !== undefined && value !== null && value !== '';
 }
 
+function transformTileCoordsToWGS84(info, object, tile) {
```
Pessimistress, line 255: You need to check if the coords are already WGS84. MVTLayer does not always load features in tile coordinates.

modules/geo-layers/src/mvt-layer/mvt-layer.js
```diff
@@ -236,5 +252,15 @@ function isFeatureIdDefined(value) {
...
+  return {
```
Pessimistress, line 256: This is potentially a very expensive operation, especially if you perform it on hover. Consider offering a callback that users can use to transform the coordinates only if they need it, similar to how query viewport features is implemented.

Pessimistress: Alternatively, you can delay the transformation to when `coordinates` is accessed:

```js
const feature = {properties, geometry: {type}};
Object.defineProperty(feature.geometry, 'coordinates', {
  get: () => {
    return transform(coordinates, tile)
  }
});
```

modules/geo-layers/src/mvt-layer/mvt-layer.js
AdriSolid: With `context.viewport.resolution` right?

modules/geo-layers/src/mvt-layer/coordinate-transform.js
AdriSolid: It works as an enum, also to dynamically call the functions

modules/geo-layers/src/mvt-layer/coordinate-transform.js
AdriSolid: Used another approach getting the bounding box of the tile. With `viewport.projectFlat` and `viewport.unprojectFlat`

modules/geo-layers/src/mvt-layer/coordinate-transform.js
AdriSolid: But you are right, there is not necessary :+1:

modules/geo-layers/src/mvt-layer/coordinate-transform.js
```diff
@@ -0,0 +1,50 @@
...
+  LineString,
+  MultiLineString,
+  Polygon,
+  MultiPolygon
+};
+
+function Point([pointX, pointY], bbox, viewport) {
+  const [minX, minY] = viewport.projectFlat([bbox.west, bbox.north]);
+  const [maxX, maxY] = viewport.projectFlat([bbox.east, bbox.south]);
+  const x = lerp(minX, maxX, pointX);
```
Pessimistress, line 15: These are the same values for the same tile. You are repeating the calculation for every point in the geometry.

modules/geo-layers/src/mvt-layer/mvt-layer.js
```diff
@@ -236,5 +250,19 @@ function isFeatureIdDefined(value) {
   return value !== undefined && value !== null && value !== '';
 }
 
+function transformTileCoordsToWGS84({object, tile}, viewport) {
+  const {properties, geometry} = object;
+  const feature = {properties, geometry: {type: geometry.type}};
```
Pessimistress, line 255: Removing featureId? Doesn't this break autoHighlight?

test/modules/geo-layers/mvt-layer.spec.js
```diff
@@ -63,6 +129,23 @@ test('ClipExtension', t => {
...
+
+  for (const tc of TRANSFORM_COORDS_DATA) {
+    const func = transform(tc.geom, bbox, viewport);
+    t.deepEqual(func, tc.result, `transform ${tc.geom.type} returned expected WGS84 coordinates`);
+  }
+
+  t.end();
+});
+
 test.skip('MVT Highlight', t => {
```
Pessimistress, line 66: Why was this test skipped?

modules/geo-layers/src/mvt-layer/mvt-layer.js
```diff
@@ -151,6 +152,19 @@ export default class MVTLayer extends TileLayer {
...
   }
 
+  getPickingInfo({info, sourceLayer}) {
+    info.sourceLayer = sourceLayer;
+    info.tile = sourceLayer.props.tile;
+
+    const isWGS84 = this.context.viewport.resolution;
+
+    if (!isWGS84 && info.object) {
+      info = {...info, object: transformTileCoordsToWGS84(info, this.context.viewport)};
```
Pessimistress, line 162: Just pass `object` and `tile` separately to `transformTileCoordsToWGS84` so that you don't have to clone the info object.

test/modules/geo-layers/mvt-layer.spec.js
AdriSolid: Just fixed!

modules/geo-layers/src/mvt-layer/mvt-layer.js
```diff
@@ -151,6 +152,17 @@ export default class MVTLayer extends TileLayer {
     return super.onHover(info, pickingEvent);
   }
 
+  getPickingInfo({info, sourceLayer}) {
+    const isWGS84 = this.context.viewport.resolution;
+
+    if (!isWGS84 && info.object) {
+      info.tile = sourceLayer.props.tile;
```
Pessimistress, line 159: `tile` must be populated regardless of what coordinate system is used

Pessimistress:
This is a breaking change. You need to add an entry to upgrade guide.

modules/geo-layers/src/mvt-layer/coordinate-transform.js
```diff
@@ -0,0 +1,56 @@
+import {lerp} from 'math.gl';
+
+const availableTransformations = {
```
Pessimistress, line 3: I do not believe the MVTLoader ever generates Multi* features:

https://github.com/visgl/loaders.gl/blob/9a62f53206db982858a42bdfde4eb1ea9606aadf/modules/mvt/src/lib/transform-to-local-range.js#L1

modules/geo-layers/src/mvt-layer/mvt-layer.js
Pessimistress: You should call `super.getPickingInfo` to avoid duplicating TileLayer code.

modules/geo-layers/src/mvt-layer/coordinate-transform.js
AdriSolid: We are rendering Multi* features.. :roll_eyes:
https://github.com/visgl/loaders.gl/blob/9a62f53206db982858a42bdfde4eb1ea9606aadf/modules/mvt/src/lib/transform-to-local-range.js#L61

Pessimistress (approved):

PR #5296, opened 2021-01-06
Fix geo-layers module peer dependencies in documentation

Pessimistress:
For #5294

#### Change List
- Document correct peer dependencies for geo-layers module


kylebarron (approved):

kylebarron:
Might not be a bad idea for the linter to check for trailing whitespace

PR #5316, opened 2021-01-11
MVTLayer: fix coordinate-transform return value

AdriSolid:
`MVTLayer`: `coordinate-transform.js` returns an object with repeated nested `geometry`, should be a standard GeoJSON.

**Actually:**
```javascript
{
  "type": "Feature",
  "geometry": {
    "type": "Point",
    "coordinates": {
      "type": "Point",
      "coordinates": [0, 0]
    }
  },
  ...
}
```

**With new changes:**
```javascript
{
  "type": "Feature",
  "geometry": {
    "type": "Point",
    "coordinates": [0, 0]
  },
  ...
}
```

#### Change List
- Return `coordinates` in `coordinate-transform.js`
- Change unit tests

### DIFF
--- a/modules/geo-layers/src/mvt-layer/coordinate-transform.js
+++ b/modules/geo-layers/src/mvt-layer/coordinate-transform.js
@@ -45,12 +45,5 @@ export function transform(geometry, bbox, viewport) {
   const se = viewport.projectFlat([bbox.east, bbox.south]);
   const projectedBbox = [nw, se];
 
-  return {
-    ...geometry,
-    coordinates: availableTransformations[geometry.type](
-      geometry.coordinates,
-      projectedBbox,
-      viewport
-    )
-  };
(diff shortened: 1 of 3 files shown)


alasarr (changes requested):
I feel a little bit strange that the method receives an input geometry in a geojson format and return only the coordinates.

According to the method name I'd expect to return the input geometry in the same format.

So instead of doing this, my suggestion is to keep this module as it's and change only the MVTLater, so you can just replace [this](https://github.com/visgl/deck.gl/blob/adrisolid/fix-mvt-coordinates-transform-return/modules/geo-layers/src/mvt-layer/mvt-layer.js#L264) with:
```javascript
      const wgs84Geom = transform(object.geometry, tile.bbox, viewport);
      return wgs84Geom.coordinates;
```

VictorVelarde:
I aggree that keeping the full geojson geometry is desirable

alasarr (approved):

PR #5332, opened 2021-01-14
MVT Binary data support

alasarr:
It allows the usage of binary data for the MVT layer to improve the performance when passing tile data between web workers and the main thread.

Pending work:

- [x] Fix multi geometry
- [x] Add tests for binary and no binary (owner: @neokore)
- [x] Performance report (owner: @alasarr)


docs/api-reference/geo-layers/mvt-layer.md
```diff
@@ -117,6 +117,11 @@ On top of the [default options](/docs/api-reference/core/layer.md#loadoptions),
 
 - [MVTLoader](https://loaders.gl/modules/mvt/docs/api-reference/mvt-loader)
 
+##### `binary` (Boolean, optional)
+
+* Default: true
+
+Reads data in binary format to improve performance. More information in the [performance optimization guide](/docs/developer-guide/performance.md#use-binary-data).
```
Pessimistress, line 124: The linked documentation doesn't contain anything specific to the MVTLayer. I think the important thing to document here is whether it affects any user-facing APIs (e.g. will GeoJSON feature still be available via callbacks?)

examples/webpack.config.local.js
```diff
@@ -53,7 +55,8 @@ function makeLocalDevConfig(EXAMPLE_DIR = LIB_DIR, linkToLuma) {
     devServer: {
       stats: {
         warnings: false
-      }
+      },
+      contentBase: [ROOT_DIR, resolve(ROOT_DIR, '..' , 'loaders.gl') ]
```
Pessimistress, line 59: Don't check this in

examples/webpack.config.local.js
```diff
@@ -127,6 +130,6 @@ module.exports = (config, exampleDir) => env => {
     config.mode = 'production';
   }
 
-  // console.warn(JSON.stringify(config, null, 2)); // uncomment to debug config
+  console.warn(JSON.stringify(config, null, 2)); // uncomment to debug config
```
Pessimistress, line 133: Same

modules/core/src/lib/composite-layer.js
```diff
@@ -113,16 +113,26 @@ export default class CompositeLayer extends Layer {
...
+            properties: this.getNonIterableProperties(i)
+          };
         }
         return accessor(x, i);
       };
     }
     return accessor;
   }
 
+  getNonIterableProperties(i) {}
```
Pessimistress, line 134: As discussed, the accessor needs to be wrapped before passing to `GeoJsonLayer`.

ocular-dev-tools.config.js
```diff
@@ -22,7 +22,10 @@ const config = {
   },
 
   aliases: {
-    'deck.gl-test': resolve(__dirname, './test')
+    'deck.gl-test': resolve(__dirname, './test'),
+    '@loaders.gl/loader-utils': `${__dirname}/../loaders.gl/modules/loader-utils/src`,
+    '@loaders.gl/mvt': `${__dirname}/../loaders.gl/modules/mvt/src`,
+    '@loaders.gl/gis': `${__dirname}/../loaders.gl/modules/gis/src`
```
kylebarron, line 28: Maybe use an environment variable for the `loaders.gl` repo location? It's not guaranteed to be in the same directory as `deck.gl`

modules/geo-layers/src/mvt-layer/mvt-layer.js
```diff
@@ -152,37 +159,84 @@ export default class MVTLayer extends TileLayer {
     return super.onHover(info, pickingEvent);
   }
 
+  // eslint-disable-next-line complexity
```
Pessimistress, line 162: Move the binary to feature conversion to a utility function?

examples/webpack.config.local.js
alasarr: It's only for debugging purpose, will be removed when draft is removed

kylebarron (commented):
If I remember correctly, there can only be a single geometry type in each set of binary data. If so, is the `GeoJsonLayer` the right place to do all the binary data handling? Might be preferable to just split the incoming binary data into parts by geometry type, and then forward each to the `LineLayer`, `ScatterplotLayer`, `PolygonLayer` etc. Then a user could pass binary data in the loaders.gl format to any of those layers and it would "just work"?

modules/layers/src/geojson-layer/geojson-layer.js
```diff
@@ -157,13 +159,90 @@ export default class GeoJsonLayer extends CompositeLayer {
...
+
+    if (isBinary) {
+      const {points, lines, polygons} = this.props.data;
+
+      layerProps.points.data = {
+        length: points.positions.value.length / points.positions.size,
+        attributes: {
+          getPosition: points.positions,
+          getProperties: points.properties,
+          getNumericProps: points.numericProps,
```
Pessimistress, line 179: These are not known attributes in the sub-layer. They should be taken out of `attributes` to avoid confusion:

```js
      data: {
        length: points.positions.value.length / points.positions.size,
        attributes: {
          getPosition: points.positions,
        },
        properties: points.properties,
        numericProps: points.numericProps,
        globalFeatureIds: points.globalFeatureIds
      }
```

modules/layers/src/geojson-layer/geojson-layer.js
```diff
@@ -338,9 +399,35 @@ export default class GeoJsonLayer extends CompositeLayer {
...
+        : null;
+    }
+    return highlightedObjectIndex;
+  }
+
+  _isBinary() {
+    const {data} = this.props;
+    return (
+      data !== null &&
+      typeof data === 'object' &&
```
Pessimistress, line 414: Simplify to `data && 'points' in data && ...`

modules/geo-layers/src/mvt-layer/mvt-layer.js
```diff
@@ -195,6 +249,45 @@ export default class MVTLayer extends TileLayer {
...
 
+  _getLayerDataGeomType(data) {
+    if (Promise.resolve(data) !== data) {
+      if (data.points.featureIds.value.length) {
+        return 'points';
+      } else if (data.lines.featureIds.value.length) {
+        return 'lines';
+      } else if (data.polygons.featureIds.value.length) {
+        return 'polygons';
+      }
```
kylebarron, line 260: I can't remember if the `data` object is guaranteed to have each of these fields, so I wouldn't be surprised if this object accessing would error at times.

kylebarron (commented):
I think this might be the first attempt at integrating loaders.gl's binary data schema into deck.gl. Let us know if you find some rough edges with the binary schema

modules/layers/src/geojson-layer/geojson-layer.js
```diff
@@ -157,13 +159,90 @@ export default class GeoJsonLayer extends CompositeLayer {
     const LineStringsLayer = this.getSubLayerClass('line-strings', PathLayer);
     const PointsLayer = this.getSubLayerClass('points', ScatterplotLayer);
 
+    const layerProps = {
```
Pessimistress, line 162: We don't need nested objects for a scoped variable. It will prevent this code from being properly minified. Use

```js
const pointLayerProps = {};
const lineLayerProps = {};
const polygonFillLayerProps = {};
const polygonStrokeLayerProps = {};
```

modules/layers/src/geojson-layer/geojson-layer.js
```diff
@@ -157,13 +159,90 @@ export default class GeoJsonLayer extends CompositeLayer {
...
+      points: {data: undefined, _dataDiff: undefined},
+      lines: {data: undefined, _dataDiff: undefined},
+      polygons: {data: undefined, _dataDiff: undefined},
+      polygonsOutline: {data: undefined, _dataDiff: undefined}
```
Pessimistress, line 166: Why initialize a prop as `undefined`?

ocular-dev-tools.config.js
alasarr: I added this only for debugging purpose on my local dev... I wanted to delete this when moving the PR out the draft. Do you think will be interested to improve this?

modules/layers/src/geojson-layer/geojson-layer.js
```diff
@@ -157,13 +159,90 @@ export default class GeoJsonLayer extends CompositeLayer {
...
+    };
+
+    const isBinary = this._isBinary();
+
+    if (isBinary) {
+      const {points, lines, polygons} = this.props.data;
+
+      layerProps.points.data = {
```
Pessimistress, line 174: This creates a new `data` object on every rerender and will recalculate all attributes if any prop changes. `data` should be created once inside `updateState` on `changeFlags.dataChange`.

Pessimistress (commented):
Comments in line

modules/layers/src/geojson-layer/geojson-layer.js
alasarr: Just to try to explain the structure the object, but open to remove it

ocular-dev-tools.config.js
kylebarron: I think it's fine to leave this out completely

modules/layers/src/geojson-layer/geojson-layer.js
alasarr: I'm going to add the structure just like a comment

docs/api-reference/geo-layers/mvt-layer.md
ibgreen: If I just set `loadOptions: {gis: { format: 'binary' }}` will the layer auto-detect that the input is binary and do the right thing? Or I also have to set `binary` flag here?

modules/geo-layers/src/mvt-layer/mvt-layer.js
ibgreen: Can we use `.?` operator in deck.gl yet? Might be a good time to introduce ES2020 - could save us a lot of exceptions here.

modules/layers/src/geojson-layer/geojson-layer.js
ibgreen, line 162: Feels like this data processing should go into separate helper file

ibgreen (commented):
Very nice to see work on binary support progressing!

alasarr:
I've converted the PR to draft... sorry guys for the noise

docs/api-reference/geo-layers/mvt-layer.md
alasarr: The GeoJSON layer auto-detect the input no flag required there.

Agree with @Pessimistress. I'll start adding the limitations in the next commits. A current know limitation is that binary mode is only supported if renderSubLayers is a GeoJSONLayer (the default).

modules/geo-layers/src/mvt-layer/mvt-layer.js
alasarr: It cannot be used yet, it throws a webpack exception now.

alasarr:
>If I remember correctly, there can only be a single geometry type in each set of binary data. If so, is the GeoJsonLayer the right place to do all the binary data handling? Might be preferable to just split the incoming binary data into parts by geometry type, and then forward each to the LineLayer, ScatterplotLayer, PolygonLayer etc. Then a user could pass binary data in the loaders.gl format to any of those layers and it would "just work"?

@kylebarron I think it's a good idea, I'll implement this tomorrow.

alasarr:
I agree with @kylebarron that we should support this binary format in multiple layers not only in GeoJsonLayer, but it requires more time and will make this PR longer, so I think we can split this into two steps:

1. **Step one**: bring full support to binary data to MVTLayer and GeoJsonLayer.
2. **Step two**: bring full support to binary data to ScatterplotLayer, SolidPolygonLayer, PathLayer, and IconLayer.

What I mean with full support is to make these layers work as they do without binary format right now. 

@Pessimistress if we follow this approach, I think we don't need to move [this](https://github.com/visgl/deck.gl/blob/alasarr/binary/modules/layers/src/geojson-layer/geojson-layer.js#L346) outside GeoJsonLayer because we need accessors working out-of-the-box with binary data.

modules/layers/src/geojson-layer/geojson-layer.js
```diff
@@ -334,13 +328,39 @@ export default class GeoJsonLayer extends CompositeLayer {
...
+
+    return (object, info) => {
+      const {data, index} = info;
+      const properties = data.properties[index];
+      const numericProps = {};
+      for (const prop in data.numericProps) {
+        numericProps[prop] = data.numericProps[prop].value[index * data.numericProps[prop].size];
+      }
+      object = {
+        properties: {...numericProps, ...properties}
```
Pessimistress, line 359: you can create one less new object here:

```js
const properties = {...data.properties[index]};
for (...) {
  properties[prop] = ...
}
```

modules/core/src/utils/binary.js
```diff
@@ -0,0 +1,48 @@
...
+  const attr = {
+    points: null,
+    lines: 'pathIndices',
+    polygons: 'primitivePolygonIndices'
+  };
+
+  return binaryToFeatureIndexAttr(geojsonBinary[geomType], index, attr[geomType]);
+}
+
+export function binaryToFeatureIndexAttr(data, index, indexAttr) {
```
Pessimistress, line 17: Add some JSDoc here on what the parameters are?

modules/core/src/utils/binary.js
```diff
@@ -0,0 +1,48 @@
+export function binaryToFeature(geojsonBinary, index) {
+  const geomType = binaryGuessGeomType(geojsonBinary);
+
+  if (!geomType) {
+    return null;
+  }
+
+  const attr = {
```
Pessimistress, line 8: Use `switch`

modules/core/src/utils/binary.js
```diff
@@ -0,0 +1,48 @@
...
+      feature.properties[prop] =
+        data.numericProps[prop].value[featureIndex * data.numericProps[prop].size];
+    }
+  }
+
+  return feature;
+}
+
+export function binaryGuessGeomType(data) {
+  if (data && Promise.resolve(data) !== data) {
```
Pessimistress, line 38: use `data instance of Promise`

modules/geo-layers/src/mvt-layer/mvt-layer.js
```diff
@@ -269,5 +288,36 @@ function transformTileCoordsToWGS84(object, tile, viewport) {
   return feature;
 }
 
+function getHightlightedObjectIndex(data, uniqueIdProperty, featureIdToHighlight) {
```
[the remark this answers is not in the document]
alasarr: I have just pushed an improvement/refactor and I changed this as part of the refactor

modules/main/src/index.js
```diff
@@ -81,7 +81,10 @@ export {
   // Utilities
   Tesselator,
   fp64LowPart,
-  createIterable
+  createIterable,
+  binaryToFeature,
```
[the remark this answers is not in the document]
alasarr: Ok, good point

modules/core/src/index.js
```diff
@@ -94,3 +94,6 @@ export {count as _count} from './utils/count';
 export {default as _memoize} from './utils/memoize';
 export {mergeShaders as _mergeShaders} from './utils/shader';
 export {compareProps as _compareProps} from './lifecycle/props';
+
+// Binary utilities
```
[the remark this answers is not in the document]
alasarr: The util class is used from two modules (geo-layers and layers) and the layer module only exports Layers... that said, you still consider is better to move to layers... It makes more sense for me, but as layers exports only layers I did like this.

I was not happy with the names and I've just pushed a commit changing the names. Do you mind to give me your feedback on the new ones?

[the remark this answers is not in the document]
Pessimistress: See /modules/aggregation-layers/src/index.js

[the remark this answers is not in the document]
alasarr: Crystal clear then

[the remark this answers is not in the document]
alasarr: Done

modules/core/src/utils/binary.js
```diff
@@ -0,0 +1,48 @@
...
+      ? data[indexAttr].value[index]
+      : data[indexAttr][index];
+  const geometryIndex = data.featureIds.value[featureIndex];
+  const feature = {};
+
+  if (featureIndex !== -1) {
+    feature.properties = {...data.properties[geometryIndex]};
+    for (const prop in data.numericProps) {
+      feature.properties[prop] =
+        data.numericProps[prop].value[featureIndex * data.numericProps[prop].size];
```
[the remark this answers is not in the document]
alasarr: According to the [doc])(https://github.com/visgl/loaders.gl/blob/master/modules/gis/docs/api-reference/geojson-to-binary.md) of the binary no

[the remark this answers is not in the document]
Pessimistress: If size is always 1 then you can drop `* data.numericProps[prop].size`

alasarr:
Moved to review. 

Pending stuff:
- Tests
- Performance report where we compare binary vs non-binary

alasarr:
I've detected we need to update the getter dataInWGS84 when binary format is provided.

test/modules/geo-layers/mvt-layer.spec.js
```diff
@@ -282,30 +297,39 @@ test('MVT dataInWGS84', async t => {
     zoom: 0
   });
 
+  const onAfterUpdate = ({layer}) => {
+    if (layer.isLoaded) {
+      const tile = layer.state.tileset.selectedTiles[0];
+      const contentWGS84 = tile.dataInWGS84;
+
+      t.deepEqual(
+        layer.props.binary ? contentWGS84 : contentWGS84[0].geometry.coordinates,
```
alasarr, line 306: Rather than adapt the test we need to make dataInWGS84 returns the same in binary vs non-binary

tile.dataInWGS84 should always return an array of geojson features

test/modules/layers/geojson-binary.spec.js
```diff
@@ -0,0 +1,121 @@
+import test from 'tape-catch';
+import {geojsonToBinary} from '@loaders.gl/gis';
+import {calculatePickingColors} from '@deck.gl/layers/geojson-layer/geojson-binary';
+import {_binaryToFeature, _findIndexBinary} from '@deck.gl/layers';
+import {Layer} from '@deck.gl/core';
+
+const geoJSONData = [
```
alasarr, line 7: Can we move this to an external file? Something like:

```javascript
import {geoJSONData} from './data/fixtures.js' 
```

test/render/test-cases/mvt-layer.js
```diff
@@ -58,6 +59,75 @@ export default [
...
+          mvt: {
+            workerUrl: null
+          }
+        }
+      })
+    ],
+    goldenImage: './test/render/golden-images/mvt-layer-highlight.png'
+  },
+  {
+    name: 'mvt-layer-binary',
```
alasarr, line 73: Would be possible to avoid this copy&paste

docs/api-reference/geo-layers/mvt-layer.md
```diff
@@ -117,6 +117,16 @@ On top of the [default options](/docs/api-reference/core/layer.md#loadoptions),
...
 
+##### `binary` (Boolean, optional)
+
+* Default: false
+
+Use tile data in [binary format](https://github.com/visgl/loaders.gl/blob/master/modules/gis/docs/api-reference/geojson-to-binary.md) to improve performance. It removes the need for serialization and deserialization of data transferred by the worker back to the main process.
+
+Remarks: 
+
+- It requires to use `GeoJsonLayer` as the `renderSubLayers` function to work.
```
Pessimistress, line 128:
```suggestion
It requires using `GeoJsonLayer` in the `renderSubLayers` callback.
```

docs/api-reference/geo-layers/mvt-layer.md
```diff
@@ -117,6 +117,16 @@ On top of the [default options](/docs/api-reference/core/layer.md#loadoptions),
...
+- In binary format, polygons or multipolygons with holes are not supported.
```
Pessimistress, line 129: This should be addressed in a follow up PR. Phrase it as work in progress.

modules/geo-layers/src/mvt-layer/mvt-layer.js
```diff
@@ -227,9 +261,16 @@ export default class MVTLayer extends TileLayer {
               return null;
             }
 
+            if (this.props.binary && Array.isArray(tile.content) && !tile.content.length) {
+              // TODO: @deck.gl/loaders returns [] when no content. It should return a valid empty binary.
```
Pessimistress, line 265: `@loaders.gl/mvt`

modules/layers/src/geojson-layer/geojson-binary.js
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  const featureIndex = 'startIndices' in data ? data.startIndices[index] : index;
+  const geometryIndex = data.featureIds.value[featureIndex];
+
+  if (featureIndex !== -1) {
+    return getPropertiesForIndex(data, geometryIndex, featureIndex);
+  }
+
+  return -1;
```
Pessimistress, line 43: `null`?

modules/layers/src/geojson-layer/geojson-binary.js
```diff
@@ -0,0 +1,124 @@
...
+}
+
+function getPropertiesForIndex(data, propertiesIndex, numericPropsIndex) {
+  const feature = {
+    properties: {...data.properties[propertiesIndex]}
+  };
+
+  for (const prop in data.numericProps) {
+    feature.properties[prop] =
+      data.numericProps[prop].value[numericPropsIndex * data.numericProps[prop].size];
```
Pessimistress, line 53: Remove `* data.numericProps[prop].size`

modules/layers/src/geojson-layer/geojson-binary.js
```diff
@@ -0,0 +1,124 @@
...
+ * @param {Object} data - The data in binary format
+ * @param {String} uniqueIdProperty - Name of the unique id property
+ * @param {Number} featureId - feature id to find
+ */
+export function findIndexBinary(data, uniqueIdProperty, featureId) {
+  if (!data) {
+    return -1;
+  }
+
+  const geomTypes = ['points', 'lines', 'polygons'];
```
Pessimistress, line 72: Define constant outside of this function

modules/layers/src/geojson-layer/geojson-binary.js
```diff
@@ -0,0 +1,124 @@
...
+
+// Custom picking color to keep binary indexes
+export function calculatePickingColors(geojsonBinary, encodePickingColor) {
+  const pickingColors = {
+    points: null,
+    lines: null,
+    polygons: null
+  };
+  for (const key in pickingColors) {
+    pickingColors[key] = new Uint8ClampedArray(geojsonBinary[key].featureIds.value.length * 3);
```
Pessimistress, line 113: Define a variable for `geojsonBinary[key].featureIds.value`. Unnecessary look ups in a big for loop.

modules/layers/src/geojson-layer/geojson-layer.js
```diff
@@ -334,13 +329,32 @@ export default class GeoJsonLayer extends CompositeLayer {
...
+    if (!this._isBinary()) {
+      return Number.isFinite(highlightedObjectIndex)
+        ? data.findIndex(d => d.__source.index === highlightedObjectIndex)
+        : null;
+    }
+    return highlightedObjectIndex;
+  }
+
+  _isBinary() {
+    const {data} = this.props;
```
Pessimistress, line 344: Save the result of this check in layer state upon `changeFlags.dataChanged`.

test/modules/layers/data/fixtures.js
```diff
@@ -0,0 +1,67 @@
...
+      ]
+    },
+    properties: {
+      cartodb_id: 149,
+      name: 'Second'
+    }
+  }
+];
+
+export const pickingColorsSample = Uint8ClampedArray.from([
```
Pessimistress, line 42: You can add `// prettier-ignore` to prevent long arrays being formatted

test/modules/geo-layers/mvt-layer.spec.js
```diff
@@ -180,15 +191,16 @@ test('MVT Highlight', async t => {
...
-          t.ok(layer.props.pickable, 'MVT Sublayer is pickable');
-          t.ok(!layer.props.autoHighlight, 'AutoHighlight should be disabled');
-          t.equal(layer.props.highlightedObjectIndex, 0, 'Feature highlighted has index 0');
-        }
-      }
+      onAfterUpdate
+    },
+    {
+      updateProps: {
+        binary: true
```
Pessimistress, line 201: What does this test? I don't see the layer re-fetch tile data when the `binary` prop changes.

test/modules/layers/data/fixtures.js
alasarr: Thanks for the tip!

test/modules/geo-layers/mvt-layer.spec.js
alasarr: You're totally right it was not called

alasarr: It tests highlight with binary data

Pessimistress (approved):

modules/geo-layers/src/mvt-layer/mvt-layer.js
```diff
@@ -154,32 +162,57 @@ export default class MVTLayer extends TileLayer {
 
     const isWGS84 = this.context.viewport.resolution;
 
-    if (!isWGS84 && info.object) {
-      info.object = transformTileCoordsToWGS84(info.object, info.tile.bbox, this.context.viewport);
+    if (info.object) {
+      if (!isWGS84 && info.object) {
+        info.object = transformTileCoordsToWGS84(info.object, info.tile, this.context.viewport);
```
Jesus89, line 167: Found an issue here while moving the function. It appears in 8.4.0-beta.2:

~info.tile~ info.tile.bbox

This breaks getting coordinates from geometries in picked objects.

PR #5339, opened 2021-01-15
Improve CARTO module coverage

alasarr:
Improve CARTO module coverage


test/modules/carto/carto-bqtiler-layer.spec.js
```diff
@@ -15,7 +15,7 @@ test('CartoBQTilerLayer', t => {
 });
 
 test('CartoBQTilerLayer#updateTileJSON', t => {
-  const fetch = mockFetchWithTileJSON();
+  const fetch = mockFetchMapsV2();
```
AdriSolid, line 18: Shall we rename this const for something different than `fetch`? It sounds like the `window.fetch` method.

test/modules/carto/carto-sql-layer.spec.js
```diff
@@ -59,3 +62,85 @@ test('CartoSQLLayer#updateTileJSON', t => {
...
+
+  const onAfterUpdate = ({layer, subLayers, subLayer}) => {
+    const {tilejson} = layer.state;
+    if (!tilejson) {
+      t.is(subLayers.length, 0, 'should no render subLayers');
+    } else {
+      t.is(subLayers.length, 1, 'should render a subLayer');
+      t.ok(Array.isArray(tilejson.tiles), 'tiles should be an array');
+      t.ok(subLayer instanceof MVTLayer, 'subLayer should be an MVT layer ');
+      t.is(subLayer.props.data, tilejson, 'data should be a tileJSON');
```
AdriSolid, line 80: Typo `'an'`
`Sublayer should be a MVT layer.`

AdriSolid (changes requested):

test/modules/carto/carto-bqtiler-layer.spec.js
alasarr: Why? We're mocking global fetch method

AdriSolid: Just suggesting not calling const names like `window` methods. It's okey anyway!

AdriSolid (approved):

alasarr:
I need to figure out why is coverage decreasing even when we're just adding tests...

test/modules/carto/carto-bqtiler-layer.spec.js
alasarr: Ok, makes sense

Pessimistress:
@alasarr The coverage number fluctuates for no reason sometimes. It's actually going up (our low bound was 81.x)

Issue #5297, opened 2021-01-06
Abort tile requests when creating a new Tileset2D


andreasg123:
#### Target Use case

When experimenting with changing z-slices quickly in [Viv](https://github.com/hms-dbmi/viv), I noticed that tiles from previously visible z-slices continued loading, delaying the loading of tiles for the currently visible z-slice. I believe that Viv uses `updateTriggersChanged.getTileData` to make `TileLayer` create a new `Tileset2D`: https://github.com/hms-dbmi/viv/issues/337

#### Proposed feature

When a new `Tileset2D` is created, the previous one just gets abandoned:
https://github.com/visgl/deck.gl/blob/8.3-release/modules/geo-layers/src/tile-layer/tile-layer.js#L78

I would like to request that all pending tile requests in the previous `Tileset2D` are aborted before the new tile set is created.

#### To Do List 
  - [ ] Add label and assign to milestone 
  - [ ] Coding
  - [ ] Doc update
  - [ ] What’s new update 
  - [ ] Test

ilan-gold:
I think we can close this now, @Pessimistress ?

Pessimistress:
Fixed in `deck.gl@8.4.0-alpha.6`

PR #5342, opened 2021-01-16
Abort Ongoing Tile Requests When New Tileset is Created

ilan-gold:
For #5297.  Adds a `finalize` function to the `Tileset2D` to handle aborting the ongoing tile requests when a new `Tileset2D` is about to be made by the layer.

I could add a test with `setTimeout` or something to guarantee tiles can actually be aborted but not sure that it's really a "unit test" at that point.  Also, I don't think I need to do anything to the visibility of the tiles since the previous (i.e the one having `finalize` called) `Tileset2D`'s tiles are all abandoned by the next render anyway (might have that wrong).

#### Background
Rapid changes on `updateTriggers` to the `TileLayer` can leave tile requests hanging, slowing down the application.  This PR adds the ability to abort ongoing requests when `updateTriggers` are triggered.

#### Change List
- Add a `finalize` function to `Tileset2D` for aborting ongoing tile requests.
- Changed the state variable from `tiles` to `tileset` in `initializeState` because `tiles` was removed some time ago.  I can remove this though, just noticed it.


chrisgervang:
Hey! This looks good. I pulled it down yesterday and started to think about how we might add a test to cover it. I was thinking we could load a couple tiles, some loaded, some pending, then finalize and check the cache to look as expected.


Do you have an ideas for how you'd test it?

Pessimistress (approved):

chrisgervang (approved):

ilan-gold:
I tried adding a test using `sleep` @chrisgervang and pushed the branch.  Let me know how that looks.

ilan-gold:
@chrisgervang  I just realized we probably want to test the layer as well.  I will add that as well.

chrisgervang:
Sounds good. I've written a second test for tileset-2d to cover a partial load, and a reload. Check it out here https://github.com/ilan-gold/deck.gl/compare/abort_tile_requests_on_update...visgl:chr/abort_tile_requests_on_update

ilan-gold:
@chrisgervang I'm realizing that because the old `Tileset2D` is discarded, there may not be any way to test this on the layer i.e we can't access the old `layer.state.tileset` to check if tiles have been aborted.

kylebarron (approved):

Pessimistress:
@ilan-gold You can query the tileset right after the initial update and hold on to the tile instances.

ilan-gold:
@Pessimistress just saw this but had the same thought. Will do.

PR #5356, opened 2021-01-19
Bump math.gl dependency => 3.4.2

igorDykhta:
Bump for math.gl 3.3.2 > 3.4.2


kylebarron (approved):

Pessimistress:
CI is failing because `@math.gl/core` is resolved to multiple versions. You can see this by running `yarn list "@math.gl/*"`. You need force the lock file to update.

PR #5360, opened 2021-01-20
Return tile objects at OnViewportLoad instead of data

alasarr:
#### Background
The callback at TileLayer doesn't return the x,y,z of the tile and it's quite difficult to work only with the tile data. For example, if the user wants to transform to WGS84 we need to also provide the reference tile together with the data.

This is a small breaking change that we need to notify in the release.

#### Change List
- Return an array of tile objects at onViewportLoad instead of the content of the file.
- Updated TileLayer doc. I've included a section for the Tile class.

### DIFF
--- a/docs/api-reference/geo-layers/tile-layer.md
+++ b/docs/api-reference/geo-layers/tile-layer.md
@@ -233,7 +233,8 @@ Affects both rendering and tile fetching to produce a transformed tile layer.  N
 
 ##### `onViewportLoad` (Function, optional)
 
-`onViewportLoad` is a function that is called when all tiles in the current viewport are loaded. The loaded content (as returned by `getTileData`) for each visible tile is passed as an array to this callback function.
+`onViewportLoad` is a function that is called when all tiles in the current viewport are loaded. The loaded content or each visible tile is passed as an array of tile objects to this callback function. Check the [Tile](#tile) class for more info.
+
 
 - Default: `data => null`
 
@@ -268,6 +269,19 @@ Receives arguments:
 
 - `tile` (Object) - the tile that has been cleared from cache.
 
+## Tile
+
+Class to hold the reading of a single tile
+
+Properties:
+
+- `x` (Number) - x index of the tile
+- `y` (Number) - y index of the tile
+- `z` (Number) - z index of the tile
+- `data` (Array) - tiles content as returned by `getTileData`. 
+
 ## Source
 
 [modules/geo-layers/src/tile-layer](https://github.com/visgl/deck.gl/tree/master/modules/geo-layers/src/tile-layer)
+
+
--- a/docs/api-reference/geo-layers/tile-layer.md
+++ b/docs/api-reference/geo-layers/tile-layer.md
@@ -233,7 +233,7 @@ Affects both rendering and tile fetching to produce a transformed tile layer.  N
 
 ##### `onViewportLoad` (Function, optional)
 
-`onViewportLoad` is a function that is called when all tiles in the current viewport are loaded. The loaded content for each visible tile is passed as an array of tile objects to this callback function. Check the [Tile](#tile) class for more info.
+`onViewportLoad` is a function that is called when all tiles in the current viewport are loaded. An array of loaded [Tile](#tile) instances are passed as argument to this function
 
 
 - Default: `data => null`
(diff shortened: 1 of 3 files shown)


docs/api-reference/geo-layers/tile-layer.md
```diff
@@ -233,7 +233,8 @@ Affects both rendering and tile fetching to produce a transformed tile layer.  N
 
 ##### `onViewportLoad` (Function, optional)
 
-`onViewportLoad` is a function that is called when all tiles in the current viewport are loaded. The loaded content (as returned by `getTileData`) for each visible tile is passed as an array to this callback function.
+`onViewportLoad` is a function that is called when all tiles in the current viewport are loaded. The loaded content for each visible tile is passed as an array of tile objects to this callback function. Check the [Tile](#tile) class for more info.
```
Pessimistress, line 236: `An array of loaded [Tile](#tile) instances are passed as argument to this function`

docs/api-reference/geo-layers/tile-layer.md
```diff
@@ -268,6 +269,19 @@ Receives arguments:
 
 - `tile` (Object) - the tile that has been cleared from cache.
 
+## Tile
+
+Class to hold the reading of a single tile
+
+Properties:
```
Pessimistress, line 276: Missing `bbox`

docs/api-reference/geo-layers/tile-layer.md
```diff
@@ -268,6 +269,19 @@ Receives arguments:
 
 - `tile` (Object) - the tile that has been cleared from cache.
 
+## Tile
```
Pessimistress, line 272: This should also be mentioned in `onTileLoad` and `onTileUnload`

Pessimistress (commented):
Add the breaking change to upgrade guide

docs/api-reference/geo-layers/tile-layer.md
alasarr: I tried to add only the most important properties, but agree this one is important

AdriSolid:
Would be great to export `transformTileCoordsToWGS84` from `MVTLayer` [function](https://github.com/visgl/deck.gl/blob/0adfe05534b73407207c73300b02856258cdfd99/modules/geo-layers/src/mvt-layer/mvt-layer.js#L253) as user could manage tile coords transformation from outside

alasarr:
>Would be great to export transformTileCoordsToWGS84 function from MVTLayer as user could manage tile coords transformation from outside.

That would be done in another PR

docs/api-reference/geo-layers/tile-layer.md
```diff
@@ -266,8 +267,22 @@ Receives arguments:
...
+
+Class to hold the reading of a single tile
+
+Properties:
+
+- `x` (Number) - x index of the tile
+- `y` (Number) - y index of the tile
+- `z` (Number) - z index of the tile
+- `bbox` (Object) - tile boundaries in WGS84
```
Pessimistress, line 281: This is incorrect. See https://deck.gl/docs/api-reference/geo-layers/tile-layer#gettiledata

alasarr: Fixed

Pessimistress (approved):

docs/api-reference/geo-layers/tile-layer.md
```diff
@@ -266,8 +267,22 @@ Receives arguments:
...
+- `y` (Number) - y index of the tile
+- `z` (Number) - z index of the tile
+- `bbox` (Object) - bounding box of the tile. When used with a geospatial view, `bbox` is in the shape of `{west: <longitude>, north: <latitude>, east: <longitude>, south: <latitude>}`. When used with a non-geospatial view, `bbox` is in the shape of `{left, top, right, bottom}`.
+- `data` (Array) - tiles content as returned by `getTileData`. 
 
 ## Source
 
 [modules/geo-layers/src/tile-layer](https://github.com/visgl/deck.gl/tree/master/modules/geo-layers/src/tile-layer)
+
+
```
kylebarron, line 288: I'm surprised the extra ending whitespace passed the linter.

kylebarron (commented):
:+1: I've also thought before that it's necessary to have the tile referencing along with the data, but never got around to submitting a PR

docs/api-reference/geo-layers/tile-layer.md
alasarr: Does the linter work in markdown?

kylebarron: I've definitely gotten linter errors in Markdown in loaders.gl; I assumed it would be the same linter in deck.gl

PR #5361, opened 2021-01-20
Tiles content in WGS84 for MVTLayer

alasarr:
#### Background
MVTLayer users receive the tiles content at `onViewportLoad` in local coordinates. This PR add a new getter `dataInWGS84` to return the tiles transformed in WGS84

Usage example:
```javascript
const onViewportLoad = tiles => {
  tiles.forEach(tile => {
    // data in local coordinates
    const dataInLocalCoords = tile.data;
    // data in world coordinates (WGS84)
    const dataInWGS84 = tile.dataInWGS84;
  });
};
new MVTLayer({
  id: "..."
  data: "..."
  onViewportLoad
})
```
#### Change List

- [x] Add `onViewportLoad` at MVTLayer doc to explain the getter and the usage.
- [x] Modify MVTLayer to dynamically add the getter to the visible tiles at updateState
- [x] Tests

### DIFF
--- a/docs/api-reference/geo-layers/mvt-layer.md
+++ b/docs/api-reference/geo-layers/mvt-layer.md
@@ -161,6 +161,30 @@ new MVTLayer({
 })
 ```
 
+##### `onViewportLoad` (Function, optional)
+
+`onViewportLoad` is a function that is called when all tiles in the current viewport are loaded. An array of loaded [Tile](/docs/api-reference/geo-layers/tile-layer.md#tile) instances are passed as argument to this function.
+
+A new getter `dataInWGS84` is added for the tile instance to retrieve the tile content in world coordinates (WGS84).
+
+- Default: `data => null`
+
+```javascript
+const onViewportLoad = tiles => {
+  tiles.forEach(tile => {
+    // data in local coordinates
+    const dataInLocalCoords = tile.data;
+    // data in world coordinates (WGS84)
+    const dataInWGS84 = tile.dataInWGS84;
+  });
+};
+new MVTLayer({
+  id: "..."
+  data: "..."
+  onViewportLoad
+})
+```
+
 ## Source
 
 [modules/geo-layers/src/mvt-layer/mvt-layer.js](https://github.com/visgl/deck.gl/tree/master/modules/geo-layers/src/mvt-layer/mvt-layer.js)
--- a/docs/api-reference/geo-layers/mvt-layer.md
+++ b/docs/api-reference/geo-layers/mvt-layer.md
@@ -163,9 +163,11 @@ new MVTLayer({
 
 ## Tile
 
-A new getter `dataInWGS84` is added to the [Tile](/docs/api-reference/geo-layers/tile-layer.md#tile) instances used on this layer. It allows to retrieve the tile content in world coordinates (WGS84).
+Aside from all members of the [Tile](/docs/api-reference/geo-layers/tile-layer.md#tile) class, tile instances from the `MVTLayer` also include the following fields:
 
-- Default: `data => null`
+##### `dataInWGS84` (Array)
+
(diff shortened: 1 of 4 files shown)


modules/geo-layers/src/mvt-layer/mvt-layer.js
```diff
@@ -228,10 +229,28 @@ export default class MVTLayer extends TileLayer {
     }
   }
 
-  _onViewportLoad() {
-    super._onViewportLoad();
-    this._onViewportChange();
+  _setWGS84PropertyForTiles() {
```
AdriSolid, line 232: - Shall we move this method to a pure function? Passing `selectedTiles`from `updateState` like:
`setWGS84PropertyForTiles(tileset.selectedTiles);`
- `propName` can be outside the function then?

docs/api-reference/geo-layers/mvt-layer.md
```diff
@@ -161,6 +161,30 @@ new MVTLayer({
 })
 ```
 
+##### `onViewportLoad` (Function, optional)
```
Pessimistress, line 164: There is no need to re-document this prop from the base TileLayer class, since it behaves the same. Instead, add a `Tile` section to this doc to describe the additional field.

modules/geo-layers/src/mvt-layer/mvt-layer.js
```diff
@@ -228,9 +229,26 @@ export default class MVTLayer extends TileLayer {
...
+    const {tileset} = this.state;
+
+    tileset.selectedTiles.forEach(tile => {
+      if (!tile.hasOwnProperty(propName)) {
+        // eslint-disable-next-line accessor-pairs
+        Object.defineProperty(tile, propName, {
+          get: () => {
+            if (!tile.contentWGS84) {
+              // Create a cache to transform only once
+              tile.contentWGS84 = tile.content.map(feature =>
```
Pessimistress, line 243: content can be `null` (still loading or encountered an error)

docs/api-reference/geo-layers/mvt-layer.md
```diff
@@ -161,6 +161,27 @@ new MVTLayer({
 })
 ```
 
+## Tile
+
+A new getter `dataInWGS84` is added to the [Tile](/docs/api-reference/geo-layers/tile-layer.md#tile) instances used on this layer. It allows to retrieve the tile content in world coordinates (WGS84).
+
+- Default: `data => null`
```
Pessimistress, line 168: Drop

docs/api-reference/geo-layers/mvt-layer.md
```diff
@@ -161,6 +161,27 @@ new MVTLayer({
 })
 ```
 
+## Tile
+
+A new getter `dataInWGS84` is added to the [Tile](/docs/api-reference/geo-layers/tile-layer.md#tile) instances used on this layer. It allows to retrieve the tile content in world coordinates (WGS84).
```
Pessimistress, line 166:
```
Aside from all members of the [Tile](/docs/api-reference/geo-layers/tile-layer.md#tile) class, tile instances from the `MVTLayer` also include the following fields:

##### `dataInWGS84` (Array)

A list of features...
```

modules/geo-layers/src/mvt-layer/mvt-layer.js
Pessimistress: - Still not checking `tile.content` before calling `map`
- Add an underscore before the name of internal cache

alasarr: Sorry I missed this comment

alasarr: Done

Pessimistress (approved):

PR #5362, opened 2021-01-20
Simpler and more robust frustum calculation

Pessimistress:
#### Background

While experimenting with additional viewport settings, I noticed that our frustum calculation is based on a bunch of assumptions of how the projection matrix is composed. For example, if `projectionMatrix` is externally supplied without the fov/near/far parameters, it will not work. After this change, `getFrustumPlanes` will support any generalized use case.

I have verified that the output is identical (aside from slight precision differences) in MapView/OrbitView/OrthographicView. This utility is also already covered by test cases.

#### Change List
- New `getFrustumPlanes` algorithm
- Drop unused `viewport` fields: `cameraDirection`, `cameraUp`, `cameraRight`, and `projectionProps`

### DIFF
--- a/modules/core/src/utils/math-utils.js
+++ b/modules/core/src/utils/math-utils.js
@@ -13,108 +13,61 @@ export function mod(value, divisor) {
 }
 
 // Extract camera vectors (move to math library?)
-export function extractCameraVectors({viewMatrix, viewMatrixInverse}) {
+export function getCameraPosition(viewMatrixInverse) {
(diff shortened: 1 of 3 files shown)


kylebarron (approved):

ibgreen (approved):
This should really go into `@math.gl/culling` module. Not sure why this code is here.

Pessimistress:
@ibgreen `@math.gl/culling` is not currently a dependency of deck.gl core. We can add it with the next math.gl release.

PR #5376, opened 2021-01-23
Add prop to specify winding direction in SolidPolygonLayer

Pessimistress:
Follow up of #5223 

This PR allows binary users to specify the winding order of their polygon data if they choose to skip normalization.

I also considered using this prop to control the default winding order during normalization but decided against it for now. Benchmark shows that perf difference is ~10% when using one vs another. I've noticed that GeoJSONs converted from Shapefiles using GDAL do not conform to the right-hand rule, so it's hard to determine what's the best order to use simply based on the file format. It might introduce more confusion than benefit if we ask the user to explicitly pick one.

#### Change List
- Add `_windingOrder` prop to `PolygonLayer` and `SolidPolygonLayer`
- A benchmark test for polygon normalization
- Docs
- Upgrade guide

### DIFF
--- a/modules/layers/src/solid-polygon-layer/polygon.js
+++ b/modules/layers/src/solid-polygon-layer/polygon.js
@@ -113,6 +113,7 @@ function copyNestedRing(target, targetStartIndex, simplePolygon, size, windingDi
 
   windingOptions.start = targetStartIndex;
   windingOptions.end = targetIndex;
+  windingOptions.size = size;
   modifyPolygonWindingDirection(target, windingDirection, windingOptions);
 
   return targetIndex;
@@ -157,6 +158,7 @@ function copyFlatRing(
 
   windingOptions.start = targetStartIndex;
   windingOptions.end = targetIndex;
+  windingOptions.size = size;
   modifyPolygonWindingDirection(target, windingDirection, windingOptions);
 
   return targetIndex;
@@ -176,7 +178,6 @@ export function normalize(polygon, positionSize) {
 
   const positions = [];
   const holeIndices = [];
-  windingOptions.size = positionSize;
 
   if (polygon.positions) {
     // complex flat
--- a/modules/layers/src/solid-polygon-layer/polygon.js
+++ b/modules/layers/src/solid-polygon-layer/polygon.js
@@ -25,6 +25,10 @@ import {modifyPolygonWindingDirection, WINDING} from '@math.gl/polygon';
 const OUTER_POLYGON_WINDING = WINDING.CLOCKWISE;
 const HOLE_POLYGON_WINDING = WINDING.COUNTER_CLOCKWISE;
 
+const windingOptions = {
+  isClosed: true
+};
+
 // 4 data formats are supported:
 // Simple Polygon: an array of points
 // Complex Polygon: an array of array of points (array of rings)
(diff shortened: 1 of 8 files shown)


modules/layers/src/solid-polygon-layer/polygon.js
```diff
@@ -107,14 +111,9 @@ function copyNestedRing(target, targetStartIndex, simplePolygon, size, windingDi
...
-    modifyPolygonWindingDirection(target, windingDirection, {
-      start: targetStartIndex,
-      end: targetIndex,
-      size,
-      isClosed: true
-    });
-  }
+  windingOptions.start = targetStartIndex;
+  windingOptions.end = targetIndex;
+  modifyPolygonWindingDirection(target, windingDirection, windingOptions);
```
igorDykhta, line 116: Would set windingOptions.size here as well as it is a bit confusing why the passed size arg is ignored. Not sure that there are any noticeable performance differences.

modules/layers/src/solid-polygon-layer/solid-polygon-layer-vertex-side.glsl.js
```diff
@@ -39,14 +39,21 @@ ${main}
 void main(void) {
   PolygonProps props;
 
-  props.positions = instancePositions;
-  props.positions64Low = instancePositions64Low;
+  #if RING_WINDING_ORDER == 1
```
igorDykhta, line 42: Maybe #if RING_WINDING_ORDER_CW/CLOCKWISE as boolean?

igorDykhta (approved):
Looks good. Do we need a render test to be sure that the _windingOrder prop works, or it can be left without one as long as it is an experimental feature?

kylebarron:
> I've noticed that GeoJSONs converted from Shapefiles using GDAL do not conform to the right-hand rule

This sounded fishy to me so I looked into it. The issue is that by default, OGR creates GeoJSON that conforms to the [original GeoJSON spec from 2008](https://geojson.org/geojson-spec.html). That spec includes no criteria on winding order; it [only says](https://geojson.org/geojson-spec.html#id4)

> The first element in the array represents the exterior ring. Any subsequent elements represent interior rings (or holes).

To tell OGR to create new-spec GeoJSON, you need to add a layer creation option, e.g.
```
ogr2ogr -f 'GeoJSON' -lco RFC7946=YES out.geojson in.shp
```

When the new spec is set, the winding order is correct.

modules/layers/src/solid-polygon-layer/solid-polygon-layer-vertex-side.glsl.js
```diff
@@ -39,7 +39,7 @@ ${main}
 void main(void) {
   PolygonProps props;
 
-  #if RING_WINDING_ORDER == 1
+  #if RING_WINDING_ORDER_CW == 1
```
igorDykhta, line 42: Just #if RING_WINDING_ORDER_CW is enough

igorDykhta (approved):

modules/layers/src/solid-polygon-layer/solid-polygon-layer-vertex-side.glsl.js
igorDykhta: Ignore, missed that RING_WINDING_ORDER_CW is defined when for both bases

PR #5394, opened 2021-01-28
Add GlobeView example

Pessimistress:
For #5373 

Screen Shot 2021-01-27 at 11 36 58 PM

#### Change List
- Add GlobeView example to the website


kylebarron (approved):
Looks like a good example of subclassing a layer with shader injections

PR #5402, opened 2021-01-29
Avoid silencing errors by default in CartoLayer

neokore:
By default errors from `_updateData` in CartoLayers were caught and output to the error log, now the errors will be uncaught by default to help with tests and debugging.

### DIFF
--- a/modules/carto/src/layers/carto-layer.js
+++ b/modules/carto/src/layers/carto-layer.js
@@ -6,8 +6,7 @@ const defaultProps = {
   data: null,
   credentials: null,
   onDataLoad: {type: 'function', value: tilejson => {}, compare: false},
-  // eslint-disable-next-line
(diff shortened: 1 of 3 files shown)


modules/carto/src/layers/carto-layer.js
```diff
@@ -6,8 +6,7 @@ const defaultProps = {
   data: null,
   credentials: null,
   onDataLoad: {type: 'function', value: tilejson => {}, compare: false},
-  // eslint-disable-next-line
-  onDataError: {type: 'function', value: err => console.error(err), compare: false}
+  onDataError: {type: 'function', value: null, compare: false}
```
Pessimistress, line 9: - You need `optional: true` to avoid prop type check error
- Update documentation?

Pessimistress (approved):

PR #5415, opened 2021-02-01
Fix FirstPersonView "matrix not invertible" error at pitch=-90

Pessimistress:
#### Background

`FirstPersonView` throws at `pitch: -90` (matrix not invertible)

#### Change List
- Add a small offset to `pitch` if it is `-90`


ibgreen (approved):
> FirstPersonView throws at pitch: 90/-90 (matrix not reversible)

As a possible idea, since this is such a recurring problem, I suppose we could filter out 90 automatically and silently replace with 89, either in the views or in the underlying mercator functions. Better being one degree off than crashing?

kylebarron (approved):

PR #5421, opened 2021-02-02
pydeck: Update custom library registration

ajduberstein:
For #5063

Trying to improve the custom library writing process in order to move 3rd party collaborations (e.g. carto+pydeck) into their own modules.

### DIFF
--- a/bindings/pydeck/pydeck/settings.py
+++ b/bindings/pydeck/pydeck/settings.py
@@ -26,9 +26,12 @@ class Settings:
         ```
     """
 
-    def __init__(self, custom_libraries=None):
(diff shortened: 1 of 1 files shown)


bindings/pydeck/pydeck/settings.py
```diff
@@ -26,9 +26,12 @@ class Settings:
         ```
     """
 
-    def __init__(self, custom_libraries=None):
+    def __init__(self, custom_libraries: list = None):
```
kylebarron, line 29: `list` with a lower case `l` is intentional? https://stackoverflow.com/q/39458193

kylebarron (approved):

Issue #5453, opened 2021-02-10
Error "Unimplemented type: 4" when using IconLayer from a MVTLayer


neokore:
#### Description
When loading the icon image `iconAtlas` for an IconLayer loaded as sublayer of a MVTLayer (or CartoLayer) it returns a `deck: Unimplemented type: 4` error and doesn't load the layer. It happens with Deck.gl v8.4 but not with v8.3 with which it works properly.

#### Expected Behavior
Icon layer loads properly.

#### Repro Steps
Try this code from Codepen: https://codepen.io/neokore/pen/NWrOpeo

#### Environment
 - Framework Version: deck.gl 8.4.0
 - Browser Version: Chrome and Firefox at least
 - OS: OSX and Linux at least

neokore:
After debugging a bit deeply it looks like it is trying to load the `iconAtlas` image in the `load` function from loaders.gl using the parser for MVT layers, and therefore throws this error code that I relate with an error parsing MVT data, but I'm not completely sure.

kylebarron:
I believe the unimplemented type is coming from the MVTLoader. I think that's a low level error in the Mapbox vector tile parsing code

neokore:
Yeah, but it seems that is deck.gl what sends the config for the loader, because when debugging I got to the `load` function with the image URL and the MVT config in options as parameters, so I thought that this could be the problem, as it may use an "Image loader" or something similar for this type of data, am I right?

Pessimistress:
I know what's going on. I'll put up a fix.

neokore:
That's great! Thank you @Pessimistress!

Pessimistress:
Fixed in 8.4.3

PR #5455, opened 2021-02-10
Fix MVTLayer loaders prop propagation

Pessimistress:
For https://github.com/visgl/deck.gl/issues/5453

When calling `load` with a non-array `loaders` argument it blocks the usage of other registered loaders.

#### Change List
- Make all `loaders` prop defaults arrays.

### DIFF
--- a/modules/geo-layers/src/mvt-layer/mvt-layer.js
+++ b/modules/geo-layers/src/mvt-layer/mvt-layer.js
@@ -15,7 +15,7 @@ const WORLD_SIZE = 512;
 const defaultProps = {
   uniqueIdProperty: {type: 'string', value: ''},
   highlightedFeatureId: null,
-  loaders: MVTLoader,
+  loaders: [MVTLoader],
   binary: false
 };
 
--- a/modules/geo-layers/src/terrain-layer/terrain-layer.js
+++ b/modules/geo-layers/src/terrain-layer/terrain-layer.js
@@ -56,7 +56,7 @@ const defaultProps = {
   wireframe: false,
   material: true,
 
-  loaders: TerrainLoader
+  loaders: [TerrainLoader]
 };
 
 // Turns array of templates into a single string to work around shallow change


ibgreen (approved):
Great, I hit this issue just yesterday, let's see if this fixes it.

kylebarron (approved):

modules/geo-layers/src/mvt-layer/mvt-layer.js
```diff
@@ -15,7 +15,7 @@ const WORLD_SIZE = 512;
 const defaultProps = {
   uniqueIdProperty: {type: 'string', value: ''},
   highlightedFeatureId: null,
-  loaders: MVTLoader,
+  loaders: [MVTLoader],
```
alasarr, line 18: After this change, CartoBQTilerLayer throws:
```
tile-layer.js:18 Error: No valid loader found data: "", contentType: "application/vnd.mapbox-vector-tile" url: https://maps-api-v2.us.carto.com/user/public/bigquery/tileset/1/1/0
    at _callee$ (select-loader.js:29)
    at tryCatch (runtime.js:45)
    at Generator.invoke [as _invoke] (runtime.js:271)
    at Generator.prototype.<computed> [as next] (runtime.js:97)
    at asyncGeneratorStep (asyncToGenerator.js:3)
    at _next (asyncToGenerator.js:25)
    at asyncToGenerator.js:32
    at new Promise (<anonymous>)
    at asyncToGenerator.js:21
    at selectLoader (select-loader.js:13)
```

Taking a look to understand why and to fix it

Pessimistress: When `loaders` is a single loader, it is automatically selected. When it's an array, it's used as one of the candidates. Your endpoint gives no indication that it is a mvt loader compatible format.

Pessimistress: Looked at your error message again, maybe you are not using the latest loader patch?

alasarr: It fails using the scripting... It looks we need to bump @loaders.gl.  https://github.com/visgl/deck.gl/blob/master/yarn.lock#L1786

alasarr: PR added: https://github.com/visgl/deck.gl/pull/5477

PR #5471, opened 2021-02-12
MVT: fix picking non-binary

alasarr:
Picking info was launching an exception when trying to transform to WGS84.


modules/geo-layers/src/mvt-layer/mvt-layer.js
```diff
@@ -164,7 +164,11 @@ export default class MVTLayer extends TileLayer {
 
     if (info.object) {
       if (!isWGS84 && info.object) {
```
Pessimistress, line 166: Excessive `&& info.object`

Pessimistress (approved):

Issue #5472, opened 2021-02-12
Bug: Arc Layer projection issue in 8.4.1


heshan0131:
#### Description
Arc is cut off when zoomed in close. This happens when upgraded to 8.4.1

![ezgif com-gif-maker (7)](https://user-images.githubusercontent.com/3605556/107740302-c1693280-6d45-11eb-82a1-8ef865530e11.gif)

UPD: The issue is in arc-layer-vertex.glsl.js
```
if (useShortestPath) {
      if (nextPos.x < 0.0) {
        currPos.x += TILE_SIZE;
        nextPos.x += TILE_SIZE;
      }
    }
```

It probably doesn’t work as expected with all projection modes. With lower zoom, we use WebMercator projection mode/cord system, with higher zoom switching to Cartesian. It looks like for cartesian the center is where the camera points and everything < 0 is cut shifted.

Should we add `project_uCoordinateSystem != COORDINATE_SYSTEM_CARTESIAN` here?

#### Repro Steps
https://codepen.io/heshan0131/pen/gOLLQzz

#### Environment
 - Framework Version: deck.gl 8.4.1

heshan0131:
@igorDykhta

igorDykhta:
Looking on project.glsl it looks like in this case the shift happens not for Cartesian coord system, but because of auto offset:

`project_uProjectionMode == PROJECTION_MODE_WEB_MERCATOR_AUTO_OFFSET && project_uCoordinateSystem == COORDINATE_SYSTEM_LNGLAT`

Pessimistress:
Fixed in 8.4.5

PR #5478, opened 2021-02-12
Fix ArcLayer with wrapLongitude:true at high zoom

Pessimistress:
For https://github.com/visgl/deck.gl/issues/5472

#### Change List
- Use offset anti meridian in `PROJECTION_MODE.WEB_MERCATOR_AUTO_OFFSET`
- Render tests


modules/layers/src/arc-layer/arc-layer-vertex.glsl.js
```diff
@@ -198,8 +198,14 @@ void main(void) {
     source = project_position(source_world, instanceSourcePositions64Low);
     target = project_position(target_world, instanceTargetPositions64Low);
 
+    // common x at longitude=-180
+    float antiMeridianX = 0.0;
+
     if (useShortestPath) {
-      float thresholdRatio = -source.x / (target.x - source.x);
+      if (project_uProjectionMode == PROJECTION_MODE_WEB_MERCATOR_AUTO_OFFSET) {
```
igorDykhta, line 205: We have the following condition in project.glsl. Not exactly sure, but do we need the same check for antiMeridianX?

```
if (project_uProjectionMode == PROJECTION_MODE_IDENTITY ||
    (project_uProjectionMode == PROJECTION_MODE_WEB_MERCATOR_AUTO_OFFSET &&
    (project_uCoordinateSystem == COORDINATE_SYSTEM_LNGLAT ||
     project_uCoordinateSystem == COORDINATE_SYSTEM_CARTESIAN))) {
    // Subtract high part of 64 bit value. Convert remainder to float32, preserving precision.
  }
```

Pessimistress: `wrapLongitude` is documented as only works with the LNGLAT coordinate system: https://deck.gl/docs/api-reference/core/layer#wraplongitude

igorDykhta (approved):

PR #5486, opened 2021-02-15
pydeck: Add _repr_html_ to the Deck object for automatic rendering in notebooks

ajfriend:
This adds a `_repr_html_()` method to `Deck` so that when a Jupyter notebook tries to display the `Deck` object, a user will see the actual Iframe rendering, instead of the JSON serialization text.

Previously, when a notebook tries to display a `Deck`, it calls its `__repr__()` method, which currently just shows the JSON text. As a user, I'd prefer the default display for the object to be the rendered map itself, rather than this text which is often a blob that can't fit on my screen:

Screen Shot 2021-02-15 at 4 39 58 PM

I'm also lazy, and don't want to have to type `deck.to_html()` if I can avoid it :)

This PR adds `_repr_html_()` ([docs](https://ipython.readthedocs.io/en/stable/config/integrating.html#rich-display)) to `Deck`. When present, Jupyter will use the `_repr_html_` representation instead of the `__repr__`, which in this case displays the HTML/IFrame rendering:

Screen Shot 2021-02-15 at 4 52 07 PM

In a text interface (rich HTML is not available) you'll still get the current text representation:

Screen Shot 2021-02-15 at 5 04 52 PM

Also, in any interface, the text representation is always available via `deck.__repr__()` or `str(deck)` (which calls the former).


ajfriend:
I didn't totally grok the data flow for the HTML, so I just kind of hacked it with `self.to_html(notebook_display=True).data`.

ajfriend:
Also, I dropped the `src="about:blank"` from the `iframe` tag. It didn't seem to be needed.

Happy to add it back in if it is needed for something.

bindings/pydeck/pydeck/io/html.py
```diff
@@ -87,14 +88,30 @@ def display_html(filename):
 
 
 def iframe_with_srcdoc(html_str, width="100%", height=500):
-    width = '"{}"'.format(width) if type(width) == str else width
-    iframe = """<iframe src="about:blank" frameborder="0" srcdoc="{}" width={} height={}></iframe>""".format(
-        html.escape(html_str), width, height
+    if isinstance(width, str):
+        width = '"{}"'.format(width)
```
kylebarron, line 92:
```suggestion
        width = f'"{width}"'
```
I think we require Python 3.6+ now, so we should default to using f-strings.

bindings/pydeck/pydeck/io/html.py
```diff
@@ -87,14 +88,30 @@ def display_html(filename):
...
+
+    iframe = """
+        <iframe
+            width={width}
+            height={height}
+            frameborder="0"
+            srcdoc="{srcdoc}"
+        ></iframe>
+    """
```
kylebarron, line 101: Likewise
```suggestion
    iframe = f"""
        <iframe
            width={width}
            height={height}
            frameborder="0"
            srcdoc="{html.escape(html_str)}"
        ></iframe>
    """
```

bindings/pydeck/pydeck/io/html.py
```diff
@@ -87,14 +88,30 @@ def display_html(filename):
...
+        width=width,
+        height=height,
+        srcdoc=html.escape(html_str),
     )
+
     from IPython.display import HTML  # noqa
 
     with warnings.catch_warnings():
-        warnings.filterwarnings("ignore", message="Consider using IPython.display.iframe instead")
+        msg = "Consider using IPython.display.iframe instead"
```
kylebarron, line 113: Should be capitalized ([doc](https://ipython.readthedocs.io/en/stable/api/generated/IPython.display.html#IPython.display.IFrame))
```suggestion
        msg = "Consider using IPython.display.IFrame instead"
```

kylebarron: Though on second read, this is catching the warning not propagating it, so this should also be fixed wherever the warning is created.

bindings/pydeck/pydeck/bindings/deck.py
```diff
@@ -188,3 +188,8 @@ def to_html(
             **kwargs,
         )
         return f
+
+    def _repr_html_(self):
+        # doesn't actually need the HTML packaging in iframe_with_srcdoc,
+        # so we just take the HTML.data part
+        return self.to_html(notebook_display=True).data
```
kylebarron, line 195: It's not clear to me why you need `.data`. Doesn't `.to_html()` work in general, so it should work inside `_repr_html_`?

bindings/pydeck/pydeck/bindings/deck.py
```diff
@@ -188,3 +188,8 @@ def to_html(
             **kwargs,
         )
         return f
+
+    def _repr_html_(self):
```
kylebarron, line 192: > Note that these are surrounded by single, not double underscores

I had to double check the docs but you're right. I guess double underscores are meant to be reserved for builtins only.

kylebarron (commented):
I think the changes look fine semantically, but this side steps the broader question of what kind of output should exist by default.

Pydeck offers multiple methods of displaying deck.gl. The "primary" method is _not_ an iframe, but rather a "native" Jupyter Widget. This brings several benefits like two-way communication, so that Python can access a clicked feature from deck.gl. This is why pydeck objects have two methods, `.show()` and `.to_html()`. If we add `.to_html()` by default, users might be confused why `pydeck_object` and `pydeck_object.show()` produce different results.

bindings/pydeck/pydeck/io/html.py
ajfriend: Agreed!

bindings/pydeck/pydeck/io/html.py
ajfriend: I couldn't do this because pasting in the html makes it so that the `dedent` no longer works as intended. But that's all just to keep the string formatting pretty. If we don't care about the string formatting, we can just use f-strings. :)

bindings/pydeck/pydeck/bindings/deck.py
ajfriend: Its more like we don't need the `HTML` object that `to_html` returns. The `HTML` object is extra wrapping around the html text, which is all we really need.

I'm just using the minimal requirement here to make it easier for a future refactoring. See my comment above; we can avoid a lot of the code in this path, including the weird warning logic. I kept it out of this PR only to minimize the changes for now.

ajfriend:
> Pydeck offers multiple methods of displaying deck.gl. The "primary" method is _not_ an iframe, but rather a "native" Jupyter Widget. This brings several benefits like [two-way communication and data updates](https://pydeck.gl/jupyter.html), so that Python can access a clicked feature from deck.gl. This is why pydeck objects have two methods, `.show()` and `.to_html()`. If we add `.to_html()` by default, users might be confused why `pydeck_object` and `pydeck_object.show()` produce different results.

Ohhh, yeah. I the widget would be nice. I wonder if there's a corresponding IPython `_repr_*_` function that would default to that in the notebook environment? Or, maybe `_repr_html_` can handle it.

Personally, I haven't used the two-way communication in a while. Does it still require the extra steps beyond just `pip install pydeck`? I'd imagine a good chunk of users don't do the extra work, in which case it might be nice to fall back to the html representation, and otherwise use the widget.

kylebarron:
> Does it still require the extra steps

It shouldn't require any extra steps, but that will only work in Jupyter environments. It won't work in Colab

bindings/pydeck/pydeck/io/html.py
ajduberstein: This is created within the IPython module itself. https://github.com/ipython/ipython/blob/6760a46eb17a32ce1c960e1de0534594bcf03267/IPython/core/display.py#L413

It's supposed to be suppressed–we want the `srcdoc`attribute, which isn't made available by the IFrame object.

bindings/pydeck/pydeck/io/html.py
ajduberstein: I agree with Kyle here, let's prefer f-strings.

bindings/pydeck/tests/io/test_html.py
```diff
@@ -44,10 +45,18 @@ def test_iframe_with_srcdoc():
...
-    iframe = """<iframe src="about:blank" frameborder="0" srcdoc="{escaped_html_str}" width="100%" height=500></iframe>""".format(
-        escaped_html_str=escaped_html_str
-    )
+
+    iframe = """
+        <iframe
+            width="100%"
+            height=500
+            frameborder="0"
+            srcdoc="{escaped_html_str}"
```
ajduberstein, line 54: Interesting that you don't need `src`, apparently called out in the spec [here](https://html.spec.whatwg.org/#an-iframe-srcdoc-document). Good find, @ajfriend

ajfriend:
> It shouldn't require any extra steps, 

Are these instructions for older versions, then?
https://pydeck.gl/installation.html#enabling-pydeck-for-jupyter

And maybe this is a topic for a separate discussion, but I've wondered if the recent [updates to Jupyter extensions](https://jupyterlab.readthedocs.io/en/stable/getting_started/changelog.html#extensions-can-be-installed-without-building-jupyterlab-with-nodejs) would allow us to avoid some of those steps.

kylebarron:
> Are these instructions for older versions, then?

I thought it was changed so those steps were run automatically on install, but I can't remember.

> recent [updates to Jupyter extensions](https://jupyterlab.readthedocs.io/en/stable/getting_started/changelog.html#extensions-can-be-installed-without-building-jupyterlab-with-nodejs) would allow us to avoid some of those steps.

That would be nice but it would mean you could only support Jupyterlab 3.0+

ajduberstein:
> It won't work in Colab, which doesn't support arbitrary Jupyter widgets.

It won't work in streamlit, either. There's a convenience function that you can use to determine if you're probably in Jupyter [here](https://github.com/visgl/deck.gl/blob/293da4b249efa9699ccc7f884637ba79121218ad/bindings/pydeck/pydeck/io/html.py#L15).

I think I'm fine with defaulting to `deck.to_html()` in Jupyter for a cell ending with `deck` and then requiring the explicit syntax (`deck.show()`) for two way communication.

ajduberstein:
Re: commands for Jupyter server, happy to discuss what's going on there in a separate issue in order to keep the conversation here relevant to the PR. The quick answer is that the widget automatically enables in some Jupyter environments but this isn't called out in the docs.

bindings/pydeck/pydeck/io/html.py
ajfriend: done!

ajduberstein (approved):

Issue #5491, opened 2021-02-15
Column-geometry - inconsistent winding order


igorDykhta:
#### Description
With culling enabled the top caps of hexbins are invisible.

#### Expected Behavior
The top cap geometry should be visible for hexbins layer without extrusion.

#### Repro Steps
1) Enable winding for Deck (cull: true)
2) Create h3 layer, looks good with and without extrusion.
3) Create hexbins layer, the top caps are invisible when the layer isn't extruded

Pessimistress:
Fix landed in 8.4.6

PR #5492, opened 2021-02-15
Fix for inconsistent column geometry's top cap winding order

igorDykhta:
For #5491

#### Background
Looks like degenerate triangles for triangle strips are used badly and swap winding order when the column isn't extruded.
The correct way to preserve winding order is to add the last vertex from the previous strip and the first vertex from the next strip. 

#### Change List
- Added one additional vertex to correctly create degenerate triangles and adjusted cap triangles generation.

- [x] update tests & render tests


modules/layers/src/column-layer/column-geometry.js
```diff
@@ -66,8 +66,18 @@ function tesselateColumn(props) {
         i += 3;
       }
     }
+
+    // duplicate the last vertex to create proper degenerate triangle.
+    positions[i + 0] = positions[i - 3];
```
Pessimistress, line 71: Why is this needed? The last two vertices were bottom v0 and top v0, and the next vertex will be top v0.

igorDykhta: > What do we need to put in between, in order to link up the triangles? We’ll need an even number of new triangles in order to preserve the winding. We can do this by repeating the last vertex of the first row, and the first vertex of the second row

https://www.learnopengles.com/tag/degenerate-triangles/

test/render/test-cases/column-layer.js
```diff
@@ -1,6 +1,12 @@
 import {ColumnLayer, GridCellLayer} from '@deck.gl/layers';
+import GL from '@luma.gl/constants';
 import {hexagons, worldGrid} from 'deck.gl-test/data';
 
+const parameters = {
```
Pessimistress, line 5: Call this something like `cullBackParameters` so that we know what it's for

test/render/test-cases/h3-layers.js
```diff
@@ -1,7 +1,13 @@
 import {H3HexagonLayer, H3ClusterLayer} from '@deck.gl/geo-layers';
+import GL from '@luma.gl/constants';
 
 import * as h3 from 'h3-js';
 
+const parameters = {
```
Pessimistress, line 6: I do not think you need to add culling to `H3HexagonLayer` and `HexagonLayer`. They eventually render the `ColumnLayer`. In the spirit of unit tests, the `ColumnLayer` render tests should include:

1. flat with stroke
2. extruded with wireframe

Each combined with

a. default disk geometry
b. externally provided vertices with CCW winding order
c. externally provided vertices with CW winding order

igorDykhta: I see. The issue with winding appeared only for hexbin layer, not for h3, probably because hexbin used the default disk geometry, and such a case wasn't covered.

Also, I would enable culling for all layers where we expect culling to work. Render tests are a good place to catch such culling issues.

examples/layer-browser/src/data-samples.js
```diff
@@ -267,3 +267,7 @@ function getMeshSampleData([xCount, yCount], spacing) {
 }
 
 export const meshSampleData = getMeshSampleData([10, 10], 120);
+
+export const polygonCW = [[1, 1, 0], [1, -1, 0], [-1, -1, 0], [-1, 1, 0]];
+
+export const polygonCCW = [[1, 0, 0], [0.3, 1, 0], [-0.6, 1, 0], [-1, -0.3, 0], [0, -1, 0]];
```
Pessimistress, line 273: Just use the same polygon reversed and they can share the same golden images

examples/layer-browser/src/data-samples.js
```diff
@@ -267,3 +267,7 @@ function getMeshSampleData([xCount, yCount], spacing) {
 }
 
 export const meshSampleData = getMeshSampleData([10, 10], 120);
+
+export const polygonCW = [[1, 1, 0], [1, -1, 0], [-1, -1, 0], [-1, 1, 0]];
```
Pessimistress, line 271: Move to the render test file since this is only used there.

examples/layer-browser/src/data-samples.js
igorDykhta: Did it to test odd/even number of verts

Pessimistress (approved):

PR #5503, opened 2021-02-19
Fix picking color values coming from a reallocated buffer

gabeboning:
When the global cache used for storing picking colors is resized, the old array is released into the pool maintained by `TypedArrayManager`, and can be reallocated for use by some other consumer. After this happens, any layer referencing an old `pickingColorCache` will read incorrect values for picking colors. 

While this PR feels like a less-than-elegant fix, it corrects the problem by ensuring that any layer reading from `pickingColorCache` will always read from the most recently allocated array, which is guaranteed to contain the correct picking colors.


modules/core/src/lib/layer.js
```diff
@@ -551,7 +554,11 @@ export default class Layer extends Component {
   restorePickingColors() {
     const {pickingColors, instancePickingColors} = this.getAttributeManager().attributes;
     const colors = pickingColors || instancePickingColors;
-    colors.updateSubBuffer({startOffset: 0});
+    // The picking color cache may have been freed and then reallocated. This ensures we read from the currently allocated cache.
+    if (this.internalState.usesPickingColorCache) {
+      colors.value = pickingColorCache;
+    }
+    colors.updateSubBuffer({startOffset: 0, endOffset: this.getNumInstances() * 3});
```
Pessimistress, line 561: `this.getNumInstances()` does not necessarily match the size of the picking buffer in all layers. I suggest you keep the `pickingColorCache.subarray` logic just to keep track of the correct size of the attribute.

Pessimistress (commented):
Thank you for digging into this issue!

modules/core/src/lib/layer.js
gabeboning: For my own edification, is there an example of a layer that doesn't provide its own picking colors where `getNumInstances` would differ from the size of the picking buffer? As I'm pretty new to the deck codebase I'm unclear on how the `numInstances` provided to `calculateInstancePickingColors` would not be the same as that returned by `getNumInstances`. 

Regardless; will update to keep the subarray and read length from that as it does certainly seem safer.

Pessimistress: A layer does not have to be instanced; when registering an attribute, it may pass in an option `noAlloc: true` to bypass the default allocation. In that case, `pickingColor` can be any size.

Pessimistress (approved):

modules/core/src/lib/layer.js
```diff
@@ -551,6 +554,10 @@ export default class Layer extends Component {
   restorePickingColors() {
     const {pickingColors, instancePickingColors} = this.getAttributeManager().attributes;
     const colors = pickingColors || instancePickingColors;
+    // The picking color cache may have been freed and then reallocated. This ensures we read from the currently allocated cache.
+    if (this.internalState.usesPickingColorCache) {
```
Pessimistress, line 558: Add a check here to avoid unnecessary `subarray` calls?

```suggestion
    if (this.internalState.usesPickingColorCache && colors.value.buffer !== pickingColorCache.buffer) {
```

Pessimistress:
Oops prettier is unhappy. You need to pull and run `yarn lint fix`.

Pessimistress:
Published 8.4.7.

PR #5507, opened 2021-02-19
Force using a single loader for MVT data

Pessimistress:
Workaround for unrecognized/empty response from the tile server

#### Change List
- Force using a single loader to parse MVT data


ibgreen (approved):
Does it need to be hardcoded so that user cannot override? Can't the layer just default its loaders prop to this single loader?

Pessimistress:
> Does it need to be hardcoded so that user cannot override?

Users can provide their own loaders array, the alternative MVT loader just needs to be the first.

> Can't the layer just default its loaders prop to this single loader?

The default `loaders` value was changed to an array to fix https://github.com/visgl/deck.gl/issues/5453 - `MVTLayer` and `TileLayer` by default passes all props down to the sub layers.

Issue #5139, opened 2020-11-17
Extend PathLayer round joint and no round cap


TextTL:
Hi,

I'm looking for a method to extend PathLayer so that the joints are rounded, but not the cap. I looked at the code, but I don't know how to inject codes to disable this if statement 

  // extend out a triangle to envelope the round cap
  if (isCap) {
    offsetVec = mix(perp * sideOfPath, dir * jointType * 4.0 * flipIfTrue(isStartCap), isJoint);
  }

Any recommendation is really appreciated.

Thanks,

Pessimistress:
This is very hacky, but you can inject into an arbitrary position in a shader:

https://codepen.io/Pessimistress/pen/dyXQBJr

txtl:
@Pessimistress 

It works great! Thank you very much for your help!

txtl:
BTW, it would be nice to turn it into a feature.

Pessimistress:
This will be available in 8.5.0. You can try it now in `8.5.0-alpha.2` with `capRounded` and `jointRounded` props.

Issue #2842, opened 2019-03-26
Path Layer - Line cap support


tomasdev:
#### Target Use case
SVG provides the attribute `stroke-linecap` which allows developers to create paths that look dotted (versus dashed)

Example output
![Screen Shot 2019-03-26 at 6 01 34 PM](https://user-images.githubusercontent.com/364033/55036510-3831a100-4ff1-11e9-90ac-00e7342a8f79.png)

Code
```
<svg width="300px" height="200px" viewBox="0 0 300 200">
    <line x1="40" x2="260" y1="100" y2="100" stroke="#5184AF" stroke-width="20" stroke-linecap="round" stroke-dasharray="1, 30"/>
</svg>
```

#### Proposed feature
Implement line cap support.

#### To Do List 
  - [ ] Add label and assign to milestone 
  - [ ] Coding
  - [ ] Doc update
  - [ ] What’s new update 
  - [ ] Test

tomasdev:
As reference, this is what happens when using `rounded=true` and `dashArray` currently.
![Screen Shot 2019-03-26 at 6 02 31 PM](https://user-images.githubusercontent.com/364033/55036541-58616000-4ff1-11e9-9ea2-1c76acbb50b9.png)

tomasdev:
Very easy to do with Mapbox GL JS API:
```
map.addLayer({
    'id': 'line-animation',
    'type': 'line',
    'source': {
        'type': 'geojson',
        'data': {
            "type": "FeatureCollection",
            "features": [{
                "type": "Feature",
                "geometry": {
                    "type": "LineString",
                    "coordinates": [...]
                }
            }]
        }
    },
    'layout': {
        'line-cap': 'round',
        'line-join': 'round'
    },
    'paint': {
        'line-color': '#ed6498',
        'line-width': 5,
        'line-opacity': .8,
        'line-dasharray': [2, 4]
    }
});
```

I guess I don't actually need to use deck.gl for what I'm doing. I took a look at the source of PathLayer, and seems very complicated to implement there.

Pessimistress:
This will be available in 8.5.0. You can try it now in `8.5.0-alpha.2` with `capRounded: true`.

PR #5508, opened 2021-02-20
PathLayer: separate joint type and cap type props

Pessimistress:
For #5139 #2842

#### Change List
- Deprecate `rounded` prop of `PathLayer`, add `jointRounded` and `capRounded`
- Add `lineCapRounded` prop to `GeoJsonLayer`
- `PathStyleExtension` renders rounded caps for dash segments if `capRounded: true`
- Documentation
- Upgrade guide
- Render tests


kylebarron (approved):

ibgreen (commented):
Late review, but looks good. Do we need a note in the upgrade guide?

PR #5527, opened 2021-02-24
Fix zero height arc rendering between coordinates with different z components

gabeboning:
When arc height is zero, every vertex on the arc will render with the source position's z component. This changes the shader to linearly interpolate when the arc height is zero.


modules/layers/src/arc-layer/arc-layer-vertex.glsl.js
```diff
@@ -63,7 +63,7 @@ float paraboloid(float distance, float sourceZ, float targetZ, float ratio) {
   float dir = step(deltaZ, 0.0);
   float z0 = mix(sourceZ, targetZ, dir);
   float r = mix(ratio, 1.0 - ratio, dir);
-  return sqrt(r * (p2 - r)) * dh + z0;
+  return dh == 0.0 ? sourceZ + deltaZ * ratio : sqrt(r * (p2 - r)) * dh + z0;
```
Pessimistress, line 66: Just add this after line 58?

```glsl
if (dh == 0.0) {
  return sourceZ + deltaZ * ratio;
}
```

gabeboning: Sure, updated to return early.

Pessimistress (approved):

Pessimistress:
Published 8.4.9.

PR #5558, opened 2021-03-04
pydeck: Update binary transport example

ajduberstein:
Binary transport example occasionally crashes because no target is specified in the viewport

#### Change List
-


ibgreen (approved):

bindings/pydeck/examples/binary_transport.py
```diff
@@ -34,13 +34,21 @@ def generate_graph_data(num_nodes, random_seed):
     return force_layout_df
 
 
-def make_renderer(nodes, use_binary_transport=False):
+def make_renderer(nodes: pd.DataFrame, use_binary_transport: bool = False) -> pdk.Deck:
```
kylebarron, line 37: 🎉 Python 3.6+!

kylebarron (approved):

Issue #5600, opened 2021-03-22
LayerManager: onError is set after call to setLayers [8.4.x]


igorDykhta:
#### Description
- An error can happen during setLayers() call, but onError isn't set at that moment and isn't handled properly.

#### Expected Behavior
- onError is set before the call to setLayers()

PR #5601, opened 2021-03-22
[fix] set onError before setLayers

igorDykhta:
For #5600 

#### Change List
- Set onError before setLayers.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,7 +9,7 @@ on:
 
 jobs:
   test-node:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-18.04
 
     steps:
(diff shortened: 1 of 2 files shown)


ibgreen (approved):
Maybe add a comment stating that this needs to be called before other things to ensure we capture all errors?

Pessimistress:
Why is this not merged into master?

igorDykhta:
@Pessimistress on master and 8.5.x we don't call setLayers from there anymore

Pessimistress:
@igorDykhta you identified an issue regarding the call order of setLayers vs onError, but you reordered the settings of onError to be before other operations too. Not merging into master creates a potential behavior discrepancy between master and release branches.

PR #5632, opened 2021-04-04
Update Carto basemap guide about Mapbox-GL v2.0

alexkreidler:
I found that I couldn't use the example as written because `mapbox-gl` and `react-map-gl` now require a token. I updated the docs to link to `react-map-gl`s recommended workaround.

I also updated the link to the Carto Terms of Service.


docs/api-reference/carto/basemap.md
```diff
@@ -2,10 +2,14 @@
 
 [CARTO basemaps](https://carto.com/basemaps/) are available and they can be used without a token.
 
-Ensure you follow the [Terms and Conditions](https://drive.google.com/file/d/0B3OBExqwT6KJNHp3U3VUamx6U1U/view) when using them.
+Ensure you follow the [Terms and Conditions](https://drive.google.com/file/d/1P7bhSE-N9iegI398QYDjKeVhnbS7-Ilk/view) when using them.
```
alasarr, line 5: 👍

docs/api-reference/carto/basemap.md
```diff
@@ -2,10 +2,14 @@
...
 
 
-## Usage
+## React
```
alasarr, line 8: I'd like to keep the usage section and have a subsection for React and Standalone

alexkreidler: Cool, fixed in latest commit.

alasarr (approved):

Issue #5576, opened 2021-03-11
The bitmap layer doesn't work with multi picking


alexandruandrei91:
#### Description
I try to get the picking information from multiple overlapping bitmap layers. When trying to use `pickMultipleObjects` on bitmap layers, I get the following error instead:
```
Uncaught TypeError: colors is undefined
    _disablePickingIndex layer.js:673
    disablePickingIndex layer.js:664
    _pickClosestObject deck-picker.js:252
    pickObject deck-picker.js:64
    _pick deck.js:364
    pickMultipleObjects deck.js:321
    pickMultipleObjects deckgl.js:89
    onClick DeckGLMap.jsx:68
```
 For more details, see this question on the discussion board: https://github.com/visgl/deck.gl/discussions/5574.

#### Expected Behavior
`pickMultipleObjects` returns all the layers, including bitmaps, together with the picking info returned by `getPickingInfo`.

#### Repro Steps
* Create a bitmap layer
* Call `pickMultipleObjects`

https://codepen.io/aalexandru/pen/yLVGgEe?editors=001

#### Environment
 - Framework Version: deck.gl 8.4.9
 - Browser Version: Firefox 86, Chrome 88
 - OS: Windows 10

Pessimistress:
Fixed in 8.4.13

PR #5640, opened 2021-04-08
Fix BitmapLayer crash during multi-depth picking

Pessimistress:
For #5576 

The `BitmapLayer` does not have a `pickingColors`/`instancePickingColors` attribute, which is required by the base `Layer` multi-picking logic.

#### Change List
- Skip drawing in multi-picking after the layer is picked


ibgreen (approved):
That was a simple fix. Nice!

PR #5656, opened 2021-04-13
pydeck: Release 0.6.2

ajduberstein:
For #5446 

#### Change List
- Updates changelog and bumps version number

PR #5657, opened 2021-04-13
pydeck: Change cell height and enable _repr_html_ in Colab

ajduberstein:
Follow up to #5656

#### Change List
-

### DIFF
--- a/bindings/pydeck/pydeck/io/html.py
+++ b/bindings/pydeck/pydeck/io/html.py
@@ -110,7 +110,10 @@ def iframe_with_srcdoc(html_str, width="100%", height=500):
 def render_for_colab(html_str, iframe_height):
     from IPython.display import HTML, Javascript  # noqa
 
-    js_height_snippet = "google.colab.output.setIframeHeight(0, true, {maxHeight: %s})" % iframe_height
+    js_height_snippet = "google.colab.output.setIframeHeight(%d, true, {minHeight: %d})" % (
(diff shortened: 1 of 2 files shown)


bindings/pydeck/pydeck/io/html.py
```diff
@@ -110,7 +110,10 @@ def iframe_with_srcdoc(html_str, width="100%", height=500):
 def render_for_colab(html_str, iframe_height):
     from IPython.display import HTML, Javascript  # noqa
 
-    js_height_snippet = "google.colab.output.setIframeHeight(0, true, {maxHeight: %s})" % iframe_height
+    js_height_snippet = "google.colab.output.setIframeHeight(%d, true, {minHeight: %d})" % (
+        iframe_height,
+        iframe_height,
+    )
```
kylebarron, line 116: Are we requiring Python 3+ yet? 

I see we don't have Python 2 classifiers anymore
https://github.com/visgl/deck.gl/blob/f726e843c8b38c8bb5a3e3e5d948750568dfd625/bindings/pydeck/setup.py#L192-L197

But there isn't a `python_requires` argument to `setup`, so there's no install-time restriction against installing on Python 2.

If we're up to 3.6+, it would be great to use f-strings for string interpolation. You could also use [`pyupgrade`](https://github.com/asottile/pyupgrade) on CI to push for newer Python syntax

kylebarron (approved):

PR #5705, opened 2021-04-28
PathStyleExtension: allow picking gaps between solid lines

xintongxia:
For #

#### Background
- There is a use case to pick dash line either interacts with gap or the solid strokes

#### Change List
- PathStyleExtension: add an option `allowPickGap`, which allows picking on the gap of dash lines

### DIFF
--- a/modules/extensions/src/path-style/path-style.js
+++ b/modules/extensions/src/path-style/path-style.js
@@ -25,7 +25,8 @@ import {dist} from 'gl-matrix/vec3';
 const defaultProps = {
   getDashArray: {type: 'accessor', value: [0, 0]},
   getOffset: {type: 'accessor', value: 0},
-  dashJustified: false
+  dashJustified: false,
+  dashGapPickable: false
 };
 
 export default class PathStyleExtension extends LayerExtension {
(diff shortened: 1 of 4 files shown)


modules/extensions/src/path-style/shaders.glsl.js
```diff
@@ -49,7 +50,9 @@ float round(float x) {
       gapLength > 0.0 &&
       mod(clamp(vPathPosition.y, 0.0, vPathLength) + offset, unitLength) > solidLength
     ) {
-      discard;
+      if (!(allowPickGap && picking_uActive)) {
```
Pessimistress, line 53: Shouldn't discard if `allowPickGap` is `true`?

modules/extensions/src/path-style/path-style.js
```diff
@@ -25,7 +25,8 @@ import {dist} from 'gl-matrix/vec3';
 const defaultProps = {
   getDashArray: {type: 'accessor', value: [0, 0]},
   getOffset: {type: 'accessor', value: 0},
-  dashJustified: false
+  dashJustified: false,
+  allowPickGap: false
```
Pessimistress, line 29: Preferably the prop starts with `dash` so that it's clear that it's from this extension. e.g. `dashGapPickable`?

modules/extensions/src/path-style/shaders.glsl.js
xintongxia: Maybe I miss something, but what I'd like to achieve is allow picking on gaps, so I updated the discard logic, such that when `dashGapPickable` with picking active, the color of gaps won't be discarded.

Pessimistress (approved):

PR #5712, opened 2021-04-30
MVT triangulation

felixpalmer:
Supporting PR for implementing triangulation of polygons for MVTLoader in loaders.gl
#### Background
[RFC](https://github.com/visgl/deck.gl/pull/5685)

#### Change List
- `triangulate` Boolean prop is added to `MVTLayer`

### DIFF
--- a/modules/layers/src/geojson-layer/geojson-layer-props.js
+++ b/modules/layers/src/geojson-layer/geojson-layer-props.js
@@ -93,6 +93,9 @@ export function createLayerPropsFromBinary(geojsonBinary, uniqueIdProperty, enco
     featureIds: polygons.featureIds
   };
   layerProps.polygons._normalize = false;
+  if (polygons.triangles) {
+    layerProps.polygons.data.attributes.indices = polygons.triangles.value;
+  }
 
   layerProps.polygonsOutline.data = {
     length: polygons.primitivePolygonIndices.value.length,


modules/core/src/utils/tesselator.js
```diff
@@ -57,6 +57,7 @@ export default class Tesselator {
     this.getGeometry = getGeometry;
     this.positionSize =
       (geometryBuffer && geometryBuffer.size) || (positionFormat === 'XY' ? 2 : 3);
+    assert(this.positionSize > 1 && this.positionSize < 4);
```
Pessimistress, line 60: In what circumstances could this be untrue? Avoid assertions that may break the app but do not provide any debugging information to the user.

modules/geo-layers/src/mvt-layer/mvt-layer.js
```diff
@@ -16,7 +16,8 @@ const defaultProps = {
   uniqueIdProperty: {type: 'string', value: ''},
   highlightedFeatureId: null,
   loaders: [MVTLoader],
-  binary: false
+  binary: false,
+  triangulate: false
```
Pessimistress, line 20: Is this necessary? Should the layer just request triangulation if `binary: true`?

Any new API must be added to the layer's documentation.

modules/layers/src/geojson-layer/geojson-layer-props.js
```diff
@@ -93,6 +93,9 @@ export function createLayerPropsFromBinary(geojsonBinary, uniqueIdProperty, enco
     featureIds: polygons.featureIds
   };
   layerProps.polygons._normalize = false;
+  if (polygons.triangles) {
+    layerProps.polygons.data.attributes.indices = new Uint32Array(polygons.triangles.value);
```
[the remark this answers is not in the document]
felixpalmer: My bad, I fixed this but forgot to push

modules/core/src/utils/tesselator.js
felixpalmer: While implementing this feature I came across a situation where this.positionSize was somehow -Infinity which led to infinite loops and out of memory crashes which were a pain to debug as the browser would freeze. It is most likely this was due to a bug in the loader I was modifying so you could argue that this assertion is unneeded. On the other hand, were it there it would have saved me a lot of time.

Given that the code assumes this value to be 2 or 3 I felt that having an assertion there to be safe and to validate the input wouldn't hurt. However if you disagree, I'm happy to remove it.

modules/geo-layers/src/mvt-layer/mvt-layer.js
felixpalmer: @alasarr what do you think? I think it also makes sense to always request triangulation if `binary: true`. If we don't then it'll be slower and the holes are broken.

modules/core/src/utils/tesselator.js
Pessimistress: Yes this should be fixed in the loader.

The criteria for adding assertions is whether they provide actionable items for the user to fix errors in their own code. You may add these during development but they should not be shipped to the end user.

felixpalmer: Fixed

modules/geo-layers/src/mvt-layer/mvt-layer.js
felixpalmer: After checking with @alasarr I believe we are all in agreement that it is simpler and logical to always request the triangulation when `binary: true`. As such I've removed this option

Pessimistress (approved):

kylebarron:
Do you need to rebase against master? Why 19 intermediate commits for a 3-line change? Or am I missing something.

kylebarron (approved):

alasarr:
>Do you need to rebase against master? Why 19 intermediate commits for a 3-line change? Or am I missing something.

Yes, we need to rebase and then squash and merge on a single commit.

alasarr (approved):

felixpalmer:
> Do you need to rebase against master? Why 19 intermediate commits for a 3-line change? Or am I missing something.

Many commits as I was learning on the way, and was planning on squashing. I did rebase, but `master` moved on since then. Rebasing now again

Issue #5717, opened 2021-05-02
Only output initialization console log message when running inside a browser


jcready:
#### Target Use case
- Unit tests are not normally run with `NODE_ENV` set to 'production'.
- Unit tests are almost always run inside node.
- Seeing this logged for every test which has a transitive dependency on deck.gl is not helpful.

#### Proposed feature
Only output initialization console log message when running inside a browser.

#### To Do List 
  - [ ] Add label and assign to milestone 
  - [ ] Coding
  - [ ] Doc update
  - [ ] What’s new update 
  - [ ] Test

PR #5718, opened 2021-05-02
Do not log init message unless in browser

jcready:
For #5717

#### Change List
- Only output initialization console log message when running inside a browser.

### DIFF
--- a/modules/core/src/lib/init.js
+++ b/modules/core/src/lib/init.js
@@ -21,7 +21,7 @@
 import {registerLoaders} from '@loaders.gl/core';
 import {ImageLoader} from '@loaders.gl/images';
 
-import {global} from 'probe.gl/env';
+import {global, isBrowserMainThread} from 'probe.gl/env';
 import log from '../utils/log';
 import {register} from '../debug';
 import jsonLoader from '../utils/json-loader';
@@ -41,7 +41,7 @@ if (existingVersion && existingVersion !== version) {
 
 if (!existingVersion) {
   // eslint-disable-next-line
-  if (process.env.NODE_ENV !== 'production') {
+  if (isBrowserMainThread() && process.env.NODE_ENV !== 'production') {
     log.log(
       0,
       `deck.gl ${version} - set deck.log.level=1 (or higher) to trace attribute updates`


modules/core/src/lib/init.js
```diff
@@ -41,7 +41,7 @@ if (existingVersion && existingVersion !== version) {
 
 if (!existingVersion) {
   // eslint-disable-next-line
-  if (process.env.NODE_ENV !== 'production') {
+  if (isBrowserMainThread() && process.env.NODE_ENV !== 'production') {
```
Pessimistress, line 44: Can you just test for `NODE_ENV==='development'` instead?

ibgreen: Rather than adding these checks, perhaps we could just change the priority of the log statement to `1`. This means we can always turn it on with probe.gl but it would not shown by default.

ibgreen (approved):

modules/core/src/lib/init.js
jcready: Happy to make either of those changes. Though setting this log priority to `1` seems to defeat the purpose of the log message since if you already had the log level set to `1` you wouldn't need to see this message.

ibgreen: I think I understood your point now. The advantage of still having the message emitted in all cases is that if someone asks me to debug an issue in a deployed deck.gl application, I can go into the console, set the log level, reload the page, and immediately see what version of deck.gl it was built with.

That works well because the probe.gl logger ensuring that log level changes are persisted in local storage across page reloads.

modules/core/src/lib/init.js
```diff
@@ -40,13 +40,10 @@ if (existingVersion && existingVersion !== version) {
...
-  // eslint-disable-next-line
-  if (process.env.NODE_ENV !== 'production') {
-    log.log(
-      0,
-      `deck.gl ${version} - set deck.log.level=1 (or higher) to trace attribute updates`
-    )();
-  }
+  log.log(
+    1,
+    `deck.gl ${version} - set deck.log.level=1 (or higher) to trace attribute updates`
```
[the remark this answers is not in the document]
ibgreen: Makes sense.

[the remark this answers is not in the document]
jcready: That was the part I was confused about earlier :)
I'll make that change.

modules/core/src/lib/init.js
```diff
@@ -40,13 +40,10 @@ if (existingVersion && existingVersion !== version) {
...
 
 if (!existingVersion) {
-  // eslint-disable-next-line
-  if (process.env.NODE_ENV !== 'production') {
-    log.log(
-      0,
-      `deck.gl ${version} - set deck.log.level=1 (or higher) to trace attribute updates`
-    )();
-  }
+  log.log(
```
ibgreen, line 43: Looks good now.

Nit: This could be a one-liner, so wonder if this will pass the formatting check (`prettier-check`) in the linter. If is doesn't, running `yarn lint fix` should take care of it.

Issue #5775, opened 2021-05-14
MVTLayer binary:true fails with CARTO MVT


alasarr:
#### Description

MVTLayer with binary `true` is not working 

#### Expected Behavior
Should work 

#### Repro Steps
Open the following codepen and set binary to `true`:
https://codepen.io/alasarr/pen/LYWNLwQ

The browser completely freezes

#### Environment
 - Framework Version: main and 8.5.0-alpha.5

felixpalmer:
The build in the example is picking up the `3.0.0-alpha.6` version of the `@loaders.gl/mvt` module. This is due to the dependency being set to that in the `yarn.lock` file: https://github.com/visgl/deck.gl/blob/4f64d78669fac7ce78bb97c1a3ca51f624bebf07/yarn.lock#L2057

The updates in the `8.5.0-alpha.5` deck.gl build require a newer version of `@loaders.gl/mvt`, specifically `alpha.13` or newer. It's not clear to me why the `8.5.0-alpha.5` deck.gl build isn't picking these up and is instead using `alpha.6` - `alpha.14`is the newest version present on npm.

I can confirm that the issue (i.e. browser freezing) is remedied by setting `"@loaders.gl/mvt": "^3.0.0-alpha.14"` in `deck.gl/modules/geo-layers/package.json` and then running your example using `yarn start-local` to create a build that targets the correct `@loaders.gl/mvt` package.

@Pessimistress I see you pushed the `8.5.0-alpha.5` deck.gl build, is there something I should have done differently here to make sure the latest `@loaders.gl/mvt` alpha package was included?

alasarr:
Can you create a PR with the bump of loaders.gl library?

alasarr:
An example of a bump of loaders.gl library can be found at https://github.com/visgl/deck.gl/pull/5559

We did that because of a similar situation.

alasarr:
>@Pessimistress I see you pushed the 8.5.0-alpha.5 deck.gl build, is there something I should have done differently here to make sure the latest @loaders.gl/mvt alpha package was included?

If you add a new feature that requires a new version of a dependency you need to bump that dependency

Pessimistress:
@felixpalmer the Codepen uses the pre-bundled version of deck.gl. The bundle is created while publishing with the dependency versions specified by yarn.lock.

@alasarr Apparently none of the existing tests are failing without upgrading to the latest loaders. You may want to update the test data to capture this.

alasarr:
[...]

PR #5780, opened 2021-05-14
Bump @loaders.gl to 3.0.0-alpha.14

felixpalmer:
Dependency bump to fix https://github.com/visgl/deck.gl/issues/5775

### DIFF
--- a/modules/layers/package.json
+++ b/modules/layers/package.json
@@ -28,13 +28,13 @@
     "prepublishOnly": "npm run build-bundle && npm run build-bundle -- --env.dev"
   },
   "dependencies": {
-    "@loaders.gl/images": "^3.0.0-alpha",
+    "@loaders.gl/images": "^3.0.0-alpha.14",
     "@mapbox/tiny-sdf": "^1.1.0",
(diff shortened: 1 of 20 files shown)


modules/layers/package.json
```diff
@@ -28,13 +28,13 @@
...
-    "@loaders.gl/images": "^3.0.0-alpha",
+    "@loaders.gl/images": "^3.0.0-alpha.14",
     "@mapbox/tiny-sdf": "^1.1.0",
     "@math.gl/polygon": "^3.4.2",
     "earcut": "^2.0.6"
   },
   "peerDependencies": {
     "@deck.gl/core": "^8.0.0",
-    "@loaders.gl/core": "^3.0.0-alpha"
+    "@loaders.gl/core": "^3.0.0-alpha.14"
```
alasarr, line 38: Peer dependencies should stay relaxed.

alasarr (changes requested):

modules/layers/package.json
felixpalmer: Fixed

alasarr (approved):

PR #5803, opened 2021-05-25
Polyfill fetch for node tests

Pessimistress:
#### Change List
- Make `fetch` available in Node tests


alasarr (approved):
Looks great! it fixes the following errors I had in my local env (node v15.12.0, Apple M1):
```
# uncaughtException
not ok 4879 TypeError: Cannot create property 'message' on string 'fetch not available in node'
```

However, for me the tests still hang but I should address that in a separate PR.

ibgreen (approved):

PR #5805, opened 2021-05-25
Bump @loaders.gl to 3.0.0-alpha.16 to pick up earcut migration to math.gl

felixpalmer:
-


alasarr:
Could you add at the description why this bump is required? I suppose it's because of the migration of earcut to math.gl, but to have it here for the changelog

alasarr (approved):

PR #5820, opened 2021-05-26
Allow CompositeLayer to filter sub layers during redraw

Pessimistress:
For #5818

This allows a composite layer to control sub layer visibility based on viewport.

In the TileLayer use case, this improves performance by avoiding the generation of new layers when tile visibilities change.

#### Change List
- Add `CompositeLayer.filterSubLayer` method (see inline JSDoc)
- `TileLayer` sub layers visibility handling

### DIFF
--- a/modules/core/src/lib/composite-layer.js
+++ b/modules/core/src/lib/composite-layer.js
@@ -76,6 +76,18 @@ export default class CompositeLayer extends Layer {
     return null;
   }
 
+  /**
+   * Filters sub layers at draw time
+   * @param {Layer} context.layer - sub layer instance
+   * @param {Viewport} context.viewport - the viewport being rendered in
+   * @param {Boolean} context.isPicking - whether it is a picking pass
+   * @param {String} context.pass - the current pass
+   * @return {Boolean} true if the sub layer should be drawn
+   */
+  filterSubLayer(context) {
+    return true;
+  }
+
   // Returns true if sub layer needs to be rendered
   shouldRenderSubLayer(id, data) {
     const {_subLayerProps: overridingProps} = this.props;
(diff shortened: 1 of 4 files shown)


ibgreen (commented):
Hmm, offer one prop that accepts a function, soon users may ask for all props to accept similar functions. If that is an API style we want to support then that is fine.

But it does seem that it is the multi view system that is pushing us into this, so maybe worth a second think about the problem and alternative, more declarative solutions?

Also I think the Tile3DLayer already handles this by caching all its layer instances and does not generate them each cycle. Not sure if that is a good blueprint for all tiled layers.

Pessimistress:
> Also I think the Tile3DLayer already handles this by caching all its layer instances and does not generate them each cycle. Not sure if that is a good blueprint for all tiled layers.

As discussed in https://github.com/visgl/deck.gl/pull/5758#issuecomment-842857130, when using `Tile3DLayer` with multiple views, the user is required to define a complex `layerFilter` to make sure that the correct sub layers are rendered in a given view. Sub layers are generated and cached once per render cycle, and there is currently no mechanism to determine which sub layers are visible in which view(s). It's an architectual decision that that we do not call `compositeLayer.renderSubLayers` for each viewport, as it would result in addition/removal of layers during each redraw, which would be very bad for perf.

Pessimistress:
> offer one prop that accepts a function, soon users may ask for all props to accept similar functions

Yes, that is a valid concern. One other prop that does this is `highlightColor`, which accepts a callback to determine color by the object picked.

It is hard to support this at infrastructure level, as each of these callbacks are invoked with different contexts at the time of use.

I have also considered keeping this a hidden (undocumented) feature and only use it internally. That would help avoid confusion among end users.

Issue #5815, opened 2021-05-26
Make IconLayer compatible with binary MVTLayer


felixpalmer:
#### Target Use case
The `binary` version of the `MVTLayer` (default as of v8.5) provides a faster way to load data, but only supports `GeoJsonLayer` in the renderSubLayers callback.

#### Proposed feature
`IconLayer` is made compatible with the `binary` `MVTLayer`

#### To Do List 
  - [ ] Add label and assign to milestone 
  - [ ] Coding
  - [ ] Doc update
  - [ ] What’s new update 
  - [ ] Test

felixpalmer:
I've managed to get the position data passed through with a fairly simple modification of `updateState` in `IconLayer`:

```javascript
updateState({oldProps, props, changeFlags}) {
  const {data} = props;
  const binary = data && 'points' in data && 'polygons' in data && 'lines' in data;

  if (binary) {
    const layerProps = createLayerPropsFromBinary(props.data, this.encodePickingColor);
    this.props.data = layerProps.points.data;
  } else {
    super.updateState({props, oldProps, changeFlags});
  }
  ...
}
```

However providing support for accessors like `getSize` is more tricky. It is not possible to simply modify the accessor props like so, as they are read-only

```javascript
updateState({oldProps, props, changeFlags}) {
  ...
  if (binary) {
    const layerProps = createLayerPropsFromBinary(props.data, this.encodePickingColor);
    this.props.data = layerProps.points.data;
    
    this.props.getSize = this._getBinaryAccessor(this.props.getSize);
  } else {
 ...
}
    
...

_getBinaryAccessor(accessor) {
  if (typeof accessor !== 'function') {
    return accessor;
  }

  return (object, info) => {
    const {data, index} = info;
    const feature = binaryToFeatureForAccesor(data, index);
    return accessor(feature, info);
  };
}
```

felixpalmer:
Any input on how to best achieve this @Pessimistress? I've verified by temporarily unfreezing the props that the data and properties end up in the correct place and the layer is drawn, but naturally this is not a solution. I'm not sure design-wise where to best place the `binaryToFeatureForAccesor` conversion function so to the consumer of the library the fact that the binary representation is being used is transparent (like we have in the `GeoJsonLayer`)

Pessimistress:
[...]

PR #5835, opened 2021-06-01
Support pointType prop to allow changing point rendering in GeoJsonLayer

felixpalmer:
Implements https://github.com/visgl/deck.gl/issues/5815

Still need to add tests & update documentation


felixpalmer:
The code has been tested for rendering `MVTLayer`s with all three `pointType` variants, e.g.

```javascript
const layer2a = new MVTLayer({
  id: 'layer2a',
  binary: true,
  data: values[1].urls,
  renderSubLayers: props => {
    return new GeoJsonLayer(props, {
      pointType: 'icon',
      _subLayerProps: {
        points: {
          iconAtlas:
            'https://raw.githubusercontent.com/visgl/deck.gl-data/master/website/icon-atlas.png',
          iconMapping: {
            marker: {x: 0, y: 0, width: 128, height: 128}
          },
          getIcon: () => 'marker',
          sizeScale: 1.5,
          getSize: (d, {index, data}) => {
            return 10 + (d && d.properties.cartodb_id);
          }
        }
      }
    });
  }
});
const layer2b = new MVTLayer({
  id: 'layer2b',
  binary: true,
  data: values[1].urls,
  renderSubLayers: props => {
    return new GeoJsonLayer(props, {
      pointRadiusUnits: 'pixels',
      pointRadiusScale: 10,
      lineWidthMinPixels: 2,
      getLineColor: [255, 0, 255, 255],
      getFillColor: [255, 255, 255, 100]
    });
  }
});
const layer2c = new MVTLayer({
  id: 'layer2c',
  binary: true,
  data: values[1].urls,
  renderSubLayers: props => {
    return new GeoJsonLayer(props, {
      pointType: 'text',
      _subLayerProps: {
        points: {
          getText: d => `cartodb_id: ${d && d.properties.cartodb_id}`,
          getSize: 12,
          getTextAnchor: 'middle'
        }
      }
    });
  }
});
```

produces:

Screen Shot 2021-06-01 at 4 56 10 PM

modules/layers/src/geojson-layer/geojson-layer.js
```diff
@@ -185,7 +189,17 @@ export default class GeoJsonLayer extends CompositeLayer {
     const PolygonFillLayer = this.getSubLayerClass('polygons-fill', SolidPolygonLayer);
     const PolygonStrokeLayer = this.getSubLayerClass('polygons-stroke', PathLayer);
     const LineStringsLayer = this.getSubLayerClass('line-strings', PathLayer);
-    const PointsLayer = this.getSubLayerClass('points', ScatterplotLayer);
+
+    const PointsLayerClassMap = {
```
Pessimistress, line 193: Define this outside as a constant

Pessimistress (commented):
This doesn't solve the problem you identified in the original issue. To fully support other point types we need to add new props to the GeoJsonLayer and drop the need for _subLayerProps.

alasarr (changes requested):
Need to implement the mapping for attributes for the three types of layers. So you can use an icon layer as follow: 

```javascript
const layer2a = new MVTLayer({
  binary: true,
  pointType: 'icon',
  iconAtlas: '...' ,
  ...other icon properties 
});
```

The current implementation uses both pointType + _subLayerProps, so we add the complexity of a new prop and we don't get rid of _subLayerProps.

If you the prop pointType at mvtLayer, it will be propagated to GeoJsonLayer, so you can read there and do the mapping.

modules/layers/src/geojson-layer/geojson-layer.js
felixpalmer: Done

felixpalmer:
I've updated the layer to support passing the props directly to the layer, avoiding `_subLayerProps`. However, a few issues with this:

- There are a large number of props to be passed through, which leads to a lot of boilerplate code
- Passing through `iconAtlas` directly does not work, I had to give the prop a different name
- The API is not consistent, as for the `ScatterplotLayer` some props are prepended with `point`, e.g.[`pointRadiusUnits`](https://deck.gl/docs/api-reference/layers/geojson-layer#pointradiusunits)
- The fact that the props are hardcoded in `GeoJsonLayer` means that if one of the sublayers is updated (a prop is added) then that changes needs to be performed in `GeoJsonLayer` also.

In order to address these I would propose:
- Dynamically loading in of the props of the sublayers, by copying from `SubLayer.defaultProps` to `GeoJsonLayer.defaultProps`
- Prepending `point` to all props that are not shared with other geometry types (e.g. `stroked` or `filled` remain unchanged)
- In a similar fashion, dynamically constructing the data structures needed to pass on the props to the sublayer, rather than having long lists of props

I didn't do this right away as it would be inconsistent with the existing code, so I wanted to sanity check this with you first. Let me know how this sounds @alasarr @Pessimistress

Pessimistress:
It is intentional that GeoJsonLayer explicitly list the props to forward. This way when a sub layer is updated, the GeoJsonLayer may be "missing" a feature, but is never broken.

I suggest that you start by making a table of prop mapping from the GeoJsonLayer to each sub layer, and we can discuss from there. Some initial thoughts:

- Some less-commonly-used, non-accessor props such as `fontSettings`, `alphaCutoff` can be hidden and easily passed through via `_subLayerProps`
- We can deprecate the `radius` props and consolidate them with the `size` props

felixpalmer:
```
GeoJsonLayer prop  | SubLayer prop (\* if same)       | Applicable SubLayer(s)    | Deprecates
--------------------------------------------------------------------------------------------------------
sizeUnits          | \*, ScatterPlot.radiusUnits      | Icon, ScatterPlot & Text  | pointRadiusUnits 
sizeScale          | \*, ScatterPlot.radiusScale      | Icon, ScatterPlot & Text  | pointRadiusScale 
sizeMinPixels      | \*, ScatterPlot.radiusMinPixels  | Icon, ScatterPlot & Text  | pointRadiusMinPixels 
sizeMaxPixels      | \*, ScatterPlot.radiusMaxPixels  | Icon, ScatterPlot & Text  | pointRadiusMaxPixels 
                   |                                  |                           |
alphaCutoff        | \*                               | Icon                      |
billboard          | \*                               | Icon & Text               |
iconAtlas          | \*                               | Icon                      |
iconMapping        | \*                               | Icon                      |
onIconError        | \*                               | Icon                      |
                   |                                  |                           |
lineWidthUnits     | \*                               | ScatterPlot               |
lineWidthScale     | \*                               | ScatterPlot               |
lineWidthMinPixels | \*                               | ScatterPlot               |
lineWidthMaxPixels | \*                               | ScatterPlot               |
stroked            | \*                               | ScatterPlot               |
filled             | \*                               | ScatterPlot               |
                   |                                  |                           |
background         | \*                               | Text                      |
backgroundPadding  | \*                               | Text                      |
characterSet       | \*                               | Text                      |
fontFamily         | \*                               | Text                      |
fontSettings       | \*                               | Text                      |
fontWeight         | \*                               | Text                      |
lineHeight         | \*                               | Text                      |
maxWidth           | \*                               | Text                      |
outlineColor       | \*                               | Text                      |
outlineWidth       | \*                               | Text                      |
wordBreak          | \*                               | Text                      |
```

For accessors the name of sublayer accessor is unchanged

```
GeoJsonLayer accessor  | Applicable SubLayer(s)     
-------------------------------------------------
getPosition            | Icon, ScatterPlot & Text
                       |
getAngle               | Icon & Text
getColor               | Icon & Text
getPixelOffset         | Icon & Text
getSize                | Icon & Text
                       |
getIcon                | Icon
                       |
getFillColor           | ScatterPlot
getLineColor           | ScatterPlot
getRadius              | ScatterPlot
getLineWidth           | ScatterPlot
                       |
getAlignmentBaseline   | Text
getBackgroundColor     | Text
getBorderColor         | Text
getBorderWidth         | Text
getText                | Text
getTextAnchor          | Text
```

felixpalmer:
How about this? I agree that the we should consolidate the naming where possible. I'm less keen on the idea of not including some props as this could trip users up.

What do you think regarding prepending the prop names with "point" e.g. `sizeUnits` would become `pointSizeUnits`? It makes it more explicit and consistent with the props for lines, but at the cost of being more verbose.

alasarr:
Looks good to me.

I'm not a fan of the idea of prepending with `point. I like the prop to be named as in the sublayer.

felixpalmer:
Still need to update docs

Pessimistress:
> I like the prop to be named as in the sublayer.

Many of these props are not clear which sublayers/feature types they apply to. `getPointSize` is much more self explanatory than `getSize` if it only applies to Point features.

Pessimistress:
- There should not be a `getPosition` accessor, position is defined by the GeoJSON spec.
- Existing prop name mappings should not be changed (`getPointRadius` can be deprecated for `getPointSize`, but no new `getRadius` prop)
- `PathLayer` also has a ` billboard` prop and it will be confusing. Since this is not an accessor it can be easily set with `_subLayerProps` instead of exposing multiple GeoJsonLayer props that apply to different feature types.
- Any prop that only apply to one point type should identify it clearly in the name, e.g. `getBackground` -> `getTextBackground`

modules/geo-layers/src/mvt-layer/mvt-layer.js
```diff
@@ -137,7 +139,13 @@ export default class MVTLayer extends TileLayer {
       props.extensions = [...(props.extensions || []), new ClipExtension()];
     }
 
-    return super.renderSubLayers(props);
+    const subLayers = super.renderSubLayers(props);
+
+    if (this.props.binary && !(subLayers instanceof GeoJsonLayer)) {
+      throw new Error('renderSubLayers() must return GeoJsonLayer when using binary:true');
```
Pessimistress, line 145: This will hard crash the layer if e.g. a different copy of `@deck.gl/layers` is imported. Consider using `log.warn` instead.

modules/layers/src/geojson-layer/geojson-layer.js
```diff
@@ -149,43 +204,76 @@ export default class GeoJsonLayer extends CompositeLayer {
...
     } = this.props;
 
     const PolygonFillLayer = this.getSubLayerClass('polygons-fill', SolidPolygonLayer);
     const PolygonStrokeLayer = this.getSubLayerClass('polygons-stroke', PathLayer);
     const LineStringsLayer = this.getSubLayerClass('line-strings', PathLayer);
-    const PointsLayer = this.getSubLayerClass('points', ScatterplotLayer);
+
+    const PointsLayerClass = PointsLayerClassMap[pointType];
+    if (!PointsLayerClass) {
+      throw new Error(`pointType must be one of: ${Object.keys(PointsLayerClassMap)}`);
```
Pessimistress, line 274: Layers usually just fall back to the default if an invalid enum is supplied.

modules/layers/src/geojson-layer/geojson-layer.js
```diff
@@ -294,48 +382,142 @@ export default class GeoJsonLayer extends CompositeLayer {
         layerProps.lines
       );
 
+    let pointLayerProps;
+    let pointLayerUpdateTriggers;
+    if (pointType === 'circle') {
```
Pessimistress, line 387: - Worth splitting into its own function
- Use `switch`

modules/layers/src/geojson-layer/geojson-layer.js
```diff
@@ -294,48 +382,142 @@ export default class GeoJsonLayer extends CompositeLayer {
...
+        radiusMinPixels: sizeMinPixels,
+        radiusMaxPixels: sizeMaxPixels,
+        lineWidthUnits,
+        lineWidthScale,
+        lineWidthMinPixels,
+        lineWidthMaxPixels,
+
+        getFillColor: this.getSubLayerAccessor(getFillColor),
+        getLineColor: this.getSubLayerAccessor(getLineColor),
+        getRadius: this.getSubLayerAccessor(getRadius),
```
Pessimistress, line 402: Why keep the term `radius` here if everything else is renamed to `size`?

modules/layers/src/geojson-layer/geojson-layer.js
```diff
@@ -294,48 +382,142 @@ export default class GeoJsonLayer extends CompositeLayer {
...
+        getRadius: updateTriggers.getRadius,
+        getLineWidth: updateTriggers.getLineWidth
+      };
+    } else if (pointType === 'icon') {
+      pointLayerProps = {
+        alphaCutoff,
+        billboard,
+        iconAtlas,
+        iconMapping,
+        onIconError,
```
Pessimistress, line 425: I suggest we drop `alphaCutoff` and `onIconError` since they are less commonly used and can be easily specified via `_subLayerProps`

modules/layers/src/geojson-layer/geojson-layer.js
```diff
@@ -294,48 +382,142 @@ export default class GeoJsonLayer extends CompositeLayer {
...
+        sizeMaxPixels,
+        sizeMinPixels,
+        sizeScale,
+        sizeUnits,
+
+        getAngle: this.getSubLayerAccessor(getAngle),
+        getColor: this.getSubLayerAccessor(getColor),
```
Pessimistress, line 432: If circle uses `getFillColor` then icon and text should too

modules/layers/src/geojson-layer/geojson-layer.js
felixpalmer: I've changed to `switch` but haven't broken up into functions as it would mean passing through a huge number of variables and would be inconsistent with how the polygon/lines layers above are constructed

modules/layers/src/geojson-layer/geojson-layer.js
felixpalmer: OK, I've removed these

modules/layers/src/geojson-layer/geojson-layer.js
felixpalmer: Done

modules/layers/src/geojson-layer/geojson-layer.js
felixpalmer: I've changed this to getSize for consistency

modules/layers/src/geojson-layer/geojson-layer.js
felixpalmer: Changed to fallback as suggested

modules/geo-layers/src/mvt-layer/mvt-layer.js
felixpalmer: Switched to log.warn

alasarr:
>Many of these props are not clear which sublayers/feature types they apply to. getPointSize is much more self explanatory than getSize if it only applies to Point features.

You're right, it's better

modules/layers/src/geojson-layer/geojson-layer.js
felixpalmer: I've changed it to getSize

felixpalmer:
@Pessimistress I've gone ahead and renamed the props and accessors as per your suggestions. Could you re-review and let me know if you are OK with the choices made? If it looks good I will proceed to updating the documentation

modules/test-utils/src/generate-layer-tests.js
```diff
@@ -146,7 +146,7 @@ function makeAltPropTestCase({propName, propTypes, defaultProps, sampleProps, as
       ];
 
     case 'accessor': {
-      if (typeof defaultProps[propName] === 'function') {
+      if (typeof defaultProps[propName] === 'function' || propDef.optional) {
```
felixpalmer, line 149: As some of the accessors in the GeoJsonLayer don't always get passed on to a sublayer (e.g. `getTextBackground` when `pointType` is not `'text'`) the updateTriggers were not getting triggered, nor were the accessors getting called. This seems like the correct behavior, so I came up with this approach to fix the tests

modules/layers/src/geojson-layer/geojson-layer.js
```diff
@@ -149,43 +192,69 @@ export default class GeoJsonLayer extends CompositeLayer {
     }
   }
 
-  /* eslint-disable complexity */
+  /* eslint-disable complexity, max-statements */
```
Pessimistress, line 195: This function is way too big. My suggestion is that we define the prop mapping in a constant, something like:

```js
const POINT_ICON_PROPS = {
  iconAtlas: 'iconAtlas',
  iconMapping: 'iconMapping',
  getPointAngle: 'getAngle',
  getFillColor: 'getColor',
  getIcon: 'getIcon'
};
```

Then we can use this in multiple places:

```js
function getDefaultProps(mapping, LayerClass) {
  const result = {};
  for (const key in mapping) {
    target[key] = LayerClass.defaultProps[mapping[key]];
  }
  return result;
}

const defaultProps = {
  ...getDefaultProps(POINT_ICON_PROPS, IconLayer),
  ...getDefaultProps(POINT_TEXT_PROPS, TextLayer),
  ...getDefaultProps(POINT_CIRCLE_PROPS, ScatterplotLayer),
  //...
};
```

```js
function getSubLayerProps(layer, mapping) {
  const props = {};
  for (const key in mapping) {
    let value = layer.props[key];
    if (key.startsWith('get')) {
      value = layer.getSubLayerAccessor(value);
    }
    result[mapping[key]] = value;
  }
  return result;
}
```

Pessimistress:
Just keeping track of remaining works:

- [ ] documentation
- [ ] unit tests
- [ ] what's new & upgrade guide
- [ ] update deprecated prop names used in examples & tests

This PR is already pretty big, feel free to split some of these tasks into a follow up PR.

modules/core/src/lib/composite-layer.js
```diff
@@ -136,6 +136,7 @@ export default class CompositeLayer extends Layer {
   }
 
   // Returns sub layer props for a specific sublayer
+  // eslint-disable-next-line complexity
```
alasarr, line 139: I think we can remove this

modules/layers/src/geojson-layer/geojson-layer.js
```diff
@@ -19,53 +19,43 @@
...
+} from './sub-layer-map';
 
 import {getGeojsonFeatures, separateGeojsonFeatures} from './geojson';
 import {createLayerPropsFromFeatures, createLayerPropsFromBinary} from './geojson-layer-props';
 
-const defaultLineColor = [0, 0, 0, 255];
-const defaultFillColor = [0, 0, 0, 255];
-
 const defaultProps = {
+  ...getDefaultProps(POINT_LAYER.circle),
```
alasarr, line 36: 👍

modules/layers/src/geojson-layer/geojson-layer.js
```diff
@@ -19,53 +19,43 @@
...
+  ...getDefaultProps(POINT_LAYER.text),
+  ...getDefaultProps(LINE_LAYER),
+  ...getDefaultProps(POLYGON_LAYER),
+
+  // Overwrite sub layer defaults
   stroked: true,
   filled: true,
   extruded: false,
   wireframe: false,
+  iconAtlas: {type: 'object', value: null},
```
alasarr, line 47: Why do you need to redefine some properties but you don't for others?

For example, you redefine `getIcon`, but you don't redefine `getFillColor`. Is this because getFillColor is already defined at `SolidPolygonLayer`?

alasarr (approved):
Looks great. We add a very nice feature and at the same time we make the code of GeoJsonLayer simpler.

Just added a couple of questions that I'd like to understand about the internals

modules/layers/src/geojson-layer/geojson-layer.js
Pessimistress: `getIcon`'s default is `d => d.icon` which definitely does not apply to GeoJSON.

`getFillColor`'s default is `[0,0,0,255]` which works fine.

`iconAtlas` and `iconMapping` are async props which cannot be directly forwarded.

Issue #5808, opened 2021-05-25
MVTLayer and TerrainLayer loading options


Pessimistress:
#### Description

TileLayer's default loading behavior merges the user-provided `loadOptions` and the [AbortSignal](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) of the tile:

https://github.com/visgl/deck.gl/blob/d5bb542d4b5d6053e61ddd38680b6ed737867a4b/modules/geo-layers/src/tile-layer/tile-layer.js#L28-L38

`MVTLayer` overrides the `getTileData` method with their own implementation. It currently does not use the abort signal - therefore it does not abort tile loading when a tile moves out of view.

`TerrainLayer` supplies `TileLayer` with a `getTileData` callback. The callback currently does not use `loadOptions` or the abort signal.

#### Expected Behavior

`MVTLayer` and `TerrainLayer`'s treatment of loading options should be consistent with that of the `TileLayer`.

Bonus if `props.fetch` is used instead of `load` so that users can override the default loading logic dynamically.

#### Repro Steps

- Open https://codepen.io/Pessimistress/pen/oNZwGLL?editors=001
- Open debug tools - network, examine request headers
- Custom header is not present

#### Environment
 - Framework Version: deck.gl 8.4.16

#### Logs

None

alasarr:
Thanks a lot for the catch @Pessimistress.

@felixpalmer can you take a look at this?

PR #5837, opened 2021-06-03
Pass through loadOptions and AbortSignal to MVTLayer & TerrainLayer

felixpalmer:
Implements https://github.com/visgl/deck.gl/issues/5808

I've tested that the extra header is present in the requests (note that this causes CORS errors in your specific example from terrarium @Pessimistress, but this is a server side problem, the OSM tiles work fine). Additionally, I've verified that when moving the map quickly tile requests do get aborted.

Where possible I rewrote the code to use the Layers `fetch` prop, but decided that it wasn't worth it for the case of loading terrain data as:

- it would be challenging to implement
- the loading of terrain tiles is closely coupled to the layer anyway
- having the `fetch` prop be invoked for two different types of request (imagery & terrain) would be error prone

### DIFF
--- a/modules/geo-layers/src/tile-layer/tile-layer.js
+++ b/modules/geo-layers/src/tile-layer/tile-layer.js
@@ -28,13 +28,13 @@ const defaultProps = {
   fetch: {
     type: 'function',
     value: (url, {layer, loaders, options, signal}) => {
-      const loadOptions = {...options, ...layer.getLoadOptions()};
+      const loadOptions = options || layer.getLoadOptions();
       loadOptions.fetch = {
         ...loadOptions.fetch,
         signal
       };
 
-      return load(url, loaders, loadOptions);
+      return loaders ? load(url, loaders, loadOptions) : load(url, loadOptions);
     },
     compare: false
   },
--- a/modules/geo-layers/src/tile-layer/tile-layer.js
+++ b/modules/geo-layers/src/tile-layer/tile-layer.js
@@ -27,14 +27,14 @@ const defaultProps = {
   // Use load directly so we don't use ResourceManager
   fetch: {
     type: 'function',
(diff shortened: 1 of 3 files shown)


modules/geo-layers/src/tile-layer/tile-layer.js
```diff
@@ -27,14 +27,14 @@ const defaultProps = {
   // Use load directly so we don't use ResourceManager
   fetch: {
     type: 'function',
-    value: (url, {layer, signal}) => {
-      const loadOptions = {...layer.getLoadOptions()};
+    value: (url, {layer, loaders, options, signal}) => {
+      const loadOptions = {...options, ...layer.getLoadOptions()};
```
Pessimistress, line 31: This will overwrite the already merged `mvt` options. Should be `options || layer.getLoadOptions()`?

modules/geo-layers/src/tile-layer/tile-layer.js
```diff
@@ -27,14 +27,14 @@ const defaultProps = {
...
       loadOptions.fetch = {
         ...loadOptions.fetch,
         signal
       };
 
-      return load(url, loadOptions);
+      return load(url, loaders, loadOptions);
```
Pessimistress, line 37: I don't think this works if `loaders` is undefined? Have you tested the terrain example?

felixpalmer: You're right, I misread how the options parsing works. Fixed

modules/geo-layers/src/tile-layer/tile-layer.js
felixpalmer: Done

Pessimistress (approved):

alasarr (approved):

Issue #5851, opened 2021-06-07
examples/website/radio is broken after MVTLayer switches to binary: true


felixpalmer:
#### Description
examples/website/radio is broken after MVTLayer switches to binary: true. An exception is hit in the `getLineColor` accessor, due to data.length having 198 elements but properties being an array of 197 elements

#### Expected Behavior
Example works

PR #5853, opened 2021-06-07
Correct data.length for geojson-layer-props polygons

felixpalmer:
Fix for https://github.com/visgl/deck.gl/issues/5851

This change does fix the issue with the radio example, but I hope I haven't missed something about how the data is then uploaded to the GPU or used elsewhere. In other words, I hope that I haven't fixed an apparent off-by-one error and inadvertently created another elsewhere

### DIFF
--- a/modules/layers/src/geojson-layer/geojson-layer-props.js
+++ b/modules/layers/src/geojson-layer/geojson-layer-props.js
@@ -79,7 +79,7 @@ export function createLayerPropsFromBinary(geojsonBinary, uniqueIdProperty, enco
   layerProps.lines._pathType = 'open';
 
   layerProps.polygons.data = {
-    length: polygons.primitivePolygonIndices.value.length,
+    length: polygons.primitivePolygonIndices.value.length - 1,
     startIndices: polygons.primitivePolygonIndices.value,
     attributes: {
       getPolygon: polygons.positions,
@@ -98,7 +98,7 @@ export function createLayerPropsFromBinary(geojsonBinary, uniqueIdProperty, enco
   }
 
   layerProps.polygonsOutline.data = {
-    length: polygons.primitivePolygonIndices.value.length,
+    length: polygons.primitivePolygonIndices.value.length - 1,
     startIndices: polygons.primitivePolygonIndices.value,
     attributes: {
       getPath: polygons.positions,
--- a/modules/layers/src/geojson-layer/geojson-layer-props.js
+++ b/modules/layers/src/geojson-layer/geojson-layer-props.js
@@ -79,7 +79,8 @@ export function createLayerPropsFromBinary(geojsonBinary, uniqueIdProperty, enco
   layerProps.lines._pathType = 'open';
 
   layerProps.polygons.data = {
-    startIndices: polygons.primitivePolygonIndices.value,
+    length: polygons.polygonIndices.value.length - 1,
+    startIndices: polygons.polygonIndices.value,
     attributes: {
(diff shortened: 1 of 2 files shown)


Pessimistress:
Can this be covered by a test case? You should be able to verify this by checking if `binary: true|false` yield the same `data.length`.

felixpalmer:
@Pessimistress I've added a test case and in doing so came across another issue where the calculation was wrong for polygons with holes

modules/layers/src/geojson-layer/geojson-layer-props.js
```diff
@@ -95,10 +94,13 @@ export function createLayerPropsFromBinary(geojsonBinary, uniqueIdProperty, enco
   layerProps.polygons._normalize = false;
   if (polygons.triangles) {
     layerProps.polygons.data.attributes.indices = polygons.triangles.value;
+    layerProps.polygons.data.length = polygons.polygonIndices.value.length - 1;
```
Pessimistress, line 97: Why is the count of polygons dependent on triangulation?

test/modules/geo-layers/mvt-layer.spec.js
```diff
@@ -457,3 +457,50 @@ test('MVTLayer#triangulation', async t => {
...
+      mvt: {
+        workerUrl: null
+      }
+    }
+  };
+  const testCases = [{props, onAfterUpdate}];
+
+  // Run as separate test runs otherwise data is cached
+  testLayerAsync({Layer: MVTLayer, viewport, testCases, onError: t.notOk});
+  testCases[0].props.binary = false;
```
Pessimistress, line 502: Just add another test case to the array

felixpalmer: Done

modules/layers/src/geojson-layer/geojson-layer-props.js
felixpalmer: Prior to triangulation the length includes the polygons which define holes (in the old code path before we added triangulation in the worker). If we've already triangulated, and thus have the indices attribute we need to set the data.length based on the number of polygon objects (where each object could contain holes)

Pessimistress: Triangulated or not it doesn't change the `properties` array. If `data.length` does not match the original polygon count in the feature collection then you'll end up looking for feature indices that do not exist.

modules/layers/src/geojson-layer/geojson-layer-props.js
```diff
@@ -79,7 +79,6 @@ export function createLayerPropsFromBinary(geojsonBinary, uniqueIdProperty, enco
   layerProps.lines._pathType = 'open';
 
   layerProps.polygons.data = {
-    length: polygons.primitivePolygonIndices.value.length,
     startIndices: polygons.primitivePolygonIndices.value,
```
Pessimistress, line 82: This should be `polygonIndices`?

felixpalmer: Yes, I've fixed it

modules/layers/src/geojson-layer/geojson-layer-props.js
felixpalmer: I was going by what the old code had, but from testing it appears that the old path had it wrong. I've fixed it so that the length is based on the number of features, which is given by the `polygonIndices.length`. The added tests confirm this

Pessimistress (approved):

Issue #5819, opened 2021-05-26
GeoJsonLayer "linestrings" sublayer has incorrectly id


tech-meppem:
#### Link

The docs say the sublayer id for the LineStrings layer is "linestrings", however after testing in 8.4.12, the layer id is actually "line-strings" (with a dash).
https://deck.gl/docs/api-reference/layers/geojson-layer#sub-layers

#### Description

Change the docs to say "line-strings".

PR #5855, opened 2021-06-08
Fix inconsistent path layer id in GeoJSONLayer

Pessimistress:
For #5819

#### Change List
- Fix inconsistent sub layer id
- Documentation

### DIFF
--- a/docs/api-reference/layers/geojson-layer.md
+++ b/docs/api-reference/layers/geojson-layer.md
@@ -280,7 +280,7 @@ The GeoJsonLayer renders the following sublayers:
 
 * `polygons-fill` - a [SolidPolygonLayer](/docs/api-reference/layers/solid-polygon-layer.md) rendering all the `Polygon` and `MultiPolygon` features.
 * `polygons-stroke` - a [PathLayer](/docs/api-reference/layers/path-layer.md) rendering the outline of all the `Polygon` and `MultiPolygon` features. Only rendered if `stroked: true` and `extruded: false`.
-* `linestrings` - a [PathLayer](/docs/api-reference/layers/path-layer.md) rendering all the `LineString` and `MultiLineString` features.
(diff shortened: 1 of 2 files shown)


docs/api-reference/layers/geojson-layer.md
```diff
@@ -280,7 +280,7 @@ The GeoJsonLayer renders the following sublayers:
 
 * `polygons-fill` - a [SolidPolygonLayer](/docs/api-reference/layers/solid-polygon-layer.md) rendering all the `Polygon` and `MultiPolygon` features.
 * `polygons-stroke` - a [PathLayer](/docs/api-reference/layers/path-layer.md) rendering the outline of all the `Polygon` and `MultiPolygon` features. Only rendered if `stroked: true` and `extruded: false`.
-* `linestrings` - a [PathLayer](/docs/api-reference/layers/path-layer.md) rendering all the `LineString` and `MultiLineString` features.
+* `line-strings` - a [PathLayer](/docs/api-reference/layers/path-layer.md) rendering all the `LineString` and `MultiLineString` features.
```
felixpalmer, line 283: This feels wrong to me. In the GeoJson spec LineString is always a single word. With the other layers (e.g. polygon-fill) the dash separates the geometry type and method of drawing. Is there a reason why we can't instead make the sublayer have the id `linestrings`?

felixpalmer (changes requested):

docs/api-reference/layers/geojson-layer.md
Pessimistress: That will be a breaking change.

felixpalmer: Perhaps something to change for a future version?

Pessimistress: Sure. I was a little surprised that it was two words also, but did not think it was important enough to justify a breaking change.

kylebarron (approved):

PR #5868, opened 2021-06-14
Fix: Initialize mapsUrl in cloud native

alasarr:


VictorVelarde (approved):
LGTM. I think it would be nice to add a small test for the case !localCreds.mapsUrl

padawannn (approved):
LGFM

PR #5878, opened 2021-06-16
Add external Timeline to Deck props

chrisgervang:
#### Background
Supply a `Timeline` on deck props to sync internal animations (such as GLTF) with external animations. If not supplied deck will use an internal timeline. This makes it easier to integrate with other animation libraries, such as hubble.gl ([see usage](https://github.com/uber/hubble.gl/pull/112/files#diff-8e8bad12de9a4cf8db92d4ab9e6cc1d372cb351ce776227f536b44a2fa113eedR95))

See https://github.com/uber/hubble.gl/pull/115

#### Change List
- Add optional `_timeline: Timeline` to Deck props.

### DIFF
--- a/docs/api-reference/core/deck.md
+++ b/docs/api-reference/core/deck.md
@@ -297,8 +297,7 @@ The array of effects to be rendered. A lighting effect will be added if an empty
 
 * Default: `null`
 
-(Experimental) Supply an external timeline for animations. If not supplied deck will use an internal timeline.
-
+(Experimental) Supply an external timeline for animations. If not supplied deck will use an internal timeline. See the [luma.gl Timeline](https://luma.gl/docs/api-reference/engine/animation/timeline) documentation.
 
 ### Interaction Settings
 
--- a/docs/api-reference/core/deck.md
+++ b/docs/api-reference/core/deck.md
@@ -293,7 +293,7 @@ The array of effects to be rendered. A lighting effect will be added if an empty
 
 (Experimental) Forces deck.gl to redraw layers every animation frame. Normally deck.gl layers are only redrawn if any change is detected.
 
-##### `_timeline` (Timeline)
+##### `_timeline` ([Timeline](https://luma.gl/docs/api-reference/engine/animation/timeline))
 
 * Default: `null`
 
--- a/docs/api-reference/core/deck.md
+++ b/docs/api-reference/core/deck.md
@@ -293,6 +293,12 @@ The array of effects to be rendered. A lighting effect will be added if an empty
 
 (Experimental) Forces deck.gl to redraw layers every animation frame. Normally deck.gl layers are only redrawn if any change is detected.
 
+##### `_timeline` (Timeline)
+
+* Default: `null`
+
+(Experimental) Supply an external timeline for animations. If not supplied deck will use an internal timeline.
+
 
 ### Interaction Settings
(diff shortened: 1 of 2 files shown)


modules/core/src/lib/deck.js
```diff
@@ -94,6 +94,8 @@ function getPropTypes(PropTypes) {
     _framebuffer: PropTypes.object,
     // Forces a redraw every animation frame
     _animate: PropTypes.bool,
+    // supply an external Timeline
+    _timeline: PropTypes.object,
```
chrisgervang, line 98: Do I need to add an underscore for the first release?

modules/core/src/lib/deck.js
```diff
@@ -287,6 +289,9 @@ export default class Deck {
 
     // Update the animation loop
     this.animationLoop.setProps(resolvedProps);
+    if ('_timeline' in props) {
+      this.animationLoop.attachTimeline(props._timeline);
```
Pessimistress, line 293: This is duplicating L645

docs/api-reference/core/deck.md
```diff
@@ -293,6 +293,12 @@ The array of effects to be rendered. A lighting effect will be added if an empty
 
 (Experimental) Forces deck.gl to redraw layers every animation frame. Normally deck.gl layers are only redrawn if any change is detected.
 
+##### `_timeline` (Timeline)
```
Pessimistress, line 296: Link to luma documentation

modules/core/src/lib/deck.js
chrisgervang: Is L645 rerun though out the lifecycle or only on initialization? The timeline can change at any time if it's supplied externally, so I'm wondering how to update the animationLoop's timeline reference in the event that happens.

Pessimistress: The LayerManager keeps a reference to the timeline which is used for attribute transition and ScenegraphLayer animation. The ViewManager also keeps a reference for view state transitions. If you want to be able to swap out the timeline at any time there will need to be much more code. What is the use case though?

chrisgervang: In my case the timeline may be nulled when a render is finished, and newly constructed when a new render job is triggered. 

Also to avoid surprises, I usually don't expect a prop needing to be treated as a singleton.

chrisgervang: I do see what you mean. Maybe what I'm trying to do can be approached in a different way, since I'm not concerned with view state or layer transitions. What I'm trying to do is set the time of the ScenegraphLayer animation, and interrupt/pause any ongoing transitions when a render is triggered.

chrisgervang: I don't think I need a timeline prop after all. I need a way to control how `this.state.animator.animate` is getting called, and understand the Scenegraph layer animation lifecycle.

https://github.com/visgl/deck.gl/blob/732a1c0ac6f0c3a47634023d69dca34b17b7d442/modules/mesh-layers/src/scenegraph-layer/scenegraph-layer.js#L247-L253

I eventually would like a way to cancel all transitions too (or document how users should approach building visualizations with both transitions and video export features). Searching for `.cancel()` shows the various managers.

chrisgervang:
Closing for now to rethink approach.

PR #5884, opened 2021-06-16
fix viewport force update

peaceshi:
Tiles layer performance issue with layer update

Detiles see: https://github.com/peaceshi/deck.gl/issues/1


Pessimistress:
This code path is only called if view states change: https://github.com/visgl/deck.gl/blob/380772653a9079758169556d620a61b15ca44189/modules/core/src/lib/view-manager.js#L186

If you don't create a new array then the React component will never be able to detect viewport changes.

Issue #5811, opened 2021-05-26
TileLayer Inconsistent Z Calculation


mcost45:
#### Description
Tile-layer uses tileset-2d, which in turn uses the getTilesIndices function within tile-layer's util.js. GetTileIndices calculates the z value from the viewport zoom, which ends up getting passed into the getTileData({x, y, z}) prop for retreiving images:
`const TILE_SIZE = 512;`
`let z = Math.round(viewport.zoom + Math.log2(TILE_SIZE / tileSize));`

Because of this, I ran into a confusing and inconsistent problem with tiled layers behaving differently with tileSizes such as 256 (as in using two separate tiled image sources that use different tile image dimensions, e.g. one has images of 256x256, and the other has images of 512x512), being given the wrong z value to fetch data at, whereas other tiled layer sources would work properly.

For example, if the TileLayer view initialises with a zoom of -1.4 (0 being fully zoomed in & max-resolution), with a tileSize of 256, z will be calculated as `Math.round(1.4 +  Math.log2(512/256))` => 0. Then the TileLayer will fetch from z 0 at a zoom of -1.4, when it should instead be fetching from z -1 - displaying incorrect/stretched tiles on screen.

Conversely, with a tileSize of 512, the TileLayer would fetch the correct level tiles for the zoom as the z calculated from a -1.4 zoom would be -1, not 0.

My work around now involves applying an offset to the z value:
`const decimals = Math.floor(this.viewStates.main.zoom) - this.viewStates.main.zoom;`
`const offset = decimals  < -0.5 ? 1 : 0;`
`const DZIResolutionLevelAtZoomZero = DZIMaxLevel - offset;`

`getTileData(DZIResolutionLevelAtZoomZero + z, x, y)`

#### Expected Behavior
Unless I'm missing something, TileLayers should be able to consistently calculate and pass the correct z value regardless of tileSize, without needing to correct with an offset.

#### Repro Steps
Use TileLayer data sources with tileSizes of 256 and 512 to see the difference in how z get's calculated.

#### Environment
Framework Version: 8.4.16
Browser Version: Chrome 90.0.4430.212
OS: Windows 10.0.19042 Build 19042

#### Logs
None

Pessimistress:
[...]

PR #5886, opened 2021-06-16
Fix `tileSize` for non-geospatial use-case

ilan-gold:
For #5811 - this PR solves the problem by basically reverting #4616 for the non-geospatial use-case.  [As I mentioned](https://github.com/visgl/deck.gl/issues/5811#issuecomment-856895740), though, the only problem now is that I have come to rely on the ability to change the zoom level at which tiles are fetched in the non-geospatial case.

Should I follow this PR up with another that allows the user to offset the zoom level in the non-geospatial instance?  Should this PR not revert the changes from #4616 and just document what people need to do with regards to the `tileSize` prop when they are using the `TileLayer` in the non-geospatial use-case.
#### Change List
- Revert #4616 in the non-geospatial use-case


Pessimistress:
Can you review the `indexing system` and `tileSize` sections of the documentation and see if this change can be explained in a way that is clear and (ideally) inconsistent with the geospatial use case? Can we just reference the DZI spec for non-geospatial the same way we use OSM?

Where `tileSize` does not intuitively define the desired "zoom offset", can we expose something else for the user to customize the layer behavior?

ilan-gold:
@Pessimistress Sounds like a plan.  I'll add the documentation for this PR and open another for the new prop then to control the "zoom offset."

ilan-gold:
Closing in favor of #5895

Issue #5836, opened 2021-06-02
Extra TextLayer fontFamily Description


mcost45:
#### Link
https://deck.gl/docs/api-reference/layers/text-layer#fontfamily

#### Description
Though it makes sense that specified font families must be fully loaded before a text layer is initialised, this is not initially apparent. I.e. if using custom font faces, or a font is not installed on a user's device, and it has not already been loaded previously, the textLayer will fall back to a generic font. From a CSS perspective I am used to setting and forgetting (i.e. letting the browser handle loading a font before swapping it in).

To help remove any confusion or debugging time, I would suggest adding an extra comment after the current description, along the lines of:
'Note that if a specified font-family font is not loaded at the initialisation of the text layer, it will fallback to the next loaded/available font. To guarantee the use of a font, assure that the font-face is loaded before initialisation - awaiting the loading of a [FontFace object](https://developer.mozilla.org/en-US/docs/Web/API/FontFace) and [adding it to the document](https://developer.mozilla.org/en-US/docs/Web/API/Document/fonts) is one way to handle this.'

PR #5888, opened 2021-06-16
Update TextLayer documentation

Pessimistress:
For #5836

#### Change List
- Add remarks regarding web fonts and unicode support


kylebarron (approved):
Thanks!

PR #5889, opened 2021-06-17
TextLayer supports characterSet:auto

Pessimistress:
For #5815 

#### Background

Make it possible to render text labels in MVTLayer without knowing the characters used.

#### Change List
- TextLayer supports `characterSet: 'auto'`
- GeoJsonLayer exposes `textCharacterSet`
- Tests
- Update TextLayer example to use new feature
- Documentation
- What's new

### DIFF
--- a/docs/api-reference/layers/geojson-layer.md
+++ b/docs/api-reference/layers/geojson-layer.md
@@ -313,7 +313,7 @@ The following props are forwarded to a `TextLayer` if `pointType` is `'text'`.
 | `textSizeScale` | `1` | [sizeScale](/docs/api-reference/layers/text-layer.md#sizescale) |
 | `textSizeMinPixels` | `0` | [sizeMinPixels](/docs/api-reference/layers/text-layer.md#sizeminpixels) |
 | `textSizeMaxPixels` | `Number.MAX_SAFE_INTEGER` | [sizeMaxPixels](/docs/api-reference/layers/text-layer.md#sizemaxpixels) |
-| `textCharacterSet` | `'auto'` | [characterSet](/docs/api-reference/layers/text-layer.md#characterset) |
+| `textCharacterSet` | ASCII chars 32-128 | [characterSet](/docs/api-reference/layers/text-layer.md#characterset) |
 | `textFontFamily` | `'Monaco, monospace'` | [fontFamily](/docs/api-reference/layers/text-layer.md#fontfamily) |
 | `textFontWeight` | `'normal'` | [fontWeight](/docs/api-reference/layers/text-layer.md#fontweight) |
 | `textLineHeight` | `1` | [lineHeight](/docs/api-reference/layers/text-layer.md#lineheight) |
--- a/docs/api-reference/layers/geojson-layer.md
+++ b/docs/api-reference/layers/geojson-layer.md
@@ -313,6 +313,7 @@ The following props are forwarded to a `TextLayer` if `pointType` is `'text'`.
 | `textSizeScale` | `1` | [sizeScale](/docs/api-reference/layers/text-layer.md#sizescale) |
 | `textSizeMinPixels` | `0` | [sizeMinPixels](/docs/api-reference/layers/text-layer.md#sizeminpixels) |
 | `textSizeMaxPixels` | `Number.MAX_SAFE_INTEGER` | [sizeMaxPixels](/docs/api-reference/layers/text-layer.md#sizemaxpixels) |
+| `textCharacterSet` | `'auto'` | [characterSet](/docs/api-reference/layers/text-layer.md#characterset) |
 | `textFontFamily` | `'Monaco, monospace'` | [fontFamily](/docs/api-reference/layers/text-layer.md#fontfamily) |
 | `textFontWeight` | `'normal'` | [fontWeight](/docs/api-reference/layers/text-layer.md#fontweight) |
 | `textLineHeight` | `1` | [lineHeight](/docs/api-reference/layers/text-layer.md#lineheight) |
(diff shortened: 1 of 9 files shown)


modules/layers/src/text-layer/text-layer.js
```diff
@@ -201,14 +203,22 @@ export default class TextLayer extends CompositeLayer {
 
       for (const object of iterable) {
         objectInfo.index++;
-        const text = getText(object, objectInfo) || '';
+        const text = Array.from(getText(object, objectInfo) || '');
+        if (autoCharacterSet) {
+          text.forEach(autoCharacterSet.add, autoCharacterSet);
+        }
         // When dealing with double-length unicode characters, `str.length` is incorrect
```
felixpalmer, line 210: Is this comment now obsolete?

docs/api-reference/layers/text-layer.md
```diff
@@ -124,9 +124,16 @@ The padding of the background, `[padding_x, padding_y]` in pixels.
 
 Specifies a prioritized list of one or more font family names and/or generic family names. Follow the specs for CSS [font-family](https://developer.mozilla.org/en-US/docs/Web/CSS/font-family).
 
-##### `characterSet` (Array | String, optional)
+##### `characterSet` (Array | Set | String, optional)
```
felixpalmer, line 127: Is it worth adding a note to `examples/website/tagmap/tagmap-layer/README.md` also?

modules/layers/src/text-layer/text-layer.js
```diff
@@ -108,7 +108,16 @@ export default class TextLayer extends CompositeLayer {
...
+    const textChanged =
+      changeFlags.dataChanged ||
+      (changeFlags.updateTriggersChanged &&
+        (changeFlags.updateTriggersChanged.all || changeFlags.updateTriggersChanged.getText));
+
+    if (textChanged) {
+      this._updateText();
+    }
+
+    const fontChanged = textChanged || this._fontChanged(oldProps, props);
```
felixpalmer, line 120: Is it not possible for the text to change in such a way so that the font is unchanged? I.e. the text grows but no new characters are used. In this case the font atlas will be updated unnecessarily.

modules/layers/src/text-layer/text-layer.js
```diff
@@ -180,17 +179,20 @@ export default class TextLayer extends CompositeLayer {
   // Text strings are variable width objects
```
felixpalmer, line 179: Is this comment still relevant? This method doesn't return anything

docs/api-reference/layers/geojson-layer.md
```diff
@@ -313,6 +313,7 @@ The following props are forwarded to a `TextLayer` if `pointType` is `'text'`.
 | `textSizeScale` | `1` | [sizeScale](/docs/api-reference/layers/text-layer.md#sizescale) |
 | `textSizeMinPixels` | `0` | [sizeMinPixels](/docs/api-reference/layers/text-layer.md#sizeminpixels) |
 | `textSizeMaxPixels` | `Number.MAX_SAFE_INTEGER` | [sizeMaxPixels](/docs/api-reference/layers/text-layer.md#sizemaxpixels) |
+| `textCharacterSet` | `'auto'` | [characterSet](/docs/api-reference/layers/text-layer.md#characterset) |
```
felixpalmer, line 316: Isn't the default value `DEFAULT_CHAR_SET`?

modules/layers/src/text-layer/text-layer.js
```diff
@@ -180,17 +179,20 @@ export default class TextLayer extends CompositeLayer {
...
 
+    const autoCharacterSet = characterSet === 'auto' && new Set(DEFAULT_CHAR_SET);
+
     if (textBuffer && startIndices) {
       const {texts, characterCount} = getTextFromBuffer({
         ...(ArrayBuffer.isView(textBuffer) ? {value: textBuffer} : textBuffer),
         length: data.length,
-        startIndices
+        startIndices,
+        characterSet: autoCharacterSet
```
felixpalmer, line 195: I find it confusing how `autoCharacterSet` is passed in and then mutated by the `getTextFromBuffer` method. I think it would be better to explicitly return a modified version so it's obvious at this point that the method adds characters to the set

modules/layers/src/text-layer/text-layer.js
```diff
@@ -180,17 +179,20 @@ export default class TextLayer extends CompositeLayer {
...
   // Returns the index at the start of each string (every character is rendered by one instance)
   _updateText() {
-    const {data} = this.props;
+    const {data, characterSet} = this.props;
     const textBuffer = data.attributes && data.attributes.getText;
     let {getText} = this.props;
     let {startIndices} = data;
     let numInstances;
 
+    const autoCharacterSet = characterSet === 'auto' && new Set(DEFAULT_CHAR_SET);
```
felixpalmer, line 188: If we are going to automatically generate a character set, do we need to include `DEFAULT_CHAR_SET`?

modules/layers/src/text-layer/text-layer.js
Pessimistress: No, we still are not using string.length

modules/layers/src/text-layer/text-layer.js
Pessimistress: The FontAtlasManager checks for the difference between the existing font map and the incoming `characterSet`. Only new characters are added.

docs/api-reference/layers/text-layer.md
Pessimistress: What kind of note? The example readmes just link to layer documentation.

modules/layers/src/text-layer/text-layer.js
Pessimistress: This is an internal utility. Unnecessarily copying Set objects will add a performance hit.

felixpalmer (approved):

PR #5895, opened 2021-06-21
Don't Use `tileSize` for Determining `z` of Tile Indexing in InfoVis

ilan-gold:
For #5811 - this PR solves the problem by basically reverting #4616 for the non-geospatial use-case.  [As I mentioned](https://github.com/visgl/deck.gl/issues/5811#issuecomment-856895740), though, the only problem now is that I have come to rely on the ability to change the zoom level at which tiles are fetched in the non-geospatial case.

Followed up with #5896 that adds a `zoomOffset` feature to replace the loss of functionality for controlling the zoom level at which tiles are fetched in the `infoVis` use-case.

#### Change List
- Revert #4616 in the non-geospatial use-case

This is a repeat of #5886 but now I can make PR's into my own branches :)


modules/geo-layers/src/tile-layer/utils.js
```diff
@@ -117,8 +117,8 @@ function getIndexingCoords(bbox, scale, modelMatrixInverse) {
   return bbox.map(i => (i * scale) / TILE_SIZE);
 }
 
-function getScale(z) {
-  return Math.pow(2, z);
+function getScale(z, tileSize = TILE_SIZE) {
```
Pessimistress, line 120: As an internal function, I recommend always explicitly setting the arguments for readability. Fallbacks may have unintended consequences.

Pessimistress (approved):

modules/geo-layers/src/tile-layer/utils.js
ilan-gold: Ok yes, seems unnecessary anyway since it is only ever used where `tileSize` would (in theory) exist anyway.

PR #5896, opened 2021-06-21
TileLayer: add zoomOffset prop 

ilan-gold:
#### Background
This adds the feature of `zoomOffset` and updates the docs as well, as mentioned [here](https://github.com/visgl/deck.gl/pull/5886#issuecomment-862863445)
#### Change List
- Add a `zoomOffset` prop to `TileLayer` for infovis applications to offset the `z` level at which tiles are fetched.


docs/api-reference/geo-layers/tile-layer.md
```diff
@@ -82,7 +82,7 @@ At each integer zoom level (`z`), the XY plane in the view space is divided into
 
 When the `TileLayer` is used with a geospatial view such as the [MapView](/docs/api-reference/core/map-view.md), x, y, and z are determined from [the OSM tile index](https://wiki.openstreetmap.org/wiki/Slippy_map_tilenames).
 
-When the `TileLayer` is used used with a non-geospatial view such as the [OrthographicView](/docs/api-reference/core/orthographic-view.md) or the [OrbitView](/docs/api-reference/core/orbit-view.md), `x` and `y` increment from the world origin, and each tile's width and height match that defined by the `tileSize` prop. For example, the tile `x: 0, y: 0` occupies the square between `[0, 0]` and `[tileSize, tileSize]`.
+When the `TileLayer` is used used with a non-geospatial view such as the [OrthographicView](/docs/api-reference/core/orthographic-view.md) or the [OrbitView](/docs/api-reference/core/orbit-view.md), `x` and `y` increment from the world origin, and each tile's width and height match that defined by the `tileSize` prop. For example, the tile `x: 0, y: 0` occupies the square between `[0, 0]` and `[tileSize, tileSize]`.  If you need to offset the `z` level at which the tiles are fetched, there is a `tileOffset` prop.
```
ilan-gold, line 85: This actually seemed more reflective of the state of things now than before.  Before both methods used the OSM index, from what I can tell, but now the infovis obeys the above.

modules/geo-layers/src/tile-layer/utils.js
```diff
@@ -177,11 +177,12 @@ export function getTileIndices({
...
   tileSize = TILE_SIZE,
   modelMatrix,
-  modelMatrixInverse
+  modelMatrixInverse,
+  tileOffset = 0
 }) {
   let z = viewport.isGeospatial
     ? Math.round(viewport.zoom + Math.log2(TILE_SIZE / tileSize))
-    : Math.ceil(viewport.zoom);
+    : Math.ceil(viewport.zoom) + tileOffset;
```
ilan-gold, line 185: Another option here would be `Math.ceil(viewport.zoom + tileOffset)` which is less error-prone for users (since it would also produce a valid `z` value) but would be less interpretable since we apply the `ceil` operation.

docs/api-reference/geo-layers/tile-layer.md
```diff
@@ -138,10 +138,17 @@ if (signal.aborted) {
 
 The pixel dimension of the tiles, usually a power of 2.
 
-The tile size represents the target pixel width and height of each tile when rendered. Smaller tile sizes display the content at higher resolution, while the layer needs to load more tiles to fill the same viewport.
+For geospatial viewports, tile size represents the target pixel width and height of each tile when rendered. Smaller tile sizes display the content at higher resolution, while the layer needs to load more tiles to fill the same viewport.
```
Pessimistress, line 141: I gave this issue some more thought. It might be simpler if we just ask the user to always set `tileSize` to the true pixel size of the tiles.

- For geospatial views, `zoom:0` means the entire Web Mercator world is fit into 512x512. Therefore, in order to display raster tiles at a specified pixel size, we should adjust `zoom` in calculating `tile.z`.
- For non-geospatial views, `zoom:0` means 1 world units ~ 1 pixel, therefore no tileSize-dependent adjustment needs be made to z - i.e. your revert is correct.

modules/geo-layers/src/tile-layer/utils.js
```diff
@@ -177,11 +177,12 @@ export function getTileIndices({
...
   tileSize = TILE_SIZE,
   modelMatrix,
-  modelMatrixInverse
+  modelMatrixInverse,
+  zoomOffset = 0
 }) {
   let z = viewport.isGeospatial
     ? Math.round(viewport.zoom + Math.log2(TILE_SIZE / tileSize))
-    : Math.ceil(viewport.zoom);
+    : Math.ceil(viewport.zoom) + zoomOffset;
```
Pessimistress, line 185: This should apply to both geospatial and non-geospatial. For example, Google Map's zoom is OSM z plus one, and although `tileSize` can be used to achieve the zoom shift, it is counter-intuitive.

docs/api-reference/geo-layers/tile-layer.md
```diff
@@ -138,10 +138,17 @@ if (signal.aborted) {
...
+
+For non-geospatial viewports, the tile size should correspond to the true pixel size of the tiles.
 
 - Default: `512`
 
+##### `zoomOffset` (Number, optional)
+
+For non-geospatial viewports, this offset changes the zoom level at which the tiles are fetched.
```
Pessimistress, line 149: Additional changes: 
- Recommend using `zoomOffset` in upgrade guide
- Update the map-tiles example (`tileSize: 256, zoomOffset: devicePixelRatio === 1 ? -1 : 0`)
- Update the image-tiles example (`data: '${ROOT_URL}/moon.image_files/{z}/{x}_{y}.jpeg', zoomOffset: 15`)

docs/api-reference/geo-layers/tile-layer.md
ilan-gold: Ok @Pessimistress this makes sense with your other comments.  I'll fix these things and re-request review.  Thanks for thinking about this!

ilan-gold: @Pessimistress could you clarify what the first points means?  You're saying similar to what you're saying below, that `zoomOffset` should be used to adjust `zoom`?

Pessimistress: I was just reasoning why we have different logic for geo and non-geo. Basically keep the current geospatial logic but add `zoomOffset` to z.

docs/api-reference/geo-layers/tile-layer.md
```diff
@@ -138,10 +138,17 @@ if (signal.aborted) {
...
+
+- Default: `0`
```
Pessimistress, line 151: Call out that this number needs to be an integer?

ilan-gold: Good call.

docs/api-reference/geo-layers/tile-layer.md
ilan-gold: The last one doesn't quite work that way, I think. I need to think about this a bit.  I think the issue might be the `extent` prop.  I think it assumes that the 0th zoom level is where the full resolution of your tiled image renders, but that wouldn't be the case here.  So this might actually be a bug in what we have at the moment.

ilan-gold: On second thought, this isn't really how I intended this prop to be used.  It is meant more for fetching tiles at a higher or lower resolution given the current zoom level.   So changing this prop independent of other things just changes the resolution of the image but this `+ 15` that currently exists in the `image-tiles` examples is meant just to reconcile their inverted (compared to ours) system.

ilan-gold: The map-tiles use case makes sense to me though, but maybe I am missing something.  There, previously, we were actually intentionally affecting the resolution.

Pessimistress: Maybe we should clamp z to `minZoom`, `maxZoom` before applying any adjustment (both from `tileSize` and `zoomOffset`)? Part of the confusion comes from mixing the terms `z` and `zoom`. Either way it would be good to explain how this works in the `minZoom`, `maxZoom` documentation.

ilan-gold: I think that is a good call.  This affects the level of fetching, not the viewport to which `maxZoom` and `minZoom` should apply.

ilan-gold: One issue this could cause would be if you zoom into zoom level 0 with `maxZoom: 0` and then offset by some number.  The `maxZoom` prop currently prevents fetching at that level but now you would fetch there even though at zoom level -1 with `zoomOffset: 1` you might want to fetch at zoom level 0.  I think it is good as it stands now, but we should update the docs to explain that this clips the fetching, which is what is currently done.

ilan-gold: Re-requesting review.  Added a note to the docs.

docs/api-reference/geo-layers/tile-layer.md
ilan-gold: @Pessimistress Could you propose a change?  I am still not clear on what is being asked here, my apologies.

Pessimistress (approved):

PR #5898, opened 2021-06-21
Fix release workflow

Pessimistress:


ibgreen (approved):
Q: Is this needed in other repos? Not clear to me...

Pessimistress:
@ibgreen Ideally we'll port it to all repos. At this point I'm just trying to make sure it works.

PR #5906, opened 2021-06-24
Add leaflet examples

Pessimistress:
Get-started examples using deck.gl-leaflet by @zakjan

### DIFF
--- a/examples/get-started/pure-js/leaflet/app.js
+++ b/examples/get-started/pure-js/leaflet/app.js
@@ -1,5 +1,6 @@
-import * as Leaflet from 'leaflet';
-import * as DeckGlLeaflet from 'deck.gl-leaflet';
+/* eslint-disable */
+import * as L from 'leaflet';
+import {LeafletLayer} from 'deck.gl-leaflet';
 import {MapView} from '@deck.gl/core';
 import {GeoJsonLayer, ArcLayer} from '@deck.gl/layers';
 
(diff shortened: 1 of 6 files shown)


examples/get-started/pure-js/leaflet/app.js
```diff
@@ -0,0 +1,54 @@
+import * as Leaflet from 'leaflet';
```
kylebarron, line 1: It seems canonical to import leaflet as `L` (or at least that's how it's exposed when included as a script, I think); don't have a strong preference whether we follow that.

kylebarron (approved):

zakjan:
@Pessimistress This is great! Did you have any issues with using deck.gl-leaflet? I know it's missing picking support so far.

I'm going to explore a possibility to move deck.gl-leaflet to a community plugin repo next week (after getting back from holiday). // cc @ibgreen 

Also it's interesting for me to compare the differences and API of existing integration plugins. It might be useful to unify them in a long run.

Pessimistress:
> Also it's interesting for me to compare the differences and API of existing integration plugins. Do you think it might be useful to unify them in a long run?

Definitely. I find myself writing very similar bootstrapping code for other integration project as well.

> it's missing picking support so far.

So I have noticed. Let me know if you need help implementing it!

Pessimistress:
@zakjan A small issue I noticed is this line in deck.gl-leaflet:

```js
import { Deck } from 'deck.gl';
```

It's better to use

```js
import { Deck } from '@deck.gl/core';
```

as well as specifying `@deck.gl/core` as peer dependency. Not all projects install the deck.gl master module.

zakjan:
Thanks, right I forgot about this. I published a new version with this change.

PR #5921, opened 2021-06-29
Add `fadeTrail` prop to TripLayer

chrisgervang:
#### Background
While developing some trip visualizations I wanted to accumulate trips that occur overtime on my screen instead of fading them away as the layer currently does.

Screen Shot 2021-06-29 at 4 06 32 PM

#### Change List
- added an optional `fadeTrail` prop to `TripLayer` to allow the trip fade out to be disabled.


Pessimistress:
Instead of adding a new prop, can we support this effect if `trailLength` is set to 0?

chrisgervang:
Setting it to 0 currently hides all trips, so I think flipping it to show all trips could be a breaking change for some applications. One I'm thinking of is transitions on trailLength where someone grows it from 0 to some value to reveal trips.

Pessimistress:
Hood point. How about negative numbers then?

chrisgervang:
I suppose, though I think negative values could useful for a "look ahead" effect instead. That would take full advantage of the number line instead of treating all negatives as a toggle. Lookahead is something I've seen in streetscape.gl, and I have situations where I can use it for two trip layers: an "actual" and "predicted" layer.
Screen Shot 2021-06-30 at 10 11 21 AM

Pessimistress:
Ok. Boolean props are usually named either `fadeTrail` or `trailFaded`.

examples/website/trips/app.js
```diff
@@ -57,6 +57,7 @@ const landCover = [[[-74.0, 40.7], [-74.02, 40.7], [-74.02, 40.72], [-74.0, 40.7
 export default function App({
   buildings = DATA_URL.BUILDINGS,
   trips = DATA_URL.TRIPS,
+  fadeTrail = false,
```
Pessimistress, line 60: Remove

modules/geo-layers/src/trips-layer/trips-layer.js
```diff
@@ -21,6 +21,7 @@
 import {PathLayer} from '@deck.gl/layers';
 
 const defaultProps = {
+  fadeTrail: {type: 'boolean', value: true},
```
Pessimistress, line 24: `fadeTrail: true` is sufficient. It will be normalized internally.

Pessimistress (approved):

examples/website/trips/app.js
chrisgervang: Oh oops, remove or set to true?

Issue #5929, opened 2021-06-30
Allow scaling of view matrix when constructing WebMercatorViewport


felixpalmer:
#### Target Use case
Match the Google Maps viewport (view & projection matrices)

#### Proposed feature
`WebMercatorViewport` is currently constructed with a `altitude` parameter which is used to construct both the view and the projection matrix:

https://github.com/visgl/deck.gl/blob/b80a46885ba8fd9bc32098050226592ebc44007f/modules/core/src/viewports/web-mercator-viewport.js#L72-L91

In the projection matrix it is used to calculate the field of view, while in the view matrix it is used to translate the camera in the z-direction.

While this is sufficient for matching the Mapbox style matrices (on which the deck.gl viewports are modeled), it makes it impossible to construct the matrices as used by Google Maps. This is because Google Maps positions the view matrix at an `altitude` of `1` (which would correspond to a field of view of 53 degrees) but then constructs a projection matrix which has a field of view of 25 degrees.

It would be good to allow configuration of these two quantities independently, as there is no inherent reason why they must be linked. One way would be to allow scaling the view matrix by a constant (`scaleMultiplier`), but other ways are possible (for example, explicitly passing the field of view)

#### To Do List 
  - [x] Coding
  - [x] Doc update
  - [x] What’s new update 
  - [x] Test

felixpalmer:
In the end a cleaner solution was adopted, where the projectionMatrix can be specified

PR #5930, opened 2021-06-30
WebMercatorViewport support external projectionMatrix

felixpalmer:
Implementation of https://github.com/visgl/deck.gl/issues/5929

### DIFF
--- a/modules/core/src/viewports/web-mercator-viewport.js
+++ b/modules/core/src/viewports/web-mercator-viewport.js
@@ -27,6 +27,7 @@ import {
   getViewMatrix,
   addMetersToLngLat,
   getProjectionParameters,
+  altitudeToFovy,
   fitBounds,
   getBounds
 } from '@math.gl/web-mercator';
@@ -34,8 +35,6 @@ import {
 // TODO - import from math.gl
 import * as vec2 from 'gl-matrix/vec2';
 import {Matrix4} from 'math.gl';
-
-const RADIANS_TO_DEGREES = 180 / Math.PI;
 
 export default class WebMercatorViewport extends Viewport {
   /**
@@ -72,11 +71,12 @@ export default class WebMercatorViewport extends Viewport {
     // TODO - just throw an Error instead?
     altitude = Math.max(0.75, altitude);
 
-    const {fov, aspect, focalDistance, near, far} = getProjectionParameters({
+    const fovy = altitudeToFovy(projectionMatrix ? projectionMatrix[5] / 2 : altitude);
+    const {aspect, fov: fovyRadians, focalDistance, near, far} = getProjectionParameters({
       width,
       height,
       pitch,
-      altitude: projectionMatrix ? projectionMatrix[5] / 2 : altitude,
+      fovy,
       nearZMultiplier,
       farZMultiplier
     });
@@ -112,7 +112,7 @@ export default class WebMercatorViewport extends Viewport {
 
       // projection matrix parameters
       orthographic,
-      fovyRadians: fov,
+      fovyRadians,
       aspect,
       focalDistance,
       near,
@@ -126,7 +126,7 @@ export default class WebMercatorViewport extends Viewport {
     this.pitch = pitch;
     this.bearing = bearing;
     this.altitude = altitude;
-    this.fovy = fov * RADIANS_TO_DEGREES;
+    this.fovy = fovy;
 
     this.orthographic = orthographic;
 
--- a/modules/core/src/viewports/web-mercator-viewport.js
+++ b/modules/core/src/viewports/web-mercator-viewport.js
@@ -53,8 +53,8 @@ export default class WebMercatorViewport extends Viewport {
       bearing = 0,
       nearZMultiplier = 0.1,
       farZMultiplier = 1.01,
-      projectionMatrix,
       orthographic = false,
+      projectionMatrix,
 
       repeat = false,
       worldOffset = 0
@@ -71,16 +71,30 @@ export default class WebMercatorViewport extends Viewport {
     // TODO - just throw an Error instead?
     altitude = Math.max(0.75, altitude);
 
-    const fovy = altitudeToFovy(projectionMatrix ? projectionMatrix[5] / 2 : altitude);
-    const {aspect, fov: fovyRadians, near, far} = getProjectionParameters({
-      width,
-      height,
-      pitch,
-      fovy,
-      nearZMultiplier,
-      farZMultiplier
-    });
-    const focalDistance = altitude;
+    let fovy;
+    let projectionParameters;
+    if (projectionMatrix) {
+      fovy = altitudeToFovy(projectionMatrix[5] / 2);
+      projectionParameters = {focalDistance: altitude, projectionMatrix};
+    } else {
+      fovy = altitudeToFovy(altitude);
+      const {aspect, fov: fovyRadians, near, far} = getProjectionParameters({
+        width,
+        height,
+        pitch,
+        fovy,
+        nearZMultiplier,
+        farZMultiplier
+      });
+      projectionParameters = {
+        orthographic,
+        fovyRadians,
+        aspect,
+        focalDistance: altitude,
+        near,
+        far
+      };
+    }
 
     // The uncentered matrix allows us two move the center addition to the
     // shader (cheap) which gives a coordinate system that has its center in
@@ -112,12 +126,7 @@ export default class WebMercatorViewport extends Viewport {
       zoom,
 
       // projection matrix parameters
-      orthographic,
-      fovyRadians,
-      aspect,
-      focalDistance,
-      near,
-      far
+      ...projectionParameters
     });
 
     // Save parameters
--- a/modules/core/src/viewports/web-mercator-viewport.js
+++ b/modules/core/src/viewports/web-mercator-viewport.js
@@ -35,6 +35,8 @@ import {
 import * as vec2 from 'gl-matrix/vec2';
 import {Matrix4} from 'math.gl';
 
+const RADIANS_TO_DEGREES = 180 / Math.PI;
+
 export default class WebMercatorViewport extends Viewport {
   /**
    * @classdesc
@@ -52,6 +54,7 @@ export default class WebMercatorViewport extends Viewport {
       bearing = 0,
       nearZMultiplier = 0.1,
       farZMultiplier = 1.01,
+      projectionMatrix,
       orthographic = false,
 
       repeat = false,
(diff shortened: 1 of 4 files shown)


felixpalmer:
Proposal for changing WebMercatorViewport construction. If you are happy with this approach I will go ahead and update the docs & add tests

Pessimistress:
Can you not achieve this by offsetting `zoom`?

Pessimistress:
I understand that you want to change camera position, however modifying scale does not just change camera z, but also x and y scaling.

https://github.com/uber-web/math.gl/blob/master/modules/web-mercator/src/web-mercator-utils.js#L176

https://github.com/uber-web/math.gl/blob/master/modules/web-mercator/src/web-mercator-utils.js#L183

`scale = 2^zoom` is used extensively across this code base in shaders and e.g. the TileLayer. Modifying it only in some places will result in many unintentional breakages.

If you just need to control camera position and fov separately, then add a fovy parameter that overrides the fovy calculation from altitude?

felixpalmer:
@Pessimistress how is this?
- I named the parameter `fov` and not `fovy` as `fovy` conflicts with a prop in `modules/core/src/views/view.js `
- `this.altitude` needs to use the same "altitude" as the projection matrix otherwise deck.gl completely crashes (tab hangs)
- As background, I wrote up the approach I took in calculating the matrices to match Google Maps [here](https://github.com/visgl/deck.gl/blob/felix/google-vector/docs/developer-guide/base-map-viewport-alignment.md)

Pessimistress:
> fovy conflicts with a prop in modules/core/src/views/view.js

It should be `fovy`. It is exactly what that prop intends to be. `WebMercatorViewport` just ignored it before.

felixpalmer:
There's an issue with that:

https://github.com/visgl/deck.gl/blob/203a658c840586fa296a91210e24737ef46fae7f/modules/core/src/views/view.js#L154

`viewState` is overwritten with `this.props` so the default `fov` of 50 is picked up from the `View`. Is it OK to switch the precedence of the props, like so:

```javascript
    return new ViewportType({...this.props, ...viewState, ...viewportDimensions}); 
```

`fovy` seems to be the only conflicting prop

Pessimistress:
No, the override order is an important feature. `fovy` is only set to 50 as a fall back in the constructor. You can change the default in the `MapView` constructor to `null` before passing it to the parent class.

felixpalmer:
OK, thanks for the pointer. The problem is that the default `fov` of the View class will be set to 50 even if I don't change it, so I can't use its presence to update the projectionMatrix. Instead, I've come up with another way, which is to simply use the `projectionMatrix`, which is `null` by default

Pessimistress:
Do you still need to make a change then? If you are supplying your own `projectionMatrix` then everything already work as is?

felixpalmer:
If I don't update `this.altitude` I'm seeing the issue I mentioned above with deck.gl freezing when the pitch is increased to 63 degrees

Pessimistress:
What causes the crash? Infinite recursive updates? Any console error? It needs to be explained otherwise future maintainers may very well break this again.

felixpalmer:
`getBounds` passes `this.altitude` to math.gl

https://github.com/visgl/deck.gl/blob/0f65a4a80b21df6e53e1b9874c01771d9fdfd8a3/modules/core/src/viewports/web-mercator-viewport.js#L182

https://github.com/uber-web/math.gl/blob/d699d26ab39f1d415b92aa69247d735efa8e56bc/modules/web-mercator/src/get-bounds.js#L20

This then uses the altitude to obtain the field of view and perform an unprojection. Thus we need the altitude to match that of the projectionMatrix.

The crash is then due an invalid bounds object in `get subViewports()`

https://github.com/visgl/deck.gl/blob/0f65a4a80b21df6e53e1b9874c01771d9fdfd8a3/modules/core/src/viewports/web-mercator-viewport.js#L138-L147

As the projection is wrong the `minOffset` generated is -1706482542 and `maxOffset` is 1706482541 which leads the method to start creating a huge number of subviewports, each of which also creates more subviewports.

Pessimistress:
`this.altitude` should always reflect `opts.altitude`. Swapping the value out after it's used for initialization is a hack and prone to future issues. Here is my proposal:

- add a field `this.fovy` to `WebMercatorViewport`, calculated from either `altitude` (default case) or `projectionMatrix` (this use case)
- math.gl's `getBounds` should look for `fovy` if defined and fall back to using `altitude` in its calculations

felixpalmer:
I like this approach. I will create a PR in math.gl and then update here

modules/core/src/viewports/web-mercator-viewport.js
```diff
@@ -68,12 +71,13 @@ export default class WebMercatorViewport extends Viewport {
...
     // TODO - just throw an Error instead?
     altitude = Math.max(0.75, altitude);
+    const altitudeFromProjection = projectionMatrix && projectionMatrix[5] / 2;
 
     const {fov, aspect, focalDistance, near, far} = getProjectionParameters({
       width,
       height,
       pitch,
-      altitude,
+      altitude: projectionMatrix ? altitudeFromProjection : altitude,
```
Pessimistress, line 80:
```suggestion
      altitude: projectionMatrix ? projectionMatrix[5] / 2 : altitude,
```

modules/core/src/viewports/web-mercator-viewport.js
```diff
@@ -73,7 +76,7 @@ export default class WebMercatorViewport extends Viewport {
       width,
       height,
       pitch,
-      altitude,
+      altitude: projectionMatrix ? projectionMatrix[5] / 2 : altitude,
```
Pessimistress, line 79: I thought this was deprecated? Let's spell out the underlying logic here:

```js
const fov = altitudeToFov(projectionMatrix ? projectionMatrix[5] / 2 : altitude);

getProjectionParameters({fov, ...});
```

Pessimistress (approved):

modules/core/src/viewports/web-mercator-viewport.js
felixpalmer: Fixed, but requires a new math.gl release as these function were not exported

modules/core/src/viewports/web-mercator-viewport.js
```diff
@@ -109,7 +112,7 @@ export default class WebMercatorViewport extends Viewport {
 
       // projection matrix parameters
       orthographic,
-      fovyRadians: fov,
+      fovyRadians,
       aspect,
       focalDistance,
```
Pessimistress, line 114: I think `focalDistance` always needs to be set to `altitude`. You can test its effect with sizeUnits/radiusUnits/widthUnits: 'pixels'.

felixpalmer: Even if I change that the size of the points is incorrect (they are too large by a factor of about 2.25) when I test it against the Google projection. `2.2553542518310286` I recognize as `fovyToAltitude(fovy)` for that projection (while `altitude` is `1`, so my feeling is that there is somewhere in deck.gl that making the assumption that `fovyToAltitude(fovy) === altitude` but I haven't yet found where

felixpalmer: I've set focalDistance to always equal altitude, as (as per offline discussion) it turns out we can make this work with the Google matrices if we adjust the near plane

modules/core/src/viewports/web-mercator-viewport.js
```diff
@@ -69,14 +71,16 @@ export default class WebMercatorViewport extends Viewport {
     // TODO - just throw an Error instead?
     altitude = Math.max(0.75, altitude);
 
-    const {fov, aspect, focalDistance, near, far} = getProjectionParameters({
+    const fovy = altitudeToFovy(projectionMatrix ? projectionMatrix[5] / 2 : altitude);
+    const {aspect, fov: fovyRadians, near, far} = getProjectionParameters({
```
Pessimistress, line 75: Only run this if `projectionMatrix` is not supplied?

```js
let fovy;
if (projectionMatrix) {
  fovy = altitudeToFovy(projectionMatrix[5] / 2);
} else {
  fovy = altitudeToFovy(altitude);
  projectionMatrix = getProjectionMatrix(...);
}
```

modules/core/src/viewports/web-mercator-viewport.js
```diff
@@ -109,7 +113,7 @@ export default class WebMercatorViewport extends Viewport {
 
       // projection matrix parameters
       orthographic,
-      fovyRadians: fov,
+      fovyRadians,
       aspect,
       focalDistance,
```
Pessimistress, line 114: Nit: `focalDistance: altitude`

modules/core/src/viewports/web-mercator-viewport.js
felixpalmer: I've done this refactor, but with a slightly different approach so as to support the orthographic code path

modules/core/src/viewports/web-mercator-viewport.js
```diff
@@ -108,12 +126,7 @@ export default class WebMercatorViewport extends Viewport {
       zoom,
 
       // projection matrix parameters
-      orthographic,
-      fovyRadians: fov,
-      aspect,
-      focalDistance,
-      near,
-      far
+      ...projectionParameters
```
Pessimistress, line 129: focalDistance is not a projection parameter. Move it here?

modules/core/src/viewports/web-mercator-viewport.js
```diff
@@ -69,14 +71,30 @@ export default class WebMercatorViewport extends Viewport {
...
-      farZMultiplier
-    });
+    let fovy;
+    let projectionParameters;
+    if (projectionMatrix) {
+      fovy = altitudeToFovy(projectionMatrix[5] / 2);
+      projectionParameters = {focalDistance: altitude, projectionMatrix};
+    } else {
+      fovy = altitudeToFovy(altitude);
+      const {aspect, fov: fovyRadians, near, far} = getProjectionParameters({
```
Pessimistress, line 81: Perf: you can avoid destructing/constructing objects here:

```js
let projectionParameters = null;
if (projectionMatrix) {
  fovy = altitudeToFovy(projectionMatrix[5] / 2);
} else {
  fovy = altitudeToFovy(altitude);
  projectionParameters = getProjectionParameters({...});
}

super({
  // prev opts
  ...projectionParameters,
  focalDistance: altitude
});
```

Note that `orthographic` and `projectionMatrix` are already passed by `...opts`.

felixpalmer: Fixed & removed `orthographic`

Issue #5797, opened 2021-05-21
Effects Break Multiple Views


mcost45:
#### Description
When applying effects (I have only tested PostProcessEffects) to Deck using multiple views, the multiple views will break. It seems to me that when applying effects, the last view in the array of views is taken, then all the 'views' are rendered within the dimensions and positions of that final view.

I have multiple views with a minimap. In the case of having my minimap as the final views array element, everything renders within the dimensions of that view:
![image](https://user-images.githubusercontent.com/17732915/119127625-d801e280-ba34-11eb-8caa-ce0cc62f249a.png)

And in the case of reversing the order, to have the main view as the final views array element: - the positioning is fixed as the main view uses 100% of the screen area, but now the depth order is wrong:
![image](https://user-images.githubusercontent.com/17732915/119128149-8f96f480-ba35-11eb-9614-313873f572e4.png)

#### Expected Behavior
Applying effects should not break views. Perhaps also effects could be applied to individual views instead as a Deck prop.

#### Repro Steps
- Set views prop to array of two views with different sizes and/or positions (i.e. minimap / main view)
- Set effects prop to (a non-empty) array of any effects (in my case PostProcessEffects brightnessContrast and/or hueSaturation)

#### Environment
 - Framework Version: 8.4.16
 - Browser Version: Chrome  90.0.4430.212
 - OS: Windows 10.0.19042 Build 19042

#### Logs
None

Pessimistress:
Fix is published in 8.4.20

PR #5935, opened 2021-07-01
Fix PostProcessEffect in multi-view

Pessimistress:
For #5797

#### Change List
- Reset viewport in `ScreenPass`

### DIFF
--- a/modules/core/src/passes/screen-pass.js
+++ b/modules/core/src/passes/screen-pass.js
@@ -4,7 +4,7 @@
 // Attribution: This class and the multipass system were inspired by
 // the THREE.js EffectComposer and *Pass classes
 
-import {ClipSpace, withParameters, clear} from '@luma.gl/core';
+import {ClipSpace, setParameters, withParameters, clear} from '@luma.gl/core';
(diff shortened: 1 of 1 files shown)


modules/core/src/passes/screen-pass.js
```diff
@@ -17,6 +17,8 @@ export default class ScreenPass extends Pass {
   render(params) {
     const gl = this.gl;
 
+    setParameters(gl, {viewport: [0, 0, gl.drawingBufferWidth, gl.drawingBufferHeight]});
+
```
felixpalmer, line 21: Why do you need to use `setParameters` rather that adding to the state changes within `withParameters`?

felixpalmer (approved):

modules/core/src/passes/screen-pass.js
Pessimistress: The idea is that `viewport` must be set before each pass any ways. Using `withParameter` it is reset after render, which result in two GPU calls instead of one. It is more important in the layers pass as an optimization, but not that important here.

PR #5937, opened 2021-07-01
Support ArcGIS globe

zakjan:
#### Background

ArcGIS supports rendering a globe with SceneView.viewingMode = global. This PR tries to add support for ArcGIS globe to @deck.gl/arcgis

#### Change List
- compute viewState depending on view.viewingMode in @deck.gl/arcgis/deck-renderer
- add new example arcgis-globe


felixpalmer:
Can you give more details on *how* the view state is incorrect? I'm doing some work on syncing the Google Maps view state and will be writing up a doc on this soon. Perhaps it'll be useful for your case also

zakjan:
It's incorrect on my side, I haven't figured out yet the correct computation from ArcGIS camera to deck globe viewState. See the four TODOs at https://github.com/zakjan/deck.gl/blob/e423ea56836eaa80ee048d49f42f04f5ac418c53/modules/arcgis/src/deck-renderer.js#L44-L56

There is a new example arcgis-globe, where it can be tested.

zakjan:
Current state: deck.gl globe overlay matches **roughly** with ArcGIS globe. ArcGIS globe features disabled in demo: tilt, rotation, sun.

It's still incorrect though, I haven't found out the correct calculations yet. The main challenge is computing deck.gl altitude and zoom from ArcGIS altitude in meters. I'm not sure why deck.gl requires both altitude and zoom.

Also, for precise match there might be an issue that Earth radiuses are different, not configurable. [deck.gl uses 6370972](https://github.com/visgl/deck.gl/blob/743e230a3ae30f4ba433ed2e9fbb99a7fba1c99f/modules/core/src/viewports/globe-viewport.js#L10), [ArcGIS uses 6378137](https://developers.arcgis.com/javascript/latest/api-reference/esri-views-3d-externalRenderers.html).

Screen Shot 2021-07-06 at 13 09 13
Screen Shot 2021-07-06 at 13 09 36
Screen Shot 2021-07-06 at 13 08 55

felixpalmer:
Altitude is used both for the projection (think of it as another way to specify FOV) and the view matrix. A similar approach is taken by Mapbox, but this isn't always what other libraries do. For background, you might want to take a look at:

https://github.com/visgl/deck.gl/blob/felix/google-vector/docs/developer-guide/base-map-viewport-alignment.md

https://github.com/visgl/deck.gl/pull/5930

zakjan:
Thanks for the fresh new guide, I'll go through it! Because ArcGIS altitude is in meters, it seems that it serves both purposes, defining FOV and map zoom level.

GemT27:
HI, is there any progress on this issue

PR #5962, opened 2021-07-07
Enforce OrbitView orbitAxis prop

Pessimistress:
For #5948 

`orbitAxis` has always been documented as a `OrbitView` prop, but in practice it could be set in the view state as well. This lead to various inconsistency bugs when we handle view state across viewport, controller and transition interpolators. 

This PR drops the default value from `OrbitController` and `OrbitViewport` for clarity. `orbitAxis` is now set once and for all in the `OrbitView` constructor.

#### Change List
- Set `orbitAxis` fall back in `OrbitView` constructor
- Upgrade guide
- Fix point cloud example

### DIFF
--- a/docs/upgrade-guide.md
+++ b/docs/upgrade-guide.md
@@ -49,6 +49,12 @@ The module entry point is now only lightly transpiled for the most commonly used
 
 `Deck`'s default `onError` callback is changed to `console.error`. Explicitly setting `onError` to `null` now silently ignores all errors, instead of logging them to console.
 
+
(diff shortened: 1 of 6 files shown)


kylebarron (approved):
lgtm though I'm not very familiar with this part of the codebase

Pessimistress:
> What would happen if you wanted to change it?

You can still change the axis by changing Deck's `views` prop. The issue here is that the OrbitView instance is used as the source of truth for all viewports.

PR #5972, opened 2021-07-15
Bump math.gl to 3.5.1

felixpalmer:

### DIFF
--- a/modules/aggregation-layers/package.json
+++ b/modules/aggregation-layers/package.json
@@ -6,23 +6,14 @@
   "publishConfig": {
     "access": "public"
   },
-  "keywords": [
-    "webgl",
-    "visualization",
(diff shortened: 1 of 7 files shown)


modules/aggregation-layers/package.json
```diff
@@ -6,31 +6,22 @@
...
     "url": "https://github.com/visgl/deck.gl.git"
   },
   "main": "dist/es5/index.js",
   "module": "dist/esm/index.js",
-  "files": [
-    "dist",
-    "src",
-    "dist.min.js"
-  ],
+  "files": ["dist", "src", "dist.min.js"],
```
Pessimistress, line 16: Can we avoid changing these unrelated lines?

Pessimistress (approved):

Issue #5980, opened 2021-07-16
Support Vector maps in @google module


felixpalmer:
#### Target Use case

Currently there is a @deck.gl/google-maps module which allows the user to synchronize the Google basemap with deck.gl so that geo-referenced deck.gl layers float on top of Google maps.

The implementation uses the raster version of Google maps, which until recently was the only one available.

Recently, Google launched a beta version of [WebGL enabled Vector Maps API](https://developers.google.com/maps/documentation/javascript/vector-map), which aside from bringing an improvement from the perspective of the user, offer a much richer API for applications to integrate with

#### Proposed feature
Extend the google module so that the vector map is supported and the deck.gl overlay is displayed within the 3D google map.

<img width="662" alt="Screen Shot 2021-07-16 at 3 40 27 PM" src="https://user-images.githubusercontent.com/453755/125957469-7090afe3-90ab-466b-8e76-5f23f04be03b.png">
<img width="667" alt="Screen Shot 2021-07-16 at 3 33 36 PM" src="https://user-images.githubusercontent.com/453755/125957477-ee2adeaf-895d-4ec3-8508-58a27dace92a.png">

https://user-images.githubusercontent.com/453755/125957809-a159ac08-3dd2-4411-b3d1-9a9e4b87c213.mov

#### To Do List 
  - [ ] Switch base map to vector
  - [ ] Migrate to new WebglOverlayView but still draw deck.gl in separate context
  - [ ] Using Mapbox overlay as an reference, move the rendering of the deck.gl layers into the gl context of Google maps
  - [ ] Use matrices from Google (view and projection) to correctly set projection with z-buffer matched
  - [ ] Confirm tilt and rotation of camera works
  - [ ] Handle mouse events (click, doubleclick, mouse move & mouse out)
  - [ ] Support interleaving (e.g. between Google buildings and deck markers)
  - [ ] Make sure animation triggered in Google maps works (deck.gl layers are also animated)
  - [ ] Doc update
  - [ ] What’s new update
  - [ ] Documentation on general basemap integration (view/projection matrix synchronization)
  - [ ] New examples
  - [ ] Test

PR #5981, opened 2021-07-16
Support vector maps in google module

felixpalmer:
Implementation of https://github.com/visgl/deck.gl/issues/5980.

I still need to add tests and examples, but I would like to get some feedback on the general approach taken before starting on this.

I would also appreciate any pointers on how to do deal with recovering from a WebGL context loss/restore


felixpalmer:
The examples in `app2.js` & `app3.js` are more as an aid for development/debugging and should not be reviewed

felixpalmer:
Changes made:

- Switch base map to vector
- Migrate to new WebglOverlayView
- Using Mapbox overlay as an reference, move the rendering of the deck.gl layers into the gl context of Google maps
- Use matrices from Google (view and projection) to correctly set projection with z-buffer matched
- Automatically fallback to old overlay if using a raster map (this can only be determined at runtime as it is configured in the Google Cloud Console)
- Confirm tilt and rotation of camera works
- Handle mouse events (click, doubleclick, mouse move & mouse out)
- Support interleaving (e.g. between Google buildings and deck markers)
- Make sure animation triggered in Google maps works (deck.gl layers are also animated)

modules/google-maps/src/utils.js
```diff
@@ -145,8 +150,75 @@ export function getViewState(map, overlay) {
...
+  });
+  const focalDistance = 0.5 * projectionMatrix[5];
+
+  return {
+    width,
+    height,
+    views: [
+      new MapView({
+        id: 'google-maps-overlay-view',
+        projectionMatrix
```
Pessimistress, line 196: Just place projectionMatrix in viewState?

examples/get-started/pure-js/google-maps/app2.js
```diff
@@ -0,0 +1,74 @@
+/* global document, google, window */
```
Pessimistress, line 1: The get-started examples are intended to be boilerplates, not showcases of library capabilities. Let's keep the dataset aligned with all other get-started examples. For dev testing purpose, you can add apps to test/apps. We will make a full standalone demo of context interleaving by updating https://deck.gl/examples/google-maps/

alasarr: Agree. I think we can remove the examples in this PR. Once we land this, we can work on a cool 3D example to replace https://deck.gl/examples/google-maps

modules/google-maps/src/utils.js
felixpalmer: That doesn't work as the default null projectionMatrix from the `MapView` overrides it in https://github.com/visgl/deck.gl/blob/2601f2869ffca4c48c18c880582631ac3435e5ca/modules/core/src/views/view.js#L154

examples/get-started/pure-js/google-maps/app2.js
felixpalmer: I've removed the examples from the PR

examples/get-started/pure-js/google-maps/app.js
```diff
@@ -6,9 +6,11 @@ import {GeoJsonLayer, ArcLayer} from '@deck.gl/layers';
 const AIR_PORTS =
   'https://d2ad6b4ur7yvpq.cloudfront.net/naturalearth-3.3.0/ne_10m_airports.geojson';
 
+const GOOGLE_MAP_ID = 'f865292e93a85c6c';
```
Pessimistress, line 9: What is this magic string? Is it something that all users would want to use?

modules/google-maps/src/google-maps-overlay.js
```diff
@@ -66,18 +73,56 @@ export default class GoogleMapsOverlay {
...
+    if (isVectorMap) {
+      overlay.onAdd = () => {};
+      overlay.onContextLost = this._onContextLost.bind(this);
+      overlay.onContextRestored = this._onContextRestored.bind(this);
+      overlay.onDraw = this._onDrawVector.bind(this);
+    } else {
+      overlay.onAdd = this._onAdd.bind(this);
+      overlay.draw = this._onDrawRaster.bind(this);
+    }
+    overlay.onRemove = this._onRemove.bind(this);
```
Pessimistress, line 96: Can we define these methods in ES6 syntax? E.g.

```js
const {VECTOR, UNINITIALIZED} = google.maps.RenderingType;

class GoogleMapsOverlay {
  _createOverlay(map) {
    this.renderingType = map.getRenderingType();
  }

  onAdd() {
    if (this.renderingType !== VECTOR) {
      this._deck = createDeckInstance(...);
    }
  }

  draw() {
    // old API - draw raster
  }

  onDraw() {
    // new API - draw vector
  }

  onRemove() {
  }
}
```

modules/google-maps/src/utils.js
```diff
@@ -29,7 +31,7 @@ export function createDeckInstance(map, overlay, deck, props) {
   deck = new Deck({
     ...props,
     style: null,
-    parent: getContainer(overlay, props.style),
+    parent: overlay && getContainer(overlay, props.style),
```
Pessimistress, line 34: When would `overlay` be null?

modules/google-maps/src/utils.js
```diff
@@ -57,7 +59,14 @@ function getContainer(overlay, style) {
   const container = document.createElement('div');
   container.style.position = 'absolute';
   Object.assign(container.style, style);
-  overlay.getPanes().overlayLayer.appendChild(container);
+  if (overlay.getPanes) {
```
Pessimistress, line 62: Add comment on old/new code path

modules/google-maps/src/utils.js
Pessimistress: Shall we just remove the `null` fallbacks in View constructor? They don't seem to provide any value.

examples/get-started/pure-js/google-maps/app.js
felixpalmer: It is defined on a per-user basis, they obtain from the Google cloud console. I've removed it an instead it is now provided via an environment variable, like we do with `GOOGLE_MAPS_API_KEY`.

modules/google-maps/src/utils.js
felixpalmer: This is a mistake, I've removed it

examples/get-started/pure-js/google-maps/app.js
Pessimistress: Is it just for styling? Is it possible to create a map without it?

felixpalmer: It is required to create a vector map, not just for styling. I have updated the README

examples/get-started/pure-js/google-maps/README.md
```diff
@@ -9,13 +9,14 @@ with [webpack-dev-server](https://webpack.js.org/guides/development/#webpack-dev
 
 ## Usage
 
-To run this example, you need a [Google Maps API key](https://developers.google.com/maps/documentation/javascript/get-api-key). You can either set an environment variable:
+To run this example, you need a [Google Maps API key](https://developers.google.com/maps/documentation/javascript/get-api-key) and a map id. You can either set an environment variable:
```
Pessimistress, line 12: Can you link to the map id documentation

modules/google-maps/src/google-maps-overlay.js
felixpalmer: Sorry, I don't follow here. Are the methods `_onAdd`, `_onContextRestored` etc not already using the correct syntax? We need to bind them to the `google.maps.WebglOverlayView` object as per Google's docs

Pessimistress (approved):
Can be a follow up PR - the module documentation should be updated to explain the compatibility/limitations with both Google Maps versions.

modules/google-maps/src/utils.js
felixpalmer: I agree, the defaults seem to get in the way. I've remove the `projectionMatrix` and `viewMatrix` defaults and updated the overlay to pass the `projectionMatrix` directly as part of the `viewState`

examples/get-started/pure-js/google-maps/README.md
felixpalmer: Done

PR #5982, opened 2021-07-16
Provide default skirt height for tiled TerrainLayer

kylebarron:
#### Background

#5979 bumped our use of loaders.gl to 3.0.2. That release includes support for terrain skirting https://github.com/visgl/loaders.gl/pull/1662, which we can now enable in the TerrainLayer. Users can enable `skirtHeight` through the user of `loadOptions.terrain.skirtHeight`, but setting a default will make it easier to use for users who don't know the loaders.gl API as well. Setting a skirt height of 2 * the max error of the mesh should cover any seams (assuming the source elevation model is continuous).

#### Change List

- Set a default `skirtHeight` when using the TerrainLayer in tiled mode.

### DIFF
--- a/modules/geo-layers/src/terrain-layer/terrain-layer.js
+++ b/modules/geo-layers/src/terrain-layer/terrain-layer.js
@@ -110,6 +110,7 @@ export default class TerrainLayer extends CompositeLayer {
     loadOptions = {
       ...loadOptions,
       terrain: {
+        skirtHeight: this.state.isTiled ? meshMaxError * 2 : 0,
         ...loadOptions?.terrain,
         bounds,
         meshMaxError,


modules/geo-layers/src/terrain-layer/terrain-layer.js
```diff
@@ -110,6 +110,7 @@ export default class TerrainLayer extends CompositeLayer {
     loadOptions = {
       ...loadOptions,
       terrain: {
+        skirtHeight: this.state.isTiled ? meshMaxError * 2 : 0,
         ...loadOptions?.terrain,
```
kylebarron, line 114: Setting `skirtHeight` before `...loadOptions` means a value in `loadOptions` will overwrite the default value of `skirtHeight`, right?

ibgreen (approved):

modules/geo-layers/src/terrain-layer/terrain-layer.js
chrisgervang: I think this does what you're hoping it does. Providing a skirtHeight in the loader options will override this default, but providing other options will not override default skirtHeight accidentally.

chrisgervang (approved):

modules/geo-layers/src/terrain-layer/terrain-layer.js
kylebarron: Yes, I wanted a default that would be overridden through `loadOptions`, but I always have to double check whether `{...a, ...b}` has `a` overwriting `b` or vice versa 😄

Pessimistress:
Add a line to what's new?

PR #6009, opened 2021-07-23
fix: MVTLayer set binary to false by default

alasarr:
Set MVTLayer binary to `false`. Related to https://github.com/visgl/deck.gl/issues/6006


Pessimistress:
Also need to update the layer documentation

alasarr:
Done, do you think it worth to mention in doc this known-issue: https://github.com/visgl/deck.gl/issues/6006#issuecomment-885635525?

Pessimistress:
I think we should fix the highlighting issue soon in a 8.5 patch.

Pessimistress (approved):

PR #6011, opened 2021-07-24
Fix picking when using binary GeoJSON

Pessimistress:
For #6006

#### Change List
- Fix picking colors when using binary GeoJSON


alasarr (approved):

modules/layers/src/geojson-layer/geojson-layer-props.js
```diff
@@ -38,21 +38,17 @@ export function createLayerPropsFromFeatures(features, featuresDiff) {
...
-    encodePickingColor
-  );
+  const customPickingColors = calculatePickingColors(geojsonBinary, encodePickingColor);
 
   layerProps.points.data = {
     length: points.positions.value.length / points.positions.size,
     attributes: {
       getPosition: points.positions,
-      pickingColors: {
+      instancePickingColors: {
```
alasarr, line 51: Just curious... why is this called `pickingColors` for SolidPolygonLayer and `instancePickingColors` for Scatterplot and PathLayer?

Pessimistress: Because picking colors do not have an accessor, you supply them using the attribute name.

https://github.com/visgl/deck.gl/blob/16006ed2a595271cf1ea5a0f8d8e8db7b00e6516/modules/layers/src/path-layer/path-layer.js#L117

https://github.com/visgl/deck.gl/blob/16006ed2a595271cf1ea5a0f8d8e8db7b00e6516/modules/layers/src/solid-polygon-layer/solid-polygon-layer.js#L183

Issue #6023, opened 2021-07-27
TileLayer no longer updates Tileset min/maxZoom opts when said props change


jcready:
#### Description
I'm trying to update from 8.4.13 to 8.5.0 and I'm noticing a difference in how the TileLayer (I'm using the MVTLayer specifically) behaves. It looks like when I change the min/max zoom props the Tileset instance is no longer seeing that change.

I believe the issue comes down to how the `updateState` method updates the Tileset options. In v8.5.0 it does this:

https://github.com/visgl/deck.gl/blob/40141c7b9209ca2497648fb0f3c470b61a85fc79/modules/geo-layers/src/tile-layer/tile-layer.js#L73-L74

Compared to 8.4.13:

https://github.com/visgl/deck.gl/blob/640bce52a72151033fbffef3f64a45dbc716bad9/modules/geo-layers/src/tile-layer/tile-layer.js#L107-L108

It seem to me like the `updateState` method need to update its `state.maxZoom` and `state.minZoom` when those props change or revert to the previous behavior where the props are passed directly to `tileset.setOptions`. For what it's worth it looks like that change happened in v8.4.15.

#### Expected Behavior
Changing the `minZoom` or `maxZoom` props will appropriately update the Tileset opts.

#### Environment
 - deck.gl 8.5.0
 - Browser Version: Firefox 90.0.2
 - OS: Mac OS 11.4

Pessimistress:
@alasarr  MVTLayer should set `state.minZoom` and `state.maxZoom` to `null` if they are not defined by tileJSON. `_updateTileData()` is only called on data change, so `props.minZoom` change is being ignored.

alasarr:
Thanks a lot for reporting, it doesn't look complicated, I'll create a PR later today or tomorrow with the fix

alasarr:
I've created a fix, @Pessimistress once approved I'll create a patch release

Pessimistress:
Fixed in 8.5.1

PR #6040, opened 2021-07-29
Fix: MVT update min/maXZoom props

alasarr:
Changes on `minZoom` or `maxZoom` props didn't work at MVTLayer. Reported at https://github.com/visgl/deck.gl/issues/6023

### DIFF
--- a/modules/geo-layers/src/mvt-layer/mvt-layer.js
+++ b/modules/geo-layers/src/mvt-layer/mvt-layer.js
@@ -47,8 +47,10 @@ export default class MVTLayer extends TileLayer {
 
   /* eslint-disable complexity */
   async _updateTileData() {
-    let {data, minZoom, maxZoom} = this.props;
+    let {data} = this.props;
     let tileJSON = null;
+    let minZoom = null;
+    let maxZoom = null;
 
     if (typeof data === 'string' && !isURLTemplate(data)) {
       const {onDataLoad, fetch} = this.props;
@@ -70,13 +72,13 @@ export default class MVTLayer extends TileLayer {
     if (tileJSON) {
       data = tileJSON.tiles;
 
-      if (Number.isFinite(tileJSON.minzoom) && tileJSON.minzoom > minZoom) {
+      if (Number.isFinite(tileJSON.minzoom) && tileJSON.minzoom > this.props.minZoom) {
         minZoom = tileJSON.minzoom;
       }
 
       if (
         Number.isFinite(tileJSON.maxzoom) &&
-        (!Number.isFinite(maxZoom) || tileJSON.maxzoom < maxZoom)
+        (!Number.isFinite(maxZoom) || tileJSON.maxzoom < this.props.maxZoom)
       ) {
         maxZoom = tileJSON.maxzoom;
       }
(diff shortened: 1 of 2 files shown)


modules/geo-layers/src/mvt-layer/mvt-layer.js
```diff
@@ -70,13 +72,13 @@ export default class MVTLayer extends TileLayer {
     if (tileJSON) {
       data = tileJSON.tiles;
 
-      if (Number.isFinite(tileJSON.minzoom) && tileJSON.minzoom > minZoom) {
+      if (Number.isFinite(tileJSON.minzoom) && tileJSON.minzoom > this.props.minZoom) {
```
Pessimistress, line 75: This is still depending on a prop that could later change. This chunk pausing should be performed in getTilesetOptions

alasarr: Good catch, thanks, another approach implemented.

Pessimistress (approved):

modules/geo-layers/src/mvt-layer/mvt-layer.js
```diff
@@ -69,21 +69,30 @@ export default class MVTLayer extends TileLayer {
...
         Number.isFinite(tileJSON.maxzoom) &&
-        (!Number.isFinite(maxZoom) || tileJSON.maxzoom < maxZoom)
+        (!Number.isFinite(props.maxZoom) || tileJSON.maxzoom < props.maxZoom)
       ) {
-        maxZoom = tileJSON.maxzoom;
+        opts.maxZoom = tileJSON.maxzoom;
       }
     }
-
-    this.setState({data, tileJSON, minZoom, maxZoom});
```
Pessimistress, line 85: Can we remove this now?

https://github.com/visgl/deck.gl/blob/5338f9abbcb253c44fcc3464676591255724431a/modules/geo-layers/src/tile-layer/tile-layer.js#L85-L86

alasarr: As we've redefined `_getTilesetOptions`, maxZoom and minZoom are directly returned no need to be in the state. 

I'm thinking now that this two lines https://github.com/visgl/deck.gl/blob/5338f9abbcb253c44fcc3464676591255724431a/modules/geo-layers/src/tile-layer/tile-layer.js#L85-L86 can be removed now.

modules/geo-layers/src/tile-layer/tile-layer.js
```diff
@@ -82,14 +82,14 @@ export default class TileLayer extends CompositeLayer {
   }
 
   _getTilesetOptions(props) {
-    const maxZoom = Number.isFinite(this.state.maxZoom) ? this.state.maxZoom : props.maxZoom;
```
alasarr, line 85: @Pessimistress do you see something wrong with this? I've tested with mvtlayer and it works.

Pessimistress: Perfect, I think this is the right thing to do (TileLayer should not reference a state that itself doesn't define).

PR #6062, opened 2021-08-04
Check for correct matching layerName when highlighting in MVTLayer

felixpalmer:
#### Background

The MVTLayer has incorrect picking behavior if two features in the source data share the same `id`, but have a different `layerName`

#### Change List
- `MVTLayer.getHighlightedObjectIndex()` checks that the hovered feature has the correct `layerName` as well as the correct `id`. This PR is only for the non-binary codepath.
- Add `MVTLayer` test cases to the `picking.spec.js` tests
- Test picking in `MVTLayer` for two features with matching `id`, but different `layerName`

### DIFF
--- a/modules/geo-layers/src/mvt-layer/mvt-layer.js
+++ b/modules/geo-layers/src/mvt-layer/mvt-layer.js
@@ -158,16 +158,25 @@ export default class MVTLayer extends TileLayer {
   _updateAutoHighlight(info) {
     const {uniqueIdProperty} = this.props;
 
-    const {hoveredFeatureId} = this.state;
+    const {hoveredFeatureId, hoveredFeatureLayerName} = this.state;
     const hoveredFeature = info.object;
     let newHoveredFeatureId;
+    let newHoveredFeatureLayerName;
 
     if (hoveredFeature) {
       newHoveredFeatureId = getFeatureUniqueId(hoveredFeature, uniqueIdProperty);
+      newHoveredFeatureLayerName = getFeatureLayerName(hoveredFeature);
     }
 
-    if (hoveredFeatureId !== newHoveredFeatureId && newHoveredFeatureId !== -1) {
-      this.setState({hoveredFeatureId: newHoveredFeatureId});
+    if (
+      (hoveredFeatureId !== newHoveredFeatureId && newHoveredFeatureId !== -1) ||
+      (hoveredFeatureLayerName !== newHoveredFeatureLayerName &&
+        newHoveredFeatureLayerName !== null)
+    ) {
+      this.setState({
+        hoveredFeatureId: newHoveredFeatureId,
+        hoveredFeatureLayerName: newHoveredFeatureLayerName
+      });
     }
   }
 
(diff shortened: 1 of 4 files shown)


felixpalmer:
https://github.com/visgl/deck.gl/issues/6006#issuecomment-891895290

modules/geo-layers/src/mvt-layer/mvt-layer.js
```diff
@@ -302,6 +312,14 @@ function getFeatureUniqueId(feature, uniqueIdProperty) {
   return -1;
 }
 
+function getFeatureLayerName(feature) {
+  if (feature.properties && feature.properties.layerName) {
+    return feature.properties.layerName;
+  }
+
+  return null;
```
Pessimistress, line 320:
```js
return feature.properties?.layerName || null;
```

modules/geo-layers/src/mvt-layer/mvt-layer.js
```diff
@@ -188,26 +197,27 @@ export default class MVTLayer extends TileLayer {
...
-      ? highlightedFeatureId
-      : hoveredFeatureId;
+    const featureIdToHighlight = isHighlighted ? highlightedFeatureId : hoveredFeatureId;
 
     // Iterable data
     if (Array.isArray(data)) {
-      return data.findIndex(
-        feature => getFeatureUniqueId(feature, uniqueIdProperty) === featureIdToHighlight
-      );
+      return data.findIndex(feature => {
```
Pessimistress, line 215: Is this called on every render when the viewport is panned? Can we cache it somehow?

modules/geo-layers/src/mvt-layer/mvt-layer.js
```diff
@@ -158,16 +158,25 @@ export default class MVTLayer extends TileLayer {
...
 
     if (hoveredFeature) {
       newHoveredFeatureId = getFeatureUniqueId(hoveredFeature, uniqueIdProperty);
+      newHoveredFeatureLayerName = getFeatureLayerName(hoveredFeature);
     }
 
-    if (hoveredFeatureId !== newHoveredFeatureId && newHoveredFeatureId !== -1) {
-      this.setState({hoveredFeatureId: newHoveredFeatureId});
+    if (
+      (hoveredFeatureId !== newHoveredFeatureId && newHoveredFeatureId !== -1) ||
```
Pessimistress, line 172: I believe null/-1 is handled by `isFeatureIdDefined`. You should still save them in that case so that pointer leaving a feature is properly handled.

modules/geo-layers/src/mvt-layer/mvt-layer.js
felixpalmer: Are you referring to just my change, or proposing an enhancement in general?

Pessimistress: An enhancement. Not in this PR.

Pessimistress (approved):

modules/geo-layers/src/mvt-layer/mvt-layer.js
```diff
@@ -302,6 +311,10 @@ function getFeatureUniqueId(feature, uniqueIdProperty) {
   return -1;
```
Pessimistress, line 311: This should return `undefined`?

felixpalmer: Changed. Note that this will alter behavior in that `const isFeatureIdPresent = isFeatureIdDefined(hoveredFeatureId)` (line 204) will return false for features without an unique id. I think this is correct as the old behavior would return true and then highlight the wrong feature

Issue #5936, opened 2021-07-01
GeoJSON/Scatterplot layer squares instead of circles in FPV


Radik24:
#### Description
I'm trying to display GeoJSON points in first person view. But points are always displayed as squares. Minimap displays the same data an there are nice points on the map.

Example:
https://codepen.io/radik24/pen/gOmoxMd

![image](https://user-images.githubusercontent.com/23722745/124083842-9f4f2300-da4e-11eb-9658-97dc0aba4337.png)

#### Expected Behavior
Display circles (points)

https://github.com/visgl/deck.gl/discussions/5813

Pessimistress:
This is caused by the way we perform antialiasing in the fragment shader. The pixel radius passed to the fragment shader is in common space, not screen space. This is fine in MapView and non-geospatial views, where the viewProjectionMatrix is constructed so that 1 common unit = 1 screen pixel at the viewport center. This assumption does not apply in FirstPersonView. In the above example, the circles are very close to the camera, yet the pixel radius seen by the fragment shader is 0.1.

We could fix this by calculating the screen size of each circle correctly in the vertex shader. However, it may introduce a negative impact on perf, as well as a change in the visual behavior. To unblock the FPV use case in the 8.5 timeframe, I propose that we introduce an opt-in switch to ScatterplotLayer that can be used to disable antialiasing. @ibgreen

The same feature is also requested by @igorDykhta in https://github.com/visgl/deck.gl/pull/5956#issuecomment-875038656

igorDykhta:
@Pessimistress I'm going add to such switch for ScatterplotLayer

Radik24:
Thanks, will try it! How often is master release to npm?

Pessimistress:
Published 8.5.3

PR #6081, opened 2021-08-10
Scatterplot layer: smooth edges prop

igorDykhta:
For #5936

#### Background
The current implementation of smoothed edges for Scatterplot layer generates artifacts when circles are rendered on different depths and doesn't work as expected in FirstPersonView.

#### Change List
- smoothEdges prop to toggle smoothed edges for Scatterplot layer


docs/api-reference/layers/scatterplot-layer.md
```diff
@@ -142,6 +142,13 @@ The maximum line width in pixels. This prop can be used to prevent the path from
 
 If `true`, rendered circles always face the camera. If `false` circles face up (i.e. are parallel with the ground plane).
 
+##### `smoothEdges` (Boolean, optional)
```
Pessimistress, line 145: - Why plural? TextLayer has a font setting called `smoothing`. Or shall we call it `antialiasing` as a commonly understood terminology?
- The documentation should explain why this prop exists and when to disable it.

modules/layers/src/scatterplot-layer/scatterplot-layer-fragment.glsl.js
```diff
@@ -36,7 +37,9 @@ void main(void) {
   geometry.uv = unitPosition;
 
   float distToCenter = length(unitPosition) * outerRadiusPixels;
-  float inCircle = smoothedge(distToCenter, outerRadiusPixels);
+  float inCircle = smoothEdges ? 
+    smoothedge(distToCenter, outerRadiusPixels) : 
+    distToCenter < outerRadiusPixels ? 1.0 : 0.0;
```
Pessimistress, line 42: `step(distToCenter, outerRadiusPixels)`

modules/layers/src/scatterplot-layer/scatterplot-layer-fragment.glsl.js
```diff
@@ -25,6 +25,7 @@ precision highp float;
 
 uniform bool filled;
 uniform float stroked;
+uniform bool smoothEdges;
```
Pessimistress, line 28: There are two usages of `smoothedge` in this shader, both need to be replaced with `step` if the toggle is off (FirstPersonView use case).

modules/layers/src/geojson-layer/sub-layer-map.js
```diff
@@ -10,6 +10,7 @@ export const POINT_LAYER = {
     props: {
       filled: 'filled',
       stroked: 'stroked',
+      smoothEdges: 'smoothEdges',
```
Pessimistress, line 13: - A prop that only apply to the ScatterplotLayer must be prefixed with `point`
- Update documentation

Pessimistress (approved):

PR #6083, opened 2021-08-10
Bug fixes in google overlay

felixpalmer:
Came across a regression with the Google library (which is updating weekly) where the rendering completely broke down. This fix makes sure the GL state is correct so that the rendering in Google Maps works

### DIFF
--- a/modules/google-maps/src/google-maps-overlay.js
+++ b/modules/google-maps/src/google-maps-overlay.js
@@ -166,7 +166,9 @@ export default class GoogleMapsOverlay {
       // the GL state can be inconsistent
       setParameters(gl, {
         scissor: [0, 0, gl.canvas.width, gl.canvas.height],
-        stencilFunc: [gl.ALWAYS, 0, 255, gl.ALWAYS, 0, 255]
+        stencilFunc: [gl.ALWAYS, 0, 255, gl.ALWAYS, 0, 255],
+        depthMask: true,
+        depthTest: true
       });
     }
   }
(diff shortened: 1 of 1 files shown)


modules/google-maps/src/google-maps-overlay.js
```diff
@@ -166,7 +166,9 @@ export default class GoogleMapsOverlay {
       // the GL state can be inconsistent
       setParameters(gl, {
         scissor: [0, 0, gl.canvas.width, gl.canvas.height],
-        stencilFunc: [gl.ALWAYS, 0, 255, gl.ALWAYS, 0, 255]
+        stencilFunc: [gl.ALWAYS, 0, 255, gl.ALWAYS, 0, 255],
+        depthMask: true,
```
Pessimistress, line 170: This does not seem right. Picking does not alter `depthMask` and `depthTest`. The places we call `setParamters` instead of `withParameters` are:

- On WebGL initialization with defaults
- Before each redraw with `deck.props.parameters`
- Before each pass with `viewport` and `framebuffer`
- For each layer with `polygonOffset`

felixpalmer: I've tested it more and you are right. The issue is not resolved by this fix, but instead occurs seemingly randomly. I think the regression is down to a change in Google's library (which is updated weekly) as I did not see this when merging&testing the Google PR a few weeks back

felixpalmer: We are working on this, but no clear fix yet. Instead I've added a few other bug fixes so we can merge this PR

felixpalmer:
Changelist:

- Blending works correctly now against Google basemap
- Overlay is correctly initialized when added later (not on pageload)
- Back-compatibility with older Google maps versions that do not support vector rendering

Pessimistress (approved):

Issue #6093, opened 2021-08-12
Add iTowns example 


aakaydin:
#### Target Use case
[iTowns](http://www.itowns-project.org/) (@iTowns) is a web map rendering framework built on top of Three js. It would be nice to have a DeckGL integration and profit from data visualization capabilities

#### Proposed feature

Add a pure-js example showing how to add a DeckGL overlay to iTowns.

#### To Do List 
  - [X] Create an example using iTowns
  - [X] Synchorinize the controls of both libraries
  - [ ] Test

PR #6094, opened 2021-08-12
Add a purejs iTowns example 

aakaydin:
Referencing issue [#6093](https://github.com/visgl/deck.gl/issues/6093)

#### Change List
- Added an example using iTowns to examples/get-started/pure-js

For the moment the background map doesn't have the same projections as DeckGL and the two maps seem to 'slide' while panning, the airports dont stay on the same place as we move the map.


Pessimistress:
Thanks for the submission. If you test against the master branch (`yarn start-local`) you can use an externally constructed projection matrix:

```js
onViewStateChange: ({viewState}) => {
  ...
  viewState.projectionMatrix = getProjectionMatrix(view);
  return viewState;
}
```

It may be easier to use your base map's controller instead:

```js
const deck = new Deck({
  controller: false
  style: {pointerEvents: 'none'},
  viewState: INITIAL_VIEW_STATE
});

// pseudo code
view.on('change', () => {
  const deckViewState = getDeckViewState(view);
  deck.setProps({viewState: deckViewState}));
});
```

ibgreen-openai:
Close due to inactivity

Issue #6096, opened 2021-08-13
Support geo_column & columns URL parameters for CartoLayer


felixpalmer:
#### Target Use case
In API v3, the CARTO platform allows configuration of the `geo_column` (the database column containing geospatial information) and `columns` (the database columns to be returned by the request). Supporting these will allow the user more flexibility when fetching data with `type: MAP_TYPES.TABLE`.

#### Proposed feature
Add `geoColumn` & `columns` props to `CartoLayer` which are then passed onto the server request.

#### To Do List 
  - [x] Add label and assign to milestone 
  - [x] Coding
  - [x] Doc update
  - [x] What’s new update 
  - [x] Test

PR #6097, opened 2021-08-13
Add geoColumn & columns props to CartoLayer

felixpalmer:
Implementation of https://github.com/visgl/deck.gl/issues/6096

### DIFF
--- a/docs/api-reference/carto/carto-layer.md
+++ b/docs/api-reference/carto/carto-layer.md
@@ -111,6 +111,14 @@ Required when apiVersion is `API_VERSIONS.V3`.
 
 Name of the connection registered in the CARTO workspace.
 
+##### `geoColumn` (String, optional)
+
+Name of the `geo_column` in the CARTO platform. Use this override the default column ('geom'), from which the geometry information should be fetched.
+
+##### `columns` (Array<String>, optional)
+
+Names of columns to fetch. By default, all columns are fetched.
+
 ##### `uniqueIdProperty` (String)
 
 * Default: `cartodb_id`
--- a/docs/api-reference/carto/carto-layer.md
+++ b/docs/api-reference/carto/carto-layer.md
@@ -113,9 +113,13 @@ Name of the connection registered in the CARTO workspace.
 
 ##### `geoColumn` (String, optional)
 
+Only supported when apiVersion is `API_VERSIONS.V3` and `type` is `MAP_TYPES.TABLE`.
+
 Name of the `geo_column` in the CARTO platform. Use this override the default column ('geom'), from which the geometry information should be fetched.
 
 ##### `columns` (Array<String>, optional)
+
+Only supported when apiVersion is `API_VERSIONS.V3` and `type` is `MAP_TYPES.TABLE`.
 
 Names of columns to fetch. By default, all columns are fetched.
 
--- a/modules/carto/src/api/maps-v3-client.js
+++ b/modules/carto/src/api/maps-v3-client.js
@@ -67,19 +67,33 @@ function dealWithError({response, error}) {
 /**
  * Build a URL with all required parameters
  */
-function getParameters({type, source}) {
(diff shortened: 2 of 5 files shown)


modules/carto/src/layers/carto-layer.js
```diff
@@ -20,7 +20,13 @@ const defaultProps = {
   /* API v3 PARAMETERS */
   /**********************/
   // (String, required): connection name at CARTO platform
-  connection: null
+  connection: null,
```
alasarr, line 23: Need to add these parameters at the doc

alasarr: Would be great to explain that default value is `geom`

alasarr (changes requested):
Looks good to me, but need to update the doc

modules/carto/src/layers/carto-layer.js
```diff
@@ -20,7 +20,13 @@ const defaultProps = {
...
+
+  // (String, optional): name of geo_column in CARTO platform
+  geoColumn: null,
+
+  // (Array<String>, optional): names of columns to fetch
+  columns: null
```
Pessimistress, line 29: If you declare `type: 'array'` here then the layer diff will block unnecessary updates unless the prop changes deeply

modules/carto/src/api/maps-v3-client.js
```diff
@@ -67,19 +67,33 @@ function dealWithError({response, error}) {
...
  */
-function getParameters({type, source}) {
+function getParameters({type, source, geoColumn, columns}) {
   const encodedClient = encodeParameter('client', 'deck-gl-carto');
   const parameters = [encodedClient];
 
   const sourceName = type === MAP_TYPES.QUERY ? 'q' : 'name';
   parameters.push(encodeParameter(sourceName, source));
 
+  if (geoColumn) {
```
alasarr, line 77: This parameter is only available for MAP_TYPES.TABLE, shouldn't be passed in other cases

docs/api-reference/carto/carto-layer.md
```diff
@@ -111,6 +111,14 @@ Required when apiVersion is `API_VERSIONS.V3`.
 
 Name of the connection registered in the CARTO workspace.
 
+##### `geoColumn` (String, optional)
```
alasarr, line 114: We need to say that this parameter is only supported for `MAP_TYPES.TABLE`, so only available in v3 because type table is only supported in V3+

docs/api-reference/carto/carto-layer.md
```diff
@@ -111,6 +111,14 @@ Required when apiVersion is `API_VERSIONS.V3`.
...
+
+Name of the `geo_column` in the CARTO platform. Use this override the default column ('geom'), from which the geometry information should be fetched.
+
+##### `columns` (Array<String>, optional)
```
alasarr, line 118: Same as before.

modules/carto/src/layers/carto-layer.js
```diff
@@ -51,12 +57,17 @@ export default class CartoLayer extends CompositeLayer {
...
       log.assert(!connection, `Connection prop is not supported for apiVersion ${apiVersion}`);
+      log.assert(!geoColumn, `geoColumn prop is not supported for apiVersion ${apiVersion}`);
+      log.assert(!columns, `columns prop is not supported for apiVersion ${apiVersion}`);
     } else if (apiVersion === API_VERSIONS.V3) {
       log.assert(connection, 'Missing mandatory connection parameter');
       log.assert(
         Object.values(MAP_TYPES).includes(type),
         `Invalid type ${type}. Use MAP_TYPES enum.`
       );
+      if (columns) {
```
alasarr, line 68: Need to check columns and geoColumn are only allow for `MAP_TYPES.TABLE`

alasarr (approved):

PR #6098, opened 2021-08-13
Fix MVTLayer autoHighlight with binary data

Pessimistress:
For #6006

#### Change List
- Fix incorrect usage of feature id (should return global feature id)
- Move `findIndexBinary` from GeoJsonLayer to MVTLayer (logic does not apply to generic GeoJSON)
- Handle default `uniqueIdProperty`
- Handle filter by layer name
- Update MVT test app
- Unit tests


felixpalmer (approved):
Looks good. I think it would be worth adding a binary version of the picking test to https://github.com/visgl/deck.gl/blob/b40f47f780682d4c87911c0642a42d4983a243e9/test/modules/core/lib/picking.spec.js#L58

@Pessimistress, I can do so if you like?

Pessimistress:
> I can do so if you like?

Please? :)

PR #6116, opened 2021-08-18
OrthographicView supports independent x/y zoom levels

Pessimistress:
#### Background

For the charting use case

deck gl-orthographic-controller

#### Change List
- `OrthographicViewport` accepts `zoom` as either a number (uniform zoom) or an array (independent zoom)
- `OrthographicController` new option `zoomAxis`
- Documentation
- Test

### DIFF
--- a/docs/api-reference/core/orthographic-controller.md
+++ b/docs/api-reference/core/orthographic-controller.md
@@ -43,6 +43,10 @@ Supports all [Controller options](/docs/api-reference/core/controller.md#options
 - `touchRotate`: not effective, this view cannot be rotated
 - `keyboard`: arrow keys to pan, +/- to zoom
 
+Also accepts additional options:
+
+- `zoomAxis` (String) - which axes to apply zoom to. Affects scroll, keyboard +/- and double tap. One of `X` (zoom along the X axis only), `Y` (zoom along the Y axis only), `all`. Default `all`. If this option is set to `X` or `Y`, `viewState.zoom` must be an array to enable independent zoom for each axis.
+
 ## Custom OrthographicController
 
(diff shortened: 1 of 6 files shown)


kylebarron (approved):

felixpalmer:
You need to supply an 2D array to the `zoom` prop in the `viewState`: https://deck.gl/docs/api-reference/core/orthographic-view#view-state. You then update a the `viewState.zoom` with the correct values based on user input

cornhundred:
Thanks @felixpalmer. I'm trying this [here](https://observablehq.com/d/358846a94ba6417b ):

```
deckgl = {
  return new deck.DeckGL({
    container,
    views: new deck.OrthographicView({controller: true}),
    initialViewState: {target: [0, 0, 0], zoom: [0, 0]},
    onViewStateChange: ({viewState}) => {
      viewState.zoom = [viewState.zoom[0], 0]
      return viewState
    }
  });
}
```
as a way of trying to limit the zooming to the X axis only and behaves as I would expect with two exceptions: 

1. The Y zooming seems to be dependent on the position of the cursor in the visualization (Y zooming does not occur if the cursor is kept in the middle of the visualization, but otherwise causes the points to 'zoom' up or down depending on whether the cursor is above or below the middle of the visual respectively)
2. The points do not increase in width as I would expect as I zoom in.

I'll read the documentation and see if I can figure out what is happening.

Pessimistress:
@cornhundred You need to set [zoomAxis](https://deck.gl/docs/api-reference/core/orthographic-controller#options) of the controller.

cornhundred:
Thanks @Pessimistress I'll start a discussion :)

PR #6143, opened 2021-08-28
CARTO: include API error at the exception message

alasarr:
It returns API error at the exception message. I've also added a bad parameter message when 400 is received


VictorVelarde (approved):
LGTM

Issue #6157, opened 2021-09-02
Heatmap - expose additional properties


igorDykhta:
#### Target Use case
Exposing several properties of Heatmap layer can be beneficial.

#### Proposed feature
- expose debounceTimeout prop to be able to control debounce timeout by adjusting it based on dataset size and radius.
- expose weightsTextureSize prop to control performance of max-weight calculation, rendering quality, memory usage.
- expose a callback to be able to modify weights transform shaders.
- move updateState logic to a separate method to be able to override it in an extended class.

#### To Do List 
  - [ ] Add label and assign to milestone 
  - [x] Coding
  - [x] Doc update
  - [ ] What’s new update 
  - [ ] Test

Pessimistress:
Published v8.5.7

PR #6158, opened 2021-09-02
Heatmap - expose additional properties

igorDykhta:
For #6157

#### Change List
- debounceTimeout prop for Heatmap layer to control debounce timeout by adjusting it based on dataset size and radius.
- weightsTextureSize prop for Heatmap to control performance of max-weight calculation, rendering quality, memory usage.
- HeatmapLayer::getShaders methods can be used to override weights transform shaders in extended layers.
- heatmap-related logic from updateState is moved to _updateHeatmapState method to be able to override it in an extended class.

### DIFF
--- a/docs/api-reference/aggregation-layers/heatmap-layer.md
+++ b/docs/api-reference/aggregation-layers/heatmap-layer.md
@@ -120,6 +120,24 @@ Operation used to aggregate all data point weights to calculate a pixel's color 
 
 The weight of each data object is distributed to all the pixels in a circle centered at the object position. The weight that a pixel receives is inversely proportional to its distance from the center. In `'SUM'` mode, pixels that fall into multiple circles will have the sum of all weights. In `'MEAN'` mode, pixels that fall into multiple circles will have their weight calculated as the weighted average from all the neighboring data points. And the weight of the pixel determines its color. 
 
+##### `weightsTextureSize` (String, optional)
+
+* Default: `2048`
+
+`weightsTextureSize` specifies the size of weight texture. Smaller texture sizes can improve rendering performance.
+
+##### `debounceTimeout` (String, optional)
+
+* Default: `500`
+
+`debounceTimeout` is an interval in milliseconds during which changes to the viewport don't trigger aggregation.
+
+##### `_modifyWeightsTransformShaders` (String, optional)
+
+* Default: `{vs: string, _fs: string} => {vs: string, _fs: string}`
+
+A callback to modify default weight transform shaders.
+
 ### Data Accessors
 
 ##### `getPosition` ([Function](/docs/developer-guide/using-layers.md#accessors), optional)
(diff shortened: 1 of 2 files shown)


modules/aggregation-layers/src/heatmap-layer/heatmap-layer.js
```diff
@@ -496,17 +501,18 @@ export default class HeatmapLayer extends AggregationLayer {
...
     let {updateTimer} = this.state;
+    const {debounceTimeout} = this.props;
 
     if (fromTimer) {
       updateTimer = null;
       // update
       this._updateBounds(true);
       this._updateTextureRenderingBounds();
-      this.setState({isWeightMapDirty: true});
+      this._updateWeightmap();
```
Pessimistress, line 511: I suggest that you keep the old code if there's no particularly good reason to make this change. Using a dirty flag in layer state instead of calling update immediately ensures that update is run at most once per render cycle.

modules/aggregation-layers/src/heatmap-layer/heatmap-layer.js
```diff
@@ -276,18 +282,18 @@ export default class HeatmapLayer extends AggregationLayer {
     }
   }
 
+  _getWeightsTransformShaders() {
```
Pessimistress, line 285: Can you use `getShaders` for both transforms? Similar to:

https://github.com/visgl/deck.gl/blob/cc225a850e343b5aadd1eb36b67b1e8fac17b66d/modules/layers/src/solid-polygon-layer/solid-polygon-layer.js#L64

modules/aggregation-layers/src/heatmap-layer/heatmap-layer.js
```diff
@@ -276,18 +276,29 @@ export default class HeatmapLayer extends AggregationLayer {
     }
   }
 
-  _createWeightsTransform(shaderOptions = {}) {
+  _getWeightsTransformShaders() {
```
Pessimistress, line 279: Delete

docs/api-reference/aggregation-layers/heatmap-layer.md
```diff
@@ -120,6 +120,18 @@ Operation used to aggregate all data point weights to calculate a pixel's color
 
 The weight of each data object is distributed to all the pixels in a circle centered at the object position. The weight that a pixel receives is inversely proportional to its distance from the center. In `'SUM'` mode, pixels that fall into multiple circles will have the sum of all weights. In `'MEAN'` mode, pixels that fall into multiple circles will have their weight calculated as the weighted average from all the neighboring data points. And the weight of the pixel determines its color. 
 
+##### `weightsTextureSize` (Number, optional)
+
+* Default: `2048`
+
+`weightsTextureSize` specifies the size of weight texture. Smaller texture sizes can improve rendering performance.
```
Pessimistress, line 127: Explain the trade off

docs/api-reference/aggregation-layers/heatmap-layer.md
```diff
@@ -120,6 +120,18 @@ Operation used to aggregate all data point weights to calculate a pixel's color
...
+
+##### `debounceTimeout` (Number, optional)
+
+* Default: `500`
+
+`debounceTimeout` is an interval in milliseconds during which changes to the viewport don't trigger aggregation.
```
Pessimistress, line 133: Explain the trade off

Pessimistress (approved):

PR #6178, opened 2021-09-09
Fix changelog 8.6.0-alpha.2

alasarr:
I forgot to add the header to the changelog when I released 8.6.0-alpha.2


kylebarron:
Is that why the release CI failed?

kylebarron (approved):

alasarr:
>Is that why the release CI failed?
Yes

PR #6194, opened 2021-09-13
TileLayer retains cache on data change

### LINKED ISSUE #6179: `TileLayer` Update Triggers
#### Target Use case
For Viv, when you make a new channel selection (i.e the drop downs above the colored sliders on avivator.gehlenborglab.org), the the whole `TileLayer` reloads which means there is a gap where there is no image on the screen from the `TileLayer`.

Our channel update mechanism is done via an update trigger: https://github.com/hms-dbmi/viv/blob/cac61bed228d28110db688fa30a3c2888d16a6d6/src/layers/MultiscaleImageLayer/MultiscaleImageLayer.js#L194-L196

#### Proposed feature

Is there a way to leave up the tiles until the `Tileset2D` has updated its tiles and is ready to swap out layers?  Is this desired behavior?  I can try to make a PR for this if it seems reasonable. 

#### To Do List 
  - [x] Coding
  - [ ] Doc update
  - [ ] What’s new update 
  - [ ] Test

ilan-gold:
Fixes #6179

#### Change List
- Add a `reload` method for `Tileset2D` to be called on updates by `updateTriggers` instead of creating a new tileset

### DIFF
--- a/modules/geo-layers/src/tile-layer/tile-2d-header.js
+++ b/modules/geo-layers/src/tile-layer/tile-2d-header.js
@@ -45,58 +45,52 @@ export default class Tile2DHeader {
 
   /* eslint-disable max-statements */
   async _loadData(getTileData, requestScheduler) {
+    const {x, y, z, bbox} = this;
+
+    this._abortController = new AbortController(); // eslint-disable-line no-undef
+    const {signal} = this._abortController;
+
+    const requestToken = await requestScheduler.scheduleRequest(this, tile => {
+      return tile.isSelected ? 1 : -1;
+    });
+
+    if (!requestToken) {
+      this._isCancelled = true;
+      return;
+    }
+    // A tile can be cancelled while being scheduled
+    if (this._isCancelled) {
+      requestToken.done();
+      return;
+    }
+
+    let tileData;
+    let error;
     try {
-      const {x, y, z, bbox} = this;
-
-      this._abortController = new AbortController(); // eslint-disable-line no-undef
-      const {signal} = this._abortController;
-      this._isLoaded = false;
-      console.log('here!', x, y, z); // eslint-disable-line
-      const requestToken = await requestScheduler.scheduleRequest(this, tile => {
-        return tile.isSelected ? 1 : -1;
-      });
-      console.log('here!', x, y, z); // eslint-disable-line
-
-      if (!requestToken) {
-        this._isCancelled = true;
-        return;
-      }
-      // A tile can be cancelled while being scheduled
-      if (this._isCancelled) {
-        requestToken.done();
-        return;
-      }
-
-      let tileData;
-      let error;
-      try {
-        tileData = await getTileData({x, y, z, bbox, signal});
-      } catch (err) {
-        error = err || true;
-      } finally {
-        requestToken.done();
-
-        if (this._isCancelled && !tileData) {
-          this._isLoaded = false;
-        } else {
-          // Consider it loaded if we tried to cancel but `getTileData` still returned data
-          this._isLoaded = true;
-          this._isCancelled = false;
-        }
-      }
-
-      if (!this._isLoaded) {
-        return;
-      }
-
-      if (error) {
-        this.onTileError(error, this);
-      } else {
-        this.content = tileData;
-        this.onTileLoad(this);
-      }
+      tileData = await getTileData({x, y, z, bbox, signal});
     } catch (err) {
-      console.log(err); // eslint-disable-line
+      error = err || true;
+    } finally {
+      requestToken.done();
+
+      if (this._isCancelled && !tileData) {
+        this._isLoaded = false;
+      } else {
+        // Consider it loaded if we tried to cancel but `getTileData` still returned data
+        this._isLoaded = true;
+        this._isCancelled = false;
+      }
+    }
+
+    if (!this._isLoaded) {
+      return;
+    }
+
+    if (error) {
+      this.onTileError(error, this);
+    } else {
+      this.content = tileData;
+      this.onTileLoad(this);
     }
   }
   /* eslint-enable max-statements */
--- a/modules/geo-layers/src/tile-layer/tile-2d-header.js
+++ b/modules/geo-layers/src/tile-layer/tile-2d-header.js
@@ -45,58 +45,52 @@ export default class Tile2DHeader {
 
   /* eslint-disable max-statements */
   async _loadData(getTileData, requestScheduler) {
+    const {x, y, z, bbox} = this;
+
+    this._abortController = new AbortController(); // eslint-disable-line no-undef
+    const {signal} = this._abortController;
+    this._isLoaded = false;
+    const requestToken = await requestScheduler.scheduleRequest(this, tile => {
+      return tile.isSelected ? 1 : -1;
+    });
+
+    if (!requestToken) {
+      this._isCancelled = true;
+      return;
+    }
+    // A tile can be cancelled while being scheduled
+    if (this._isCancelled) {
+      requestToken.done();
+      return;
+    }
+
+    let tileData;
+    let error;
     try {
-      const {x, y, z, bbox} = this;
-
-      this._abortController = new AbortController(); // eslint-disable-line no-undef
-      const {signal} = this._abortController;
-      this._isLoaded = false;
-      console.log('here!', x, y, z); // eslint-disable-line
-      const requestToken = await requestScheduler.scheduleRequest(this, tile => {
-        return tile.isSelected ? 1 : -1;
-      });
-      console.log('here!', x, y, z); // eslint-disable-line
-
-      if (!requestToken) {
-        this._isCancelled = true;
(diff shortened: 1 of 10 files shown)


modules/geo-layers/src/tile-layer/tile-layer.js
```diff
@@ -57,28 +57,42 @@ export default class TileLayer extends CompositeLayer {
...
+      tileset.tiles.forEach(tile => {
+        tile.updateLayers = true;
       });
-      this.setState({tileset});
     } else if (changeFlags.propsChanged || changeFlags.updateTriggersChanged) {
       tileset.setOptions(this._getTilesetOptions(props));
       // if any props changed, delete the cached layers
-      this.state.tileset.tiles.forEach(tile => {
+      // Should we do the same as the above?  Not delete but use a boolean to cause an update?
+      tileset.tiles.forEach(tile => {
```
ilan-gold, line 89: @Pessimistress Not sure what the origin of this behavior was but seems similar to the above?  We should just set them to have `updateLayers` be `true`

modules/geo-layers/src/tile-layer/tile-layer.js
```diff
@@ -107,17 +121,19 @@ export default class TileLayer extends CompositeLayer {
     };
   }
 
-  _updateTileset() {
-    const {tileset} = this.state;
+  _updateTileset(tileset, reload = false) {
```
ilan-gold, line 124: Open to just setting up a new method but it feels like it would need to look similar i.e change frame number, check loading, call `onViewportLoad` etc.

modules/geo-layers/src/tile-layer/tileset-2d.js
```diff
@@ -45,6 +45,38 @@ export default class Tileset2D {
...
+
+    // Cache the last processed viewport
+    this._viewport = null;
+    this._selectedTiles = null;
+    this._frameNumber = 0;
+  }
+
+  loadOptions(opts) {
+    this._loadOptions(opts);
+  }
```
ilan-gold, line 64: There's already a `setOptions` function - should these be merged?

modules/geo-layers/src/tile-layer/tileset-2d.js
```diff
@@ -323,7 +344,7 @@ export default class Tileset2D {
   }
   /* eslint-enable complexity */
 
-  _getTile({x, y, z}, create) {
+  _getTile({x, y, z}, create, reload) {
```
ilan-gold, line 347: Better way to do forced reloads?

ilan-gold (commented):
Some comments.

ilan-gold:
Also, the `requestScheduler` appears to be hanging.  Whenever I comment out all the bits about it, the tests pass (obviously the `Tile2DHeader` tests fail though).

ilan-gold:
This is what I was talking about with the scheduler - I just pushed a change with some debugging changes.  I log out before and after the scheduler is called, and then look at the output:

```
# Custom getTileData
waiting for request token!
waiting for request token!
ok 4134 Rendered sublayers
not ok 4135 Fetched tile data
  ---
    operator: equal
    expected: 2
    actual:   0
    at: Object.onAfterUpdate (/Users/ilan_gold/deck.gl/test/modules/geo-layers/tile-layer/tile-layer.spec.js:99:13)
    stack: |-
      Error: Fetched tile data
          at Test.assert [as _assert] (/Users/ilan_gold/deck.gl/node_modules/tape/lib/test.js:228:54)
          at Test.bound [as _assert] (/Users/ilan_gold/deck.gl/node_modules/tape/lib/test.js:80:32)
          at Test.equal (/Users/ilan_gold/deck.gl/node_modules/tape/lib/test.js:389:10)
          at Test.is (/Users/ilan_gold/deck.gl/node_modules/tape/lib/test.js:80:32)
          at Object.onAfterUpdate (/Users/ilan_gold/deck.gl/test/modules/geo-layers/tile-layer/tile-layer.spec.js:99:13)
          at runLayerTestPostUpdateCheck (/Users/ilan_gold/deck.gl/modules/test-utils/src/lifecycle-test.js:187:14)
          at testLayerAsync (/Users/ilan_gold/deck.gl/modules/test-utils/src/lifecycle-test.js:103:5)
          at Test.<anonymous> (/Users/ilan_gold/deck.gl/test/modules/geo-layers/tile-layer/tile-layer.spec.js:168:3)
  ...
ok 4136 Layer is loaded
ok 4137 Sublayers at z=2 are visible
waiting for request token!
waiting for request token!
got request token!
got request token!
got request token!
got request token!
```
(Assuming the logging is right) The tokens come back after the tests.  But this isn't the case on the current master branch so there must be something going on here that I'm missing:

```
# Custom getTileData
waiting for request token!
waiting for request token!
ok 4136 should be truthy
got request token!
got request token!
ok 4137 Rendered sublayers
ok 4138 Fetched tile data
ok 4139 Layer is loaded
ok 4140 Sublayers at z=2 are visible
```

One thing I notice is that https://github.com/visgl/deck.gl/blob/e98df82c1bbcce31c9763726af090175ee3a6558/test/modules/geo-layers/tile-layer/tile-layer.spec.js#L94-L97 is being called on master and not this branch which would make sense if the scheduler is hanging...going to keep digging

Pessimistress:
@ilan-gold I pushed a different approach:

- `Tile2DHeader.loadData` can be called multiple times and the result from the last call is used
- `Tileset2D.reloadAll` flags all cached tiles as needs reload
- When a tile is selected during `Tileset2D.update`, call `tile.loadData` if it's flagged as needs reload

Also made some internal API changes to `tile.loadData` to improve readability. Unit tests are updated to reflect this API change.

modules/geo-layers/src/tile-layer/tile-2d-header.js
```diff
@@ -64,47 +69,56 @@ export default class Tile2DHeader {
...
-      return;
-    }
-
+  loadData(opts) {
     this._isCancelled = false;
-    this._loader = this._loadData(getTileData, requestScheduler);
-    this._loader.finally(() => {
+    this._needsReload = false;
+    this._loaderId++;
+    this._loader = this._loadData(opts);
```
ilan-gold, line 112:
```suggestion
    this._isLoaded = false;
    this._needsReload = false;
    this._loaderId++;
    this._loader = this._loadData(opts);
```

I think?

Since the state of loading upstream depends on all tiles being loaded, I think when the update triggers for the `getTileData` change, the tile will remain loaded when this method is called again since `isLoaded` has never been set to `false`.  For example. I think the `onViewportLoad` method will be called immediately when `getTileData` changes not when the new data has loaded.

Pessimistress: I left the `isLoaded` flag on because it is used in `renderLayers` to determine if the tile should be rendered. But you are right that it doesn't work for the `onViewportLoad` callback. I think it's right to reset it here and we need a different check for layer rendering.

ilan-gold:
Screen Shot 2021-09-29 at 3 33 00 PM
Changing channels (i.e `updateTriggers`) too quickly yields this result.  I'll look into it.

modules/geo-layers/src/tile-layer/tileset-2d.js
```diff
@@ -107,6 +104,15 @@ export default class Tileset2D {
...
     }
+    this._cache.clear();
+    this._tiles = [];
+    this._selectedTiles = null;
+  }
+
+  reloadAll() {
+    for (const tile of this._cache.values()) {
+      tile.setNeedsReload();
+    }
```
ilan-gold, line 115: Do you think it makes sense to have two different loops here, one for reloading `_selectedTiles` i.e those in the viewport, and then another for clearing the cache of those outside the viewport?  For example if you zoom out after the load has occurred, you will get old tiles sometimes.  I think it would be better to have just no tiles outside the viewport instead of old data.  I think that was the idea originally with my (very bad) code: https://github.com/visgl/deck.gl/pull/6194/commits/8f4cdd2b7fdaa6d4ac4591c8e0026874d96f5cdd#diff-ebfc47135e3d8fe2bee8b3d33fae4ebbbc295d01887c751d4cb77f71fc85a87aR70-R77

ilan-gold:
```suggestion
  reloadAll() {
    for (const [tileId, tile] of this._cache) {
      if (!this._selectedTiles.includes(tile)) {
        this._cache.delete(tileId);
      } else {
        tile.setNeedsReload();
      }
    }
```
Something like this maybe?

Pessimistress: Yes the proposed behavior makes sense. Feel free to push a commit.

I'm always a little wary about deleting items in a traversal. `for (const tileId of this._cache.keys())` maybe?

ilan-gold: updated!

Pessimistress (approved):

PR #6242, opened 2021-09-27
H3HexagonLayer: force low precision; early exit for data analysis loop

igorDykhta:
For #6238

#### Change List
- Update highPrecision prop to expect 'auto' | true | false values. Default is now 'auto'. (Breaking change)
- The h3 data is fully analyzed only for highPrecision: auto mode.

### DIFF
--- a/docs/api-reference/geo-layers/h3-hexagon-layer.md
+++ b/docs/api-reference/geo-layers/h3-hexagon-layer.md
@@ -102,7 +102,7 @@ Possible values:
 
 * `'auto'`: The layer chooses the mode automatically. High-precision rendering is only used if an edge case is encountered in the data.
 * `true`: Always use high-precision rendering.
-* `false`: Always use low-precision rendering regardless of the resolution of the data.
+* `false`: Always use instanced rendering, regardless of the characteristics of the data.
 
 ##### `resolution` (number, optional)
 
--- a/docs/api-reference/geo-layers/h3-hexagon-layer.md
+++ b/docs/api-reference/geo-layers/h3-hexagon-layer.md
@@ -104,28 +104,6 @@ Possible values:
 * `true`: Always use high-precision rendering.
 * `false`: Always use instanced rendering, regardless of the characteristics of the data.
 
-##### `resolution` (number, optional)
-
-* Default: `-1`
-
-Resolution of the data. It's possible to save some initialization time by passing the resolution, especially for large datasets. Values other than `-1` indicate that the data has only one resolution.
-
-Possible values:
-* `-1`: Indicates that the resolution is unknown and should be calculated.
-* `[0, 15]`: Valid h3 resolution.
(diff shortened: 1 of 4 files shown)


docs/api-reference/geo-layers/h3-hexagon-layer.md
```diff
@@ -103,6 +103,38 @@ Possible values:
 * `false`: The layer chooses the mode automatically. High-precision rendering is only used if an edge case is encountered in the data.
 * `true`: always use high-precision rendering.
 
+##### `lowPrecision` (Boolean, optional)
```
Pessimistress, line 106: It doesn't make sense to have two props for controlling the same thing. What if `highPrecision` and `lowPrecision` are both true? Can you just make `highPrecision` accept `true|false|'auto'` and default to `'auto'`?

docs/api-reference/geo-layers/h3-hexagon-layer.md
```diff
@@ -103,6 +103,38 @@ Possible values:
...
+
+* Default: `-1`
+
+Resolution of the data. It's possible to save some initialization time by passing the resolution, especially for large datasets. Values other than `-1` indicate that the data has only one resolution.
+
+Possible values:
+* `-1`: Indicates that the resolution is unknown and should be calculated.
+* `[0, 15]`: Valid h3 resolution.
+
+##### `hasPentagon` (boolean | null, optional)
```
Pessimistress, line 127: I don't think you need these two props if you:

- make the layer defer calculating these flags until it needs to
- always force high/low res rendering from the user application

Pessimistress (changes requested):

docs/api-reference/geo-layers/h3-hexagon-layer.md
igorDykhta: Is it going to be a breaking change in such case? As atm false indicates that precision should be selected automatically.

Pessimistress: Yes it will be breaking. Still better than adding confusing API IMO

docs/api-reference/geo-layers/h3-hexagon-layer.md
nrabinowitz: I think we'd still need `resolution`, which is currently inferred from data and used to get the cell used for instanced rendering. Otherwise you would need to infer that `lowPrecision: true` implied a single resolution, and loop through the data for the first valid cell to determine what that resolution was.

igorDykhta: Removed resolution and hasPentagon props. 
For highPrecision: false - select first resolution and exit the loop early.

Pessimistress (approved):
Add the breaking change to upgrade guide please

PR #6287, opened 2021-10-10
Fix remove html comments at whats-new

alasarr:
Gatsby doesn't support comments at whats new, so website cannot be built


alasarr:
Closing in favor of https://github.com/visgl/deck.gl/pull/6288

PR #6329, opened 2021-10-27
fix(mapbox) leave additional views intact

chrisgervang:
#### Background
The deck/mapbox interleaved rendering needs to reproject the viewport to match mapbox when an external deck object is provided. Currently the extension assumes the first View is a `MapView` and deletes any other viewports, however deletion is unnecessary and not expected if provided multiple views (e.g. a minimap or info viz overlay).

Screen Shot 2021-10-23 at 10 38 49 PM

**Result After Fix**
TextLayer on OrthographicView renders on top of interleaved MapView.
Screen Shot 2021-10-23 at 10 42 05 PM

#### Change List
Related to interleaved mapbox rendering:
- No `viewId` provided - same behavior as today
- `viewId` provided and found - replace specified viewport
- `viewId` provided and not found - do nothing


docs/api-reference/mapbox/mapbox-layer.md
```diff
@@ -101,6 +101,7 @@ Parameters:
 - `props` (Object)
   + `props.id` (String) - an unique id is required for each layer.
   + `props.deck` (`Deck`, optional) - a `Deck` instance that controls the rendering of this layer. If provided, the layer will be looked up from its layer stack by `id` at render time, and all other props are ignored.
+  + `props.viewId` (String, optional) - If provided, the view will be looked up from the `Deck` instance's view stack by `id` at render time. Intended to only be used when a `Deck` instance is provided that has multiple `views`. If it is not provided or cannot be found, the first view will be used instead.
```
Pessimistress, line 104: If the prop is not provided, then we should keep the current behavior (only render a single viewport from Map).

If the prop is provided but no matching view is found, then we shouldn't replace any viewport.

chrisgervang: That is how it will behave. Should I replace the doc with your comment's phrasing?

Pessimistress (approved):

docs/api-reference/mapbox/mapbox-layer.md
```diff
@@ -101,6 +101,7 @@ Parameters:
 - `props` (Object)
   + `props.id` (String) - an unique id is required for each layer.
   + `props.deck` (`Deck`, optional) - a `Deck` instance that controls the rendering of this layer. If provided, the layer will be looked up from its layer stack by `id` at render time, and all other props are ignored.
+  + `props.viewId` (String, optional) - Provide the `id` of a `MapView` instance to render the Map to (e.g. if using multiple `views` on a `Deck` instance). If it is not provided, the Map will be rendered on a single viewport. If a viewId is not found, nothing will be done to the viewports.
```
Pessimistress, line 104: Required if this layer should be rendered into multiple views. Specify the id of a view that corresponds to the base map.

- If not provided, the layer will only be drawn into the base map's viewport.
- If provided, the viewport with this id will be replace to match the base map.

docs/api-reference/mapbox/overview.md
```diff
@@ -14,7 +14,7 @@ Use deck.gl layers as custom Mapbox layers, enabling seamless interleaving of Ma
 
 ### Limitations
 
-* deck.gl's multi-view system cannot be used.
+* deck.gl's multi-view system only supports one MapView (e.g. minimaps are not supported).
```
Pessimistress, line 17: Technically speaking you can have multiple `MapView`s, the limitation is that you will always have a MapView that fills the entire canvas (what Mapbox draws into).

modules/mapbox/src/deck-utils.js
```diff
@@ -28,7 +33,8 @@ export function getDeckInstance({map, gl, deck}) {
     userData: {
       isExternal: false,
       mapboxLayers: new Set()
-    }
+    },
+    views: deck.props.views || [new MapView({id: 'mapbox'})]
```
Pessimistress, line 37: This will be overwritten if you use React and the component rerenders. `views` must be resolved before `drawLayer` is called, not at initialization.

modules/mapbox/src/deck-utils.js
```diff
@@ -88,9 +94,12 @@ export function drawLayer(deck, map, layer) {
...
     return;
   }
+
   deck._drawLayers('mapbox-repaint', {
-    viewports: [currentViewport],
-    layerFilter: ({layer: deckLayer}) => layer.id === deckLayer.id,
+    viewports: applyViewport(deck, map, currentViewport),
+    layerFilter: props =>
+      (deck.props.layerFilter && deck.props.layerFilter(props)) ||
+      (props.viewport.id === 'mapbox' && layer.id === props.layer.id),
```
Pessimistress, line 102: The two filters must both pass. This function is called once for every Mapbox Layer.

```js
params =>
      (!deck.props.layerFilter || deck.props.layerFilter(params)) &&
      (params.viewport.id === 'mapbox' && layer.id === params .layer.id),
```

modules/mapbox/src/deck-utils.js
```diff
@@ -159,8 +178,10 @@ function afterRender(deck, map) {
     const hasNonMapboxLayers = deck.props.layers.some(layer => !mapboxLayerIds.includes(layer.id));
     if (hasNonMapboxLayers) {
       deck._drawLayers('mapbox-repaint', {
-        viewports: [getViewport(deck, map, false)],
-        layerFilter: ({layer}) => !mapboxLayerIds.includes(layer.id),
+        viewports: applyViewport(deck, map),
+        layerFilter: props =>
+          (deck.props.layerFilter && deck.props.layerFilter(props)) ||
```
Pessimistress, line 183: Ditto

modules/mapbox/src/deck-utils.js
chrisgervang: Should I memoize `new MapView({id: 'mapbox'})` in userData, or is it fine to re-initialize on drawLayer calls?

modules/mapbox/src/deck-utils.js
```diff
@@ -8,6 +8,11 @@ export function getDeckInstance({map, gl, deck}) {
 
   const customRender = deck && deck.props._customRender;
 
+  if (deck && deck.props.views && !deck.props.views.some(view => view.id === 'mapbox')) {
```
Pessimistress, line 11: You can't manipulate views during initialization. If the user defines views inline in React then it will be replaced.

modules/mapbox/src/deck-utils.js
Pessimistress: You don't need to mutate the prop at all. In applyViewport you can just use the viewport ids to figure out where to position the map viewport.

chrisgervang: That would work if the "mapbox" view is present, right? Would we insert the view if it's missing somewhere?

Pessimistress: You don't need the view since deck is not providing the controller here.

chrisgervang: Ok, I see two `_drawLayers` calls but it isn't clear how to only manipulate views for external deck instances. 

In `afterRender`, `applyViewport` is only called when deck is external - that is clear. But `drawLayer` it is called as long as `deck.layerManager` is defined - there is no check for deck being external, so it is unclear.

modules/mapbox/src/deck-utils.js
chrisgervang: Ok, that makes sense. This can be moved into the `_drawLayers` calls

Pessimistress (approved):

docs/api-reference/mapbox/overview.md
chrisgervang: Ah, is that why minimap views can't use mapbox? I'd like to explain the limitation better in the overview in case its useful to someone hacking around it.

docs/api-reference/mapbox/mapbox-layer.md
chrisgervang: We decided against this in the end, resolving.

Issue #6335, opened 2021-10-28
Some props are missing from GeoJsonLayer


urdeveloper:
#### Target Use case
While working with GeoJsonLayer, I've noticed some of the properties related to point features are missing:

Missing props from TextLayer:
- `billboard`
- `fontSettings`

Missing props from IconLayer:
- `billboard`
- `alphaCuttoff`

Missing props from ScatterPlotLayer:
- `billboard`

It would be great to make the available GeoJsonLayer props consistent with what's available in sub layers.

#### Proposed feature
Map the missing props from GeoJsonLayer to corresponding sub layers props.

#### To Do List 
  - [ ] Add label and assign to milestone 
  - [ ] Coding
  - [ ] Doc update
  - [ ] What’s new update 
  - [ ] Test

Pessimistress:
Thank you for your contribution! The new props are now available in v8.6.3.

PR #6336, opened 2021-10-28
add-missing-props-geojsonlayer

urdeveloper:
For #6335

#### Background
Some of the sublayer properties are missing from GeoJsonLayer. This PR will add the necessary mapping to support those properties.

#### Change List
- Add property mappings
- Update documentation


Pessimistress (approved):

urdeveloper:
> Thanks. Looks good. Could you perhaps add a test to `test/render/test-cases/geojson-layer.js` to demonstrate this fix is working?

Thanks, been very busy. Will work on the tests this week.

Issue #6337, opened 2021-10-29
Easily create deck.gl layers from a map created in CARTO Builder


felixpalmer:
#### Target Use case
Currently when a map is created in the CARTO Builder interface, there is not easy way to export it to be used within an application, with all the styles mapped across. 

#### Proposed feature
Addition of a `getMap` function to the CARTO module, which returns a set of pre-configured deck.gl layers, which a developer can then use to build an application, without having to worry about synchronizing the styles that are created in the CARTO cloud. The proposed API is:

```javascript
import {getMap, setDefaultCredentials, API_VERSIONS} from '@deck.gl/carto';

setDefaultCredentials({
  apiVersion: API_VERSIONS.V3,
  apiBaseUrl: 'https://gcp-us-east1.api.carto.com'
});

const deck = new Deck({canvas: 'deck-canvas'});
const {mapState, mapStyle, layers} = await getMap({mapId: '<UNIQUE_CARTO_MAP_ID>');
deck.setProps({initialViewState: mapState, layers});
```

The API is kept as simple, but by exposing the layers in this manner, the consumer of the library can modify them at will, to add interactivity/animation etc.

#### To Do List 
  - [x] Add label and assign to milestone 
  - [x] Coding
  - [ ] Doc update
  - [ ] What’s new update 
  - [x] Test

PR #6341, opened 2021-10-29
Add `fetchMap` function to carto module

felixpalmer:
For https://github.com/visgl/deck.gl/issues/6337

#### Change List
- Add `getMap` function for fetching a set of predefined layers from CARTO
- Example app to compare output for a given map id
- Implementation of autoRefresh to periodically update layers when data from datasources changes

Test updates and documentation to follow in separate PR

### DIFF
--- a/modules/carto/src/api/layer-map.js
+++ b/modules/carto/src/api/layer-map.js
@@ -1,5 +1,6 @@
 import {extent} from 'd3-array';
-import {scaleSqrt} from 'd3-scale';
+import {rgb} from 'd3-color';
+import {scaleSqrt, scaleQuantize} from 'd3-scale';
 import CartoLayer from '../layers/carto-layer';
 
 const RADIUS_DOWNSCALE = 4;
@@ -65,8 +66,17 @@ export function getLayerMap() {
   return layerMap;
 }
 
-// Manually transform data
-// Range of input data
+export function getColorAccessor({name}, scale, range, {data}) {
+  const dataExtent = extent(data.features, ({properties}) => properties[name]);
+  const scaler = scaleQuantize()
+    .domain(dataExtent)
+    .range(range);
+  return ({properties}) => {
+    const rgba = rgb(scaler(properties[name]));
+    return [rgba.r, rgba.g, rgba.b, 255 * rgba.a];
+  };
+}
+
 export function getSizeAccessor({name}, scale, range, {data}) {
   const dataExtent = extent(data.features, ({properties}) => properties[name]);
   const radiusRange = range.map(r => r / RADIUS_DOWNSCALE);
--- a/modules/carto/src/api/layer-map.js
+++ b/modules/carto/src/api/layer-map.js
@@ -6,10 +6,13 @@ const sharedPropMap = {
   filled: 'filled',
   fixedRadius: {pointRadiusUnits: v => (v ? 'meters' : 'pixels')},
   getLineColor: 'strokeColor',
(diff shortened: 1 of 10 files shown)


modules/carto/src/api/maps-v3-client.js
```diff
@@ -167,6 +183,70 @@ export async function getData({type, source, connection, credentials, geoColumn,
...
+  const {connectionName: connection, source, type} = dataset;
+  const {url, mapFormat} = await _getDataUrl({
+    credentials: {accessToken},
+    connection,
+    source,
+    type
+  });
+
+  // Extract the last time the data changed
+  const cache = parseInt(new URLSearchParams(url).get('cache'), 10);
```
alasarr, line 201: Do we really need this cache system? Data will be cached by the CDN and the browser.

If the data changes the URL will be different, if the data doesn't change the URL will be the same but with a cache-control of 1 year:

`cache-control: public,max-age=31536000,must-revalidate`

So I think we can use the default cache system of the browser and remove these lines

modules/carto/src/api/maps-v3-client.js
```diff
@@ -124,6 +125,21 @@ function getUrlFromMetadata(metadata, format) {
 }
 
 export async function getData({type, source, connection, credentials, geoColumn, columns, format}) {
+  // Internally we split data fetching into two parts to allow us to
```
alasarr, line 128: I think we should trust the standard browser cache system to fetch the actual. In any case, I stille like the idea of split getData method 👍

modules/carto/src/api/maps-v3-client.js
```diff
@@ -167,6 +183,70 @@ export async function getData({type, source, connection, credentials, geoColumn,
...
+  }
+
+  if (autoRefresh || onNewData) {
+    log.assert(onNewData, 'Must define `onNewData` when using autoRefresh');
+    log.assert(typeof onNewData === 'function', '`onNewData` must be a function');
+    log.assert(autoRefresh > 0, '`autoRefresh` must be a positive number');
+  }
+
+  const url = `${localCreds.mapsUrl}/public/${mapId}`;
+  const map = await request({url});
```
alasarr, line 236: Would be great if we could cache here the status response to say this functionality only works for public maps

modules/carto/src/api/layer-map.js
```diff
@@ -0,0 +1,174 @@
...
+  wrapLongitude: false
+};
+
+function mergePropMaps(a, b) {
+  return {...a, ...b, visConfig: {...a.visConfig, ...b.visConfig}};
+}
+
+export const LAYER_MAP = {
+  point: {
+    Layer: GeoJsonLayer,
```
alasarr, line 67: Builder maps point Layer with ScatterplotLayer

alasarr (changes requested):
Looks amazing dude! It's a killer feature to make easy the creation of maps in deck.gl.

Congrats on the API and how you've resolved such a complex task

@AnaDiaz85 can you help here with QA? Ping @felixpalmer to let you know how to setup the local environment and start the sample app he has just built

modules/carto/src/api/maps-v3-client.js
felixpalmer: The reason to grab the cache parameter is to avoid doing the second request at all. It is true that if the cache is unchanged the browser will return the response from the local cache, but the response will still get processed, which means the deck.gl layers will all get recreated

alasarr (approved):

Issue #6230, opened 2021-09-24
CartoLayer v3 is not rendered when setting local credentials


Jesus89:
#### Description
In CartoLayer, passing `"apiVersion": "v3"` to the layer credentials gets the wrong `"mapUrl"` and the layer is not rendered.

#### Expected Behavior
It should get the correct `"mapUrl"` and the layer should be rendered.

#### Repro Steps
Load the following JSON in deck.gl

```
{
    "initialViewState": {
        "longitude": 0,
        "latitude": 0,
        "zoom": 0
    },
    "views": [{
        "@@type": "MapView",
        "controller": true,
        "mapStyle": "https://basemaps.cartocdn.com/gl/dark-matter-nolabels-gl-style/style.json"
    }],
    "layers": [{
        "@@type": "CartoLayer",
        "data": "SELECT * FROM cartobq.public_account.world_population_2015",
        "type": "query",
        "connection": "bigquery",
        "credentials": {
            "apiVersion": "v3",
            "apiBaseUrl": "https://gcp-us-east1.api.carto.com",
            "accessToken": "eyJhbGciOiJIUzI1NiJ9.eyJhIjoiYWNfN3hoZnd5bWwiLCJqdGkiOiJlMzhhMzliNyJ9.Qm0I3bJ-676TMJx6ZoVYWg4iuKuHqYyeKU5xNGU82y0"
        },
        "pointRadiusMinPixels": 2,
        "getLineColor": [0, 0, 0, 125],
        "getFillColor": [238, 77, 90],
        "lineWidthMinPixels": 1
    }]
}
```

Note that passing `"mapsUrl": "https://gcp-us-east1.api.carto.com/v3/maps"` to the `"credentials"` object solves the issue because it sets the correct value.

#### Environment
 - Framework Version: deck.gl 8.5.9

#### Logs
```
GET https://maps-api-v2.%7Bregion%7D.carto.com/user/%7Buser%7D/bigquery/query?client=deck-gl-carto&q=SELECT%20*%20FROM%20cartobq.public_account.world_population_2015 net::ERR_NAME_NOT_RESOLVED
```

alasarr:
Fixed at 8.6.2

PR #6349, opened 2021-11-02
CARTO: fix layer credentials overwrite

felixpalmer:
For https://github.com/visgl/deck.gl/issues/6230

#### Change List
- Add tests to capture case where CARTO is initialized without `setDefaultCredentials` (repro for bug)
- Modify `getData` to only use default credentials if they have been defined for correct API version

Source of bug was fact that V1 & V2 API uses a format string for the `mapsUrl` so it is by default not null. Sanity checking we have the right API version in the default credentials is good anyway and removes the source of this bug

### DIFF
--- a/modules/carto/src/api/maps-v3-client.js
+++ b/modules/carto/src/api/maps-v3-client.js
@@ -68,9 +68,11 @@ function dealWithError({response, error}) {
 /**
  * Build a URL with all required parameters
  */
-function getParameters({type, source, geoColumn, columns}) {
-  const encodedClient = encodeParameter('client', 'deck-gl-carto');
-  const parameters = [encodedClient];
+function getParameters({type, source, geoColumn, columns, schema}) {
+  const parameters = [encodeParameter('client', 'deck-gl-carto')];
+  if (schema) {
+    parameters.push(encodeParameter('schema', true));
+  }
 
   const sourceName = type === MAP_TYPES.QUERY ? 'q' : 'name';
   parameters.push(encodeParameter(sourceName, source));
@@ -93,10 +95,11 @@ export async function mapInstantiation({
   connection,
   credentials,
   geoColumn,
-  columns
+  columns,
+  schema
 }) {
   const baseUrl = `${credentials.mapsUrl}/${connection}/${type}`;
-  const url = `${baseUrl}?${getParameters({type, source, geoColumn, columns})}`;
+  const url = `${baseUrl}?${getParameters({type, source, geoColumn, columns, schema})}`;
   const {accessToken} = credentials;
 
   const format = 'json';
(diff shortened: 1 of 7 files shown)


modules/carto/src/api/maps-v3-client.js
```diff
@@ -133,7 +139,6 @@ export async function getData({type, source, connection, credentials, geoColumn,
   log.assert(localCreds.apiVersion === API_VERSIONS.V3, 'Method only available for v3');
   log.assert(localCreds.apiBaseUrl, 'Must define apiBaseUrl');
   log.assert(localCreds.accessToken, 'Must define an accessToken');
-  log.assert(localCreds.mapsUrl, 'mapsUrl cannot be undefined');
 
   if (!localCreds.mapsUrl) {
```
alasarr, line 143: Not 100% sure, but I think localCreds.mapsUrl is not overwritten. As we defined mapsUrl at setDefaultCredentials, https://github.com/visgl/deck.gl/blob/master/modules/carto/src/config.js#L52.

Could you try this?

```
setDefaultCredentials({ 
    apiVersion: 'v3',
    apiBaseUrl: 'api1'
})

CartoLayer({
    ...
    credentials: { apiBaseUrl: 'api2'} 
})
```

I think api2 will be ignored

alasarr (changes requested):

modules/carto/src/api/maps-v3-client.js
```diff
@@ -124,7 +124,13 @@ function getUrlFromMetadata(metadata, format) {
 }
 
 export async function getData({type, source, connection, credentials, geoColumn, columns, format}) {
-  const localCreds = {...getDefaultCredentials(), ...credentials};
+  const defaultCredentials = getDefaultCredentials();
+  // Only pick up default credentials if they have been defined for
+  // correct API version
+  const localCreds = {
```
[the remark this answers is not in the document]
felixpalmer: Done

modules/carto/src/api/maps-v3-client.js
felixpalmer: I feel that the current behavior of having multiple places where credentials can be set and validates makes the code overcomplicated, especially as `setDefaultCredentials` goes and backfills options which are missing. This makes it hard to see from the outside what is going on and leads to situations which I'm not even sure we can reconcile. For example:

```
setDefaultCredentials({ 
    apiVersion: 'v3',
    apiBaseUrl: 'api1',
    mapsUrl: 'http://custom'
})

CartoLayer({
    ...
    credentials: { apiBaseUrl: 'api2'} 
})
```

It don't think it is clear to the user what is going to take precendence here. Should the explicitly specified `mapsUrl` win? Or should the later `apiBaseUrl: 'api2'` be used to overwrite it? I would say maybe we should take a step back and make `setDefaultCredentials` less clever, and rather validate all the parameters at the point at which we need them. We need to do that anyway as we allow partial overriding of the credentials.

`setDefaultCredentials` then just holds onto whatever is passed to it to help the user simplify this pattern: 

```
// Credentials common to whole app
const defaultCredentials = {...}

// Create layers with optional credential over-rides
CartoLayer({..., credentials: { apiBaseUrl: 'api1', ...defaultCredentials} })
CartoLayer({..., credentials: { apiBaseUrl: 'api2', ...defaultCredentials} })
```

To this:

```
// Credentials common to whole app
const defaultCredentials = {...}
setDefaultCredentials(defaultCredentials);

// Create layers with optional credential over-rides
CartoLayer({..., credentials: { apiBaseUrl: 'api1'} })
CartoLayer({..., credentials: { apiBaseUrl: 'api2} })
```

alasarr: >I would say maybe we should take a step back and make setDefaultCredentials less clever, and rather validate all the parameters at the point at which we need them.
Agree

alasarr (approved):

PR #6356, opened 2021-11-03
Hooks names must be uppercase

afonsof:
This doc with lower case chars was leading to misunderstandings


Pessimistress (approved):
Thank you!

Issue #6362, opened 2021-11-04
deck.pickMultipleObjects returning duplicate features in binary MVTLayer


felixpalmer:
#### Description
`MVTLayer` is returning duplicate features from `deck.pickMultipleObjects` when in fact there is only one in the dataset. The issue only occurs when `binary: true`.

#### Expected Behavior
Correct picking as per `binary: false`

#### Repro Steps
https://stackblitz.com/edit/deckgl-pickmultipleobjects?file=index.js
Hover over polygon and observe console output. Compare with same layer, but with `binary: false`.

Clebal:
@felixpalmer If you zoom in, the picked feature is repeated even with binary as false.

reececasey:
I am seeing the same issue on the GeoJsonLayer is this related?

Pessimistress:
The issue is present when `layerFilter` or `CompositeLayer.filterSubLayer` is involved. I will put up a fix.

Pessimistress:
Fixed in v8.6.4.

PR #6380, opened 2021-11-10
Fix multi-depth picking with layerFilter

Pessimistress:
For #6362

Fallout of #6184

#### Change List
- Remove ad-hoc picking color decoding
- Unit test

### DIFF
--- a/modules/core/src/lib/deck-picker.js
+++ b/modules/core/src/lib/deck-picker.js
@@ -177,7 +177,7 @@ export default class DeckPicker {
 
     let infos;
     const result = [];
-    const affectedLayers = {};
+    const affectedLayers = new Set();
 
     for (let i = 0; i < depth; i++) {
       const pickedResult =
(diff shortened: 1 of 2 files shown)


felixpalmer (approved):
Looks good. 

Just to check my understanding: the functional change is `if (pickInfo.pickedColor && i + 1 < depth) {` becomes `if (pickInfo.pickedLayer && i + 1 < depth) {`? 

This is done as old code could return a layer but without a picking color (due to an applied filter), thus causing the `break` to be invoked too soon. With this change we simply disable the picking index in that filtered layer and repeat the picking for layers at a lower depth, thus picking up the previously dropped features.

Pessimistress:
`pickedColor` always exists if something is picked. The issue is that after #6184 we no longer use `pickedColor[3]` to encode the index in the layers array. `pickedLayer` comes from proper decoding of the picked color.

PR #6398, opened 2021-11-15
[WIP] Documentation for CARTO `fetchMap`

felixpalmer:
Followup PR adding documentation for #6337

#### Change List
- Documentation for `fetchMap`
- Updates to docs to give priority to `apiVersion` `v3`
- Updates to examples to use new API


felixpalmer:
@borja-munoz @alasarr could you take a look at the updated example in `examples/get-started/pure-js/carto/app.js`? In the overview we link to 3 flavours (scripting, React & pure-js) of an example and I would like to update these to all be the same simple example using `fetchMap`. These will all reside in the deck.gl repo. 

I've tried to keep it as simple as possible, just setting up a map using `fetchMap` and attaching it to a basemap. It's a bit more verbose than the "magic one-liner" that I included in the overview. This has a good wow-effect, but it isn't as clear what is going on.

How do you feel about this approach?

borja-munoz (commented):
Looks good to me but I have two comments:

- We might need to have a section in the overview page or somewhere else to document the input and output objects accepted/returned by the `fetchMap` function. 

- I am not sure about changing the example to use just the `fetchMap` function. I like the simplicity but we lose an example showing how to use the `CartoLayer`, the CARTO basemaps, the color style helpers...

examples/get-started/pure-js/carto/app.js
```diff
@@ -1,98 +1,22 @@
...
-    visibleLayer = e.value;
-    render();
+import {fetchMap} from '@deck.gl/carto';
+
+const cartoMapId = 'ff6ac53f-741a-49fb-b615-d040bc5a96b8';
+const deck = new Deck({canvas: 'deck-canvas'});
+
+// Get map info from CARTO and update deck
+fetchMap({cartoMapId}).then(({initialViewState, mapStyle, layers}) => {
+  deck.setProps({initialViewState, layers});
```
Pessimistress, line 10: You should set `controller: true` here. The comment below implies that the next `setProps` call is optional, but this statement alone does not enable interaction.

felixpalmer:
@borja-munoz 

- I plan on adding a entire page (like we have for `CartoLayer` for `fetchMap`
- The `CartoLayer` doc page will remain with examples of use. However, I can see it being nice having the `CartoLayer` example than can just be launched. How about if I move the current example (upgraded to v3) into `get-started/pure-js/cartolayer` ?

docs/api-reference/carto/carto-bqtiler-layer.md
```diff
@@ -1,10 +1,11 @@
 # CartoBQTilerLayer
```
alasarr, line 1: We announced the deprecation of CartoBQTilerLayer in favor of  8.5, at 8.7 we should remove it completely, maybe in a separated PR, but just to not forget about it

docs/api-reference/carto/carto-sql-layer.md
```diff
@@ -70,14 +70,14 @@ new CartoSQLLayer({});
 To use pre-bundled scripts:
 
 ```html
-<script src="https://unpkg.com/deck.gl@^8.5.0/dist.min.js"></script>
-<script src="https://unpkg.com/@deck.gl/carto@^8.5.0/dist.min.js"></script>
+<script src="https://unpkg.com/deck.gl@^8.7.0/dist.min.js"></script>
```
alasarr, line 73: Same here, we should remove this layer in 8.7

docs/api-reference/carto/fetch-map.md
```diff
@@ -0,0 +1,82 @@
+# fetchMap
+
+`fetchMap` is an API that instantiates layers configured in [CARTO Builder](https://carto.com/builder/) for use with deck.gl. It is available starting with CARTO Maps API version v3.
```
alasarr, line 3: This part is more a work for product than engineering. @borja-munoz could you help here? 

This is a little bit confusing, we're linking to the CARTO builder landing page, so people can think it will work in the builder version of CARTO 2, have in mind lots of users don't know about API versions.

We need to rethink what's going to be the product strategy around this feature in our side and doc. 

In the meantime, to avoid blocking this PR, can we add here a quick gif to demo the feature? And remove the link

alasarr (changes requested):

docs/api-reference/carto/carto-sql-layer.md
felixpalmer: Split out into separate PR: https://github.com/visgl/deck.gl/pull/6402

borja-munoz:
> @borja-munoz
> 
> * I plan on adding a entire page (like we have for `CartoLayer` for `fetchMap`
> * The `CartoLayer` doc page will remain with examples of use. However, I can see it being nice having the `CartoLayer` example than can just be launched. How about if I move the current example (upgraded to v3) into `get-started/pure-js/cartolayer` ?

I agree

docs/api-reference/carto/fetch-map.md
borja-munoz: I agree, we should remove the link to the current Builder landing page until we update it in the public website. I think adding a video is a good idea.

alasarr (approved):

PR #6410, opened 2021-11-17
Pass through CARTO label

felixpalmer:
For https://github.com/visgl/deck.gl/issues/6337

#### Change List
- Pass through CARTO layer label


borja-munoz (approved):
Looks good to me!

Issue #6415, opened 2021-11-19
CartoLayer doesn't support `format` prop


felixpalmer:
#### Target Use case
In order to support dynamic tiling CartoLayer should allow setting the `format` prop
#### To Do List 
  - [x] Coding
  - [x] Test

PR #6416, opened 2021-11-19
CartoLayer support `format` prop

felixpalmer:
For https://github.com/visgl/deck.gl/issues/6415

#### Change List
- Prioritize tilejson data format
- Choose MVTLayer based on data format on, not map type
- Fix bug with credentials not passed through correctly in fetchMap


felixpalmer:
@alasarr based on the conversation, I would propose to change this PR so:
- geojson is still top priority (so revert my change https://github.com/visgl/deck.gl/pull/6416/files#diff-c10ab0eaee5657ffca0e684656f2fd6767b7e47b2cd6af7e347b0d357cdae448L213-R213)
- add the `format` prop to `CartoLayer` to let us test going forward

alasarr:
Yes, it makes sense to me, but as a first step not as the definitive one. We're doing an important change here...

In the current version, the layer used is defined by the type, i.e., if we use a type tileset we use MVTLayer, if type is table or query the type is geojson.

With this change, the layer used will be defined by the format instead and it's a bit confusing. 

Not sure about the solution, but definitely web need to think more about this @borja-munoz

alasarr:
What about adding a parameter `renderDeckLayer`? This parameter lets the user specify the deck.gl layer to use since we're breaking the map between type and deck.layer because now a table might be available in geojson or tileset

```javascript
new CartoLayer({
  id: 'osm_buildings',
  connection: 'bigquery',
  type: MAP_TYPES.TABLE,
  data: 'cartobq.testtables.table',
  renderDeckLayer: MVTLayer // (optional)
}),

new CartoLayer({
  id: 'osm_buildings',
  connection: 'bigquery',
  type: MAP_TYPES.TABLE,
  data: 'cartobq.testtables.table',
})

new CartoLayer({
  id: 'osm_buildings',
  connection: 'bigquery',
  type: MAP_TYPES.TILESET,
  data: 'cartobq.testtables.tileset',
  // renderDeckLayer no need to specify, we'll default to tileset
})

new CartoLayer({
  id: 'osm_buildings',
  connection: 'bigquery',
  type: MAP_TYPES.QUERY,
  data: 'cartobq.testtables.tileset',
  // renderDeckLayer no need to specify, we'll default to tileset
})

```

felixpalmer:
> 

Interesting idea, my first thought is why not just use `fetchLayerData` in this case? Also, there is a danger here that users will expect that they can provide any arbitrary layer here and expect it to work. Which it might, but then we might change something and it could break. 

My feeling is that we should have the separation as follows:

- `fetchLayerData` is only for data fetching and allows the user to specify what datasource to use and in which format
- `CartoLayer` provides the same datasource specification, but does not burden the user with the choice of format (allowing us to change them under the covers without causing breaking changes)
- `fetchMap` instantiates layers that have been created in Builder without the user even having to think about the concept of datasources

alasarr:
I like this approach more than what I suggested.

alasarr (approved):

alasarr:
For the next time, it would be great to avoid mixing bug fixing with new features in the same PR. We use the commit message to track the changelog and it makes thing harder for maintenance (specially in patch releases)

Issue #6433, opened 2021-11-27
Links on pydeck docs pointing to wrong pages


dpshelio:
#### Description

Various links in the [Layer Overview and Examples page](https://deckgl.readthedocs.io/en/latest/layer.html) and [Deck](https://deckgl.readthedocs.io/en/latest/deck.html#deck) are pointing to non-existing locations.

#### Expected Behavior

Going to the right page :sweat_smile: 

#### Repro Steps

- Visit [Layer Overview and Examples page](https://deckgl.readthedocs.io/en/latest/layer.html) and click on the first link refered as `deck.gl layer catalog`, this points to a RTD non-existent page. 
- Visit the [Understanding keyword arguments in pydeck layers section](https://deckgl.readthedocs.io/en/latest/layer.html#understanding-keyword-arguments-in-pydeck-layers) and click on the `deck.gl layer catalog` link. This point to a non-existent page in github.
- Visit [Deck section](https://deckgl.readthedocs.io/en/latest/deck.html#deck). Click on the `Deck` Class link, this takes to the wrong page (i.e., no the documentation)

#### Environment
 - ~Framework Version: [e.g. deck.gl 8.3.0 / pydeck 0.5.0]~
 - Browser Version: Any browser
 - OS: Any OS

PR #6434, opened 2021-11-27
pydeck docs: Update links pointing to deck.gl

dpshelio:
For #6433

#### Change List
- Fix links pointing to deck.gl documentation on pydeck docs


Pessimistress (approved):
Thank you!

Issue #6423, opened 2021-11-25
GeoJsonLayer on GlobeView wrong lighting


Baseform:
#### Description
When using GeoJsonLayer on GlobeView lighting effects are wrong. Normals seem not to be calculated for Polygons in GlobeView.

#### Expected Behavior
With the right normals, GeoJsonLayer/Polygon layers should normally, behave as SimpleMesh does.

#### Repro Steps
https://codepen.io/diogovit/full/OJjeeJG

#### Environment
Using the latest build of deckgl from unpkg.

<img width="483" alt="Screenshot 2021-11-25 at 10 40 49" src="https://user-images.githubusercontent.com/1381857/143426798-53bd37ba-3899-4acf-ae08-61fedce5e74a.png">

Pessimistress:
The fix will be landed in v8.7.0 since it introduces visual breaking changes. You can try it now with `deck.gl@beta`.

PR #6435, opened 2021-11-29
Fixing point lighting on globe view

Baseform:
For #6423
#### Change List
-adding normals for top vertex shader


Pessimistress:
You can't just rewrite normal for all projection modes. A similar technique to https://github.com/visgl/deck.gl/blob/master/modules/core/src/shaderlib/project32/project32.js#L28 is required.

Pessimistress:
Redirect to #6445

PR #6446, opened 2021-11-30
Update text-layer.md

brianshano:
Fix link to Unit System docs

For #

#### Change List
-


Pessimistress:
I think this blob of text was copy&pasted for all layer docs. The same link is possibly broken in more places :(

ibgreen (approved):

Issue #6451, opened 2021-11-30
TerrainLayer is limited to Web Mercator Projection


warflash:
#### Description
Currently TerrainLayer is limited to only supporting Web Mercator Projection as mentioned here: https://github.com/visgl/deck.gl/discussions/6431

#### Expected Behavior
Terrain Layer has usecases that don't involve geospatial projections, so adding support for those would be nice!

Pessimistress:
Fixed in v8.6.5.

PR #6461, opened 2021-12-01
TerrainLayer handle non-WebMercator viewports

Pessimistress:
For https://github.com/visgl/deck.gl/issues/6451

Use more generic viewport handling.

This change unblocks the `TerrainLayer` from being used in `OrthographicView` and `OrbitView`. The layer still does not work in `GlobeView` (would require the `TerrainLoader` transform Mercator coordinates that the tiles are encoded in back to lngLat).

cc @ibgreen @igorDykhta to see if `GlobeView` support is of interest.

#### Change List
- Remove WebMercatorViewport specific code from `getTerrainData`


chrisgervang (approved):

kylebarron (approved):
I think a more standard way for the TerrainLayer to support a GlobeView is to support global geodetic tiling, so that the tiles are loaded in lng/lat coordinates.

PR #6466, opened 2021-12-02
feat(geo-layers): Tile3DLayer pass coordinateSystem as prop

belom88:
#### Background
I3S positions are stored as "lnglat" offsets. We need ability to pass coordinate system as property that allows to make positions transformation on the GPU level. This approach saves precision and increases performance.
#### Change List
- Tile3DLayer pass `coordinateSystem` as prop


ibgreen (approved):
Seems reasonable

Pessimistress (changes requested):
Is the current release broken then, since it force set the coordinate system to METER_OFFSET?

The offset systems are for local scale only. The precision  error is NOT acceptable beyond a ~30km radius. Before this layer moved to its own MeshLayer implementation, the sub layer is rendered using `CORDINATE_SYSTEM.LNGLAT` and adding `cartographicOrigin` and `positions` together in the shader.

ibgreen:
Hmm, 95% + of tiles are on local scales. Even if there are potential issues with large tiles, is there a really reason to prevent the layer to be configured this way.

Pessimistress:
The seams between tiles caused by projection error was an issue that needed fixing after we originally shipped i3s support. SimpleMeshLayer's `_instanced` prop was added specifically for this.

ibgreen:
Hmm yes I recall, we never did close that out properly. FWIW I still don't believe that we addressed the true issue. Those tiles are not even close to 30 km. I believe that there was something else going on that we haven't addressed. Other systems like Cesium do use local coordinates successfully.

belom88:
Before:
http://ec2-3-130-251-129.us-east-2.compute.amazonaws.com/sprint-3-demo/examples/i3s :
image

After:
http://ec2-3-130-251-129.us-east-2.compute.amazonaws.com/sprint-5-demo/examples/i3s
image

belom88:
Precision is lost here:
https://github.com/visgl/deck.gl/blob/969cc4808c2b6dd3ebe1dff86a3343c3b445f8ce/modules/geo-layers/src/tile-3d-layer/tile-3d-layer.js#L330 .
Float64 values are cut to Float32 . We have Float64 because of this transformation:
https://github.com/visgl/loaders.gl/blob/4378e975584cc8e99aa9f9a871088dd0c304d34e/modules/i3s/src/lib/parsers/parse-i3s-tile-content.ts#L418 .
If we use LNGLAT_OFFSETS we don't need this transformation and don't have Float64 anymore.

Pessimistress:
@belom88 `METER_OFFSETS` and `LNGLAT_OFFSETS` are not interchangable, so does your before/after contain changes in loaders.gl as well? If a new version of loaders.gl now returns `coordinateSystem` and different positions then this must be controlled by a loader option. This package only locks its loaders dependency to major versions, so an older release of deck can potentially pick up a newer version of loaders and break.

belom88:
> @belom88 `METER_OFFSETS` and `LNGLAT_OFFSETS` are not interchangable, so does your before/after contain changes in loaders.gl as well? If a new version of loaders.gl now returns `coordinateSystem` and different positions then this must be controlled by a loader option. This package only locks its loaders dependency to major versions, so an older release of deck can potentially pick up a newer version of loaders and break.

Yes. I've already added corresponding option in the loader:
https://github.com/visgl/loaders.gl/blob/4378e975584cc8e99aa9f9a871088dd0c304d34e/modules/i3s/src/i3s-loader.ts#L37
The default value is still `METER_OFFSETS` so this won't be breaking change.
Implementation for supporting both `METER_OFFSETS` and `LNGLAT_OFFSETS` is done and published in beta releases.

Pessimistress (approved):

Issue #6473, opened 2021-12-03
TypeError: Cannot read properties of undefined (reading 'length')


NameFILIP:
#### Description
The error happens here: https://github.com/visgl/deck.gl/blob/master/modules/core/src/lifecycle/props.js#L224 . The oldExtensions for some reason is `undefined`.

Using `PathMarkerLayer` from `@nebula.gl/layers`:

```
new PathMarkerLayer({
  id: 'my-layer-id',
  data: props.traces
    .filter((trace) => <some_condition>)
    .map((trace) => <some_transformation>),
  getColor: HIGHTLIGHT_COLOR,
  widthMinPixels: TRACE_WIDTH,
  widthMaxPixels: TRACE_WIDTH,
  sizeScale: 10,
})
```

A potential solution can be to add the following check to the `diffExtensions`:

```
if (extensions && !oldExtensions || !extensions && oldExtensions) {
  return true;
}
```

#### Environment
 - Framework Version: [deck.gl 8.3.15, nebula.gl 0.23.3]

Pessimistress:
Fixed in v8.6.5.

PR #6474, opened 2021-12-04
Fix "Cannot read properties of undefined (reading 'length')"

NameFILIP:
Fixes https://github.com/visgl/deck.gl/issues/6473


modules/core/src/lifecycle/props.js
```diff
@@ -221,6 +221,9 @@ function diffExtensions(props, oldProps) {
   if (extensions === oldExtensions) {
     return false;
   }
+  if ((extensions && !oldExtensions) || (!extensions && oldExtensions)) {
```
Pessimistress, line 224:
```suggestion
  if (!oldExtensions || !extensions) {
```

Pessimistress (approved):

PR #6478, opened 2021-12-06
feat(geo): controller position

belom88:
#### Background
In I3S almost all datasets have floating buildings. We need to set elevation for viewport to rotate about some custom z-position.
It is also could be helpful for TerrainLayer.
#### Change List
- pass position through MapStateProps;
- add `postition` to transition props.


test/modules/core/controllers/controllers.spec.js
```diff
@@ -15,7 +15,8 @@ test('MapController', async t => {
     latitude: 37.78,
     zoom: 10,
     pitch: 30,
-    bearing: -45
+    bearing: -45,
+    position: [0, 0, 0]
```
Pessimistress, line 19: If you have to add this for tests to pass then this will break every existing application that does not explicitly set `position` in its view state.

kylebarron:
> It is also could be helpful for TerrainLayer.

Yes, it would be a great enhancement for terrain!

chrisgervang:
Ohh yes, this can unlock some exciting views! 

It might be good to note somewhere that basemap libraries might not support position offset. From my quick check, Maplibre is currently not supporting it, but others might support position offset.

test/modules/core/controllers/controllers.spec.js
belom88: Done

modules/core/src/controllers/transition-manager.js
```diff
@@ -13,7 +13,8 @@ const DEFAULT_PROPS = {
   transitionInterruption: TRANSITION_EVENTS.BREAK,
   onTransitionStart: noop,
   onTransitionInterrupt: noop,
-  onTransitionEnd: noop
+  onTransitionEnd: noop,
+  position: [0, 0, 0]
```
Pessimistress, line 17: We shouldn't add props that are not related to transition to this object.

belom88: Done

modules/core/src/controllers/map-controller.ts
```diff
@@ -471,6 +477,7 @@ export class MapState extends ViewState {
 export default class MapController extends Controller {
   constructor(props) {
     props.dragMode = props.dragMode || 'pan';
+    props.position = props.position || [0, 0, 0];
```
Pessimistress, line 480: You are requiring `position` to be present in any props object, not just the one used on initial construction. Move this to [setProps](https://deck.gl/docs/api-reference/core/controller#setpropsprops)

Pessimistress (approved):

modules/core/src/controllers/map-controller.ts
belom88: Done

Issue #6432, opened 2021-11-26
[RFC] Supply attributes directly in binary GeoJsonLayer


Clebal:
#### Proposed feature
I want to be able to [supply attributes directly](https://deck.gl/docs/developer-guide/performance#supply-attributes-directly) to GPU as it's possible in others layers. Due to GeoJsonLayer nature, it isn't possible right now. But this is one of the most interesting features to avoid bottlenecks in deck.gl and I think it should be covered in such a relevant layer like GeoJsonLayer is.

#### Target Use case
Being able to supply attributes directly in the binary GeoJsonLayer would improve performance in advanced use cases. In my case, for example, I want to supply getFilterValue attribute to GeoJsonLayer, because its value comes from one of the TypedArray contained in the `numericProps` data.

#### How? - First proposal
My proposal is very simple:

The attribute would be pass seperately for each geometry type:

```
  return new GeoJsonLayer(props, {
    data: {
      pointsAttributes: {
        getFilterValue: { value: Uint16Array, size: 1 }
      },
      polygonsAttributes: {
        getFilterValue: { value: Uint16Array, size: 1 }
      },
      linesAttributes: {
        getFilterValue: { value: Uint16Array, size: 1 }
      },
      ...data
    }
  });
```

And in `createLayerPropsFromBinary` from [geojson-layer-props](https://github.com/visgl/deck.gl/blob/master/modules/layers/src/geojson-layer/geojson-layer-props.js#L41):

At line 43:
```
const {points, pointsAttributes = {}, lines, linesAttributes = {}, polygons, polygonsAttributes = {}} = geojsonBinary;
```

And I would spread these objects for each data:

![image](https://user-images.githubusercontent.com/9151432/143543205-c6ff76c3-2016-46b5-a913-ad942e97c956.png)

(Also polygonsOutline, but it didn't fit in the image)

#### How? - Second proposal

Same as first proposal but instead of summiting separately to binary data, I would create a property called `attributes` inside `data.points`, `data.lines` and `data.polygons`.

```
  return new GeoJsonLayer(props, {
    data: {
      points: {
        ...data.points,
        attributes: {
           getFilterValue: { value: Uint16Array, size: 1 }
        }
      },
      // others geometry types
    }
  });
```

#### To Do List
  - [ ] Add label and assign to milestone
  - [ ] Coding
  - [ ] Doc update
  - [ ] What’s new update
  - [ ] Test

felixpalmer:
[...]

PR #6492, opened 2021-12-09
Supply attributes directly in binary GeoJson layer

Clebal:
For #6432 

#### Change list

- Pass binary attributes to GeoJsonLayer sublayers in [createLayerPropsFromBinary](https://github.com/visgl/deck.gl/pull/6492/files#diff-d1664e13173d9407c92e7ee39e9ff8aae2e325ab4c3a34cc1b091f8c44f7c4d4).
- Add tests to check that binary attributes are correctly being passed.


test/modules/layers/geojson-layer.spec.js
```diff
@@ -126,7 +128,107 @@ test('GeoJsonLayer#tests', t => {
...
+  });
+
+  const binaryDataWithAttributes = {
+    ...binaryData,
+    points: {
+      ...binaryData.points,
+      attributes: {
+        getRadius: {
+          size: 1,
+          value: binaryData.points.featureIds.value.map(_ => Math.round(Math.random() * 10))
```
felixpalmer, line 142: I would avoid having random behavior in tests as errors can then be unreproducible. If you want a range of values you could always do `(_, i) => 10 *i`

test/modules/layers/geojson-layer.spec.js
```diff
@@ -126,7 +128,107 @@ test('GeoJsonLayer#tests', t => {
...
+        }
+      }
+    },
+    lines: {
+      ...binaryData.lines,
+      attributes: {
+        getWidth: {
+          size: 1,
+          value: binaryData.lines.featureIds.value.map(_ => Math.random())
```
felixpalmer, line 151: Same as above

test/modules/layers/geojson-layer.spec.js
```diff
@@ -126,7 +128,107 @@ test('GeoJsonLayer#tests', t => {
...
+      );
+    },
+    props: {
+      // TODO: Set a right geojson example as the provided from 'deck.gl-data' contains 'GeometryCollection' types that are not compatible with geojsonToBinary
+      data: binaryDataWithAttributes
+    }
+  });
+
+  const binaryDataWithGetFilterValue = {
+    ...binaryDataWithAttributes,
```
felixpalmer, line 198: Why extend `binaryDataWithAttributes` here rather than just `binaryData`? Makes for a more isolated test

felixpalmer (changes requested):

test/modules/layers/geojson-layer.spec.js
Clebal: Ok, better 👌

test/modules/layers/geojson-layer.spec.js
Clebal: Fixed.

test/modules/layers/geojson-layer.spec.js
Clebal: Fixed.

Pessimistress (approved):
I do not believe the binary GeoJSON format is documented for this layer. Now that this format is diverging from the one that what MVT loader returns, we should at least add some comments in the binary handling code, linking to  `@loaders.gl/gis` docs and noting the extended fields.

felixpalmer:
@Pessimistress I'm working on a PR (https://github.com/visgl/loaders.gl/issues/1984) in loaders.gl to unify the MVT & geojson-to-binary codepaths. As part of this I'll update the docs there to make sure they are up to date, and then also update here as requested

Issue #6508, opened 2021-12-15
CARTO Dynamic Tiling support


felixpalmer:
#### Target Use case
Large datasets in CARTO can currently only be visualized when the map type is `TILESET`, which requires a tileset to be generated. A new experimental feature on the API allows `TABLE` maps to be tiled dynamically, without the creation of a tileset.

#### Proposed feature
Add support to the `CartoLayer` for map type `TABLE` with format `TILEJSON`, enabling the dynamic tiling

#### To Do List 
  - [x] Add label and assign to milestone 
  - [x] Coding
    - [x] Support `'geojson'` tile format (intermediate format for testing)
    - [x] Support `'binary'` tile format
  - [ ] Doc update
  - [ ] What’s new update
  - [ ] Test

PR #6509, opened 2021-12-15
CartoLayer support for dynamic tiled maps (geojson)

felixpalmer:
For https://github.com/visgl/deck.gl/issues/6508 - `'geojson'` tile format support

#### Change List
- Test app demonstrating how to create a `CartoLayer` for a dynamic tiled map
- Add `CartoDynamicTileLayer` & `CartoDynamicTileLoader`
- Extend `CartoLayer` to create a `CartoDynamicTileLayer` when `type` is `TABLE` and `format` is `TILEJSON`


modules/carto/src/layers/carto-binary-tile-layer.js
```diff
@@ -0,0 +1,56 @@
...
+  mimeTypes: ['application/x-protobuf'],
+  category: 'geometry',
+  worker: false,
+  parse: async (arrayBuffer, options) => parseCartoBinaryTile(arrayBuffer, options),
+  parseSync: parseCartoBinaryTile
+};
+
+// Currently we only support loading via geojson, but in future the data
+// format will be binary, as such keep `binary` in layer name
+export default class CartoBinaryTileLayer extends MVTLayer {
```
alasarr, line 34: Can we rename this to `CartoDynamicTileLayer`?

alasarr: I think it fits better the purpose, the format of the tiles could be a prop of this layer

test/apps/carto-binary-tile/app.js
```diff
@@ -0,0 +1,78 @@
+/* global document */
```
alasarr, line 1: Can we rename this examples to carto-dynamic-tile?

modules/carto/src/layers/carto-binary-tile-layer.js
```diff
@@ -0,0 +1,56 @@
...
+  renderSubLayers(props) {
+    if (props.data === null) {
+      return null;
+    }
+
+    props.autoHighlight = false;
```
alasarr, line 40: It's temporal, but if it's easy would be great if we can warn the user when the autoHighlight is changed to false

alasarr (changes requested):
Looks great to me, just a suggestion on the naming of the layer

modules/carto/src/layers/carto-binary-tile-layer.js
felixpalmer: I did think that, but have we also not considered that in the future we might use the same data format for tile layers? As such, the tile might not always be dynamic. At any rate, the client doesn't really care if the backend is generating the tiles dynamically or not. So perhaps `CartoTileLayer` is better? 

I will add the format of the tiles as a prop.

alasarr: Sounds good.

modules/carto/src/layers/carto-binary-tile-layer.js
felixpalmer: I'm just mimicing what the MVTLayer already does here, so there is no change in behavior

modules/carto/src/layers/carto-binary-tile-layer.js
alasarr: After a second thought, I've changed my mind :). CartoTileLayer could be confused with CartoLayer + tileset. As this layer is not exposed externally, it's only used by us internally, I think CartoDynamicTyleLayer makes sense because it has the format and limitations (as the clipping) of our dynamic tiling backend.

modules/carto/src/layers/carto-dynamic-tile-layer.js
```diff
@@ -0,0 +1,87 @@
...
+
+    return fetch(url, {propName: 'data', layer: this, loadOptions, signal});
+  }
+
+  renderSubLayers(props) {
+    if (props.data === null) {
+      return null;
+    }
+
+    props.autoHighlight = false;
```
alasarr, line 71: Why are disabling autoHighlight? The backend is returning the properties as we're using the geojson

alasarr (changes requested):

modules/carto/src/layers/carto-dynamic-tile-layer.js
felixpalmer: I based the method `renderLayers` method on `MVTLayer` which does the same: https://github.com/visgl/deck.gl/blob/433c82367009e123fb2866fd6622d15ef7e4335e/modules/geo-layers/src/mvt-layer/mvt-layer.js#L144

alasarr: True, it's because mvt manage highlight on their own via https://github.com/visgl/deck.gl/blob/433c82367009e123fb2866fd6622d15ef7e4335e/modules/geo-layers/src/mvt-layer/mvt-layer.js#L162

alasarr (approved):

Issue #6527, opened 2021-12-22
[Feature] CARTO - Maps API client


eamador:
### Target Use Case

`maps-v3-client` and `maps-client` are attaching a `client` parameter to the urls before sending any request, the problem is that the `client` value is hardcoded `deck-gl-carto`.

As this `client` value could be useful for metrics, ideally it could be set dynamically.

### Proposal

In order to allow the developers to specify the `client` value for the Maps API (v1, v2 & v3) we should add this as argument to the exported functions found at `modules/carto/src/api/maps-v3-client.js` & `modules/carto/src/api/maps-client.js`

alasarr:
I think the best approach is to add a new prop to CartoLayer with the default value to the one which is hardcoded right now. As you point out your need to add a parameter to maps api client

PR #6534, opened 2021-12-27
CARTO - add clientId parameter for internal tracing

eamador:
For #6527
#### Change List
- `client` parameter added to `fetchMap`, `fetchLayerData` & `CartoLayer`

### DIFF
--- a/examples/get-started/pure-js/carto/app.js
+++ b/examples/get-started/pure-js/carto/app.js
@@ -3,9 +3,10 @@ import {Deck} from '@deck.gl/core';
 import {fetchMap} from '@deck.gl/carto';
 
 const cartoMapId = '52dcbaa4-f84c-4fc8-93a9-628ed859f937';
+const appClient = 'deck-gl-pure-js-carto';
 
 // Get map info from CARTO and update deck
-fetchMap({cartoMapId}).then(({initialViewState, mapStyle, layers}) => {
+fetchMap({cartoMapId, client: appClient}).then(({initialViewState, mapStyle, layers}) => {
   const deck = new Deck({canvas: 'deck-canvas', controller: true, initialViewState, layers});
 
   // Add Mapbox GL for the basemap. It's not a requirement if you don't need a basemap.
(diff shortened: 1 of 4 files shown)


examples/get-started/pure-js/carto/app.js
```diff
@@ -3,9 +3,10 @@ import {Deck} from '@deck.gl/core';
 import {fetchMap} from '@deck.gl/carto';
 
 const cartoMapId = '52dcbaa4-f84c-4fc8-93a9-628ed859f937';
+const appClient = 'deck-gl-pure-js-carto';
 
 // Get map info from CARTO and update deck
-fetchMap({cartoMapId}).then(({initialViewState, mapStyle, layers}) => {
+fetchMap({cartoMapId, client: appClient}).then(({initialViewState, mapStyle, layers}) => {
```
alasarr, line 9: As it's for internal purpose we don't need to expose it at our more visible demo. I prefer to keep the demo as it's right now

modules/carto/src/layers/carto-layer.js
```diff
@@ -27,6 +27,9 @@ const defaultProps = {
   // (String, optional): format of data
   format: null,
 
+  // (String, optional): client, if not provided, the assigned value will be 'deck-gl-carto'
```
alasarr, line 30:
```suggestion
  // (String, optional): client identifier used for internal tracing, place here a string to identify the client who is doing the request. 
```

modules/carto/src/api/maps-v3-client.js
```diff
@@ -69,8 +70,8 @@ function dealWithError({response, error}) {
 /**
  * Build a URL with all required parameters
  */
-function getParameters({type, source, geoColumn, columns, schema}) {
-  const parameters = [encodeParameter('client', 'deck-gl-carto')];
+function getParameters({type, source, geoColumn, columns, schema, client}) {
```
alasarr, line 73: I prefer to call it clientId rather than client in all the places. client could be confusing because it can be understood as a class

alasarr (changes requested):

alasarr:
@Josmorsot as @eamador is in out this week, could you take care of the requested changes?

alasarr (approved):

alasarr:
fetchLayerData is not working

alasarr:
>fetchLayerData is not working. it doesn't propagate the clientId

Sorry my bad, it works

PR #6553, opened 2022-01-10
Correctly calculate GLViewport with Framebuffer

felixpalmer:

#### Background

When passing a `Framebuffer` as a `target` the GLViewport calculated assumed the width and height to be equal to the GL canvas

#### Change List
- When a custom target is passed to the layers pass, use its width&height rather than that of the canvas
- When the `moduleParameters` specify a pixel ratio, respect this when calculating the GL viewport

### DIFF
--- a/modules/core/src/passes/layers-pass.js
+++ b/modules/core/src/passes/layers-pass.js
@@ -282,19 +282,15 @@ function getGLViewport(gl, {moduleParameters, target, viewport}) {
   const useTarget = target && target.id !== 'default-framebuffer';
   const pixelRatio =
     (moduleParameters && moduleParameters.devicePixelRatio) || cssToDeviceRatio(gl);
-  const targetHeight = target ? target.height / pixelRatio : 100;
-
-  // Fallback to width/height when clientWidth/clientHeight are 0 or undefined.
-  const canvasHeight = gl.canvas ? gl.canvas.clientHeight || gl.canvas.height : 100;
 
   // Default framebuffer is used when writing to canvas
-  const height = useTarget ? targetHeight : canvasHeight;
+  const height = useTarget ? target.height : gl.drawingBufferHeight;
 
   // Convert viewport top-left CSS coordinates to bottom up WebGL coordinates
   const dimensions = viewport;
   return [
     dimensions.x * pixelRatio,
-    (height - dimensions.y - dimensions.height) * pixelRatio,
+    height - (dimensions.y + dimensions.height) * pixelRatio,
     dimensions.width * pixelRatio,
     dimensions.height * pixelRatio
   ];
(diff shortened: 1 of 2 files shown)


Pessimistress:
If I understand this approach correctly, it expects the application to calculate and pass a `devicePixelRatio`, which is not straightforward, yet it still doesn't account for the case where the framebuffer's aspect ratio does not match the canvas.

The canvas's aspect ratio is never wrong because we base the dimensions of `viewport`  on the canvas's size. The same cannot be said for an external framebuffer.

I think it would more robust to take the dimensions of the canvas and "stretch to fit" into the dimensions of the external framebuffer.

felixpalmer:
This does work for a external framebuffer as the dimensions (and thus aspect) are based on the `viewport` that is passed in. I have it working with a square framebuffer in https://github.com/visgl/deck.gl/pull/6554

The `moduleParameters.devicePixelRatio` is optional and is used when rendering into the framebuffer. I need this when rendering to a fixed size off-screen framebuffer as otherwise the y-axis inversion is wrong. If I cannot override this then it becomes difficult to construct a `projectionMatrix` that I can use when reading the framebuffer later.

I think it is reasonable to expect that the following works

```javascript
const target = new Framebuffer(gl, {width, height});
const viewport = new WebMercatorViewport({width, height});
deck.deckRenderer.renderLayers({
  target,
  viewports: [viewport],
  moduleParameters: {devicePixelRatio: 1},
  ...
});
```

`_getModuleParameters` already accepts the passed in `moduleParameters` as an override and these then get passed onto the shaders. It just happens to be the case that `getGLViewport` ignores this and hardcodes the pixel ratio to the `cssToDeviceRatio` default.

I can see that the stretch-to-fit approach would work for the case of rendering the current canvas into a framebuffer, but if I just want to render a fixed view into a fixed framebuffer why should this be dependent on the canvas pixel ratio & size?

Does this make sense? Perhaps I'm misunderstanding what you mean by "stretch-to-fit"?

modules/core/src/passes/layers-pass.js
```diff
@@ -270,13 +278,20 @@ export function layerIndexResolver(startIndex = 0, layerIndices = {}) {
...
+  const pixelRatio =
+    (moduleParameters && moduleParameters.devicePixelRatio) || cssToDeviceRatio(gl);
+  const targetHeight = target ? target.height / pixelRatio : 100;
+
   // Fallback to width/height when clientWidth/clientHeight are 0 or undefined.
-  const height = gl.canvas ? gl.canvas.clientHeight || gl.canvas.height : 100;
+  const canvasHeight = gl.canvas ? gl.canvas.clientHeight || gl.canvas.height : 100;
+
+  // Default framebuffer is used when writing to canvas
+  const height = useTarget ? targetHeight : canvasHeight;
```
Pessimistress, line 291: This seems more complicated than it needs to be. You should be able to use 

```js
const height = useTarget ? target.height : gl.drawingBufferHeight;
```

Then

```js
height - (dimensions.y + dimensions.height) * pixelRatio
```

Pessimistress (commented):
Yes it makes sense. Thank you for the clarification.

modules/core/src/passes/layers-pass.js
felixpalmer: Fixed

PR #6554, opened 2022-01-10
Mask Extension [WIP]

felixpalmer:

#### Change List
- Implementation of MaskExtension
- Example app for website

Tests & documentation in future PR, publishing this minimal WIP to gather feedback on API and to allow downstream implementation in CARTO for React


felixpalmer:
@Pessimistress I followed the code in the other extension as well as the shadow module to implement this. 

One thing I would appreciate you commenting on the the maths around the offset projection. I think what I have is close, but not quite right, and I see precision errors when zoomed in close. Relevant code: https://github.com/visgl/deck.gl/pull/6554/files#diff-e7aa5378c14ee05c714d147abc0b53e860e4f422f9fc85cae7d44bb4f3e9971aR56-R93

modules/core/src/lib/deck.js
```diff
@@ -705,9 +705,14 @@ export default class Deck {
 
     this.props.onBeforeRender({gl});
 
+    const layers = this.layerManager.getLayers();
+    const drawLayers = layers.filter(l => l.props.operation === 'draw');
+    const maskLayers = layers.filter(l => l.props.operation === 'mask');
+    maskLayers.forEach(layer => layer.setNeedsRedraw());
```
Pessimistress, line 711: `setNeedsRedraw` should never be called outside of the layer lifecycle / layer manager.

Filtering logic is in passes/layers-pass.js. The Deck instance should not contain any specific knowledge of how to draw layers.

Pessimistress (commented):
An extension is the wrong place to hold a FBO because it is run per layer. Mask(s) are global and shared and should be implemented in an Effect, which can be executed before and after any render and inject resources into all draw calls. Since we are adding a core `operation` prop it is not inappropriate to always load a `MaskEffect` by default. See how shadow FBOs are handled in https://github.com/visgl/deck.gl/blob/master/modules/core/src/effects/lighting/lighting-effect.js#L62

examples/website/mask/README.md
```diff
@@ -0,0 +1,26 @@
+This is a minimal standalone version of the MaskExtension example
```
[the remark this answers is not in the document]
felixpalmer: It is primarily for testing. I've moved it to test/apps

felixpalmer:
@Pessimistress

### Updated Change list
- `Layer` class has an added prop, `operation` which determines how it is drawn (defaults to `operation: OPERATION.DRAW`)
- `DrawLayersPass` modified to only render layers with `OPERATION.DRAW`
- Added `MaskEffect` to render layers marked with `operation: OPERATION.MASK` to a FBO, which are then used as masks for layers with the `maskId` prop set
- Added `MaskPass` (invoked by `MaskEffect`) which renders `OPERATION.MASK` layers to an FBO
- Added a `mask` shader module which performs the masking in a manner similar to the shadow module
- `MaskEffect` passed to `deck-picker` (when enabled) to allow picking respect mask operation
- Test app showing different ways masking can be used
- Unit tests for `MaskEffect`

### Limitations
- Masking layer must be performed with a SolidPolygonLayer
- Only one masking layer supported

There are some commented out snippets of test code which I will remove - I've left these in case they are helpful

modules/core/src/effects/mask/mask-effect.js
```diff
@@ -0,0 +1,146 @@
...
+    if (!this.dummyMaskMap) {
+      this.dummyMaskMap = new Texture2D(gl, {
+        width: 1,
+        height: 1
+      });
+    }
+
+    // Using the 'positions' attribute will work for the PolygonLayer,
+    // but not for all layers
+    const {positions} = maskLayer.state.attributeManager.attributes;
```
Pessimistress, line 55: use `layer.getAttributeManager()`

modules/core/src/effects/mask/utils.js
```diff
@@ -0,0 +1,75 @@
...
+}
+
+/*
+ * Compute viewport to render the mask into, such that it covers an
+ * area currently visible (extended by a buffer) or the area of the masking
+ * data, whichever is smaller
+ */
+export function getMaskViewport(positions, layerViewport, {width, height}) {
+  const bounds = layerViewport.getBounds();
+  let viewport = new WebMercatorViewport({width, height}).fitBounds([
```
Pessimistress, line 29: Is there a reason to hard code the viewport type here?

modules/core/src/effects/mask/utils.js
```diff
@@ -0,0 +1,75 @@
...
+function getBounds({startIndices, size, value}) {
+  const bounds = [
+    [Infinity, Infinity],
+    [-Infinity, -Infinity]
+  ];
+
+  const start = startIndices[0];
+  const end = startIndices[startIndices.length - 1];
+  for (let i = start; i < end; i++) {
+    const coord = value.subarray(size * i);
```
Pessimistress, line 53: This is extremely expensive (creating a small array for every vertex). Why not just calculate the index in the original type array?

modules/core/src/effects/mask/utils.js
```diff
@@ -0,0 +1,75 @@
...
+export function getMaskViewport(positions, layerViewport, {width, height}) {
+  const bounds = layerViewport.getBounds();
+  let viewport = new WebMercatorViewport({width, height}).fitBounds([
+    [bounds[0], bounds[1]],
+    [bounds[2], bounds[3]]
+  ]);
+  const {longitude, latitude, zoom} = viewport;
+  viewport = new WebMercatorViewport({width, height, longitude, latitude, zoom: zoom - 1});
+
+  const dataBounds = getBounds(positions);
```
Pessimistress, line 36: Is this done every render frame without the data changing?

modules/core/src/effects/mask/utils.js
```diff
@@ -0,0 +1,75 @@
...
+  const dataViewport = new WebMercatorViewport({width, height}).fitBounds(dataBounds, {
+    padding: 2
+  });
+
+  return dataViewport.zoom > viewport.zoom ? dataViewport : viewport;
```
Pessimistress, line 41: Some comments here on what this does

modules/core/src/effects/mask/mask-effect.js
```diff
@@ -0,0 +1,146 @@
...
+    const layerViewport = viewports[0];
+    const maskViewport = getMaskViewport(positions, layerViewport, maskMap);
+    this.maskProjectionMatrix = getMaskProjectionMatrix(maskViewport);
+
+    maskPass.render({
```
Pessimistress, line 60: Skip rerendering the mask framebuffer if using the same viewport and the mask layer has not changed.

Example: if you move your mouse across the map, picking is performed on every animation frame. There is no reason to rerender the mask.

modules/core/src/passes/mask-pass.js
```diff
@@ -0,0 +1,66 @@
...
+        [gl.COLOR_ATTACHMENT0]: this.maskMap
+      }
+    });
+  }
+
+  render(params) {
+    const target = this.fbo;
+
+    withParameters(
+      this.gl,
```
Pessimistress, line 35: Disable blending/depth test for better perf?

modules/core/src/effects/mask/utils.js
felixpalmer: Does the function comment not cover it? It is basically picking the smallest necessary area to render into to avoid pixelation artifacts. Which bit is unclear

modules/core/src/effects/mask/utils.js
felixpalmer: The other viewports don't have the `fitBounds` methods and the rest of the code doesn't support other projections. Would it be acceptable to restrict this PR to the `WebMercatorViewport` as it the mainline case and add the other viewports in a followup PR?

felixpalmer: I've fixed this now, such that `WebMercatorViewport` isn't hardcoded. If `fitBounds` is available on the `Viewport`, then it is used for the calculation, otherwise a fallback method is used

modules/core/src/passes/draw-layers-pass.ts
```diff
@@ -1,3 +1,8 @@
+import {OPERATION} from '../lib/constants';
 import LayersPass from './layers-pass';
 
-export default class DrawLayersPass extends LayersPass {}
+export default class DrawLayersPass extends LayersPass {
+  shouldDrawLayer(layer) {
```
Pessimistress, line 5: This is also needed in PickLayersPass

modules/core/src/passes/mask-pass.ts
```diff
@@ -0,0 +1,69 @@
...
+    );
+  }
+
+  shouldDrawLayer(layer) {
+    return layer.props.operation === OPERATION.MASK;
+  }
+
+  getModuleParameters() {
+    return {
+      drawToMaskMap: true
```
Pessimistress, line 61: Is this used anywhere?

Pessimistress (approved):

modules/core/src/effects/mask/utils.js
```diff
@@ -0,0 +1,101 @@
...
+    }
+  }
+  const viewportBounds = viewport.getBounds();
+  if (!bounds) {
+    return viewportBounds;
+  }
+
+  // Expand viewport bounds by 2X. Heurestically chosen to avoid masking
+  // errors when mask is partially out of view
+  const paddedBounds = _doubleBounds(viewportBounds);
```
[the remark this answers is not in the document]
felixpalmer: As we do an intersection with the bounds, I believe it is safe

modules/core/src/effects/mask/mask-effect.js
```diff
@@ -0,0 +1,135 @@
...
+
+    if (maskChanged || !this.lastViewport || !viewport.equals(this.lastViewport)) {
+      // Update mask FBO
+      const {maskPass, maskMap} = this;
+      this.lastViewport = viewport;
+      this.maskLayers = maskLayers;
+
+      const bounds = getMaskBounds({layers: maskLayers, viewport});
+
+      if (!equals(bounds, this.lastBounds)) {
```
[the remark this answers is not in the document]
felixpalmer: Added a comment. At worst the mask will be wrong only until the viewport is changed

modules/core/src/passes/mask-pass.ts
felixpalmer: No, I added it for consistency with `drawToShadowMap` to inform shader modules. I'll remove it as it isn't used

Issue #6528, opened 2021-12-22
[Bug] GlobeView MVT layers not showing 


Baseform:
### Description

MVT Layers don't show up; no error reported on console. 
alpha.3 and previous versions do not have the same problem.

### Expected Behavior

For MVT Layer data to show on Globe

### Steps to Reproduce

Two codepens:
a) with alpha.7 (not working):
https://codepen.io/diogovit/pen/rNGGWQz

b) with alpha.3 (working)
https://codepen.io/diogovit/pen/dyVVOBw

### Environment

- Framework version: deck.gl@8.7.0-alpha.7
- Browser: safari, chrome
- OS: mac

Pessimistress:
@felixpalmer Looks like MVTParser returns `NaN` coordinates when using `coordinates: 'wgs84'`.

felixpalmer:
@Pessimistress thanks for the report, will look into it this week

felixpalmer:
Underlying issue fixed in https://github.com/visgl/loaders.gl/pull/2052

PR #6562, opened 2022-01-17
MVT GlobeView render test

felixpalmer:
For https://github.com/visgl/deck.gl/issues/6528

#### Change List
- Adds render test for binary & non-binary `MVTLayer` when rendered with `GlobeView`


alasarr (approved):
Thanks for the initiative and the implementation!

PR #6573, opened 2022-01-20
CARTO fetchMap:  Fix no layer shown if one of the map layers fails

padawannn:
#### Background
When you are going to load a map with the fechMap Carto module method if any of the layers fail, neither layer is loaded. It's easy to test if you try to load a no-supported layer type.
#### Change List
- Added try-catch block in parseMap Carto module method to allow loading the map even if some layers fail


Pessimistress:
Can you rename the title of this PR? Keep in mind the title will go in to the release note here: https://github.com/visgl/deck.gl/releases and think what will make sense for a user.

felixpalmer (approved):
Looks great. Thanks for adding the test also :)

padawannn:
> Can you rename this PR? Keep in mind the title will go in to the release note here: https://github.com/visgl/deck.gl/releases and think what will make sense for a user.

Sure!

Issue #6575, opened 2022-01-23
[Feat] Allow setting FontAtlasManager LRU Cache Limit


urdeveloper:
### Target Use Case

FontAtlasManager uses LRU Cache to keep 3 most used font atlases in memory. This is fine for typical use cases but it becomes  performance bottleneck when much more than 3 fonts (or combination of font-family, font-weight ...) is needed.

If you need for example 15 different fonts in your application, FontAtlasManager keeps recreating them because it can't keep all of them in memory. 

### Proposal

Introduce a function that allows setting the LRU cache limit in such use cases.

PR #6576, opened 2022-01-23
Allow setting FontAtlasManager LRU Cache limit

urdeveloper:
For #6575 

#### Background
FontAtlasManager uses LRU Cache to keep 3 most used font atlases in memory. This is fine for typical use cases but it becomes performance bottleneck when much more than 3 fonts (or combination of font-family, font-weight ...) is needed.

#### Change List
- Introduce TextLayer.fontAtlasCacheLimit static setter to recreate the LRU cache with new limit
- Add documentation to TexLayer
- Add tests


docs/api-reference/layers/text-layer.md
```diff
@@ -463,6 +463,22 @@ The TextLayer has full support for Unicode characters. To reference a Unicode ch
...
 
+### Font Atlas Cache Limit
+
+To conserve memory, DeckGL caches the most 3 used `fontAtlas` by default. Creating a `fontAtlas` is a CPU intesive operation specially if `fontSettings.sdf` is set to `true`. 
+If you are using much more than 3 fonts, you might experience perforamnce hits because DeckGL constantly tries to evict the least most used `fontAtlas` from cache and recreate them when needed.
+
+To mitigate the potential performance degradation, you can override the `fontAtlas` default cache limit by calling `setFontAtlasCacheLimit` function:
+
+```js
+import {setFontAtlasCacheLimit} from '@deck.gl/react';
```
Pessimistress, line 474: `@deck.gl/layers`?

modules/layers/src/text-layer/font-atlas-manager.js
```diff
@@ -97,6 +97,22 @@ function setTextStyle(ctx, fontFamily, fontSize, fontWeight) {
   ctx.textAlign = 'left';
 }
 
+/**
+ * Sets the Font Atlas LRU Cache Limit
+ * @param {number} limit LRU Cache limit
+ */
+export function setFontAtlasCacheLimit(limit) {
```
Pessimistress, line 104: Nit: we try not to use verbose error messages because it bloats the bundle size. Maybe consider something like

```js

log.assert(Number.isFinite(limit) && limit >= CACHE_LIMIT, 'Invalid cache limit');
```

modules/layers/src/index.ts
```diff
@@ -37,3 +37,6 @@ export {default as SolidPolygonLayer} from './solid-polygon-layer/solid-polygon-
 // Experimental layer exports
 export {default as _MultiIconLayer} from './text-layer/multi-icon-layer/multi-icon-layer';
 export {default as _TextBackgroundLayer} from './text-layer/text-background-layer/text-background-layer';
+
+// Font Atlas Manager
+export {setFontAtlasCacheLimit} from './text-layer/font-atlas-manager';
```
Pessimistress, line 42: Regarding this API - we don't have other setters like this in the library. My personal preference is to expose it as a static member of `TextLayer`, i.e.

```js
TextLayer.fontAtlasCacheLimit = 5;
```

docs/api-reference/layers/text-layer.md
urdeveloper: Fixed in new commit

modules/layers/src/text-layer/font-atlas-manager.js
urdeveloper: Replaced it with your suggestion.

modules/layers/src/index.ts
urdeveloper: As you suggested, I don't expose the function anymore and instead added a new static setter to TextLayer.

Pessimistress (approved):
Thank you!

PR #6580, opened 2022-01-24
Support TILE_FORMAT.BINARY in CartoDynamicTileLayer

felixpalmer:
For https://github.com/visgl/deck.gl/issues/6508

#### Change List
- Add pbf dependency
- Schema for binary data format
- Parse pbf and display using CartoLayer


modules/carto/src/layers/carto-layer.js
```diff
@@ -137,8 +140,12 @@ export default class CartoLayer extends CompositeLayer {
       apiVersion === API_VERSIONS.V3 &&
       format === FORMATS.TILEJSON &&
       new URLSearchParams(data.tiles[0]).get('format');
+    const nonMvtTiles = tilesFormat && tilesFormat !== TILE_FORMATS.MVT;
+
+    const dynamicTiles =
```
alasarr, line 145: Not for this PR, but I feel we can make it simpler if maps API always returns the format at tiles url. 

Don't you think?

alasarr: I feel this part is getting complicated

alasarr (approved):

modules/carto/src/layers/carto-layer.js
alasarr: if (
      apiVersion === API_VERSIONS.V1 ||
      apiVersion === API_VERSIONS.V2 || 
    )
   {
      // backward compatibility CARTO 2
      layer = MVTLayer;
    } else if (format === FORMATS.TILEJSON) {
       layer = tilesFormat === TILE_FORMATS.MVT ? MVTLayer : CartoDynamicTileLayer
    }
     else {
      layer = GeoJsonLayer;
    }

padawannn: If the backend always going to return the format this suggested code is easier to understand.

felixpalmer: Yes I would much prefer if the backend returned the format it was sending so we could simplify here

PR #6591, opened 2022-01-27
Respect mask coordinate system & origin

felixpalmer:

#### Background

When projecting the bounds of the mask, the layer was assuming that the mask had the same coordinate system and origin as the layer being rendered. This is often true, but not always - in particular for the `MVTLayer`, where the coordinates are Cartesian and relative to the tile.

#### Change List
- Pass coordinate system and origin as mask moduleParameters
- Project using mask coordinate system explicitly in mask extension
- Add MVTLayer to Mask test app
- Tests

### DIFF
--- a/modules/core/src/effects/mask/mask-effect.js
+++ b/modules/core/src/effects/mask/mask-effect.js
@@ -163,8 +163,8 @@ export default class MaskEffect extends Effect {
       }
       channelInfo.layers.push(layer);
       channelInfo.layerBounds.push(layer.getBounds());
-      channelInfo.coordinateOrigin = layer.props.coordinateOrigin;
-      channelInfo.coordinateSystem = layer.props.coordinateSystem;
+      channelInfo.coordinateOrigin = layer.root.props.coordinateOrigin;
+      channelInfo.coordinateSystem = layer.root.props.coordinateSystem;
     }
 
     for (let i = 0; i < 4; i++) {
--- a/modules/core/src/effects/mask/mask-effect.js
+++ b/modules/core/src/effects/mask/mask-effect.js
@@ -127,7 +127,13 @@ export default class MaskEffect extends Effect {
         });
       }
     }
-    this.masks[channelInfo.id] = {index: channelInfo.index, bounds: channelInfo.maskBounds};
+
+    this.masks[channelInfo.id] = {
+      index: channelInfo.index,
+      bounds: channelInfo.maskBounds,
+      coordinateOrigin: channelInfo.coordinateOrigin,
+      coordinateSystem: channelInfo.coordinateSystem
+    };
   }
 
   /**
@@ -157,6 +163,8 @@ export default class MaskEffect extends Effect {
       }
       channelInfo.layers.push(layer);
       channelInfo.layerBounds.push(layer.getBounds());
+      channelInfo.coordinateOrigin = layer.props.coordinateOrigin;
+      channelInfo.coordinateSystem = layer.props.coordinateSystem;
     }
 
     for (let i = 0; i < 4; i++) {
--- a/modules/core/src/effects/mask/mask-effect.js
+++ b/modules/core/src/effects/mask/mask-effect.js
@@ -157,14 +157,14 @@ export default class MaskEffect extends Effect {
           id,
           index: this.channels.findIndex(c => c?.id === id),
           layers: [],
-          layerBounds: []
+          layerBounds: [],
+          coordinateOrigin: layer.root.props.coordinateOrigin,
+          coordinateSystem: layer.root.props.coordinateSystem
         };
         channelMap[id] = channelInfo;
       }
       channelInfo.layers.push(layer);
       channelInfo.layerBounds.push(layer.getBounds());
-      channelInfo.coordinateOrigin = layer.root.props.coordinateOrigin;
-      channelInfo.coordinateSystem = layer.root.props.coordinateSystem;
     }
 
     for (let i = 0; i < 4; i++) {
--- a/modules/core/src/lib/layer.js
+++ b/modules/core/src/lib/layer.js
@@ -255,13 +255,16 @@ export default class Layer extends Component {
     return viewport.unproject(xy);
   }
 
-  projectPosition(xyz) {
-    return projectPosition(xyz, {
+  projectPosition(
+    xyz,
+    params = {
       viewport: this.context.viewport,
       modelMatrix: this.props.modelMatrix,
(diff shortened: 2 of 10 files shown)


felixpalmer:
@Pessimistress I've based this off your work on multiple masks in order to avoid merge conflicts once you close https://github.com/visgl/deck.gl/pull/6589

modules/core/src/effects/mask/mask-effect.js
```diff
@@ -66,7 +66,12 @@ export default class MaskEffect extends Effect {
 
         channelInfo.bounds = getMaskBounds({layers: channelInfo.layers, viewport});
 
-        if (maskChanged || !equals(channelInfo.bounds, oldChannelInfo.bounds)) {
+        if (
+          maskChanged ||
+          !equals(channelInfo.bounds, oldChannelInfo.bounds) ||
+          !equals(channelInfo.coordinateOrigin, oldChannelInfo.coordinateOrigin) ||
```
Pessimistress, line 72: I don't think this is necessary. There is virtually no chance of a layer changing its coordinate system without changing its data.

modules/core/src/effects/mask/mask-effect.js
```diff
@@ -149,6 +160,8 @@ export default class MaskEffect extends Effect {
       }
       channelInfo.layers.push(layer);
       channelInfo.layerBounds.push(layer.getBounds());
+      channelInfo.coordinateOrigin = layer.props.coordinateOrigin;
```
Pessimistress, line 163: Set this only once when the layer is created (the root layer's props)?

Technically a parent layer can have a different coordinate system from its children (e.g. MVT). To support that scenario we'll need to finish implementing the `getBounds` method for individual composite layers, instead of merging all layer bounds in this effect.

modules/extensions/src/mask/mask.js
```diff
@@ -1,5 +1,6 @@
 import {LayerExtension, log} from '@deck.gl/core';
 import mask from './shader-module';
+import {projectPosition} from '@deck.gl/core/shaderlib/project/project-functions';
```
Pessimistress, line 3: You can't import functions like this. When published the files will be in dist/es5 or dist/esm. This only works in test because we set module aliases in the test harness.

This function is available under the exported `project` shader module.

modules/extensions/src/mask/mask.js
```diff
@@ -20,17 +21,19 @@ export default class MaskExtension extends LayerExtension {
     };
   }
 
-  draw({uniforms, moduleParameters}) {
+  draw({uniforms, context, moduleParameters}) {
     uniforms.mask_maskByInstance = this.state.maskByInstance;
-    const {maskId} = this.props;
+    const {maskId} = this.root.props;
```
Pessimistress, line 26: maskId should be automatically passed down via `LayerExtension.getSubLayerProps`.

modules/extensions/src/mask/mask.js
felixpalmer: I couldn't find where it was exported. I added it to the core module exports

modules/core/src/effects/mask/mask-effect.js
felixpalmer: I'm not sure if I correctly understood you here. Did you mean this: https://github.com/visgl/deck.gl/pull/6591/commits/1686fcd33ef913af07280aa77992f5c725038347

Pessimistress: I meant move this into the block `channelInfo = {...}`. You are re-assigning these two fields again and again for each descendant of any composite layer.

modules/core/src/shaderlib/index.js
```diff
@@ -47,4 +48,4 @@ export function createProgramManager(gl) {
   return programManager;
 }
 
-export {picking, project, project32, gouraudLighting, phongLighting, shadow};
+export {picking, project, project32, projectPosition, gouraudLighting, phongLighting, shadow};
```
Pessimistress, line 51: Ok I thought this function was exposed via `project.projectPosition` but apparently not. I think we removed it when we moved the functionality to `layer.projectPosition`.

I prefer to avoid exposing another API. Maybe add a second parameter to `layer.projectPosition` to allow overriding the project options?

modules/extensions/src/mask/mask.js
```diff
@@ -20,17 +20,19 @@ export default class MaskExtension extends LayerExtension {
...
-
-      const bl = this.projectPosition([bounds[0], bounds[1], 0]);
-      const tr = this.projectPosition([bounds[2], bounds[3], 0]);
+      const {index, bounds, coordinateOrigin, coordinateSystem} = maskChannels[maskId];
       uniforms.mask_enabled = true;
       uniforms.mask_channel = index;
+
+      const opts = {viewport, coordinateOrigin, coordinateSystem};
+      const bl = projectPosition([bounds[0], bounds[1], 0], opts);
+      const tr = projectPosition([bounds[2], bounds[3], 0], opts);
```
Pessimistress, line 35: You need `fromCoordinate*`: https://github.com/visgl/deck.gl/blob/master/modules/core/src/shaderlib/project/project-functions.js#L105

modules/core/src/lib/layer.js
```diff
@@ -255,13 +255,16 @@ export default class Layer extends Component {
     return viewport.unproject(xy);
   }
 
-  projectPosition(xyz) {
-    return projectPosition(xyz, {
+  projectPosition(
+    xyz,
+    params = {
```
Pessimistress, line 260: Use this argument to override not replace?

modules/extensions/src/mask/mask.js
felixpalmer: A bit confused here. The bounds are specified in the coordinate system of the mask. What should I be passing into the `fromCoordinate*` parameters?

Pessimistress: `fromCoordinateSystem` (source) is from the mask layer's props.
`coordinateSystem` (target) is from the current layer's props.

Otherwise you are projecting to the source layer's common space, which isn't necessarily the same as the one you are using it in.

felixpalmer: I've made the change, however I had to pass in a `null` `modelMatrix` to make it work. Does that mean that we should also pass through the `modelMatrix` of the mask layer

Pessimistress: Let's leave the model matrix as null for now. I think `layer.getBounds()` should apply the matrix of defined.

Pessimistress (approved):

PR #6596, opened 2022-01-28
CARTO fetchMap: Fix h3 hexagon layers are not shown

padawannn:
#### Background
H3 hexagon layers are not shown when you load a map using the fechMap Carto module method. It's because we are assuming that the hexagon column is always 'h3'. Instead, we should read this column from the map config. To get this, I've replaced the `LAYER_MAP` constant with a function that receives the layer type and the config as params. This also allows us to read more configuration parameters if would be necessary.

#### Change List
- Replaced the `LAYER_MAP` constant with a function that receives the layer type and the config as params

### DIFF
--- a/modules/carto/src/api/layer-map.js
+++ b/modules/carto/src/api/layer-map.js
@@ -64,6 +64,13 @@ function mergePropMaps(a, b) {
 
 export function getLayer(type, config) {
   return {
+    hexagonId: {
+      Layer: H3HexagonLayer,
+      propMap: mergePropMaps(sharedPropMap, {
+        visConfig: {coverage: 'coverage', elevationScale: 'elevationScale'}
+      }),
+      defaultProps: {...defaultProps, getHexagon: d => d[config.columns.hex_id]}
+    },
     point: {
       Layer: GeoJsonLayer,
       propMap: mergePropMaps(sharedPropMap, {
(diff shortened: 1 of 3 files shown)


modules/carto/src/api/parseMap.js
```diff
@@ -24,8 +24,10 @@ export function parseMap(json) {
     mapStyle,
     layers: extractTextLayers(layers.reverse()).map(({id, type, config, visualChannels}) => {
       try {
-        log.assert(type in LAYER_MAP, `Unsupported layer type: ${type}`);
-        const {Layer, propMap, defaultProps} = LAYER_MAP[type];
+        const layer = getLayer(type, config);
+
+        log.assert(getLayer(type, config), `Unsupported layer type: ${type}`);
```
felixpalmer, line 29: Move this assertion into `getLayer` to avoid double calling the function?

modules/carto/src/api/layer-map.js
```diff
@@ -62,37 +62,39 @@ function mergePropMaps(a, b) {
...
+      Layer: GeoJsonLayer,
+      propMap: sharedPropMap,
+      defaultProps: {...defaultProps, lineWidthScale: 2}
+    },
+    hexagonId: {
+      Layer: H3HexagonLayer,
+      propMap: mergePropMaps(sharedPropMap, {
+        visConfig: {coverage: 'coverage', elevationScale: 'elevationScale'}
+      }),
+      defaultProps: {...defaultProps, getHexagon: d => d[config.columns.hex_id]}
```
felixpalmer, line 84: I'd pull the `getHexagon` function definition to the top of the `getLayer` function to make it easier to see which parts of the code are using the `config` object, and which are just the static mappings

felixpalmer (changes requested):

modules/carto/src/api/parseMap.js
padawannn: Ups, thank you!

modules/carto/src/api/layer-map.js
padawannn: Done

felixpalmer (approved):

PR #6602, opened 2022-01-28
Fix `update` Behavior in `Tileset2D`

ilan-gold:
#### Background
Previously, if an `updateTrigger` prompted the `TileLayer` to call `tileset.reloadAll()` (as is intended), the subsequent `tileset.update()` call had no guarantee of actually doing the reloading.  This PR forces the `update` function to do a reload.

Furthermore, previously, because of the `&&` in the changed lines, with a `null` `modelMatrix` argument, `this._modelMatrix` would be `null` so `const isModelMatrixNew = !modelMatrixAsMatrix4.equals(this._modelMatrix);` was always true.  This resolves that by ensuring everything is always a matrix, even if just the identity.

#### Change List
- add a condition to `update` in `Tileset2D` to check for reloading tiles in the absence of a viewport or model matrix change
- Ensure `this._modelMatrix` and `this._inverseModelMatrix` are both matrices so that they pass the `equals` test correctly in `update` of tileset.


ilan-gold:
The reason this isn't two PR's is that the `modelMatrix` fix alone was causing the tests to stall.  Apologies.

Pessimistress (approved):

PR #6609, opened 2022-01-31
CARTO module: Support tiles in different formats

padawannn:
#### Background
Currently, the Carto module always uses the MVT layer to load any tileset. Now, we read the tile format from the Maps API and we use the MVT layer or the Carto Dynamic Tile Layer that support tiles in other formats like geojson or binary
#### Change List
- Read the tile format from the Maps API to decide to load an MVT layer or a Carto Dynamic Tile Layer


modules/carto/src/api/maps-api-common.js
```diff
@@ -26,6 +26,12 @@ export const FORMATS = {
   JSON: 'json'
 };
 
+export const TILES_FORMATS = {
```
felixpalmer, line 29: I think `TILE_FORMATS` is a better name

felixpalmer (changes requested):
Nice, thanks. Just a few nits

modules/carto/src/api/maps-api-common.js
padawannn: Done

modules/carto/src/layers/carto-layer.js
```diff
@@ -128,12 +128,14 @@ export default class CartoLayer extends CompositeLayer {
...
     if (!data) return null;
 
-    const {type, updateTriggers} = this.props;
+    const {updateTriggers} = this.props;
 
     let layer;
 
-    if (apiVersion === API_VERSIONS.V3 && type === MAP_TYPES.TABLE && format === FORMATS.TILEJSON) {
-      layer = CartoDynamicTileLayer;
+    if (apiVersion === API_VERSIONS.V3 && format === FORMATS.TILEJSON) {
```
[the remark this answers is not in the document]
padawannn: Done

felixpalmer (approved):

PR #6623, opened 2022-02-07
CARTO module: Fix extensions don't work with CartoDynamicTileLayer

padawannn:
#### Background
If you try to use an extension with the CartoDynamicTileLayer it doesn't work because the CartoDynamicTileLayer is overwriting the `props.extensions`
#### Change List
- Include  extensions sent in properties in the CartoDynamicTileLayer


alasarr (approved):
Good catch!

Issue #6628, opened 2022-02-08
[Feature] CARTO - Maps API fetchMap including accessToken


eamador:
### Target Use Case

Update maps-v3-client to work with the latests API changes.

### Proposal

The `fetchMap` method from `maps-v3-client` should include the `accessToken` as `headers.Authorization` since it is needed for the latests API changes.

eamador:
Done [here](https://github.com/visgl/deck.gl/pull/6629)

PR #6629, opened 2022-02-08
CARTO - Maps API - accessToken in fetchMap

eamador:
For #6628

#### Change List
- `fetchMap` including `accessToken` in the request as `headers.Authorization`


modules/carto/src/api/maps-v3-client.js
```diff
@@ -307,7 +308,7 @@ export async function fetchMap({cartoMapId, clientId, credentials, autoRefresh,
   }
 
   const url = `${localCreds.mapsUrl}/public/${cartoMapId}`;
-  const map = await request({url});
+  const map = await request({url, accessToken});
```
VictorVelarde, line 311: @eamador token is received, but it is not being included in the fetch headers, right?

alasarr (approved):

PR #6635, opened 2022-02-09
CARTO: Simplify format tiles logic

alasarr:
Simplify how format tiles logic works, we trust now on the parameter returned by the backend instead of apply a complex logic here.

### DIFF
--- a/modules/carto/src/layers/carto-dynamic-tile-layer.js
+++ b/modules/carto/src/layers/carto-dynamic-tile-layer.js
@@ -6,7 +6,7 @@ import {MVTLayer, _getURLFromTemplate} from '@deck.gl/geo-layers';
 import {GeoJsonLayer} from '@deck.gl/layers';
 import {geojsonToBinary} from '@loaders.gl/gis';
 import {Tile} from './schema/carto-dynamic-tile';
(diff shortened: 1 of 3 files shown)


modules/carto/src/layers/carto-dynamic-tile-layer.js
```diff
@@ -86,7 +86,10 @@ export default class CartoDynamicTileLayer extends MVTLayer {
         Object.values(TILE_FORMATS).includes(formatTiles),
         `Invalid value for formatTiles: ${formatTiles}. Use value from TILE_FORMATS`
       );
-      url += `&${encodeParameter('formatTiles', formatTiles)}`;
+      // eslint-disable-next-line no-undef
+      const newUrl = new URL(url);
```
padawannn, line 90: Better use URLSearchParams instead of URL

padawannn (changes requested):

modules/carto/src/layers/carto-dynamic-tile-layer.js
alasarr: I think in this case URL is better because I'm actually modifying the URL

padawannn (approved):

PR #6636, opened 2022-02-09
Update fetchMap docs and example

borja-munoz:
#### Background

We need to improve the documentation for the fetchMap function in the CARTO module and update the CARTO map ID used in the get started examples.

#### Change List
- Improve documentation for fetchMap function
- Update cartoMapID in examples


docs/api-reference/carto/fetch-map.md
```diff
@@ -41,30 +41,67 @@ fetchMap({cartoMapId}).then(({initialViewState, mapStyle, layers}) => {
 });
 ```
 
-## Usage
+## Parameters
```
alasarr, line 44: I think usage is more clear. What about keeping `Usage` and adding a new `##Parameters` section?

alasarr (approved):

PR #6637, opened 2022-02-09
CARTO fetchMap: Support tiles in different formats 

padawannn:
#### Background
With this PR the `fetchMap` method support tile layers in geojson and binary format
#### Change List
- The fetchMap will load a CartoDynamicTileLayer if the tile format is not MVT

### DIFF
--- a/modules/carto/src/api/layer-map.js
+++ b/modules/carto/src/api/layer-map.js
@@ -104,7 +104,7 @@ function getTileLayer(dataset) {
     }
   } = dataset;
   /* global URLSearchParams */
-  const formatTiles = new URLSearchParams(tileUrl).get('format');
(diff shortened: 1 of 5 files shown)


modules/carto/src/api/layer-map.js
```diff
@@ -103,6 +97,28 @@ export function getLayer(type, config) {
...
 }
 
+function getTileLayer(dataset) {
+  const {
+    data: {
+      tiles: [tileUrl]
+    }
+  } = dataset;
+  /* global URLSearchParams */
+  const formatTiles = new URLSearchParams(tileUrl).get('format');
```
alasarr, line 107: It will be changed to formatTiles

PR #6638, opened 2022-02-09
Default format tiles to MVT to guarantee compatilibity with data cached in CDN

alasarr:
Some data in the CDN is cached and it has not the `formatTiles` parameter, we're setting the default to MVT to guarantee compatibility with public maps.

It cannot be done by purging CDN because some customer could have it at the local browser cache.

### DIFF
--- a/modules/carto/src/api/layer-map.js
+++ b/modules/carto/src/api/layer-map.js
@@ -107,7 +107,7 @@ function getTileLayer(dataset) {
   const formatTiles = new URLSearchParams(tileUrl).get('formatTiles');
 
   return {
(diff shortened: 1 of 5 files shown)


padawannn (approved):

alasarr:
I've changed two extra things:

- Fix the tests, they were working in dynamic tiling because the formatTiles was null and it always used CartoDynamicTileLayer
- URLSearchParams was used badly in many places. URLSearchParams params expect to receive the URL params not the full URL. I've replaced it by `URL(myurl).searchParams.get...`

Issue #6457, opened 2021-12-01
MVTLayer does not render labels using pointType: text in binary mode


AdriSolid:
#### Description
`CartoLayer` does not render labels using `pointType` = `text`.

Setting `binary` to `false` fixes the problem apparently.

For non-Mac devices everything works as expected, even with `binary` prop set to `true`.

#### Expected Behavior
Labels are rendered as expected. As using `TextLayer`.

#### Repro Steps
Codepen: https://codepen.io/AdriSolid/pen/zYEYyea

#### Environment
 - Framework Version: deck.gl ^8.6.0
 - Browser Version: Chrome 96.0.4664.55, Safari 15.1, Firefox 94.0.2
 - OS: macOS Monterey. Version 12.0.1

#### Logs
Console warning: `[.WebGL-0x14d828e00]GL ERROR:GL_INVALID_OPERATION :glDrawArraysInstancedANGLE: attempt to access out of range vertices in attribute 4`

PR #6650, opened 2022-02-15
TextLayer: expand picking color attribute from per-point to per-icon

felixpalmer:
For https://github.com/visgl/deck.gl/issues/6457

#### Background

`MVTLayer` passes an explicit `instancePickingColors` attribute for points when using binary mode. For `pointType: 'circle'` this works fine, but for `pointType: 'text'` it breaks because the number of characters is not generally equal to the number of points rendered.

#### Change List
- Expand `instancePickingColors` attribute such that it matches the other attributes
- Pass expanded attribute to `TextBackgroundLayer` to fix background rendering on binary `MVTLayer`
- Test

### DIFF
--- a/modules/layers/src/text-layer/text-layer.js
+++ b/modules/layers/src/text-layer/text-layer.js
@@ -221,6 +221,22 @@ export default class TextLayer extends CompositeLayer {
     });
   }
 
+  // Expands picking color attribute from per-point to per-icon
+  _expandPickingColors() {
+    const {startIndices, numInstances} = this.state;
+    const {instancePickingColors} = this.props.data.attributes;
+    const {value, size} = instancePickingColors;
+
+    const newValue = new value.constructor(numInstances * size);
+    for (let i = 0; i < startIndices.length - 1; i++) {
+      const color = value.subarray(size * i, size * (i + 1));
+      for (let j = startIndices[i]; j < startIndices[i + 1]; j++) {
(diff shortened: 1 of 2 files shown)


modules/layers/src/text-layer/text-layer.js
```diff
@@ -320,6 +336,10 @@ export default class TextLayer extends CompositeLayer {
     const CharactersLayerClass = this.getSubLayerClass('characters', MultiIconLayer);
     const BackgroundLayerClass = this.getSubLayerClass('background', TextBackgroundLayer);
 
+    if (data.attributes && data.attributes.instancePickingColors) {
+      this._expandPickingColors();
```
Pessimistress, line 340: TextLayer is behaving as expected. If you manipulate the attribute it needs to be done outside of the layer.

felixpalmer: For the manipulation I need the `startIndices` that are calculated by `TextLayer`. Are you saying that I should duplicate this calculation inside `GeojsonLayer`?

Pessimistress: What you are doing here will break the apps that are currently providing `instancePickingColors` correctly.

felixpalmer: I've moved the conversion to GeoJsonLayer

Pessimistress (approved):

PR #6655, opened 2022-02-17
Migrate carto config & basemap modules to Typescript 

zbigg:
For #5589
#### Change List
- Migrate Carto config & basemap modules to Typescript


modules/carto/src/config.ts
```diff
@@ -4,17 +4,33 @@ export const defaultClassicCredentials = {
...
+}
 
-let credentials = {};
+type Credentials = ClassicCredentials | CloudNativeCredentials;
+let credentials: Credentials = {} as Credentials;
 
 setDefaultCredentials({});
 
-export function setDefaultCredentials({apiVersion, ...rest}) {
+export function setDefaultCredentials({apiVersion, ...rest}: Partial<Credentials>) {
```
felixpalmer, line 33: Does this have to be Partial? I would instead mark the properties that are optional as optional (that is, everything except `apiVersion`).

modules/carto/src/api/maps-api-common.ts
```diff
@@ -5,37 +5,37 @@ export const API_VERSIONS = {
...
   GEOJSON: 'geojson',
   BINARY: 'binary'
-};
+} as const;
 
 /**
  * Simple encode parameter
  */
-export function encodeParameter(name, value) {
+export function encodeParameter(name: string, value: string) {
```
felixpalmer, line 39: Include return value here

felixpalmer (changes requested):
Great job! Just a few minor changes

modules/carto/src/config.ts
zbigg: In this signature everything is optional, including `apiVersion`. So yes, this is partial.

With those typings, i tried to convey fact that `setDefaultCredentials` can accept almost everything (partial), but `getDefaultCredentials` will return actual "Credentials" for which some of them are actually mandatory and/or filled using defaults.

So, it depends.

With this typings, instantiating `Credentials` is close-to typesafe.

modules/carto/src/api/maps-api-common.ts
zbigg: appied

modules/carto/src/config.ts
felixpalmer: Ideally it shouldn't accept everything. E.g. this mix of parameters is invalid:

```javascript
{
  apiVersion: 'v1',
  accessToken: 'XXXX'
}
```

Would `Partial<ClassicCredentials> | Partial<CloudNativeCredentials>` work?

felixpalmer (approved):

PR #6659, opened 2022-02-18
Migrate carto api modules to Typescript

zbigg:

Change List
 * Migrate Carto api modules to Typescript


modules/carto/src/api/maps-v3-client.ts
```diff
@@ -224,10 +297,11 @@ async function _fetchDataUrl({
         break;
       }
     }
+    log.assert(url!, 'Unsupported data formats received from backend.');
   }
 
   const {accessToken} = localCreds;
-  return {url, accessToken, mapFormat, metadata};
+  return {url: url!, accessToken, mapFormat: mapFormat!, metadata};
```
zbigg, line 304: General note w.r.t migration to Typescript.

We use log.assert extensively and we expect that it throws. We know it, but with current typings, typescript don't, so we have to "teach it" using ! or other ugly casts.

Best way, would be to tell that log.assert throws if url is non-thruthy - [it's possible](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions).

This would require either our own local assert function with signature like this:

```
function assert(condition, message) asserts condition {}
```
Or update deck.gl to properly export log object ... only after patching probe.gl.

zbigg: Now the only problem with using `assert` is that asserts are typically used for "development mode/time" checks and frequently, there is expectation that in production builds they are disabled and/or completly removed from code. They are in several runtimes/libraries.

Question is whether deck.gl/probe.gl wants to rely on asserts as normal runtime checks or maybe there should be new concept of `runtimeAssert` that has same semantics but won't be removed by some future tooling.

zbigg: For record: https://github.com/uber-web/probe.gl/pull/188

modules/carto/src/api/maps-api-common.ts
```diff
@@ -1,11 +1,13 @@
 export const DEFAULT_USER_COMPONENT_IN_URL = '{user}';
 export const DEFAULT_REGION_COMPONENT_IN_URL = '{region}';
 
+export type ValuesOf<T extends {}> = T[keyof T];
 export const API_VERSIONS = {
   V1: 'v1',
   V2: 'v2',
   V3: 'v3'
 } as const;
+export type AoiVersion = ValuesOf<typeof API_VERSIONS>;
```
felixpalmer, line 10: Seems to be a typo?

modules/carto/src/api/maps-v3-client.ts
```diff
@@ -268,32 +334,50 @@ async function _fetchMapDataset(dataset, accessToken, credentials, clientId) {
...
     ...(defaultCredentials.apiVersion === API_VERSIONS.V3 && defaultCredentials),
     ...credentials
-  };
+  } as CloudNativeCredentials;
   const {accessToken} = localCreds;
 
-  log.assert(cartoMapId, `Must define CARTO map id: fetchMap({cartoMapId: 'XXXX-XXXX-XXXX'})`);
+  // TODO: prettier conflicts with eslint here
+  // eslint-disable-next-line quotes
+  log.assert(cartoMapId, "Must define CARTO map id: fetchMap({cartoMapId: 'XXXX-XXXX-XXXX'})");
```
felixpalmer, line 380: As it is just an error message I would remove the single quotes around XXXX

zbigg: Hmm, i've changed to `"XXXX-XXXX-XXXX"` and both eslint and prettier are happy, i think it's good trade.

modules/carto/src/api/maps-v3-client.ts
zbigg: Resolved temporarily for carto module by introducing carto-module local `assert` helper that has correct typings.

felixpalmer (approved):

PR #6663, opened 2022-02-21
fix(test): include Typescript sources in test coverage

zbigg:
For #5589

#### Background

Improve coverage analysis to include also Typescript sources.

#### Change List
- Instruct `nyc` to include `*.ts` files in coverage report


felixpalmer:
@ibgreen @Pessimistress we noticed the coverage dropping when migrating to Typescript. Are we OK to make this change, or is there some reason `.ts` files are not being included in the coverage report?

alasarr:
Right now, master CI is broken because of coverage decrease and it fixes it. If there is no special reason I suggest merging it

Pessimistress (approved):

PR #6667, opened 2022-02-21
Migrate carto style functionality to Typescript

zbigg:
For #5589

#### Change List
- Migrate carto style functionality to Typescript


modules/carto/src/style/utils.ts
```diff
@@ -1,23 +1,26 @@
...
     return attr(d);
   }
 
   // Is string
-  if (typeof attr === ALLOWED_ATTR_TYPES[1]) {
+  if (typeof attr === 'string') {
     return d.properties[attr];
   }
 
   return {};
```
felixpalmer, line 20: Can this ever be hit?

felixpalmer (approved):

modules/carto/src/style/utils.ts
zbigg: nope, fixed

Issue #6677, opened 2022-02-22
[Feat] QuadkeyLayer


felixpalmer:
### Target Use Case

Deck.gl already has [S2Layer](https://deck.gl/docs/api-reference/geo-layers/s2-layer) and [H3HexagonLayer](https://deck.gl/docs/api-reference/geo-layers/h3-hexagon-layer) for easily rendering data based on a geospatial key.

Quadkeys are common way of encoding geospatial data and can be rendered in a similar fashion. A `QuadkeyLayer` would render datasets encoded using quadkeys in a similar fashion.

### Proposal

Usage is similar to the `S2Layer`. The `QuadkeyLayer` inherits all of the props of the `PolygonLayer`, with the addition of the `getQuadkey` accessor.

```javascript
import DeckGL from '@deck.gl/react';
import {QuadkeyLayer} from '@deck.gl/geo-layers';

function App({data, viewState}) {
  /**
   * Data format:
   * [
   *   {
   *     // Quadkey in SF Bay Area
   *     quadkey: "023010203321",
   *     value: 1.2
   *   },
   *   {
   *     quadkey: "023010203321",
   *     value: 0.5446256069712141
   *   },
   *   ...
   * ]
   */
  const layer = new QuadkeyLayer({
    id: 'quadkey-layer',
    data,
    pickable: true,
    wireframe: false,
    filled: true,
    extruded: true,
    elevationScale: 1000,
    getQuadkey: d => d.quadkey,
    getFillColor: d => [d.value * 255, (1 - d.value) * 255, (1 - d.value) * 128],
    getElevation: d => d.value
  });

  return <DeckGL viewState={viewState}
    layers={[layer]}
    getTooltip={({object}) => object && `${object.quadkey} value: ${object.value}`} />;
}
```
<img width="981" alt="Screenshot 2022-02-22 at 15 25 43" src="https://user-images.githubusercontent.com/453755/155152079-92ba8200-304a-4c2d-b503-664cc3bc5770.png">

PR #6678, opened 2022-02-22
QuadkeyLayer implementation & tests

felixpalmer:
For https://github.com/visgl/deck.gl/issues/6677

#### Change List
- Introduction of private abstract `GeoCellLayer` to share common code between `H3ClusterLayer` & `S2Layer`
- Addition of `QuadkeyLayer` (also based on `GeoCellLayer`)
- Unit tests
- Documentation including live example
- `QuadkeyLayer` added to layer-browser


modules/geo-layers/src/quadkey-layer/quadkey-layer.js
```diff
@@ -0,0 +1,87 @@
...
+import {getQuadkeyPolygon} from './quadkey-utils';
+
+const defaultProps = {
+  ...PolygonLayer.defaultProps,
+  getQuadkey: {type: 'accessor', value: d => d.quadkey}
+};
+
+export default class QuadkeyLayer extends CompositeLayer {
+  renderLayers() {
+    // Layer prop
```
felixpalmer, line 12: A lot of this code is duplicated with `S2Layer`. It may be worth considering introducing a common subclass

modules/geo-layers/src/quadkey-layer/quadkey-utils.js
```diff
@@ -0,0 +1,28 @@
...
+    [x / scale, TILE_SIZE - y / scale],
+    [(x + 1) / scale, TILE_SIZE - (y + 1) / scale]
+  ];
+}
+
+export function getQuadkeyPolygon(quadkey) {
+  const [topLeft, bottomRight] = quadkeyToWorldBounds(quadkey);
+  const [w, n] = worldToLngLat(topLeft);
+  const [e, s] = worldToLngLat(bottomRight);
+  return new Float64Array([e, n, e, s, w, s, w, n, e, n]);
```
Pessimistress, line 27: It is not necessary to convert the coordinates to a typed array. It only adds performance overheads because of extra GC.

modules/geo-layers/src/quadkey-layer/quadkey-layer.js
Pessimistress: Agreed

modules/geo-layers/src/quadkey-layer/quadkey-utils.js
felixpalmer: Fixed. Is there a reason why a `Float64Array` is created in `s2-utils.js`?

https://github.com/visgl/deck.gl/blob/master/modules/geo-layers/src/s2-layer/s2-utils.js#L42

modules/geo-layers/src/quadkey-layer/quadkey-layer.js
felixpalmer: As the shared class wouldn't be ever exposed I instead used a utility method as it felt more consistent with what with do with `GeoJsonLayer`

modules/geo-layers/src/quadkey-layer/quadkey-layer.js
```diff
@@ -0,0 +1,45 @@
...
+
+import {createCellLayer} from '../common/cell-layer-utils';
+import {getQuadkeyPolygon} from './quadkey-utils';
+
+const defaultProps = {
+  ...PolygonLayer.defaultProps,
+  getQuadkey: {type: 'accessor', value: d => d.quadkey}
+};
+
+export default class QuadkeyLayer extends CompositeLayer {
```
Pessimistress, line 32: It will be easier to read if you create a generic `GeoCellLayer` that has a method `indexToBounds `. Then `S2Layer`, `QuadkeyLayer` and `H3ClusterLayer` can just extend this base layer and implement their own `indexToBounds`.

felixpalmer: Done

felixpalmer:
I've added the following example to the layer-browser and the documentation

Screenshot 2022-02-23 at 12 24 00

The test data is synthetic and needs to be uploaded to deck.gl-data: https://github.com/visgl/deck.gl/pull/6678/files#diff-202bd1a464188d5c7bfadc858bca0e579de2527f08217648e28aba7c31bef569

docs/api-reference/geo-layers/quadkey-layer.md
```diff
@@ -0,0 +1,107 @@
...
+```
+
+
+## Properties
+
+Inherits from all [Base Layer](/docs/api-reference/core/layer.md), [CompositeLayer](/docs/api-reference/core/composite-layer.md), and [PolygonLayer](/docs/api-reference/layers/polygon-layer.md) properties, plus the following:
+
+### Data Accessors
+
+##### `getQuadkey` ([Function](/docs/developer-guide/using-layers.md#accessors), optional) ![transition-enabled](https://img.shields.io/badge/transition-enabled-green.svg?style=flat-square")
```
Pessimistress, line 91: Remove the transition flag

modules/geo-layers/src/geo-cell-layer/GeoCellLayer.js
```diff
@@ -0,0 +1,103 @@
+// Copyright (c) 2015 - 2017 Uber Technologies, Inc.
```
Pessimistress, line 1: Remove the Uber header

modules/geo-layers/src/geo-cell-layer/GeoCellLayer.js
```diff
@@ -0,0 +1,103 @@
...
+        getLineDashArray
+      },
+      this.getSubLayerProps({
+        id: 'cell',
+        updateTriggers: {
+          getElevation: updateTriggers.getElevation,
+          getFillColor: updateTriggers.getFillColor,
+          getLineColor: updateTriggers.getLineColor,
+          getLineWidth: updateTriggers.getLineWidth,
+          getLineDashArray: updateTriggers.getLineDashArray
```
Pessimistress, line 94: This is no longer a PolygonLayer accessor

modules/geo-layers/src/h3-layers/h3-cluster-layer.js
```diff
@@ -1,14 +1,14 @@
 import {h3SetToMultiPolygon} from 'h3-js';
 
-import {CompositeLayer, createIterable} from '@deck.gl/core';
-import {PolygonLayer} from '@deck.gl/layers';
+import {createIterable} from '@deck.gl/core';
+import GeoCellLayer from '../geo-cell-layer/GeoCellLayer';
 
 const defaultProps = {
-  ...PolygonLayer.defaultProps,
+  ...GeoCellLayer.defaultProps,
```
Pessimistress, line 7: Sub class automatically gets all base class props. You do not have to merge them.

modules/geo-layers/src/quadkey-layer/quadkey-layer.js
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) 2015 - 2017 Uber Technologies, Inc.
```
Pessimistress, line 1: Remove

test/modules/geo-layers/quadkey-layer.spec.js
```diff
@@ -0,0 +1,98 @@
+// Copyright (c) 2015 - 2017 Uber Technologies, Inc.
```
Pessimistress, line 1: Remove

Pessimistress (approved):
The PR is already pretty large and I think it's ok to merge as is. 

Follow up, the GeoCellLayer needs to handle update trigger changes of the index accessor, and ideally avoid passing the expensive index to geo bounds conversion to `getPolygon` - it will be called twice for each object in the tessellator. This is an existing perf issue in some of the current layers and not introduced by this PR.

PR #6687, opened 2022-02-23
CARTO: Fix formatTiles propagation at CartoLayer

alasarr:
CartoLayer was not setting the formatTiles prop for CartoTileLayer. So, if the backend returns the formatTiles to geojson it was ignoring it and try to load it as binary tiles


felixpalmer (approved):
Nice catch!

PR #6730, opened 2022-03-04
Merge GeoJsonLayer default props into MVTLayer

felixpalmer:
#### Background

`MVTLayer` inherits `TileLayer`‘s default props. It does not understand `GeoJsonLayer` props, so when a `GeoJsonLayer` accessor shallowly changes (e.g. inline function in React) it will force purge all cached sub layers, causing unnecessary re-renders.

#### Change List
- Test to expose issue
- Merge `GeoJsonLayer` props into `MVTLayer`


test/modules/geo-layers/mvt-layer.spec.js
```diff
@@ -678,3 +678,53 @@ test('findIndexBinary', t => {
 
   t.end();
 });
+
+test.only('MVTLayer#GeoJsonLayer.defaultProps', t => {
```
Pessimistress, line 682: Remove

Pessimistress (approved):

PR #6733, opened 2022-03-04
Cull TileLayer sub layers during picking

Pessimistress:
Proof of concept for discussion #6729

Picking on hover draws 20 tiles before this PR, and only 1 tile after.

#### Change List
- Add `cullRect` to layer filter context
- `TileLayer.filterSubLayer` culls sub layers based on the scissor rectangle

#### TODO
- Test
- Culling for generic layers?

### DIFF
--- a/docs/api-reference/core/deck.md
+++ b/docs/api-reference/core/deck.md
@@ -214,6 +214,7 @@ Receives arguments:
 - `layer` (Layer) - the layer to be drawn
 - `viewport` (Viewport) - the current viewport
 - `isPicking` (Boolean) - whether this is a picking pass
+- `cullRect` (Object) - if defined, indicates that only the content rendered to the given rectangle is needed.
 - `renderPass` (String) - the name of the current render pass. Some standard passes are:
   + `'screen'` - drawing to screen
   + `'picking:hover'` - drawing to offscreen picking buffer due to pointer move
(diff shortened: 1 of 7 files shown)


felixpalmer (commented):
I worry that using memoize is going to consume a lot of memory. If a user is panning and zooming around a map, how often are the rectangles going to be the same again?

ibgreen (approved):
> I worry that using memoize is going to consume a lot of memory. If a user is panning and zooming around a map, how often are the rectangles going to be the same again?

I would agree. I feel that the main benefit comes from avoiding the calculation if the viewport hasn't changed. Saving old viewport calculations does not seem likely to improve perf.

The other nit is that I feel that such culling should be the job of the Tileset class (`Tileset2D`?), but I haven't been paying enough attention to that for a while to have a good sense for whether the getCullBounds and filterSubLayer could call corresponding methods on that class.

The final nit is that while this is nice optimization, we really should make a proper color picking module that doesn't suffer from these limitations. I might be able to take a look at creating a luma.gl picking module that can handle "unlimited" layers and objects, I am a bit stuck with luma.gl 9.0 being ahead of deck.gl right now, and I don't want to put more work into luma.gl 8.x, but once deck.gl upgrades to luma v9 I think that would be a good step

Pessimistress:
The memoize is needed during the same render cycle (`filterSubLayer` is called for every sub layer). I agree this feels like something `Tileset2D` should handle instead of the layer itself.

It will be definitely nice to be able to encode more layers with the picking module, and we should continue to pursue that. This optimization reduces draw calls and improves performance as well as functionality, so i feel that it should be done with or without the picking color limitation.

felixpalmer (approved):

PR #6743, opened 2022-03-07
MVTLayer: force binary to false for GlobeView

felixpalmer:
For https://github.com/visgl/deck.gl/discussions/6717#discussioncomment-2269006

#### Change List
- Store `binary` in `state` in `MVTLayer`
- Override `prop` value when `viewport` is `GlobeViewport`


modules/geo-layers/src/mvt-layer/mvt-layer.js
```diff
@@ -25,7 +26,9 @@ const defaultProps = {
 export default class MVTLayer extends TileLayer {
   initializeState() {
     super.initializeState();
+    const binary = this.context.viewport instanceof GlobeViewport ? false : this.props.binary;
```
Pessimistress, line 29: Use `viewport.resolution`. Layers should not depend on a particular viewport constructor.

Pessimistress (approved):

Issue #6731, opened 2022-03-04
Googlemaps GeoJsonLayer Not Resizing Correctly


MikeKeane:
### Description

I am using the GeoJsonLayer with Google Maps and when zooming, the layer does not resize smoothly like it used to, rather it snaps to size.

### Flavors

- [ ] React
- [ ] Python/Jupyter notebook
- [ ] MapboxLayer
- [X] GoogleMapsOverlay
- [ ] CartoLayer
- [ ] DeckLayer/DeckRenderer for ArcGIS

### Expected Behavior

I expect the layer to zoom with a smooth transition.  

### Steps to Reproduce

Create a geoJsonLayer with GoogleMapsOverlay and zoom the map. 

### Environment

- Framework version: @latest
- Browser: Chrome 98.0.4758.102
- OS: Windows 10

felixpalmer:
This is a regression introduced by https://github.com/visgl/deck.gl/pull/6653. I will work on a fix. @MikeKeane it is only an issue for raster maps, so if you can switch your Google map to a vector map the smooth zooming will work now.

felixpalmer:
@MikeKeane fix deployed in https://github.com/visgl/deck.gl/releases/tag/v8.7.1

PR #6744, opened 2022-03-07
Handle undefined heading for raster Google maps

felixpalmer:
For https://github.com/visgl/deck.gl/issues/6731

#### Background

[getHeading](https://developers.google.com/maps/documentation/javascript/reference/map?hl=en#Map.getHeading) sometimes returns undefined instead of 0.

#### Change List
- Handle case of `heading = undefined`
- Correct zoom calculation for mulitple world copies (use map height rather than diagonal)

### DIFF
--- a/modules/google-maps/src/utils.js
+++ b/modules/google-maps/src/utils.js
@@ -139,10 +139,13 @@ export function getViewPropsFromOverlay(map, overlay) {
   let bearing = (180 * delta.verticalAngle()) / Math.PI;
   if (bearing < 0) bearing += 360;
 
+  // Maps sometimes returns undefined instead of 0
+  const heading = map.getHeading() || 0;
+
   let zoom = map.getZoom() - 1;
 
   // Fractional zoom calculation only correct when bearing is not animating
-  if (bearing === map.getHeading()) {
+  if (bearing === heading) {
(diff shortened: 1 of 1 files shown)


modules/google-maps/src/utils.js
```diff
@@ -139,21 +139,29 @@ export function getViewPropsFromOverlay(map, overlay) {
...
   if (bearing < 0) bearing += 360;
 
-  let zoom = map.getZoom() - 1;
+  // Maps sometimes returns undefined instead of 0
+  const heading = map.getHeading() || 0;
 
-  // Fractional zoom calculation only correct when bearing is not animating
-  if (bearing === map.getHeading()) {
+  let zoom = map.getZoom() - 1;
+  let scale = 1;
```
Pessimistress, line 146: Nit: this initial value is never used

modules/google-maps/src/utils.js
```diff
@@ -139,21 +139,29 @@ export function getViewPropsFromOverlay(map, overlay) {
...
+
+  if (bearing === 0) {
+    // At full world view (always unrotated) simply compare height, as diagonal
```
Pessimistress, line 149: Are multiple world copies only rendered when heading is 0?

Pessimistress (approved):

Issue #6760, opened 2022-03-17
[Feat] Support aggregation layers in CARTO fetchMap


felixpalmer:
### Target Use Case

Expand the number of layers supported in https://deck.gl/docs/api-reference/carto/fetch-map to include `CPUGridLayer`, `HeatmapLayer` and `HexagonLayer`

### Proposal

Usage is the same as with existing layers

PR #6761, opened 2022-03-18
Support aggregation layers in CARTO fetchMap

felixpalmer:
For https://github.com/visgl/deck.gl/issues/6760

#### Change List
- Implement binding for `CPUGridLayer`, `HeatmapLayer` and `HexagonLayer` in `fetchMap`
- Tests


modules/carto/src/api/maps-v3-client.ts
```diff
@@ -338,16 +338,17 @@ async function _fetchMapDataset(
...
   const {url, mapFormat} = await _fetchDataUrl({
+    clientId,
     credentials: {...credentials, accessToken},
     connection,
     columns,
+    format,
     geoColumn,
     source,
-    type,
-    clientId
```
alasarr, line 350: Why is clientId removed? If we do that we'll lose the tracing at metrics

alasarr (changes requested):
Not sure why are we removing clientId

modules/carto/src/api/maps-v3-client.ts
felixpalmer: It's not removed, I just moved it to the top of the params list to have it alphabetical

alasarr (approved):

modules/carto/src/api/maps-v3-client.ts
alasarr: My bad :)

PR #6767, opened 2022-03-22
Force GeoJSON for geojson/point fetchMap layers

felixpalmer:
Fixes regression in https://github.com/visgl/deck.gl/pull/6761

### DIFF
--- a/modules/carto/src/api/maps-v3-client.ts
+++ b/modules/carto/src/api/maps-v3-client.ts
@@ -429,6 +429,16 @@ export async function fetchMap({
     };
   }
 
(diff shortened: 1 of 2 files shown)


padawannn (changes requested):
I'm ok with the solution but for some reason, we've lost the color for line layers and polygons borders. For example `bfd965e8-99c1-4b52-ae4b-bcec6cfbde76`

Captura de pantalla 2022-03-23 a las 9 04 15

It only happens in new maps

padawannn (approved):
LGFM. Thank you man

PR #6777, opened 2022-03-25
Fix Python CI

kylebarron:
The latest version of `setuptools` broke our CI, see: https://github.com/visgl/deck.gl/runs/5687413943?check_suite_focus=true, but it turns out our `pyproject.toml` file is invalid anyways. The `requires` array should be under the `[build-system]` header.

#### Change List

- Update `pyproject.toml` in line with [PEP 621](https://peps.python.org/pep-0621/)

### DIFF
--- a/bindings/pydeck/pyproject.toml
+++ b/bindings/pydeck/pyproject.toml
@@ -1,10 +1,12 @@
 [project]
 name = "pydeck"
-version = "0.3.0"
+version = "0.7.1"
+requires-python = ">=3.7"
 
+[build-system]
 requires = [
-    "setuptools",
-    "wheel",
+  "setuptools",
+  "wheel",
 ]
 
 [tool.black]


ajduberstein (approved):

abravalheri:
Hi @kylebarron thank you very much for having a look on this.

From the setuptools perspective, I would recommend that if you guys are not using the `[project]` table on the `pyproject.toml` then the best is to remove it for the time being (alternatively you can embrace `pyproject.toml` for all the static configuration).

The [standards](https://peps.python.org/pep-0621/) this file is based on [^1], imply a series of validations and additional semantics that `pydeck` might not be prepared for.

For example, when `setuptools` adds full support to PEP 621, it will have to ignore any configuration that is covered by the standard but is not specified via `pyproject.toml`, unless explicitly listed in the `dynamic` field[^2].

Examples of existing configurations that would ideally have to be ignored right now are:
- `install_requires`  in your `setup.py`
- `description_file` in your `setup.cfg`
- ...


[^1]: There is a "living" version of this standard that is kept up to date with Python's PEP process https://packaging.python.org/en/latest/specifications/declaring-project-metadata/
[^2]: There is an open issue in the setuptools repository pypa/setuptools#3195, and the fact that `setuptools` is not yet honouring completely [PEP 621](https://peps.python.org/pep-0621/#dynamic) is technically a bug.

donmccurdy:
Related issue in https://github.com/visgl/deck.gl/issues/8469. It seems that local builds of pydeck are not currently working, I've had to (locally) delete `pyproject.toml` to build, but there are other remaining issues in Jupyter Notebooks v7+, described at https://github.com/visgl/deck.gl/pull/7030#issuecomment-2027367256. 

@abravalheri or @kylebarron do I understand correctly that we have two overlapping configs for pydeck, and do you have any reference on how we should resolve that?

abravalheri:
Hi @donmccurdy , I am having a look on this from my phone, so not sure what is going on... But have you had a look on my previous comment? https://github.com/visgl/deck.gl/pull/6777#issuecomment-1079652376

If the configuration is done via `setup.py`, then you have 2 options. Configure `[project] dynamic` correctly on `pyproject.toml`, or remove `[project]` completely from the file.

donmccurdy:
@abravalheri Thanks! I don't doubt you're right about `pyproject.toml`, but I think our local builds may be broken for more reasons than just that. I'm having trouble figuring out what else is needed to run build pydeck from source again, but have posted some debugging progress in https://github.com/visgl/deck.gl/issues/8469#issuecomment-2030480564.

PR #6789, opened 2022-03-31
Remove all scales from CARTO fetchMap 

padawannn:
#### Background
Cloud-native is going to start styling using pixels instead of meters so the scales are no longer necessary
#### Change List
- Remove lineWidthScale and pointRadiusScale from CARTO fetchMap


modules/carto/src/api/layer-map.ts
```diff
@@ -110,7 +107,7 @@ export function getLayer(
     },
     geojson: {
       Layer: GeoJsonLayer,
-      defaultProps: {lineWidthScale: 2}
+      defaultProps: {}
```
felixpalmer, line 110: As `defaultProps` is empty you can simply remove it

felixpalmer (changes requested):

felixpalmer:
A number of tests are also failing

felixpalmer:
Comparing using the `test/apps/carto-map` app with map id `ba2ef0ba-e7bb-4a9a-a2a0-e8ade556b3d2` shows a difference. Is this because the Builder update is yet to be deployed?
Screenshot 2022-03-31 at 9 50 10

padawannn:
> Comparing using the `test/apps/carto-map` app with map id `ba2ef0ba-e7bb-4a9a-a2a0-e8ade556b3d2` shows a difference. Is this because the Builder update is yet to be deployed? Screenshot 2022-03-31 at 9 50 10

Yes, exactly. The Builder PRs have not been approved yet. I'll deploy in a dedicated to check the differences.

modules/carto/src/api/layer-map.ts
padawannn: I was not sure if I could remove it. Done

felixpalmer (approved):
Looks good to me. However I would say hold off on merging until we've done full QA and agreed on a deployment plan

felixpalmer:
@padawannn this PR: https://github.com/visgl/deck.gl/pull/6800 fixes the issues found in QA. I would recommend we review and merge that first and then update this one and do a final round of QA

Issue #6296, opened 2021-10-12
BitmapLayer world copies don't blend with vector GoogleMapsOverlay


zakjan:
#### Description
Steps:
- vector GoogleMapsOverlay
- BitmapLayer with opacity <1
- zoomed out viewState so that multiple world copies are visible

The first world copy of BitmapLayer blends well. The other world copies don't blend.

![Screen Shot 2021-10-12 at 21 31 12](https://user-images.githubusercontent.com/173585/137017595-3dff46ca-89ef-4601-b8f6-762ee3fea758.png)

#### Expected Behavior
All BitmapLayer world copies blend well.

#### Repro Steps
https://codepen.io/zakjan/pen/yLXPWNe

#### Environment
 - Framework Version: 8.6.0
 - Browser Version: Chrome 94

Pessimistress:
`MapView({ repeat: true })` is the default view of `GoogleMapsOverlay`. It's not clear to me how Google Maps handles the repeat, maybe they duplicate the custom overlay as part of a "copy"? Is the custom overlay drawn once for the main copy, or once for each copy, or after all copies are drawn? Should we just draw one copy instead of repeating? What about 3D objects then?

alasarr:
Update: this issue has been reported to Google Maps team, they're going to investigate further, we'll keep you posted once we've more info

felixpalmer:
I've done some experimenting and it isn't just the `BitmapLayer`, I also see this with the `ScenegraphLayer`, i.e. with 3D objects. It doesn't appear to be the case that there is some copy happening, but the opacity is different

https://user-images.githubusercontent.com/453755/157007797-1b827967-effe-404c-bee0-55c1f0c88b5f.mov

.

felixpalmer:
@amuramoto you can find a demo of the above bug here: https://jsfiddle.net/felixp/ac20sfqd/

zakjan:
Considering that this is outside of deck control, is it possible to un-deprecate the overlaid mode from <=8.5.0, and allow developers to choose? Related to https://github.com/visgl/deck.gl/issues/6773

alasarr:
Google has just confirmed it's been fixed, I've tested the codepen and it works (depthMask true or false) for me. @zakjan feel free to re-open if you see something wrong

zakjan:
Thanks @alasarr for tracking this, but it seems that the fix from Google is not released yet?

[...]

PR #6804, opened 2022-04-05
GoogleMapsOverlay support non-interleaved mode for vector maps

felixpalmer:
For https://github.com/visgl/deck.gl/issues/6296

#### Change List
- Add `interleaved` prop to `GoogleMapsOverlay` (default `false` to provide back-compatibility)
- Implement rendering path for vector maps on separate overlaid canvas
- Test updates


modules/google-maps/src/google-maps-overlay.js
```diff
@@ -81,6 +84,7 @@ export default class GoogleMapsOverlay {
 
   /* Private API */
   _createOverlay(map) {
+    const {interleaved} = this.props;
```
Pessimistress, line 87:
```suggestion
    const {interleaved = true} = this.props;
```

And you can drop the mutation in the constructor.

modules/google-maps/src/utils.js
```diff
@@ -204,9 +204,8 @@ export function getViewPropsFromCoordinateTransformer(map, transformer) {
   const focalDistance = 0.5 * projectionMatrix[5];
 
   return {
-    // Using external gl context - do not set css size
-    width: false,
-    height: false,
```
Pessimistress, line 209: Why remove this? It prevents Deck from scaling the gl context size to device pixel ratio. Removing ii=t will break context sharing.

zakjan:
Thanks, I can confirm this works.

`pointerEvents: 'none'` is unfortunate (because it prevents mouse events and tooltips from working), but it's in the Mapbox overlay as well, so I guess it's acceptable

felixpalmer:
> Thanks, I can confirm this works.
> 
> `pointerEvents: 'none'` is unfortunate (because it prevents mouse events and tooltips from working), but it's in the Mapbox overlay as well, so I guess it's acceptable

I tested it and mouse events do work, as we proxy them: https://github.com/visgl/deck.gl/blob/f0649b69d6a91e0aec146da9c6d30d60ccf4d1ef/modules/google-maps/src/utils.js#L252-L289

Did you come across something that was broken?

modules/google-maps/src/google-maps-overlay.js
felixpalmer: As I also read the value out in utils.js I wanted it to be defaulted in one place. I've removed the mutation.

Pessimistress (approved):

zakjan:
> Did you come across something that was broken?

Ok, sorry my mistake, the canvas I tested it with was overlayed by another element. Events works :+1:

Issue #6813, opened 2022-04-11
[Bug] CPUGridLayer misses some columns when cellSize === 1


fifzteen:
### Description

CPUGridLayer some columns are missing.
I found this when `props.cellSize === 1`.
I’m not sure if this bug raises with other values.

See my Codepen in `Steps to Reproduce` section.
this image is a screen capture of that codepen.

![CPUGridLayer_column_missing](https://user-images.githubusercontent.com/25410805/162687184-0197a652-415f-4423-9b2a-5c9a7c4b5db1.png)

The length of `data` passed from CPUGridLayer is 12. (should be 4 x 4 = 16)
And the values of `lonIdx` are (`28818328`, `28818330`, `28818331`). 
`28818329` is missing, even though data points exist in those cell area.

In every case I faced, missed cells are column-ordered.

### Flavors

- [ ] React
- [ ] Python/Jupyter notebook
- [ ] MapboxLayer
- [ ] GoogleMapsOverlay
- [ ] CartoLayer
- [ ] DeckLayer/DeckRenderer for ArcGIS

### Expected Behavior

CPUGridLayer fills all grid cells(columns) contains some data point.

### Steps to Reproduce

https://codepen.io/fifzteen/pen/GRyGzNw

### Environment

- Framework version: deck.gl v8.7.3
- Browser: Chrome 100.0.4896.75
- OS: macOS 11.6.2

felixpalmer:
Seems to be some sort of precision issue, I tried a few different values and `1.04` is broken while `1.05` is fine

felixpalmer:
As your data is very fine-grained, it is better if you use a [different coordinate system](https://deck.gl/docs/developer-guide/coordinate-systems). Could you try defining your data in `COORDINATE_SYSTEM.METER_OFFSETS` or `COORDINATE_SYSTEM.LNGLAT_OFFSETS`?

fifzteen:
@felixpalmer 
> it is better if you use a [different coordinate system](https://deck.gl/docs/developer-guide/coordinate-systems).

I add `LNGLAT_OFFSETS` codes to my Codepen... That's works fine!
Thanks a lot!
![スクリーンショット 2022-04-11 19 27 50](https://user-images.githubusercontent.com/25410805/162721883-11eef266-914a-4305-b37f-5a95896d378a.png)

Pessimistress:
Looks like positions are stored in a `Float32Array` in the `CPUGridLayer`. I'll put up a fix.

[...]

PR #6815, opened 2022-04-12
Fix position precision in CPU aggregation layers

Pessimistress:
For #6813

#### Change List
- Use double precision for positions in `CPUGridLayer` and `HexagonLayer`


felixpalmer (approved):

kylebarron (approved):

fifzteen:
How about this PR? I eagerly await this fix!

PR #6819, opened 2022-04-13
CARTO: fetchMap support for custom breaks

padawannn:
Background
Add support for layers colors custom breaks

Change List
- Read the colorMap from Kepler config
- For ordinal scales, order the categories by frequency instead of alphabetical

### DIFF
--- a/modules/carto/src/api/layer-map.ts
+++ b/modules/carto/src/api/layer-map.ts
@@ -7,7 +7,8 @@ import {
   scalePoint,
   scaleQuantile,
   scaleQuantize,
-  scaleSqrt
+  scaleSqrt,
+  scaleThreshold
 } from 'd3-scale';
 import {format as d3Format} from 'd3-format';
 import moment from 'moment-timezone';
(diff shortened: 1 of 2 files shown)


padawannn:
Tests for `colorMap` pending to add

modules/carto/src/api/layer-map.ts
```diff
@@ -189,6 +193,20 @@ function domainFromValues(values, scaleType: SCALE_TYPE) {
   return extent(values);
 }
 
+function uniqueSortedByFrequency(values) {
```
felixpalmer, line 196: You can replace `uniqueSortedByFrequency` with `d3.groupSort(values, g => -g.length, d => d)`

felixpalmer (changes requested):
You can replace

modules/carto/src/api/layer-map.ts
padawannn: Thx, I didn't know about this d3 function

padawannn:
@felixpalmer tests added

felixpalmer (approved):

PR #6822, opened 2022-04-14
[ts] QuadkeyLayer and GeoCellLayer translated to typescript.

zbigg:
For #5589

Change List:
* Convert PolygonLayer to Typescript (as dependency of GeoCellLayer)
* Convert GeoCellLayer to ts (as dependency of QuadkeyLayer)
* Convert QuadkeyLayer to Typescript


modules/geo-layers/src/quadkey-layer/quadkey-layer.ts
```diff
@@ -0,0 +1,39 @@
...
+  Required<QuadkeyLayerProps> & ExtraProps
+> {
+  static layerName = 'QuadkeyLayer';
+  static defaultProps: any = defaultProps;
+
+  indexToBounds(): Partial<GeoCellLayer['props']> | null {
+    const {data, getQuadkey} = this.props;
+
+    return {
+      data,
```
Pessimistress, line 32: Not related to this PR, why do you need to pass data here?

modules/layers/src/polygon-layer/polygon-layer.ts
```diff
@@ -18,12 +18,186 @@
...
+  log,
+  Unit,
+  UpdateParameters
+} from '@deck.gl/core';
 import SolidPolygonLayer from '../solid-polygon-layer/solid-polygon-layer';
 import PathLayer from '../path-layer/path-layer';
 import * as Polygon from '../solid-polygon-layer/polygon';
 import {replaceInRange} from '../utils';
 
+// TODO: not sure where this type belongs to
```
Pessimistress, line 38: This should be defined by core's `lighting` shader module. You can leave it as is for now.

modules/layers/src/polygon-layer/polygon-layer.ts
```diff
@@ -18,12 +18,186 @@
...
+};
+
+/**
+ * Properties for `PolygonLayer`.
+ */
+export type PolygonLayerProps<DataT = any> = {
+  /**
+   * Whether to draw an outline around the polygon (solid fill).
+   *
+   * Note that both the outer polygon as well the outlines of any holes will be drawn.
```
Pessimistress, line 53: Add default value to the docs?

modules/layers/src/polygon-layer/polygon-layer.ts
```diff
@@ -18,12 +18,186 @@
...
+
+  /** Called on each object in the data stream to retrieve its corresponding polygon. */
+  getPolygon?: AccessorFunction<DataT, any>;
+
+  /**
+   * Fill collor value or accessor.
+   *
+   * @default [0, 0, 0, 255]
+   */
+  getFillColor?: Accessor<DataT, Color> | null;
```
Pessimistress, line 143: Why the `| null` here?

modules/geo-layers/src/quadkey-layer/quadkey-layer.ts
zbigg: That's probably question to @felixpalmer 

It appears that while extracting `GeoCellLayer` in #6678 it was useful to leave `data` as responsibility of concrete class: `H3ClusterLayer` needs to pre-process data and keep it in `state` and `QuadKeyLayer` just forward data as is.

modules/layers/src/polygon-layer/polygon-layer.ts
zbigg: Added to all props.

modules/layers/src/polygon-layer/polygon-layer.ts
zbigg: Mistake, fixed this and alike.

Pessimistress (approved):

PR #6835, opened 2022-04-19
[ts] GoogleMapsOverlay converted to Typescript.

zbigg:

Change List:

* Convert GoogleMapsOverlay to Typescript

### DIFF
--- a/modules/google-maps/src/google-maps-overlay.ts
+++ b/modules/google-maps/src/google-maps-overlay.ts
@@ -7,6 +7,7 @@ import {
   getViewPropsFromOverlay,
   getViewPropsFromCoordinateTransformer
 } from './utils';
+import {Deck, Layer} from '@deck.gl/core';
 
 const HIDE_ALL_LAYERS = () => false;
 const GL_STATE = {
@@ -17,27 +18,34 @@ const GL_STATE = {
   blendEquation: GL.FUNC_ADD
 };
 
+// eslint-disable-next-line @typescript-eslint/no-empty-function
 function noop() {}
 
 const defaultProps = {
   interleaved: true
 };
 
+// TODO: this should be alias to Deck Props.
+export type GoogleMapsOverlayProps = Record<string, any>;
 export default class GoogleMapsOverlay {
-  constructor(props) {
-    this.props = {};
-    this._map = null;
+  private props: GoogleMapsOverlayProps = {};
+  private _map: google.maps.Map | null = null;
+  private _deck: Deck | null = null;
+  private _overlay: google.maps.WebGLOverlayView | google.maps.OverlayView | null = null;
+
+  constructor(props: GoogleMapsOverlayProps) {
     this.setProps({...defaultProps, ...props});
   }
 
   /* Public API */
 
-  setMap(map) {
+  /** Add/remove the overlay from a map. */
+  setMap(map: google.maps.Map | null): void {
     if (map === this._map) {
       return;
     }
     if (this._map) {
-      this._overlay.setMap(null);
+      this._overlay?.setMap(null);
       this._map = null;
     }
     if (map) {
@@ -54,7 +62,10 @@ export default class GoogleMapsOverlay {
     }
   }
 
-  setProps(props) {
+  /**
+   * Update (partial) props.
+   */
+  setProps(props: Partial<GoogleMapsOverlayProps>): void {
     Object.assign(this.props, props);
     if (this._deck) {
       if (props.style) {
@@ -65,18 +76,22 @@ export default class GoogleMapsOverlay {
     }
   }
 
+  /** Equivalent of `deck.pickObject`. */
   pickObject(params) {
     return this._deck && this._deck.pickObject(params);
   }
 
+  /** Equivalent of `deck.pickObjects`.  */
   pickMultipleObjects(params) {
     return this._deck && this._deck.pickMultipleObjects(params);
   }
 
+  /** Equivalent of `deck.pickMultipleObjects`. */
   pickObjects(params) {
     return this._deck && this._deck.pickObjects(params);
   }
 
+  /** Remove the overlay and release all underlying resources. */
   finalize() {
     this.setMap(null);
     if (this._deck) {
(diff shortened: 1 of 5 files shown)


modules/google-maps/src/utils.ts
```diff
@@ -287,3 +310,9 @@ function handleMouseEvent(deck, type, event) {
       return;
   }
 }
+
```
zbigg, line 313: I'm really not sure whether we should export `assert` from [core](https://github.com/visgl/deck.gl/pull/6822/files#diff-c7af049053d7ecbaeda943f96b0bc615a658be293117ce9185afc87593742ef8R90) or redefine in each of module.

modules/google-maps/src/utils.ts
```diff
@@ -287,3 +310,9 @@ function handleMouseEvent(deck, type, event) {
...
+export default function assert(condition: unknown, message?: string): asserts condition {
```
zbigg, line 314: I'm really not sure whether we should export `assert` from [core](https://github.com/visgl/deck.gl/pull/6822/files#diff-c7af049053d7ecbaeda943f96b0bc615a658be293117ce9185afc87593742ef8R90) or redefine in each of module.

modules/google-maps/src/utils.ts
```diff
@@ -99,17 +108,20 @@ export function getViewPropsFromOverlay(map, overlay) {
   const projection = overlay.getProjection();
 
   const bounds = map.getBounds();
+  assert(bounds);
```
zbigg, line 111: Now, this other usages of `assert` may be controversial, so small explanation:

1) `map.getBounds` and similar APIs _according_ to definition can always return `null` or other non-value
2) so, with old code we would have "NullPointerException" here

So, potential bug was lurking, but i wasn't sure how to fix it.

However, typescript complains and we have few ways of addressing this:
 - call `assert` which unfortunately contain _some_ runtime and bundle cos
 - use typescript `!`, but in my opinion, it's ugly and indistigushable from lazy programming
 - spam `// @ts-expect-error` in every usage
 - change code, in a way that we return some kind of no-value (null/undefined) and act accordingly in upper layer like ... skip render in invalid state?
 
I've chosen `assert` as this is fail-safe and most verbose and don't change current behavior, but i understand that we may have other priorities.

modules/google-maps/src/utils.ts
```diff
@@ -142,7 +155,8 @@ export function getViewPropsFromOverlay(map, overlay) {
   // Maps sometimes returns undefined instead of 0
   const heading = map.getHeading() || 0;
 
-  let zoom = map.getZoom() - 1;
+  let zoom = (map.getZoom() ?? 0) - 1;
```
zbigg, line 158: @felixpalmer 
I'm not sure how act here. `getZoom` can return `undefined` and that then?

(see also previous comment)

felixpalmer: `getZoom` should only return undefined if the map hasn't been initialized yet. As the code above (see my other comment) will catch this case when accessing `getBounds` I would say we should assume `getZoom` always returns a valid value and use `!`.

modules/google-maps/src/utils.ts
felixpalmer: We have `log.assert` in core, can you use that instead?

felixpalmer (changes requested):

modules/google-maps/src/google-maps-overlay.ts
```diff
@@ -17,27 +18,34 @@ const GL_STATE = {
...
 };
 
+// eslint-disable-next-line @typescript-eslint/no-empty-function
 function noop() {}
 
 const defaultProps = {
   interleaved: true
 };
 
+// TODO: this should be alias to Deck Props.
```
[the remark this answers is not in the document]
zbigg: AFAIR, we don't have typings for `Deck` nor its props adding them now, would quite big undertaking...
@felixpalmer

modules/google-maps/src/google-maps-overlay.ts
```diff
@@ -86,46 +101,56 @@ export default class GoogleMapsOverlay {
...
-      if (interleaved) {
-        overlay.onAdd = noop;
-        overlay.onContextRestored = this._onContextRestored.bind(this);
-        overlay.onDraw = this._onDrawVectorInterleaved.bind(this);
-      } else {
-        overlay.onAdd = this._onAdd.bind(this);
-        overlay.onContextRestored = noop;
-        overlay.onDraw = this._onDrawVectorOverlay.bind(this);
+
+    const overlay = (() => {
```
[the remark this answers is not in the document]
zbigg: Original code called multiple methods of sum type, resulting with type errors.
I found it _easiest_ way to clean up code without many casts to tell ts on which type we're actually working.

felixpalmer (approved):

modules/main/src/index.ts
```diff
@@ -68,6 +68,7 @@ export {
   DeckRenderer,
   // Logging
   log,
+  assert,
```
Pessimistress, line 71: Prefix internal utilities with `_` so that you wouldn't have to re-export it here. This function is not intended to be used by the users of this library.

modules/google-maps/src/google-maps-overlay.ts
```diff
@@ -179,11 +204,8 @@ export default class GoogleMapsOverlay {
 
   // Vector code path
   _onDrawVectorInterleaved({gl, transformer}) {
-    if (!this._deck || !this._map) {
-      return;
-    }
-
     const deck = this._deck;
+    assert(this._map && this._overlay && deck);
```
Pessimistress, line 208: This is a breaking change. Where it silently skipped before you are now crashing the app. Is it necessary?

modules/google-maps/src/google-maps-overlay.ts
```diff
@@ -221,11 +243,9 @@ export default class GoogleMapsOverlay {
   }
 
   _onDrawVectorOverlay({gl, transformer}) {
-    if (!this._deck || !this._map) {
-      return;
-    }
-
     const deck = this._deck;
+    assert(this._map && this._overlay && deck);
```
Pessimistress, line 247: Ditto

modules/google-maps/src/utils.ts
```diff
@@ -99,17 +108,23 @@ export function getViewPropsFromOverlay(map, overlay) {
...
   const bounds = map.getBounds();
+  if (!bounds) {
+    return {width, height, left: 0, top: 0};
+  }
+
   const ne = bounds.getNorthEast();
   const sw = bounds.getSouthWest();
   const topRight = projection.fromLatLngToDivPixel(ne);
   const bottomLeft = projection.fromLatLngToDivPixel(sw);
+  assert(topRight && bottomLeft);
```
Pessimistress, line 119: In this code base we only use `assert`s to notify the user that there is something wrong with their usage pattern. In this context there is no clear action item for the user to fix the error if and when it occurs. `assert` is not a convenience to work around TypeScript errors. Consider casting the output of `fromLatLngToDivPixel` to a non-null type unless there is a legitimate reason that this might break in runtime.

Pessimistress (changes requested):
Reduce the usage of asserts to avoid breaking changes in this PR. If necessary, we can introduce them in a follow up PR.

modules/google-maps/src/google-maps-overlay.ts
zbigg: Btw, I've discussed it in [already collapsed comment](https://github.com/visgl/deck.gl/pull/6835#discussion_r853017297).

I see that in #6803 You've chose `// @ts-ignore (TS2531) state is always defined` method -  i will adapt to this way of working-around Typescript strictness.

Rest of comment only for reference.

> This is a breaking change (..) Where it silently skipped before ...

I know that excessive usage of `assert` have consequences and i'll try to reduce it, but it's not _hard_ breaking change i.e would  `this._map && this._overlay && deck` be falsy, this method would crash by with old "cannot access 'x' of undefined/null" at first random access to `deck` or to `this._map`.
So, it only changes signature of crash.

(Yes, `this._overlay` check was excessive copy-pasta, sorry)

Pessimistress: I'm not sure I follow. Your previous comment was about the use of assert in the `getViewPropsFromOverlay` utility. This particular conversation is regarding you deleting

```js
if (!this._deck || !this._map) {
      return;
    }
```

And adding

```js
assert(this._map && this._overlay && deck);
```

How is it not a breaking change?

Referencing original PR #6340

zbigg: @Pessimistress 
I am sorry for deleting comment, it was mistake from my side and i deleted to prevent "misinformation and wasted effort" it as soon as i've realized it's bogus.

---

You're absolutely right, this particular assertion was changing behavior. I've managed to remove most all of them by changing approach to "no-op" in callbacks.

Here's my original comment for reference:

---
Btw, I've discussed it in already collapsed comment.

I see that in #6803 You've chose // @ts-ignore (TS2531) state is always defined method - i will adapt to this way of working-around Typescript strictness.

Rest of comment only for reference.

This is a breaking change (..) Where it silently skipped before ...

I know that excessive usage of assert have consequences and i'll try to reduce it, but it's not hard breaking change i.e would this._map && this._overlay && deck be falsy, this method would crash by with old "cannot access 'x' of undefined/null" at first random access to deck or to this._map.
So, it only changes signature of crash.

(Yes, this._overlay check was excessive copy-pasta, sorry)

modules/google-maps/src/utils.ts
zbigg: Note taken. Changing it in all pending PRs of mine.

zbigg: In this case, i'll follow @felixpalmer [suggestion](https://github.com/visgl/deck.gl/pull/6835#discussion_r853088637) and return _minimal_ props that should work.

Pessimistress (approved):

Issue #6839, opened 2022-04-20
[Bug/codedoc] function comment references parameter that doesn't exist 


mitresthen:
### Description

Commit https://github.com/visgl/deck.gl/commit/95ac009d05acae4b0a2a69b09d5a565bb28322ac Removed param vertexCount from function normalize, without removing the corresponding param reference in the documentation. 

Thus the function definition ( https://github.com/visgl/deck.gl/blob/8.7-release/modules/layers/src/solid-polygon-layer/polygon.js#L176 ) is now:

`/**
 * Normalize any polygon representation into the "complex flat" format
 * @param {Array|Object} polygon
 * @param {Number} positionSize - size of a position, 2 (xy) or 3 (xyz)
 * @param {Number} [vertexCount] - pre-computed vertex count in the polygon.
 *   If provided, will skip counting.
 * @return {Object} - {positions: <Float64Array>, holeIndices: <Array|null>}
 */
/* eslint-disable max-statements */
export function normalize(polygon, positionSize) {`

This might cause issues when generating types based on the function docs.

PR #6840, opened 2022-04-20
Remove non-existent param from function doc in polygon.js

mitresthen:
Remove non-existent param from function doc, as described in: https://github.com/visgl/deck.gl/issues/6839

For #6839

#### Change List
- Removed [vertexCount] from doc, as it is not a param of the function


Pessimistress (approved):
Thank you!

PR #6842, opened 2022-04-20
[ts] MVTLayer

zbigg:
For https://github.com/visgl/deck.gl/issues/5589

Change List:

* Convert MVTLayer to Typescript
* TileLayer props alignment after #6848
* TileLayer: cleanup extra asserts

PR #6847, opened 2022-04-21
Base Layer type signature

Pessimistress:
For #5589 

Continue discussion from https://github.com/visgl/deck.gl/pull/6803#discussion_r846417620

In this proposal, we no longer require the generic type `PropsT` to extend `LayerProps`. The type of `layer.props` is instead enforced with `PropsT & Required<LayerProps>`. This removes TypeScript's complaint that `PropsT` may override some of `LayerProps` fields.

See proof-of-concept in [this Playground snippet](https://www.typescriptlang.org/play?#code/C4TwDgpgBAMghiCAnACkg9mAzlAvFAbwCgpSpM4BjAS1AH4AuKAOwFcBbAI2QG4iBfIkQD0AKlFQAQnCzQANgmRRRwopQVYc8REgA8aTFgAqAPkIkyYDNiYHsRqADIoAJQgBHVtSQQAJru1kOywTITIoSnRmLGAkVkpgdCQACgA6dKtDWzgkYGo4OX1rYxMAbQBdAEpzcPDgAAtqLFTM7DwoAHlOACsIBNSZLGoAc2Zkgn4AGih0luLKi1JBRahhiGAOsCpaEGTKpjYuJWJash9gViRmKAamucNUihpQFcFl0EgoI2o5CEDUYrtE5kdjUZgALXQ6HYjBYHG4SD4yzEEgAyqxOBENDgAIzKVTqQZfH5-RR6ACiAA9YnBgg58BMzBBqRBmL4tGTdFSaXSnK4PF4fP5vr9-sETGZgaQ1sAALJgyHQvYHeHHFbhc6Xa63ZqtZqgiFQ9ivARCD7QWUANSMYsBDJW1F8dmQ9CYMSQYOGSKEKKg6MxhM0UAATPi1NioFabZzuUhacV6YR+EyWWycCLSToudS47znG5PN4-Loo7bDBKamQZQBJJ3WF27fZQd2eyunTVXG6NXXFVKO525EAm97gC0gADCrBi0LLbXt4UiciSsMOCIqppE4j9GKxRIAzFAALQsdDAKDM4CsoacX5hwM4WUTqeJdj-c+p9mR63-XQFwXFx9J2nV8yXFSUVhlcd0CXFIm1XZAKjbWoO21bt7mwVJFySPhwjeIggA)

`ScatterplotLayerProps` still extends `LayerProps` in case the user needs to use the type outside of the constructor. However, it is no longer necessary.

### DIFF
--- a/modules/geo-layers/src/tile-layer/tile-layer.ts
+++ b/modules/geo-layers/src/tile-layer/tile-layer.ts
@@ -1,10 +1,10 @@
 import {
   assert,
-  ChangeFlags,
   CompositeLayer,
   CompositeLayerProps,
   Layer,
   LayerProps,
+  UpdateParameters,
   _flatten as flatten
 } from '@deck.gl/core';
 import {GeoJsonLayer} from '@deck.gl/layers';
@@ -39,12 +39,20 @@ const defaultProps = {
   zoomOffset: 0
 };
 
-export type TileLayerProps<DataT = any> = CompositeLayerProps<DataT> & {
+/** All props supported by the TileLayer */
+export type TileLayerProps<DataT> = _TileLayerProps<DataT> & CompositeLayerProps<DataT>;
+
+/** Props added by the TileLayer */
+type _TileLayerProps<DataT> = {
+  /**
+   * Optionally implement a custom indexing scheme.
+   */
+  TilesetClass: typeof Tileset2D;
   /**
    * Renders one or an array of Layer instances.
    */
-  renderSubLayers: (
-    props: TileLayerProps & {
+  renderSubLayers?: (
+    props: TileLayerProps<DataT> & {
       id: string;
       data: any;
       _offset: number;
@@ -54,58 +62,62 @@ export type TileLayerProps<DataT = any> = CompositeLayerProps<DataT> & {
   /**
    * If supplied, `getTileData` is called to retrieve the data of each tile.
    */
-  getTileData: (props: TileLoadProps) => Promise<any> | any;
+  getTileData?: (props: TileLoadProps) => Promise<DataT> | DataT;
 
   /** Called when all tiles in the current viewport are loaded. */
   onViewportLoad?: (tiles: Tile2DHeader[]) => void;
 
   /** Called when a tile successfully loads. */
-  onTileLoad: (tile: Tile2DHeader) => void;
+  onTileLoad?: (tile: Tile2DHeader) => void;
 
   /** Called when a tile is cleared from cache. */
-  onTileUnload: (tile: Tile2DHeader) => void;
+  onTileUnload?: (tile: Tile2DHeader) => void;
 
   /** Called when a tile failed to load. */
-  onTileError: (err: any) => void;
+  onTileError?: (err: any) => void;
 
   /** The bounding box of the layer's data. */
-  extent: number[] | undefined | null;
+  extent?: number[] | null;
 
   /** The pixel dimension of the tiles, usually a power of 2. */
-  tileSize: number;
+  tileSize?: number;
 
-  /** The max zoom level of the layer's data.  */
-  maxZoom: number;
+  /** The max zoom level of the layer's data.
+   * @default null
+   */
+  maxZoom?: number | null;
 
-  /** The min zoom level of the layer's data.  */
-  minZoom: number;
+  /** The min zoom level of the layer's data.
+   * @default 0
+   */
+  minZoom?: number | null;
 
   /** The maximum number of tiles that can be cached. */
-  maxCacheSize: number | undefined | null;
+  maxCacheSize?: number | null;
 
   /**
    * The maximum memory used for caching tiles.
    *
    * @default null
    */
-  maxCacheByteSize: number | undefined;
+  maxCacheByteSize?: number | null;
 
   /**
    * How the tile layer refines the visibility of tiles.
    *
    * @default 'best-available'
    */
-  refinementStrategy: RefinementStrategy;
+  refinementStrategy?: RefinementStrategy;
 
   /** Range of minimum and maximum heights in the tile. */
-  zRange: ZRange | null;
+  zRange?: ZRange | null;
 
   /**
    * The maximum number of concurrent getTileData calls.
    *
    * @default 6
    */
-  maxRequests: number;
+  maxRequests?: number;
 
   /**
    * This offset changes the zoom level at which the tiles are fetched.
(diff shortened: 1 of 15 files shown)


zbigg (commented):
Honestly i have not enough background in project to give strong opinions ...

In general, i think we're working-around definiciencies of Typescript expressivity and fact
that component/layer architecture is _very_ hard if not not-economical to type properly.

But that statement is not very helpful, unfortunately i don't know if proposed direction is good
or wrong.

I've played a little with this approach in #6822 and i'll comment about issues later.

Anyway,

I like spirit of having only _extra_ props as part of type signature like here:

    class TileLayer<ExtraPropsT = {}> extends Layer<ExtraPropsT & Required<TileLayerProps>> {

However, i find lack of proper `PropsT` in scope of classes
problematic and boilerplate-prone. Example:

   > updateState<T extends _ScatterplotLayerProps<DataT>>(params: UpdateParameters<T>) {

It doesn't even compile (don't know why ...).
Following PRs would introduce more of similarly "over"-typed methods.
Those and similar methods have by no means "another dimension" of genericity.
This really should depend on good-old `PropsT`, which is
class parameter. 

> updateState<(params: UpdateParameters<PropsT>) {

Unfortunately typescript doesn't allow class-local `type PropsT = boilerplate` to reuse in scope :/

Pessimistress:
@zbigg I have updated the TileLayer typings. Please take another look.

modules/geo-layers/src/tile-layer/tile-layer.ts
```diff
@@ -150,15 +161,7 @@ export default class TileLayer<
...
-  updateState({
-    props,
-    changeFlags
-  }: {
-    props: TileLayerProps;
-    oldProps: TileLayerProps;
-    context: any;
-    changeFlags: ChangeFlags;
-  }) {
+  updateState({changeFlags}: UpdateParameters<ExtraPropsT & Required<_TileLayerProps<DataT>>>) {
```
zbigg, line 164: Technically this signature is incorrect as we're missing fact that we also have _CompositeLayerProps included.

It's not used, but that makes me struggle to write proper signature for MVTLayer.

modules/layers/src/scatterplot-layer/scatterplot-layer.ts
```diff
@@ -134,7 +145,7 @@ export default class ScatterplotLayer<
     });
   }
 
-  updateState(params) {
+  updateState(params: UpdateParameters<ExtraPropsT & Required<_ScatterplotLayerProps<DataT>>>) {
```
zbigg, line 148: I've found workaround, that enables _not_ to repeat this monstrous signature:

```
export type UpdateParameters2<LayerT extends Layer> = {
  props: LayerT['props'];
  oldProps: LayerT['props'];
  context: LayerContext;
  changeFlags: ChangeFlags;
};
updateState({props, oldProps, context, changeFlags}: UpdateParameters2<this>) {
```

modules/geo-layers/src/tile-layer/tile-layer.ts
zbigg: MVTLayer signature will have too look like this:
```
updateState({
    props,
    oldProps,
    context,
    changeFlags
  }: UpdateParameters<ExtraPropsT & Required<_MVTLayerProps> & Required<_TileLayerProps<DataT>>>) {
  ```
  
  (that requires to export remember all _parent_, _partial_ parent classes ...)

File: modules/geo-layers/src/tile-layer/tile-layer.ts [hunk 3]
```diff
@@ -290,7 +293,7 @@ export default class TileLayer<
   }
 
   renderSubLayers(
-    props: TileLayerProps & {
+    props: _TileLayerProps<DataT> & {
```
zbigg, line 296: This is technically incorrect as _TileLayerProps has lots of optional stuff.

And _in_ this stage of processing `props` are resolved.

I this we should invent something like this:

```
export type ComponentProps<T extends Compoennt> = Component["props"]
```


and this line would be:
```
renderSubLayers(
    props: ComponentProps<this>
 ) {
     props.tileSize // is of type `number` with optionals removed
}
```
This would of course remove assume `DataT=any`, but if someone needs genericty, he can write:

```
props: ComponentProps<TileLayer<DataT,ExtraPropsT>> 
```
and have proper signature.

modules/layers/src/scatterplot-layer/scatterplot-layer.ts
```diff
@@ -80,10 +92,9 @@ const defaultProps = {
   getColor: {deprecatedFor: ['getFillColor', 'getLineColor']}
 };
 
-export default class ScatterplotLayer<
-  DataT = any,
-  PropsT extends ScatterplotLayerProps<DataT> = ScatterplotLayerProps<DataT>
-> extends Layer<PropsT> {
+export default class ScatterplotLayer<DataT = any, ExtraPropsT = {}> extends Layer<
+  ExtraPropsT & Required<_ScatterplotLayerProps<DataT>>
```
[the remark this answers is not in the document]
zbigg: Pity, didn't expected that. 

Note, this [stack answer](https://stackoverflow.com/a/71483186) sheds some light on this. Looks like we have to remove `undefined` twice.

And this `FullRequired` does trick at least in [derived sandbox](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgELIN4FgBQz-IBGhA-AFzIDOYUoA5gNy4HIBeFhA9pwDYRwhcAX1yhIsRCgCCmZgQAe5ZCACuAW0LQmOETlxgAngAcUAMRU8eAJQgBHFcCgQAJgB4AKsgjzIIZ5WROQgArCAQwAD5kAF5kG3tHF1dsPGQAbQBpZFBkAGsIA04YZHcAXQoAUXkEHhVnCA9M0oAaZBU-CBhQFwjtIV7cXAB6IeQYdvDgThBkRDAPLx8IPwD0WNQIgApOIzBKCniHJzdPADJkKQiASlkccZBJ6dnwhe9ffzQYtC2dvYpzSyHRInZDnS43FIsABucCgyAMFFUGmgX1+lAAdPJhIM7hMwFMZnMAEweH67fZxOxHJJnSkJY6uS7XW7Q2HwxHqTRw2JozHYnBAA)

zbigg:
I've tried to use presented approach for MVTLayer migration (#6842), but pushed to separate branch [here](https://github.com/visgl/deck.gl/compare/zbigg/ts-test-mvt-layer?expand=1#diff-7718cc277780a038e3a94c20e4833adc596791fa96fe87780161c606993d1c45R100).

I would say, that current typings are way to scary and i would risk saying impossible to maintain and understand (i admit as moderately experienced typescripter).

However, if we simplify types and deduce types from `Layer["props"]` as proposed above this approach can work.

zbigg (approved):

PR #6848, opened 2022-04-22
Expose Generic Tileset2D class for use with TileLayer

felixpalmer:
For https://github.com/visgl/deck.gl/issues/6816 See proposal for new API there


modules/geo-layers/src/tile-layer/tile-layer.ts
```diff
@@ -272,11 +274,11 @@ export default class TileLayer<
   // Methods for subclass to override
 
   getTileData(tile: TileLoadProps) {
```
Pessimistress, line 276: Nit: Just pass `TileIndex` as argument?

modules/geo-layers/src/tile-layer/tileset-2d.ts
```diff
@@ -261,11 +259,20 @@ export default class Tileset2D {
     });
   }
 
+  // Returns unique string key for a tile index
```
Pessimistress, line 262: Use `/** */` comments to add TypeScript annotation to the public methods

Pessimistress (commented):
To carry on with the discussion, it would be very helpful to see the updated documentation for this PR. As commented inline, it seems that there are breaking changes to the current accessor APIs (`tile.x` etc are no longer present).

Pessimistress:
Let's upload test data to https://github.com/visgl/deck.gl-data/tree/master/test-data instead

modules/geo-layers/src/tile-layer/utils.ts
```diff
@@ -57,19 +57,16 @@ function transformBox(bbox: Bounds, modelMatrix: Matrix4): Bounds {
...
   }
   if (Array.isArray(template)) {
-    const index = Math.abs(properties.x + properties.y) % template.length;
-    template = template[index];
+    const i = Math.abs(index.x + index.y) % template.length;
+    template = template[i];
   }
 
-  const {x, y, z} = properties;
+  const {x, y, z} = index;
```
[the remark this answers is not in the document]
felixpalmer: Originally I envisaged the user implementing a custom `getURLFromTemplate` function, but have switched to your suggestion of iterating over the keys

docs/api-reference/geo-layers/tile-layer.md
```diff
@@ -112,9 +112,8 @@ If the value is an array: multiple URL templates. Each endpoint must return the
 
 If supplied, `getTileData` is called to retrieve the data of each tile. It receives one argument `tile` which contains the following fields:
 
-- `x` (Number) - x index of the tile
-- `y` (Number) - y index of the tile
-- `z` (Number) - z index of the tile
+- `index` (Object) - index of the tile. `index` is in the shape of `{x, y, z}`, corresponding to the integer values specifying the tile.
```
Pessimistress, line 115: Add the breaking change to upgrade guide

docs/api-reference/geo-layers/tile-layer.md
```diff
@@ -112,9 +112,8 @@ If the value is an array: multiple URL templates. Each endpoint must return the
...
+- `cacheKey` (String) - unique string representation of index, as 'x-y-z', e.g. '0-2-3'.
```
Pessimistress, line 116: The rest of the code base would just name this kind of property `id`

Pessimistress (approved):

docs/api-reference/geo-layers/tile-layer.md
felixpalmer: I've renamed to `id`

docs/api-reference/geo-layers/tile-layer.md
```diff
@@ -112,9 +112,8 @@ If the value is an array: multiple URL templates. Each endpoint must return the
 
 If supplied, `getTileData` is called to retrieve the data of each tile. It receives one argument `tile` which contains the following fields:
 
-- `x` (Number) - x index of the tile
-- `y` (Number) - y index of the tile
-- `z` (Number) - z index of the tile
+- `index` (Object) - index of the tile. `index` is in the shape of `{x, y, z}`, corresponding to the integer values specifying the tile.
+- `id` (String) - unique string representation of index, as 'x-y-z', e.g. '0-2-3'.
```
Pessimistress, line 116:
```suggestion
- `id` (String) - unique string representation of index.
```

just leave it like that since `index` can be in other forms.

docs/api-reference/geo-layers/tile-layer.md
```diff
@@ -1,4 +1,4 @@
-import {TileLayerDemo} from 'website-components/doc-demos/geo-layers';
+id {TileLayerDemo} from 'website-components/doc-demos/geo-layers';
```
kylebarron, line 1:
```suggestion
import {TileLayerDemo} from 'website-components/doc-demos/geo-layers';
```
?

docs/api-reference/geo-layers/tile-layer.md
```diff
@@ -228,7 +232,7 @@ Renders one or an array of Layer instances with all the `TileLayer` props and th
 
 * `id`: An unique id for this sublayer
 * `data`: Resolved from `getTileData`. As of deck.gl 8.2, this prop is always the data resolved from the Promise and is never a Promise itself.
-* `tile`: An object containing tile index `x`, `y`, `z`, and `bbox` of the tile.
+* `tile`: An object containing `index` and `bbox` and `id` of the tile.
```
kylebarron, line 235: Nit:
```suggestion
* `tile`: An object containing `index`, `bbox`, and `id` of the tile.
```

modules/geo-layers/src/index.ts
```diff
@@ -31,6 +31,7 @@ export {default as TerrainLayer} from './terrain-layer/terrain-layer';
 export {default as MVTLayer} from './mvt-layer/mvt-layer';
 
 export {getURLFromTemplate as _getURLFromTemplate} from './tile-layer/utils';
+export {default as _Tileset2D} from './tile-layer/tileset-2d';
```
kylebarron, line 34: Should we document in the `tile-layer.md` doc that this is currently exported with an underscore?

modules/geo-layers/src/tile-layer/utils.ts
```diff
@@ -57,24 +58,32 @@ function transformBox(bbox: Bounds, modelMatrix: Matrix4): Bounds {
   return transformedBox;
 }
 
-export function getURLFromTemplate(
-  template: string | string[],
-  properties: TileIndex
-): string | null {
+function stringHash(s) {
+  return s.split('').reduce((a, b) => ((a << 5) - a + b.charCodeAt(0)) | 0, 0);
+}
```
kylebarron, line 63: I'm not familiar with deck.gl's TS style rules. When do we use literal input and/or output types?

Just noting that in the case of this function, I think `stringHash(s: string): number` would've made the intent of the function easier to parse at a glance.

docs/api-reference/geo-layers/tile-layer.md
```diff
@@ -307,8 +310,43 @@ Properties:
...
+  }
+
+  getTileZoom({q}) {
+    return q.length;
+  }
+
+  getParentIndex(index) {
+    index.q = index.q.slice(0, -1);
+    return index;
+  }
```
kylebarron, line 346: Do you think it would be helpful to add short one-line explanations of how this works in the context of quadkeys? Or should we expect that readers would be familiar with how quadkeys work?

modules/geo-layers/src/index.ts
felixpalmer: Add to the code example

modules/geo-layers/src/tile-layer/utils.ts
felixpalmer: Good catch

docs/api-reference/geo-layers/tile-layer.md
felixpalmer: I would rather avoid getting into the details as the whole point of exposing the `Tileset2D` class is to allow an arbitrary indexing scheme and so it is sufficient to state what interface must be implemented without specific details on the example. That said, I'll add a link to an article on quadkeys to help readers

PR #6849, opened 2022-04-22
fix doc

YousefED:
default is 0 (https://github.com/visgl/deck.gl/blob/12546f90973a565be9c8c847e73fb273110f874d/modules/aggregation-layers/src/hexagon-layer/hexagon-layer.js#L52)


felixpalmer (approved):
Thanks!

PR #6862, opened 2022-04-26
pydeck: Rename `pydeck.Layer` param `type`

Jesus89:
#### Background
Rename `pydeck.Layer` param `type` to `class_type` to avoid collision with layer params called `type`, for example, with the [CartoLayer](https://deck.gl/docs/api-reference/carto/carto-layer).

#### Change List
- Rename `pydeck.Layer` param `type` to `class_type`

### DIFF
--- a/bindings/pydeck/pydeck/bindings/layer.py
+++ b/bindings/pydeck/pydeck/bindings/layer.py
@@ -15,7 +15,7 @@ QUOTE_CHARS = {"'", '"', "`"}
 
 
 class Layer(JSONMixin):
-    def __init__(self, type, data=None, id=None, use_binary_transport=None, **kwargs):
+    def __init__(self, class_type, data=None, id=None, use_binary_transport=None, **kwargs):
         """Configures a deck.gl layer for rendering on a map. Parameters passed
         here will be specific to the particular deck.gl layer that you are choosing to use.
 
@@ -27,7 +27,7 @@ class Layer(JSONMixin):
         Parameters
         ==========
 
-        type : str
+        class_type : str
             Type of layer to render, e.g., `HexagonLayer`
         id : str, default None
             Unique name for layer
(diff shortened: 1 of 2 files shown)


kylebarron:
Hmm. This would be a backwards incompatible change. What would you think about instead naming the param in the pydeck `CartoLayer` `type_`, and then handling `type_` to `type` in the pydeck `CartoLayer`?

felixpalmer:
I agree with @kylebarron here, the `type` parameter is pretty integral and is used all over the documentation. The collision is unfortunate, could we change `type` to `map_type` in `CartoLayer`? https://deck.gl/docs/api-reference/carto/carto-layer#type

alasarr:
> could we change type to map_type in CartoLayer?

+1

alasarr:
Even I'd like to evaluate to remove type parameter at cartolayer. It's a conversation we've had multiple times and we should make a decision on that. I would like to avoid change the name and then remove.

Jesus89:
Actually `type_` is a valid workaround because it avoids the collision, and then it's converted to `type` in the serialization ([to_camel_case](https://github.com/visgl/deck.gl/blob/776ddfa9aa1f763548e1dd52e28840e1e031e756/bindings/pydeck/pydeck/bindings/json_tools.py#L19)).

I commented this solution some time ago with @ajduberstein, but it's true that introduces a breaking change. I'm OK with renaming or removing (even better) the parameter in `CartoLayer` for deck.gl. In the meanwhile, we can handle with this in the `CartoLayer` for pydeck.

kylebarron:
I don't think you need to change the name of the parameter in the JS layer. You could just intercept `type_` in the Python layer and then change that name to be `type` when passed to JS.

Jesus89:
Actually, this is happening out of the box in pydeck because of the camel conversion of the parameters: `type_` -> `type`. However, we have talked in the past to rename or even remove this parameter to improve CartoLayer API in general.

PR #6864, opened 2022-04-27
CARTO: fetchLayerData added support for spatial index parameters

padawannn:
#### Background
The Carto Maps API now supports spatial indices adding two new parameters: `aggregationExp` and `aggregationResLevel` in the map instantiation. This PR adds these parameters to the fetchLayerData method
#### Change List
- Added `aggregationExp` and `aggregationResLevel` to fetchLayerData method

### DIFF
--- a/modules/carto/src/api/maps-v3-client.ts
+++ b/modules/carto/src/api/maps-v3-client.ts
@@ -144,12 +144,12 @@ function getParameters({
   if (GEO_COLUMN_SUPPORT.includes(type)) {
     if (geoColumn) {
       parameters.push(encodeParameter('geo_column', geoColumn));
-    }
-    if (aggregationExp) {
-      parameters.push(encodeParameter('aggregationExp', aggregationExp));
-    }
-    if (aggregationResLevel) {
-      parameters.push(encodeParameter('aggregationResLevel', aggregationResLevel));
+      if (aggregationExp) {
+        parameters.push(encodeParameter('aggregationExp', aggregationExp));
+      }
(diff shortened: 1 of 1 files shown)


modules/carto/src/api/maps-v3-client.ts
```diff
@@ -137,8 +141,16 @@ function getParameters({
   const sourceName = type === MAP_TYPES.QUERY ? 'q' : 'name';
   parameters.push(encodeParameter(sourceName, source));
 
-  if (GEO_COLUMN_SUPPORT.includes(type) && geoColumn) {
-    parameters.push(encodeParameter('geo_column', geoColumn));
+  if (GEO_COLUMN_SUPPORT.includes(type)) {
+    if (geoColumn) {
```
felixpalmer, line 145: This method is getting quite complex. Can we move this logic into `checkFetchLayerData` and add some tests to check invalid combinations throw an error? Then this method can be simplified to just encode & append parameters when they are present

felixpalmer (changes requested):

modules/carto/src/api/maps-v3-client.ts
padawannn: Done, let me know if you're ok with that

felixpalmer: Perfect, thanks. Could you also add relevant test cases here: https://github.com/visgl/deck.gl/blob/master/test/modules/carto/api/maps-api-client.spec.js#L390-L394

felixpalmer (approved):

PR #6866, opened 2022-04-27
Export CartoLayerLibrary in carto module for pydeck integration

Jesus89:
#### Background

I'm working on supporting `CartoLayer` in a new module `pydeck_carto`. This module will use the [custom layers](https://deckgl.readthedocs.io/en/latest/custom_layers.html) system to load `@deck.gl/carto` module, but this requires a `CartoLayerLibrary` object to be exposed in global.

This PR adds the line to expose `CartoLayerLibrary` in the `dist.min.js` bundle of the carto module, so it can be directly used as a custom layer in pydeck and there is no need to create an extra bundle.

More details in this [proof of concept](https://github.com/visgl/deck.gl/discussions/6861#discussioncomment-2648190).

NOTE: any suggestion about the best place to put a test to cover this code is welcome :)

#### Change List
- Export CartoLayerLibrary in carto module for pydeck integration


modules/carto/bundle.ts
```diff
@@ -7,4 +7,7 @@ if (!deck.LineLayer) {
   throw new Error('@deck.gl/layers is not found');
 }
 
+// Export carto layer library for pydeck integration
+globalThis.CartoLayerLibrary = {CartoLayer: CartoUtils.CartoLayer};
```
felixpalmer, line 11: I must say I'm a bit lost as to why this is necessary for pydeck. Why doesn't pydeck need similar exports to use other deck Layers?

Jesus89:
Sorry, I removed a duplicated comment, but it seems it was the same. Restoring there

> OK, thanks for clarifying. So the alternative here would be to create a wrapper npm package that did nothing but created this export? It's a bit unfortunate that we have to pollute the global scope for everyone who uses the JavaScript module even when they aren't using pydeck. Is there no way to check if we are running in pydeck? Something like:
`if(globalThis.pydeckInstalled) globalThis.CartoLayerLibrary = {CartoLayer: CartoUtils.CartoLayer};`

 I think at the very least it wouldn't hurt to link to https://github.com/ajduberstein/pydeck_custom_layer

Jesus89:
Pydeck imports the file "as is" so I think can not do a global check to expose the library object. But, it is a reference, so it does not increase the size of the bundle.

Another option is to create a new bundle, just for `@deckl.gl/carto` module. This would require a custom webpack configuration.

Regarding the docs, I think it's a good idea to add the link in the code for more context.

Jesus89:
What about the option of creating a new bundle for the carto module with the global export with a custom webpack config?. @felixpalmer @alasarr @ajduberstein.

Jesus89:
Adding @kylebarron for advice and suggestions on this topic.

alasarr (approved):

PR #6880, opened 2022-05-04
Add pydeck_carto alpha version

Jesus89:
#### Background

Add alpha `pydeck_carto` package in `bindings`. This version contains a couple of functions to manage session credentials and register the CartoLayer. This is related to the PR to generate the JS bundle (https://github.com/visgl/deck.gl/pull/6866). More context here: https://github.com/visgl/deck.gl/discussions/6861#discussioncomment-2648190.

This PR adds:
- pydeck_carto source code
- tooling: venv, flake8, black, pytest
- unit tests
- examples

#### Change List
-  Add `pydeck_carto` version 0.1.0a1.


kylebarron (commented):
I'm not the one who makes a decision on this, but I don't see why this should live in the deck.gl repository.

This has its own repo setup, its own Python packaging setup, and by using Pydeck's extension registry there's no static linking to either the Python or JS sides of Pydeck. It loads a hosted version of the `@deck.gl/carto module` https://github.com/visgl/deck.gl/blob/52806653e4992cb2a3f3cf2a0afebb7cefd6693b/bindings/pydeck-carto/pydeck_carto/layer.py#L3

What's the benefit of having this live in the deck.gl repo over an outside repo?

ajduberstein:
> I'm not the one who makes a decision on this, but I don't see why this should live in the deck.gl repository.

@kylebarron Given that this is a comment about the project rather than the code, let's consider this on the related discussion (#6861) rather than this PR.

bindings/pydeck-carto/examples/basic-examples/hello_world.py
```diff
@@ -0,0 +1,30 @@
...
+layer = pdk.Layer(
+    "CartoLayer",
+    data="SELECT geom, name FROM cartobq.public_account.populated_places",
+    type_=pdk.types.String("query"),
+    connection=pdk.types.String("bqconn"),
+    credentials={
+        "apiBaseUrl": "https://gcp-us-east1.api.carto.com",
+        "accessToken": "eyJhbGciOiJIUzI1NiJ9.eyJhIjoiYWNfbHFlM3p3Z3UiLCJqdGkiOiI1YjI0OWE2ZCJ9.Y7zB30NJFzq5fPv8W5nkoH5lPXFWQP0uywDtqUg8y8c"},
+    get_fill_color=[238, 77, 90],
+    pointRadiusMinPixels=2.5)
```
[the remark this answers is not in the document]
Jesus89: This credentials object goes directly to the JS CartoLayer (https://deck.gl/docs/api-reference/carto/carto-layer), so it requires the attributes to be in camelCase.

A possible solution is that pydeck formats the attributes inside dictionaries, but I think this is out of the scope of this PR, so it can be done in a new one.

Jesus89:
Hi @kylebarron, the package was added next to pydeck at the suggestion of @ajduberstein and I think it makes sense because the `@deck.gl/carto` module also lives in the repo, and this is a python binding using that code.

The main friction point, regarding the hosted version of the bundle, is a temporary solution until this topic about the bundle is solved https://github.com/visgl/deck.gl/pull/6866, so the layer could be loaded directly from npm via jsdelivr or unpkg, for example.

As @ajduberstein comments, I think we could use this discussion https://github.com/visgl/deck.gl/discussions/6861 to talk about this topic.

Jesus89:
Hi, after a conversation with @alasarr, we can move forward with this PR as is (using the current bundle for the alpha version). But after that, we will create another PR to deploy a beta release of the package pointing to the npm bundle: `https://cdn.jsdelivr.net/npm/@deck.gl/carto@~8.8.*/dist.min.js`.

I've included @volaya to do an extra review before merging.

volaya:
It all looks good to me. 

I would suggest adding a "Installing from source" entry under the "Development" section of the readme, to be in line with what is found in the pydeck instructions. Kinda trivial, but for the sake of completeness.

Great work!

volaya (approved):

bindings/pydeck-carto/pydeck_carto/credentials.py
```diff
@@ -0,0 +1,38 @@
+import json
+import requests
+
+
+def load_carto_credentials(filepath):
+    """Load credentials from file. It must contain the following attributes:
+    - api_base_url
```
alasarr, line 7: Not for right now, but a great feature would be that a user could get the accessToken through the browser without requiring to create an app at CARTO dashboard.

As google collab does:
image

bindings/pydeck-carto/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+### Installing from source
+
+```bash
+git clone https://github.com/visgl/deck.gl
+cd deck.gl/bindings/pydeck-carto
+pip install .
+```
+
+## Usage
```
alasarr, line 21: Need to say how to get the client_id and secret

Jesus89: Done

Issue #6871, opened 2022-04-29
[Bug] Tile3DLayer tests intermittently failing due to race condition


felixpalmer:
### Description

A number of tests are intermittently failing due https://github.com/visgl/deck.gl/commit/6c7eab32ff818df212930f18b4936e50070dca19

To get CI working again, I have marked the tests to be skipped for now:

 https://github.com/visgl/deck.gl/blob/master/test/modules/geo-layers/tile-3d-layer/tile-3d-layer.spec.js#L26-L27

### Flavors

- [ ] React
- [ ] Python/Jupyter notebook
- [ ] MapboxLayer
- [ ] GoogleMapsOverlay
- [ ] CartoLayer
- [ ] DeckLayer/DeckRenderer for ArcGIS

### Steps to Reproduce

Failing test https://github.com/visgl/deck.gl/runs/6226836907?check_suite_focus=true

Tests pass after reverting 6c7eab32ff818df212930f18b4936e50070dca19: https://github.com/visgl/deck.gl/runs/6227065856?check_suite_focus=true

### Environment

- Framework version: latest
- Browser: all
- OS:

PR #6882, opened 2022-05-04
fix(geo-layers): unskip fixed test

belom88:
For #6871


kylebarron (approved):
Remove the comment as well?

Issue #6912, opened 2022-05-17
[Feat] CartoLayer support for spatial indices


felixpalmer:
### Target Use Case

The [recent changes to TileLayer](https://github.com/visgl/deck.gl/pull/6848) make it possible to use data that has its geography encoded using a spatial index, rather than with latitude/longitude. 

The upcoming API in the CARTO platform will support tiled data using quadkeys and h3 indices.

### Proposal

`CartoLayer` is extended to support these datasets:

```js
return new CartoLayer({
    // Usual data props
    connection,
    data,

    // Aggregation configuration
    aggregationExp: 'avg(population) as value, 0.1*avg(population) as elevation',
    aggregationResLevel: 6,
    geoColumn: 'geom',
    getQuadkey: d => d.id,

    // Styling
    getFillColor: d => [255 * d.properties.value, 0, 0],
    getElevation: d => d.properties.elevation,
    extruded: true,
    opacity: 0.3,
    elevationScale: 100
  })
```

As the spatially indexed datasets are aggregated, the following props are added:

- `aggregationExp` a SQL expression to specify how to aggregate
- `aggregationResLevel` the resolution at which to aggregate

These parameters are passed directly to the API and are opaque to the client code.

felixpalmer:
## Checklist

- [ ] Implement quadkey support
- [ ] Implement h3 support
- [ ] Tests
- [ ] Doc update
- [ ] QA

alasarr:
After creating a couple of maps I've just realized that we shouldn't require the usage of  `getQuadkey: d => d.id`. As the format is always like that I'd like to remove that complexity for the user.

PR #6915, opened 2022-05-18
CARTO: Support H3TileLayer in CartoLayer

felixpalmer:
## Change List
   - Implementation of H3TileLayer
   - Instantatiate H3TileLayer when 'h3' scheme is detected

### DIFF
--- a/modules/carto/src/layers/h3-tile-layer.js
+++ b/modules/carto/src/layers/h3-tile-layer.js
@@ -33,6 +33,7 @@ export default class H3TileLayer extends CompositeLayer {
       new TileLayer(this.props, {
         id: 'h3-tile-layer',
         data,
+        getHexagon: d => d.id,
         TilesetClass: H3Tileset2D,
         renderSubLayers
       })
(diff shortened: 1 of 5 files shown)


alasarr (changes requested):
I think overzooming is not implemented

modules/carto/src/layers/h3-tile-layer.js
```diff
@@ -0,0 +1,43 @@
...
+import {H3HexagonLayer, TileLayer} from '@deck.gl/geo-layers';
+import H3Tileset2D from './h3-tileset-2d';
+
+const renderSubLayers = props => {
+  const {data} = props;
+  const {index} = props.tile;
+  if (!data || !data.length) return null;
+
+  return new H3HexagonLayer(props, {
+    centerHexagon: index,
```
alasarr, line 11: Would be great to have a defaultProp to avoid the user to write at CartoLayer:
`getHexagon: d => d.id`

alasarr (approved):

PR #6928, opened 2022-05-19
CARTO: Fix autohighlighting in spatial index tile layers

felixpalmer:

#### Change List
- Map spatial index to data index for autoHighlighting

### DIFF
--- a/modules/carto/src/layers/spatial-index-tile-layer.js
+++ b/modules/carto/src/layers/spatial-index-tile-layer.js
@@ -13,12 +13,13 @@ export default class SpatialIndexTileLayer extends TileLayer {
     if (hoveredFeature) {
       newHoveredFeatureId = hoveredFeature.id;
     }
-    let {highlightColor} = this.props;
-    if (typeof highlightColor === 'function') {
-      highlightColor = highlightColor(info);
-    }
 
     if (hoveredFeatureId !== newHoveredFeatureId) {
+      let {highlightColor} = this.props;
+      if (typeof highlightColor === 'function') {
+        highlightColor = highlightColor(info);
+      }
+
       this.setState({
         highlightColor,
         hoveredFeatureId: newHoveredFeatureId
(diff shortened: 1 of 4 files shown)


modules/carto/src/layers/spatial-index-tile-layer.js
```diff
@@ -0,0 +1,47 @@
+import {TileLayer} from '@deck.gl/geo-layers';
```
Pessimistress, line 1: Write new classes in TypeScript?

Pessimistress (approved):

modules/carto/src/layers/spatial-index-tile-layer.js
felixpalmer: As the other files were in JS in this directory I stuck with that, but the plan is to migrate this next week

alasarr: Our plan is to do it right after this PR. I know in the future we should develop directly in typescript, but we were in hurry for this to do QA in other pieces of our platform.

Added to the TODO of spatial indexes: https://github.com/visgl/deck.gl/issues/6912#issuecomment-1129682952 to make sure we don't forget about it

modules/carto/src/layers/spatial-index-tile-layer.js
```diff
@@ -0,0 +1,47 @@
...
+  _updateAutoHighlight(info) {
+    const {hoveredFeatureId} = this.state;
+    const hoveredFeature = info.object;
+    let newHoveredFeatureId;
+
+    if (hoveredFeature) {
+      newHoveredFeatureId = hoveredFeature.id;
+    }
+    let {highlightColor} = this.props;
+    if (typeof highlightColor === 'function') {
```
alasarr, line 17: I feel we can move this piece of code inside of the next if `(hoveredFeatureId !== newHoveredFeatureId)` to save some performance

alasarr (approved):
Looks good to me, just a small perf thing

PR #6937, opened 2022-05-20
Carto: fix spatial index tileLayer id

padawannn:
#### Change List
- Read the SpatialIndexTileLayer id from props


modules/carto/src/layers/quadkey-tile-layer.js
```diff
@@ -32,7 +32,7 @@ export default class QuadkeyTileLayer extends CompositeLayer {
     const maxZoom = parseInt(tileJSON.maxresolution);
     return [
       new SpatialIndexTileLayer(this.props, {
-        id: 'quadkey-tile-layer',
+        id: `quadkey-tile-layer_${this.props.id}`,
```
felixpalmer, line 35:
```suggestion
        id: `quadkey-tile-layer-${this.props.id}`,
```

felixpalmer (changes requested):

felixpalmer (approved):

PR #6940, opened 2022-05-20
[ts] GeoJsonLayer

zbigg:

Change List:

* Convert GeoJsonLayer to Typescript

Prerequisite for #6842


modules/layers/src/geojson-layer/types.ts
```diff
@@ -0,0 +1,100 @@
...
+   *
+   * @default false
+   */
+  lineBillboard?: boolean;
+
+  /** @deprecated */
+  getRadius?: Accessor<DataT, number>;
+} & FillProps<DataT> &
+  StrokeProps<DataT> &
+  Polygon3DProps<DataT> &
```
Pessimistress, line 39: I would recommend you define all props explicitly here, just like how props are explicitly forwarded when GeoJsonLayer renders its sub-layers. Sharing prop types among layers will introduce the risk of declaring props that the layer does not actually support, and there is no way for the compiler to catch it.

modules/layers/src/polygon-layer/polygon-layer.ts
```diff
@@ -19,198 +19,56 @@
...
+     * Specifies the winding order of rings in the polygon data.
+     *
+     * **Note**: This prop is experimental
+     *
+     * @default 'CW'
+     */
+    _windingOrder?: 'CW' | 'CCW';
+
+    /** @deprecated */
+    lineDashJustified?: boolean;
```
Pessimistress, line 70: Just drop the deprecated props to encourage following the latest API

modules/layers/src/geojson-layer/types.ts
zbigg: Ok, just wanted to avoid repetition.

OTOH, there is a way to ensure that Layer explicitly supports all properties defined in interface: properly typed defaultProps. Something like:
```ts
// some-layer.ts
type PropTypeDefsFor<PropT> = Record<keyof PropT, PropTypeDef>>;

const defaultProps = PropTypeDefsFor<_LayerProperties> = {
   // you're enforced to define _every_ property defined in _LayerProperties
   // of course layer implementation can ignore some props, but has to be explicit at least in 
   // one in layer source code
}
```
But that's probably story for other PR ...

modules/layers/src/polygon-layer/polygon-layer.ts
zbigg: Well, i would discuss. I think that typing something as `@deprecated` is enough signal to discourage use of this feature.

My understanding of deprecation is feature _still_ works, but is planned to be removed or discouraged.

Implementation still supports this prop (brief look, correct me if i'm wrong) so i would rather have it explicitly typed as deprecated that removed.

zbigg: Second thing, is that impl _uses_ this like this:

const {
     ...
      lineDashJustified
    } = this.props;
    
and to remove this prop from `_PolygonLayerProps` i would need to add unsafe workarounds (like `as any` or @ts-expect-error) or some very explit typings.

Pessimistress: `lineDashJustified` no longer works after we move the dash styling from `PathLayer` to `PathStyleExtension`. I think these are code left over from the refactor.

modules/layers/src/geojson-layer/geojson-layer.ts
```diff
@@ -58,24 +312,36 @@ const defaultProps = {
...
+  static defaultProps = defaultProps;
+
+  initializeState(): void {
     this.state = {
       layerProps: {},
       features: {}
     };
 
-    if (this.props.getLineDashArray) {
+    if ((this.props as any).getLineDashArray) {
```
Pessimistress, line 333: We can drop this check. The prop was removed in v8.0

modules/layers/src/types.ts
```diff
@@ -0,0 +1,6 @@
+export type MaterialProps = {
+  ambient: number;
+  diffuse: number;
+  shininess: number;
+  specularColor: [r: number, g: number, b: number];
```
Pessimistress, line 5: `@babel/preset-typescript` does not support named tuples and this breaks local debugging with the examples.

Pessimistress (approved):

PR #6946, opened 2022-05-23
Add optional types entry point

Pessimistress:
For #5589 

This PR creates the following structure in the published packages:

```
@deck.gl/*
  package.json
- dist
  + es5
  + esm
+ src
- typed
    package.json
``` 

Here is the content of `typed/package.json`:

```json
{
  "internal": true,
  "main": "../dist/es5/index.js",
  "module": "../dist/esm/index.js",
  "types": "index.d.ts"
}
```

When installed as a dependency, apps can import it as an untyped library, consistent with the behavior of v8.7:

```js
```

Or as a fully typed library:

```js
import {Deck} from '@deck.gl/core/typed';
```

What is working right now:
- Build succeeds
- Importing from `*/typed` endpoint gets type suggestions and checks

What is NOT working:
- Not all submodules are being compiled (see root `tsconfig.build.json`) due to TS errors.
- If one tries to build an application with the `*/typed` imports, lib check fails due to the not-yet-typed classes.

#### Change List
- Output ts declarations to /typed directory
- Script to properly expose types for publish

#### Todo
- We should continue to type the remaining code base and test them as we go.
- Documentation


zbigg (approved):
Works well with demo/test app.

PR #6947, opened 2022-05-24
[ts] Deck class

Pessimistress:
For #5589 

#### Change List
- Deck class methods and props
- Update type references in GoogleMapsOverlay


modules/google-maps/src/google-maps-overlay.ts
```diff
@@ -26,7 +28,10 @@ const defaultProps = {
 };
 
 // TODO: this should be alias to Deck Props.
-export type GoogleMapsOverlayProps = Record<string, any>;
```
zbigg, line 29: Minor, TODO is no longer needed

zbigg (approved):
Looks good, minor comments only.

PR #6957, opened 2022-05-27
[ts] mapbox module

Pessimistress:
For #5589 

#### Change List
- Convert the mapbox modue to TypeScript

### DIFF
--- a/modules/mapbox/src/deck-utils.ts
+++ b/modules/mapbox/src/deck-utils.ts
@@ -1,22 +1,40 @@
 import {Deck, WebMercatorViewport, MapView} from '@deck.gl/core';
+import type {DeckProps, MapViewState, Layer} from '@deck.gl/core';
+import type MapboxLayer from './mapbox-layer';
+import type {Map} from 'mapbox-gl';
 
-export function getDeckInstance({map, gl, deck}) {
+type UserData = {
+  isExternal: boolean;
+  currentViewport?: WebMercatorViewport | null;
+  mapboxLayers: Set<MapboxLayer<any>>;
+  mapboxVersion: {minor: number; major: number};
+};
+
+export function getDeckInstance({
+  map,
+  gl,
+  deck
+}: {
+  map: Map;
+  gl: WebGLRenderingContext;
+  deck?: Deck;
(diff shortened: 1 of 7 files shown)


modules/mapbox/src/deck-utils.ts
```diff
@@ -1,22 +1,40 @@
...
+  mapboxLayers: Set<MapboxLayer<any>>;
+  mapboxVersion: {minor: number; major: number};
+};
+
+export function getDeckInstance({
+  map,
+  gl,
+  deck
+}: {
+  map: Map;
```
felixpalmer, line 18:
```suggestion
  map: Map & {__deck?: Deck | null};
```

The annotation is repeated in many places, I would say it is better to extend the `Map` type.

modules/mapbox/src/deck-utils.ts
```diff
@@ -158,13 +180,18 @@ function getViewport(deck, map, useMapboxProjection = true) {
...
 
   if (isExternal) {
     // Draw non-Mapbox layers
     const mapboxLayerIds = Array.from(mapboxLayers, layer => layer.id);
-    const hasNonMapboxLayers = deck.props.layers.some(layer => !mapboxLayerIds.includes(layer.id));
+    const hasNonMapboxLayers = deck.props.layers.some(
+      layer =>
+        layer &&
+        // @ts-ignore (2339) id does not exist on Layer[]
+        !mapboxLayerIds.includes(layer.id)
```
felixpalmer, line 193: Will this not fail if `layers` has nested layers? Instead of the ts-ignore could we flatten the layers?

modules/mapbox/src/mapbox-overlay.ts
```diff
@@ -105,7 +113,7 @@ export default class MapboxOverlay implements IControl {
     }
   }
 
-  _updateContainerSize = () => {
+  private _updateContainerSize = () => {
```
felixpalmer, line 116: Worth being consistent with method syntax?
```suggestion
  private _updateContainerSize() {
```

felixpalmer (approved):

PR #6968, opened 2022-06-02
chore: upgrade @loaders.gl version

belom88:
#### Background
@loaders.gl@3.2.0-alpha.4 has bug in `tiles` module that is included in the script version of @deck.gl.
Use latest version of @loaders.gl in the script version of @deck.gl
#### Change List
- upgrade `@loaders.gl` version in `yarn.lock`


zbigg (commented):
Shouldn't we upgrade respective entries in `package.json`s too?

Pessimistress (approved):

belom88:
> Shouldn't we upgrade respective entries in `package.json`s too?

Major version is the same. Agreed with @ibgreen doing this way

Issue #6967, opened 2022-06-02
8.8 Release Tracker


felixpalmer:
## PRs/work remaining

- [x] Typescript migration, remaining modules:
  - [x] layers. Owner: @zbigg 
  - [x] core. Owner: @Pessimistress 
Publish first BETA
  - [x] aggregation-layers. Owner: @zbigg 
  - [x] carto. Owner: @zbigg  
  - [x] geo-layers. Owner: @felixpalmer 
  - [x] main. Owner: @zbigg 
  - [x] extensions. Owner: @Pessimistress 
- [x] [CartoLayer support for spatial indices](https://github.com/visgl/deck.gl/issues/6912)

## Release checklist

Target release date 28/06

- [x] Publish alpha/beta
- [x] Bump all examples to latest alpha/beta
- [x] Stage the website
- [x] Verify whats-new
- [x] Verify upgrade-guide
- [x] Verify all issues in milestone are closed or move out
- [x] Verify all API changes are documented
- [ ] Website (docs and example) testing: https://pessimistress.github.io/deck.gl @AnaDiaz85
    - [x] Testing on Chrome
    - [x] Testing on Safari
    - [x] Testing on Firefox
    - [x] Testing on Edge
    - [x] Testing on iOS
    - [x] Testing on Android
- [x] Cut 8.8-release branch after all fixes
- [x] Update website doc urls to point to `8.8-release` branch
- [ ] Publish 8.8 prod version.

AnaDiaz85:
Hi @Pessimistress , I am Ana from QA. Just to let you know that I am testing all the examples and the documentation in the website with the new deck.gl version. Ping me if you need clarification on any of the issues I have reported. Thank you

PR #6971, opened 2022-06-02
[ts]: Carto spatial layers

zbigg:
For https://github.com/visgl/deck.gl/issues/5589 and #6967

Change List:

* carto: convert `SpatialIndexTileLayer`, `H3TileLayer`, `QuadkeyTileLayer` to typescript
* enable carto & geo-layers in tsbuild


zbigg:
@felixpalmer @Pessimistress 

Now this PR shows problem with exposing Tileset2d as it was originally designed for `{x,y,z}` tile indices.

Without making `TileIndex`, `TileHeader`, `Tileset2D` generic over "TileIndexT" we can't have properly typed "customizations". 
We either have to use "any" or skip errors reported by compiler as this new code doesn't work on `{x,y,z}` but on something completly different.

Nevetheless, i would like to resolve this problem in other PR, so we don't mix different types of work.

Pessimistress:
I find it helpful to add a definition for layer state in each layer:

```js
state!: {
  model?: Model;
  ...
}
```

tsconfig.build.json
```diff
@@ -8,10 +8,10 @@
   "references": [
     // {"path": "modules/aggregation-layers"},
     // {"path": "modules/arcgis"},
-    // {"path": "modules/carto"},
+    {"path": "modules/carto"},
     {"path": "modules/core"},
     {"path": "modules/extensions"},
-    // {"path": "modules/geo-layers"},
+    {"path": "modules/geo-layers"},
```
felixpalmer, line 14: Are you sure you want to be uncommenting geo-layers in this PR?

felixpalmer (changes requested):

modules/carto/package.json
```diff
@@ -43,6 +43,7 @@
     "d3-color": "^2.0.0",
     "d3-format": "^2.0.0",
     "d3-scale": "^3.2.3",
+    "h3-js": "^3.6.0",
```
alasarr, line 46: I think adding h3 as a full dependency will make the bundle considerable bigger.

felixpalmer: I think this will prevent it from being bundled, like is currently the case in `geo-layers`

https://github.com/visgl/deck.gl/blob/master/scripts/bundle.config.js#L18-L21

tsconfig.build.json
zbigg: `geo-layers` are ready to be part of tsbuild, it's agreed with @Pessimistress

modules/carto/package.json
zbigg: `dependencies` in `package.json` doesn't affect bundle size, it's webpack config and as @felixpalmer pointed out, h3-js is already configured as external dependency.

This patch only "sanitizes" dependencies as _we_ truly depend h3-js and it's good practice not to rely on transitive dependencies (from geo-layers in this case), also ... eslint complains about using deps not specified in package.json.

modules/carto/src/layers/carto-layer.ts
```diff
@@ -276,7 +282,7 @@ export default class CartoLayer<DataT = any> extends CompositeLayer<CartoLayerPr
     const {uniqueIdProperty} = defaultProps;
     // eslint-disable-next-line @typescript-eslint/no-unused-vars
     const {data: _notUsed, ...propsNoData} = this.props;
-    const props = {uniqueIdProperty, ...propsNoData};
+    const props = {...propsNoData, uniqueIdProperty};
```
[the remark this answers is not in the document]
zbigg: Great catch, fixing, sorry that it even landed in review.

modules/carto/src/api/maps-v3-client.ts
```diff
@@ -242,7 +242,7 @@ function checkFetchLayerDataParameters({
 export interface FetchLayerDataResult {
   data: any;
   format?: Format;
-  schema: SchemaField[];
+  schema?: SchemaField[];
```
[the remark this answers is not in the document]
zbigg: Accidental change, reverted

modules/carto/src/layers/carto-layer.ts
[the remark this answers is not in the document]
felixpalmer: I agree that this is surprising. In order to land this PR I would propose you create a new issue with the above for visibility and we can discuss further there

felixpalmer (approved):

PR #6973, opened 2022-06-02
[ts] Tesselators

Pessimistress:
For #5589 

#### Change List
- Tesselator
- PathTesselator
- PolygonTesselator


modules/core/src/utils/tesselator.ts
```diff
@@ -169,28 +213,36 @@ export default class Tesselator {
...
+          vertexStarts[dataIndex + 1] =
+            vertexStarts[dataIndex] +
+            (normalizedGeometry ? this.getGeometrySize(normalizedGeometry) : 0);
         },
         startRow,
         endRow
       );
       // count instances
       instanceCount = vertexStarts[vertexStarts.length - 1];
-    } else if (geometryBuffer.buffer instanceof Buffer) {
```
zbigg, line 182: Again, using `isBinaryAttribute()` type guard (to be written) would make it more readable and help saving original, less nested control flow.

zbigg (approved):
LGTM, only minor comments.

Issue #6970, opened 2022-06-02
[Bug] DeckGL component doesn't sync view


felixpalmer:
### Description

When the user drags the map, only the deck layers move, the basemap remains static. 

https://user-images.githubusercontent.com/453755/171666795-e150a72a-a4f9-4e83-8216-72a5ed80b47d.mov

With the core module forced to 8.8.0-alpha.4 it works. The cause is that the `_customRender` function is not being invoked. It is only called on initial render, but never after

### Flavors

- [X] React
- [ ] Python/Jupyter notebook
- [ ] MapboxLayer
- [ ] GoogleMapsOverlay
- [ ] CartoLayer
- [ ] DeckLayer/DeckRenderer for ArcGIS

### Steps to Reproduce

Unfortunately don't have a public repro scenario, can work on that if needed

### Environment

- Framework version: core/8.8.0-alpha.5 (works in core/8.8.0-alpha.4)
- Browser: Chrome/Firefox
- OS: OSX

PR #6977, opened 2022-06-03
Fix regression in DeckGL React component not syncing base map

felixpalmer:
For https://github.com/visgl/deck.gl/issues/6970

#### Background

https://github.com/visgl/deck.gl/pull/6947/files#diff-d41f5114076a8be933e2409af8aa332e45ecb5e146ac71c18375229995f8ba31 introduced a change in the `Deck` class which gives the `_customRender` prop a default value of null, whereas previously there was no default value. This caused the `setProps` call in the `DeckGL` component

https://github.com/visgl/deck.gl/blob/master/modules/react/src/deckgl.js#L152

to overwrite the `_customRender` prop with null, breaking the render cycle. 

@Pessimistress I assume that defining a default value for `_customRender` was intentional, and thus the correct fix here make sure we always decoratively pass in the `_customRender` prop rather than relying on the state of `Deck.props` leaking through when `setProps` is called.

#### Change List
- Move definition of `_customRender` to ensure it is always set


modules/react/src/deckgl.js
```diff
@@ -142,6 +123,25 @@ const DeckGL = forwardRef((props, ref) => {
       style: null,
       width: '100%',
       height: '100%',
+      // The Deck's animation loop is independent from React's render cycle, causing potential
+      // synchronization issues. We provide this custom render function to make sure that React
+      // and Deck update on the same schedule.
+      _customRender: redrawReason => {
```
Pessimistress, line 129: This will be called frequently and recreating a new method all the time. Maybe just `delete forwardProps._customRender` and document why?

Pessimistress (approved):

PR #6984, opened 2022-06-04
[ts] Shader modules

Pessimistress:
For #5589 

#### Change List
- Core shaderlib


modules/core/src/effects/lighting/camera-light.ts
```diff
@@ -7,7 +7,6 @@ export default class CameraLight extends PointLight {
     const {projectedLight} = this;
     const viewport = layer.context.viewport;
     const {coordinateSystem, coordinateOrigin, modelMatrix} = layer.props;
-    // @ts-expect-error
     const {project_uCameraPosition} = getUniformsFromViewport({
```
felixpalmer, line 10: Is it worth configuring eslint to allow some non-camelCase names? https://eslint.org/docs/rules/camelcase support a regex based allow-list

zbigg (approved):

Issue #6985, opened 2022-06-04
[Bug] Heatmap Layer example on website does not render heatmap on iOS mobile safari 15


puopg:
### Description

Im noticing that on the documentation for the heatmap layer here: https://deck.gl/docs/api-reference/aggregation-layers/heatmap-layer.

The heatmap does not render on my iPhone running.

iOS 15.4.1
iPhone 8

The example works however on:

iOS 14.7.1
iPhone 7

### Flavors

- [ ] React
- [ ] Python/Jupyter notebook
- [ ] MapboxLayer
- [ ] GoogleMapsOverlay
- [ ] CartoLayer
- [ ] DeckLayer/DeckRenderer for ArcGIS

### Expected Behavior

I would expect the heatmap to render something, even given some WebGL limitations on mobile safari.

### Steps to Reproduce

Visit https://deck.gl/docs/api-reference/aggregation-layers/heatmap-layer on a similar iPhone device.

### Environment

iOS 15.4.1
iPhone 8

### Logs
![143C0E13-4AD3-4D15-9C41-8BCFEB6CD104](https://user-images.githubusercontent.com/7741481/172020126-a1851d72-2218-4900-ac7b-6b7c7ebb1316.png)

_No response_

cjwang18:
I'm noticing the same with an iPhone 11 running iOS 15.5. The example page shows up just like the screenshot above.

puopg:
Update:

I tried turning off both webgl 2.0 and webgl via Metal on safari and afterwards reloading safari, the heat map shows up.

Does this point to an issue around Safari’s implementation of webgl 2?

puopg:
Some more info:

Tested passing in a WebGL 1.0 context to the deck instance also fails, however disabling "WebGL via Metal" on Safari works. I guess it's not really any different than the above since its not trying to use webgl 2.0.

Pessimistress:
Fixed in 8.7.11.

PR #6988, opened 2022-06-06
Fix HeatmapLayer on iOS

Pessimistress:
For #6985

See https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API/WebGL_best_practices#dont_assume_you_can_render_into_float_textures

This was not a problem before because iOS used to fail the `isWebGL2` and/or `COLOR_ATTACHMENT_RGBA32F` checks.

#### Change List
- Check `EXT_float_blend` for float blending support


puopg:
@Pessimistress Thank you!

kylebarron (approved):

PR #6990, opened 2022-06-07
CARTO: Add tests for spatial index layers

felixpalmer:
Tests for https://github.com/visgl/deck.gl/issues/6912


test/modules/carto/layers/h3-tile-layer.spec.js
```diff
@@ -0,0 +1,92 @@
...
+      props: {
+        data: TILEJSON,
+        getTileData: () => []
+      },
+      assert: t.ok,
+      onAfterUpdate({layer, subLayers}) {
+        if (!layer.isLoaded) {
+          t.equal(subLayers.length, 1, 'Rendered sublayers');
+          t.deepEqual(subLayers[0].props.data, TILES, 'Extract tiles from tilejson');
+          t.deepEqual(subLayers[0].props.maxZoom, 10, 'Extract maxZoom from tilejson');
```
alasarr, line 37: I'm confused here, in h3 maxZoom should be the zoom that matches with the resolution 10.

alasarr: Just realized that backend already does the conversion

alasarr (approved):

test/modules/carto/layers/h3-tile-layer.spec.js
felixpalmer: Yes it is a little messy with the variable names, as the base `Tileset2D` class uses `maxZoom` - even if it treats it as an abstract "resolution" parameter

PR #6997, opened 2022-06-08
[ts] TripsLayer

felixpalmer:

#### Change List
- TripsLayer


modules/geo-layers/src/trips-layer/trips-layer.ts
```diff
@@ -27,7 +28,39 @@ const defaultProps = {
...
+  trailLength?: number;
+  /**
+   * The current time of the frame.
+   * @default 0
+   */
+  currentTime?: number;
+  /**
+   * Timestamp accessor.
+   */
+  getTimestamps?: AccessorFunction<DataT, number>;
```
Pessimistress, line 54: Should return `NumericArray`

Pessimistress (approved):

zbigg (approved):

PR #6998, opened 2022-06-08
[ts] _MeshLayer

felixpalmer:

#### Change List
- _MeshLayer converted to TS


modules/geo-layers/src/mesh-layer/mesh-layer.ts
```diff
@@ -14,12 +22,33 @@ function validateGeometryAttributes(attributes) {
...
   pbrMaterial: {type: 'object', value: null},
   featureIds: {type: 'array', value: null, optional: true}
 };
 
-export default class _MeshLayer extends SimpleMeshLayer {
+/** All properties supported by _MeshLayer. */
+export type _MeshLayerProps<DataT = any> = __MeshLayerProps<DataT> & SimpleMeshLayerProps<DataT>;
+
+/** Properties added by _MeshLayer. */
+type __MeshLayerProps<DataT = any> = {
```
Pessimistress, line 33: Too many underscores... this class is not exposed to the user, we could just name stuff normally in this file.

Pessimistress (approved):

PR #6999, opened 2022-06-08
[ts] S2Layer

felixpalmer:
For #5589

#### Change List
- S2Layer & utils


modules/geo-layers/src/s2-layer/s2-utils.ts
```diff
@@ -54,7 +62,7 @@ function getGeoBounds({face, ij, level}) {
       offset[1] += stepJ;
       // Cell can be represented by coordinates IJ, ST, UV, XYZ
       // http://s2geometry.io/devguide/s2cell_hierarchy#coordinate-systems
-      const st = IJToST(ij, level, offset);
+      const st = IJToST(ij, level, offset) as [number, number];
```
Pessimistress, line 65: Just type the function's return type as `[number, number]`?

Pessimistress (approved):

Pessimistress:
S2LayerProps should be exported by the root index.ts

PR #7005, opened 2022-06-08
[ts] ArcLayer

zbigg:
For https://github.com/visgl/deck.gl/issues/5589 & https://github.com/visgl/deck.gl/issues/6967

Change List
* ArcLayer converted to Typescript


modules/layers/src/arc-layer/arc-layer.ts
```diff
@@ -140,8 +240,8 @@ export default class ArcLayer extends Layer {
       .draw();
   }
 
-  _getModel(gl) {
-    let positions = [];
+  private _getModel(gl: WebGLRenderingContext): Model {
```
Pessimistress, line 243: `protected` to unblock subclassing

Pessimistress (approved):

Issue #6989, opened 2022-06-07
[Feat] [pydeck] Update for JupyterLab 3


dakoop:
### Target Use Case

First, there is an issue with the jupyter-widget module in JupyterLab 3 (#5615).

Second, using a [prebuilt extension](https://jupyterlab.readthedocs.io/en/stable/extension/extension_dev.html#prebuilt-extensions) in JupyterLab 3 would allow a more standard installation for pydeck. A user could just run `pip install pydeck` to install the JupyterLab extension and not require a rebuild.

### Proposal

I think I was able to make the update for JupyterLab 3 here by following the advice in the [ipywidgets CHANGELOG](https://ipywidgets.readthedocs.io/en/latest/changelog.html#updates-for-widget-maintainers):

<https://github.com/dakoop/deck.gl/tree/pydeck-8.7>

Then, I tried to create a prebuilt extension. Since jupyter-builder uses webpack 5, and pydeck uses 4, there were some headaches, including something that seems to be related to the acorn package. pydeck specifies a resolution for acorn to 7.4.1, and `jupyter labextension build` just hangs. If I remove the resolution (which I think allows acorn versions to be different for different packages), the build suceeds. This should allow a `python -m build` in the bindings/pydeck directory, which produces a wheel that can be used for pip. That branch is available here:

<https://github.com/dakoop/deck.gl/tree/pydeck-8.7-prebuilt>

dakoop:
#7011 contains the ipywidgets fixes. Want to review the prebuilt code but will plan to submit a PR for that too.

PR #7011, opened 2022-06-09
[pydeck] Update jupyter-widget to work with JupyterLab 3

dakoop:
For #6989

#### Change List
- Make changes for JupyterLab 3 as suggested by [ipywidgets](https://ipywidgets.readthedocs.io/en/stable/changelog.html#updates-for-widget-maintainers)
- Bump the linked version of deck.gl to 8.7
- Update pyproject.toml to add header

#### Notes
- Not sure if this should be based on the 8.7 release branch or master. Should be easy to change.
- This does not create a prebuilt extension (see <https://github.com/dakoop/deck.gl/tree/pydeck-8.7-prebuilt> for work on that)


modules/jupyter-widget/package.json
```diff
@@ -36,14 +36,20 @@
     "@deck.gl/json": "8.7.11",
     "@deck.gl/layers": "8.7.11",
     "@deck.gl/mesh-layers": "8.7.11",
-    "@jupyter-widgets/base": "^1.1.10 || ^2 || ^3",
+    "@jupyter-widgets/base": "^2 || ^3 || ^4",
```
kylebarron, line 39: Is there a reason to remove `^1`? Or, is it known which version of jupyter notebook started to move to `^2`? Is this suggested by one of the current dev references?

kylebarron (approved):

kylebarron:
> * Not sure if this should be based on the 8.7 release branch or master. Should be easy to change

@Pessimistress I would think this should go into master; thoughts?

ajduberstein:
What's the case for not merging this into master?

@dakoop before merging let me test that this works locally.

kylebarron:
> What's the case for not merging this into master?

I think it would better to be merged into master and then cherry picked onto the 8.7 release branch. (`@deck.gl/jupyter-widget` would need another release for these changes to take effect in published pydeck)

dakoop:
Re: branches, I wasn't sure how the version updates happen. Since pydeck ties itself to a particular deck.gl release in `pydeck/frontend_semver.py` (currently `"~8.5.*"`), wasn't sure which version to set if it goes to master and figured this was clear on `8.7-release`. The changes should be the same on any recent branch.

Pessimistress:
This should go into master and then cherry-picked to the 8.7-release branch. There shouldn't be any conflict.

modules/jupyter-widget/package.json
dakoop: The [webpage](https://ipywidgets.readthedocs.io/en/stable/changelog.html#updates-for-widget-maintainers) says

> Update the @jupyter-widgets/base dependency version to include ^4 to work in JupyterLab 3.0. For example, if you had a dependency on @jupyter-widgets/base version ^2 || ^3, update to ^2 || ^3 || ^4 for your widget to work in classic Jupyter Notebook, JupyterLab 1, JupyterLab 2, and JupyterLab 3. See [#2472](https://github.com/jupyter-widgets/ipywidgets/pull/2472) for background.

However, it seems like keeping ^1.1.10 would be fine.

kylebarron: I'd prefer to keep `^1.1.10` as well unless there's a conflicgt.

kylebarron:
@Pessimistress did you mean to move the base branch back to `8.7-release`?

Pessimistress:
The current PR is done on top of 8.7-release. The release branch cannot be merged  cleanly into master.

kylebarron:
@dakoop Can you make a PR for this off of the `master` branch?

dakoop:
Yes, will do

dakoop:
Ok, I recall more about why I went with 8.7. `frontend_semver.py` is not automatically kept in sync with the current jupyter-widget version. This means that if I check in a change to master which ties pydeck to, e.g., 8.8.0-alpha.6, this reference breaks the next time jupyter-widget's version gets bumped. More in #7025

dakoop:
Replaced by #7026

PR #7016, opened 2022-06-09
[ts] react module

Pessimistress:
For #5589


kylebarron (approved):
Seems fine at a glance but I really don't know the react module, or react itself, well enough to give a great review of this.

modules/react/src/deckgl.ts
```diff
@@ -187,8 +223,7 @@ const DeckGL = forwardRef((props, ref) => {
 
   useImperativeHandle(ref, () => getRefHandles(thisRef), []);
 
-  const {viewManager} = thisRef.deck || {};
-  const currentViewports = viewManager && viewManager.getViewports();
```
felixpalmer, line 191: @Pessimistress This change has introduced a regression which causes React apps to crash on initialization. It happens because there is no guarantee that `viewManager` is initialized when `thisRef.deck` is. The simplest fix would be to add back the old check, but this would require exposing `viewManager` as public.

felixpalmer: Fix without exposing `viewManager`: https://github.com/visgl/deck.gl/pull/7049

Issue #7025, opened 2022-06-10
[Feat] [pydeck] Update frontend_semver.py when jupyter-widget version changes


dakoop:
### Target Use Case

There are methods to bump versions across the various package.json files in deck.gl. However, pydeck's `frontend_semver.py` is not updated when these versions change. This may be because pydeck releases are not synchronized with deck.gl releases, but I could not find a documented way to run pydeck with the jupyter-widget module when building/running from source because the semver link between pydeck and jupyter-widget is broken, leading to errors:
```
Unhandled Promise Rejection: Error: Module @deck.gl/jupyter-widget, semver range ~8.5.* is not registered as a widget module
```
There is a `bump_version.py` script that pulls the version info from deck.gl and updates `frontend_semver.py`, but this is lumped in with setting the pydeck release (one has to choose major/minor, etc.)

### Proposal

I see three potential solutions:
 1.  Modify the `bump_version.py` script so it can be called to only run `rewrite_frontend_version_file` and put this step into the Makefile
 2. Create a separate script (e.g. `link_frontend.py`) that calls `rewrite_frontend_version_file` and document this in the installing from source docs
 3. Automatically update the `frontend_semver.py` file when deck.gl versions change

Note that these could be combined so that the `link_frontend.py` script is created, used in the Makefile, and also called as part of the global version update action.

PR #7026, opened 2022-06-10
[pydeck] Update jupyter-widget to work with JupyterLab 3

dakoop:
For #6989

#### Change List

- Make changes for JupyterLab 3 as suggested by [ipywidgets](https://ipywidgets.readthedocs.io/en/stable/changelog.html#updates-for-widget-maintainers)
- Bump the linked version of deck.gl (in frontend_semver.py) to 8.8.0-alpha.6 (see #7025)

#### Notes

 - base is master
 - Replaces #7011


bindings/pydeck/pydeck/frontend_semver.py
```diff
@@ -1 +1 @@
-DECKGL_SEMVER = "~8.5.*"
+DECKGL_SEMVER = '8.8.0-alpha.6'
```
kylebarron, line 1: @ajduberstein can confirm, but I thought we usually had this file point to the current non-alpha release, which in this case would be
```suggestion
DECKGL_SEMVER = '8.7.*'
```

bindings/pydeck/pydeck/_version.py
```diff
@@ -1 +1 @@
-__version__ = "0.7.1"
+__version__ = '0.7.2'
```
kylebarron, line 1: This doesn't look like correct `black` formatting. You should run `make pre-commit` in the local dir: https://github.com/visgl/deck.gl/blob/e1b18d42245922714bf6752ff1819f3279cbc39a/bindings/pydeck/Makefile#L27-L28

kylebarron: @ajduberstein we should probably add `black` to CI; it doesn't look like it's currently run on CI

bindings/pydeck/pydeck/frontend_semver.py
dakoop: If I do that, I see:
```
Unhandled Promise Rejection: Error: Module @deck.gl/jupyter-widget, semver range 8.7.* is not registered as a widget module
```

bindings/pydeck/pydeck/_version.py
dakoop: Yep, there are a number of files that are not consistent with black currently. `_version.py` and `frontend_semver.py` were generated by the `bump_version.py` script.

bindings/pydeck/pydeck/frontend_semver.py
dakoop: Set this to `8.7.*` as suggested, but I believe anyone running from source would need to update this to be compatible with the jupyter-widget version (see #7025).

kylebarron: Hmm. I think you're right that on the development branch, this has to match the development version of `@deck.gl/jupyter-widget`. So I think we need either:

- This should leave as-is or change to `8.8.*` if the star works with alpha versions.
- Cherry-pick these changes to the 8.7-release branch, but changing this to `8.7.*` there
- Publish the next version of pydeck from that branch, so that it uses the currently-released version of deck

Note that since `@deck.gl/jupyter-widgets` is a JS packaged published at the same time as every other JS package, this would require a patch release of the JS packages as well. cc @Pessimistress

dakoop: Right, forgot to mention that `~8.8.*` doesn't work either as alphas are strictly less than `8.8.0`. That was the reason I initially moved the work to the `8.7-release` branch.

kylebarron: Makes sense. At least this way we have the code on the master branch for the future, and we can cherry pick for a release now.

kylebarron (approved):

davidbrochart:
Looks like this only needs a rebase?

Jesus89:
Hi @ajduberstein we are preparing the next pydeck release using deck.gl 8.8. I think we can merge this one and include it in the next release too. Could I go ahead, solve the conflicts and merge? Thanks.

kylebarron:
I'm 👍 on merging

PR #7027, opened 2022-06-10
Add interleaved prop to MapboxOverlay

Pessimistress:
For #6773

#### Change List
- Support both overlaid and interleaved modes in `MapboxOverlay`
- `View` class: do not overwrite `padding` in view state unless `padding` is defined on the view
- `MapboxLayer` and `MapboxOverlay` support `padding` of the base map

### DIFF
--- a/modules/mapbox/src/mapbox-overlay.ts
+++ b/modules/mapbox/src/mapbox-overlay.ts
@@ -5,27 +5,49 @@ import type {Map, IControl, MapMouseEvent} from 'mapbox-gl';
 import type {MjolnirGestureEvent, MjolnirPointerEvent} from 'mjolnir.js';
 import type {DeckProps} from '@deck.gl/core';
 
+import {resolveLayers} from './resolve-layers';
+
+export type MapboxOverlayProps = Omit<
+  DeckProps,
+  | 'width'
+  | 'height'
+  | 'gl'
+  | 'parent'
+  | 'canvas'
+  | '_customRender'
+  | 'viewState'
+  | 'initialViewState'
+  | 'controller'
+>;
+
 /**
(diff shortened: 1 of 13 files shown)


felixpalmer (approved):

modules/mapbox/src/mapbox-overlay.ts
```diff
@@ -5,28 +5,50 @@ import type {Map, IControl, MapMouseEvent} from 'mapbox-gl';
...
+  private _interleaved: boolean;
 
-  constructor(props: DeckProps) {
-    this._props = {...props};
+  constructor(
+    props: MapboxOverlayProps & {
+      interleaved?: boolean;
+    }
+  ) {
+    const {interleaved = false, ...otherProps} = props;
```
felixpalmer, line 39: The default for `GoogleMapsOverlay` is `interleaved=true`. Would it not be better to be consistent?

chrisgervang: Is google's WebGL2? I thought one reason we didn't do it by default was mapbox is WebGL, which for some cases could be limiting (though I don't notice which cases, so this could be an old assumption).

Are we positioning MapboxOverlay as the de-facto way of using mapboxgl-js (or react-map-gl) with deck now? If yes, I'm hesitant to default interleaved rendering for the reason stated above. If no, then default interleaved would make sense to me since the user is opting in to a tight mapbox ecosystem integration.

Pessimistress: Mapbox's interleaving implementation is less reliable than Google Map's. mapbox-gl's WebGLContext is WebGL1 and there are known issues of z-fighting between native layers and deck layers. Since we are promoting this solution I think it's safer to have the option off for now. We can revisit the decision in the next release, after getting some feedback from users.

PR #7032, opened 2022-06-12
Update defaultProps typings (PR 2/2)

Pessimistress:
For #5589 

#### Change List
- Update defaultProps typing for all layers
- Fix mistyped and missing props


zbigg (approved):
LGTM, thanks for clearing inconsistencies and mistakes i've introduced.

PR #7035, opened 2022-06-13
Improve performance for `visible: false` layers

juneidysoo:
#### Background
I have a map with 600+ `visible: false` tile layers which resulted in almost 20k of sublayers where the visibility is `false`.

The performance is very slow due to some checks in the deck lifecycle.

These changes have shown 20-40% performance improvements for each render frame.

#### Change List
- Prioritise checking visibility flag first
- Do not calculate render index if layer is not drawn

Edit: This is related to discussion #7021

### DIFF
--- a/modules/core/src/effects/mask/mask-effect.ts
+++ b/modules/core/src/effects/mask/mask-effect.ts
@@ -55,7 +55,7 @@ export default class MaskEffect implements Effect {
       });
     }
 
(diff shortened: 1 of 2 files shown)


Pessimistress (approved):
Nice. How much improvement do you get in your use case?

juneidysoo:
> How much improvement do you get in your use case?

From 75-110ms to 50-70ms a frame. Not the best, but I will add additional optimisation in my own code.

PR #7049, opened 2022-06-15
Do not access viewports before viewManager initialization

felixpalmer:
Fix for regression introduced in https://github.com/visgl/deck.gl/pull/7016 

#### Background

https://github.com/visgl/deck.gl/pull/7016#discussion_r897819653

#### Change List
- Do not access `getViewports` before `viewManager` is initialized

### DIFF
--- a/modules/react/src/deckgl.ts
+++ b/modules/react/src/deckgl.ts
@@ -195,11 +195,18 @@ const DeckGL = forwardRef<DeckGLRef, DeckGLProps>((props, ref) => {
     return forwardProps;
   }, [props]);
 
+  const [isWebGLInitialized, setIsWebGLInitialized] = useState(false);
+
   useEffect(() => {
     thisRef.deck = createDeckInstance(thisRef, {
       ...deckProps,
       parent: containerRef.current,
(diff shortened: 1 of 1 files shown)


modules/react/src/deckgl.ts
```diff
@@ -195,11 +195,18 @@ const DeckGL = forwardRef<DeckGLRef, DeckGLProps>((props, ref) => {
...
 
+  const [isWebGLInitialized, setIsWebGLInitialized] = useState(false);
+
   useEffect(() => {
     thisRef.deck = createDeckInstance(thisRef, {
       ...deckProps,
       parent: containerRef.current,
-      canvas: canvasRef.current
+      canvas: canvasRef.current,
+      onWebGLInitialized: () => {
```
Pessimistress, line 205: Use `onLoad`?

Pessimistress:
An alternative is to introduce a public `isLoaded` flag to the Deck class. We have been using now-protected properties (layerManager/viewManager) to check its state in a few other integration modules.

Pessimistress (approved):

PR #7065, opened 2022-06-20
CARTO: fetchMap support spatial index layers

felixpalmer:

#### Change List
- Unify layer selection based on tileset definition between `CartoLayer` & `fetchMap`
- Generate spatial index tile layers in fetchMap
- Map aggregation parameters from layer style
- Support stying for dynamic tiling layers via the stats API
- Tests


padawannn (changes requested):
@felixpalmer I think that you're not building the `aggregationExp` from the layer style? Have you tried to style a map by a property in builder and check if it works with the fetchMap?

felixpalmer:
> `aggregationExp`

@padawannn No, you're right - that is missing. Could you point me at where that is done in builder? I can't seem to find the code

alasarr (dismissed):

modules/carto/src/api/layer-map.ts
```diff
@@ -163,17 +165,36 @@ export function getLayer(
   };
 }
 
+export function layerFromTileDataset(
```
alasarr, line 168: 👍

alasarr:
>@padawannn No, you're right - that is missing. Could you point me at where that is done in builder? I can't seem to find the code

You can try at dedicated server 05, click on layer and fill by color based on a property

padawannn (changes requested):

padawannn (approved):

Issue #6993, opened 2022-06-07
[Bug] Google Map in VECTOR rendering mode gets distorted when GoogleMapOverlay gets set.


puopg:
### Description

When using `useDevicePixels: false` in the config passed to the `<GoogleMapsOverlay>`, the map gets distorted and anchor positions of markers no longer look correct.

Seems to happen with the vector rendered google map, once the overlay is added to the map calling setMap.

Setting `useDevicePixels: true` (or unset) seems to resolve this.

Uploading Screen Recording 2022-06-07 at 2.14.58 PM.mov…

### Flavors

- [X] React
- [ ] Python/Jupyter notebook
- [ ] MapboxLayer
- [X] GoogleMapsOverlay
- [ ] CartoLayer
- [ ] DeckLayer/DeckRenderer for ArcGIS

### Expected Behavior

I guess I would expect no distortion and markers are still correctly positioned.

https://user-images.githubusercontent.com/7741481/172492079-db2a7bff-f1af-412c-ac8f-dd09ef9d66f6.mov

### Steps to Reproduce

This one is tough to show a codepen for since it requires a google map id to use the vector based rendering.

The setup though is essentially the following. This will need to run on iOS Safari.

```
import { GoogleMap, LoadScriptNext } from "@react-google-maps/api";

const mapIdToUse = "INSERT_VECTOR_MAP_ID_HERE";

<LoadScriptNext
       googleMapsApiKey="INSERT_API_KEY"
       mapIds={[mapIdToUse]}
       libraries={[]}
       version="quarterly"
>
      <GoogleMap
         options={{ mapId: mapIdToUse }
         onLoad={(map) => { this.setState({ googleMapInstance: map }}
         onIdle={() => { this.setState({ mapReady: true }); this.attachDeckOverlay(this.state.googleMapInstance) }}
      />
</LoadScriptNext>
```

Then:
```
attachDeckOverlay = (mapInstance) => {
  const deckOverlay = new DeckOverlay({ useDevicePixels: false });
  deckOverlay.setMap(mapInstance);
}
```

### Environment

- Framework version: 8.7.11
- Browser: Desktop Chrome
- OS: OSX

Pessimistress:
@felixpalmer I think there is indeed a bug. When using external gl context, framebuffer resizing should be disabled by setting width/height to null. `useDevicePixels` value would not matter if you do so.

https://github.com/visgl/deck.gl/blob/master/modules/google-maps/src/google-maps-overlay.ts#L157

PR #7066, opened 2022-06-21
GoogleMapsOverlay: Force useDevicePixels to true in interleaved mode

felixpalmer:
For https://github.com/visgl/deck.gl/issues/6993

#### Change List
- Override `useDevicePixels` in `GoogleMapsOverlay` (as is done in `MapboxOverlay`)
- Tests

### DIFF
--- a/modules/google-maps/src/utils.ts
+++ b/modules/google-maps/src/utils.ts
@@ -35,6 +35,7 @@ export function createDeckInstance(
 
   const newDeck = new Deck({
     ...props,
+    useDevicePixels: props.interleaved ? true : props.useDevicePixels,
     style: props.interleaved ? null : {pointerEvents: 'none'},
     parent: getContainer(overlay, props.style),
     initialViewState: {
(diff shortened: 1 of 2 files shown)


alasarr (approved):

felixpalmer:
> See [#6993 (comment)](https://github.com/visgl/deck.gl/issues/6993#issuecomment-1161463370) - this prop should not matter in the first place

@Pessimistress  The issue is that the `useDevicePixels` prop is passed onto luma.gl which uses it to automatically resize the canvas:

https://github.com/visgl/luma.gl/blob/98097c6047710837713efa0efccb527af3bbbede/modules/engine/src/lib/animation-loop.ts#L482-L486

The `MapboxOverlay` is doing exactly the same thing (hardcoding `useDevicePixels: true`):

https://github.com/visgl/deck.gl/blob/master/modules/mapbox/src/deck-utils.ts#L30

Pessimistress (approved):

Pessimistress:
I think the proper fix would be preventing `AnimationLoop` from auto resizing an externally supplied GL context. However, considering that we are very close to the next release, this PR is a less risky patch for the immediate issue.

Issue #7081, opened 2022-06-24
[Bug] Website Examples: Integrations ArcGIS does not work well


AnaDiaz85:
### Description

I can’t visualize properly the integration of ArcGIS.

It happens in all browsers, with the lastest version:

<img width="795" alt="Screenshot 2022-06-24 at 10 10 33" src="https://user-images.githubusercontent.com/49390946/175492952-075bd7db-72d8-4b8b-aef4-54d54f800e55.png">

Check this video:

[deck_bug_integration_arcgis.mov.zip](https://github.com/visgl/deck.gl/files/8974554/deck_bug_integration_arcgis.mov.zip)

### Flavors

- [ ] React
- [ ] Python/Jupyter notebook
- [ ] MapboxLayer
- [ ] GoogleMapsOverlay
- [ ] CartoLayer
- [X] DeckLayer/DeckRenderer for ArcGIS

### Expected Behavior

You should be able to visualize the example correctly

### Steps to Reproduce

1. Go to https://pessimistress.github.io/deck.gl/examples/arcgis/
2. Move around the map, do zoom in/out

### Environment

- Framework version: Latest version (check image in the description)
- Browser: All browsers
- OS: macOS Monterey version 12.4

Pessimistress:
If you are testing the staging website for #6967 please mark framework version as `8.8.0-beta`. `latest` refers to the production release.

AnaDiaz85:
> If you are testing the staging website for #6967 please mark framework version as `8.8.0-beta`. `latest` refers to the production release.

Ok, thank you

PR #7092, opened 2022-06-25
Fix arcgis integration

Pessimistress:
For #7081

This is a regression due to a change in `redraw` from

```js
const redrawReason = reason || this.needsRedraw({clearRedrawFlags: true});
```

to 

```js
const redrawReason = this.needsRedraw({clearRedrawFlags: true}) || reason;
```

The intention is to always clear the dirty flags when redraw is forced by the user. However, this changed the `reason` received by `_customRender`, which the ArcGIS integration is relying on.

#### Change List
- `redraw` always invoke `_customRender` with user-supplied reason, if any

### DIFF
--- a/modules/core/src/lib/deck.ts
+++ b/modules/core/src/lib/deck.ts
@@ -520,7 +520,9 @@ export default class Deck {
       return;
     }
     // Check if we need to redraw
-    const redrawReason = this.needsRedraw({clearRedrawFlags: true}) || reason;
+    let redrawReason = this.needsRedraw({clearRedrawFlags: true});
+    // User-supplied should take precedent, however the redraw flags get cleared regardless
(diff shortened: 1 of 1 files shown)


Gaurav-Bhole (approved):

felixpalmer (approved):

kylebarron (approved):

ibgreen (approved):
No code should rely on the returned reason string. It should be considered a boolean (we just return a string for debugging purposes). 

This sounds like there is a hack somewhere in arcgis code? 

Nit: I suppose there is some modest perf implication with this change, as unnecessary checking may now be done in certain cases (at the same it could be considered better for debugging purposes as the more redreaw underlying cause is reported).

PR #7095, opened 2022-06-27
CARTO: Add QuadbinTileLayer

felixpalmer:

#### Background

As an optimization the CARTO backend is switching to a different data format for the Quadkey. This PR implements that change. 

#### Change List
- Add `QuadbinLayer`, `QuadbinTileLayer` and `QuadbinTileset2D` classes
- Modify `CartoLayer` to use `QuadbinTileLayer` rather than `QuadkeyTileLayer`
- Update `carto-dynamic-tile` test app
- Tests


modules/carto/src/layers/carto-layer.ts
```diff
@@ -304,7 +304,7 @@ export default class CartoLayer<ExtraProps = {}> extends CompositeLayer<
       return [H3TileLayer, props];
     }
     if (data.scheme === 'quadkey') {
```
alasarr, line 306: Backend should return the scheme quadbin

modules/carto/src/layers/quadbin-tile-layer.ts
```diff
@@ -0,0 +1,73 @@
...
+  DefaultProps
+} from '@deck.gl/core';
+import {QuadkeyLayer, QuadkeyLayerProps} from '@deck.gl/geo-layers';
+import QuadbinTileset2D, {quadbinToTile, tileToQuadkey} from './quadbin-tileset-2d';
+import SpatialIndexTileLayer from './spatial-index-tile-layer';
+
+const renderSubLayers = props => {
+  const {data} = props;
+  if (!data || !data.length) return null;
+  return new QuadkeyLayer(props, {
```
alasarr, line 16: I'm fine keeping the QuadkeyLayer in a first implementation, but as you suggested in the PR description we've room for optimization here

alasarr: The main drawback is that the user will receive a Quadkey and they're working with Quadbin, so agree would be better to have the QuadBinLayer and also save some conversions

alasarr (commented):
Looks good! Pending things for me, similar to what you pointed out:
- get rid of quadkey layer in favor of quadbin.
- remove all references to quadkey: quadkey-tile-layer and quadkey-tileset-2d

modules/carto/src/layers/carto-layer.ts
charly22: Roger, I'll change it to `quadbin` only when `geo_column` is also `quadbin`

felixpalmer: I've temporarily added support for both

modules/carto/src/layers/carto-layer.ts
```diff
@@ -303,8 +303,8 @@ export default class CartoLayer<ExtraProps = {}> extends CompositeLayer<
     if (data.scheme === 'h3') {
       return [H3TileLayer, props];
     }
-    if (data.scheme === 'quadkey') {
-      return [QuadkeyTileLayer, props];
+    if (data.scheme === 'quadbin' || data.scheme === 'quadkey' /* TODO remove quadkey */) {
```
alasarr, line 306: How about remove quadkey before merging?

modules/carto/src/index.ts
```diff
@@ -2,7 +2,8 @@ export {getDefaultCredentials, setDefaultCredentials} from './config';
 export {default as CartoLayer} from './layers/carto-layer';
 export {default as _CartoTileLayer} from './layers/carto-tile-layer';
 export {default as _H3TileLayer} from './layers/h3-tile-layer';
-export {default as _QuadkeyTileLayer} from './layers/quadkey-tile-layer';
+export {default as _QuadbinLayer} from './layers/quadbin-layer';
```
alasarr, line 5: Do we really need to export _QuadbinLayer?

alasarr (approved):

modules/carto/src/index.ts
felixpalmer: @padawannn needs it in Builder I think

modules/carto/src/layers/carto-layer.ts
felixpalmer: Yes, agreed. Waiting on deploy from @charly22

charly22: Already deployed

modules/carto/src/index.ts
padawannn: @alasarr I need it for builder as the rest of the layer exported

alasarr: I mean @padawannn needs _QuadbinTileLayer not _QuadbinLayer

padawannn: Sorry I really need the `_QuadbinTileLayer`

PR #7106, opened 2022-06-29
CARTO: Remove CARTO 2 for documentation

felixpalmer:
For https://github.com/visgl/deck.gl/issues/6967

#### Change List
- Remove mention of CARTO 2

### DIFF
--- a/docs/api-reference/carto/carto-layer.md
+++ b/docs/api-reference/carto/carto-layer.md
@@ -8,7 +8,10 @@
 import DeckGL from '@deck.gl/react';
 import {CartoLayer, setDefaultCredentials, MAP_TYPES, API_VERSIONS} from '@deck.gl/carto';
 
-setDefaultCredentials({accessToken: 'XXX'});
+setDefaultCredentials({
+  accessToken: 'XXX'
+  apiBaseUrl: 'https://gcp-us-east1.api.carto.com' // Default value (optional)
+});
 
 function App({viewState}) {
   const layer = new CartoLayer({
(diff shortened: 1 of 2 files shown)


docs/api-reference/carto/carto-layer.md
```diff
@@ -2,17 +2,13 @@
...
 ```js
 import {CartoLayer, setDefaultCredentials, MAP_TYPES, API_VERSIONS} from '@deck.gl/carto';
 
-setDefaultCredentials({
-  apiVersion: API_VERSIONS.V3
-  apiBaseUrl: 'https://gcp-us-east1.api.carto.com',
-  accessToken: 'XXX',
-});
+setDefaultCredentials({accessToken: 'XXX'});
```
borja-munoz, line 11: It might be useful to keep at least the apiBaseUrl prop because customers with self-hosted or SaaS in a different instance will need to change it.

docs/api-reference/carto/carto-layer.md
```diff
@@ -107,23 +77,21 @@ Required. Data type. Possible values are:
 
 ##### `connection` (String)
 
-Required when `apiVersion` is `API_VERSIONS.V3`.
-
-Name of the connection registered in the CARTO workspace.
+Required. Name of the connection registered in the CARTO workspace.
```
borja-munoz, line 80: Not sure about this. This property is optional. Someone can come here and believe they need to use it for CARTO 2. I agree with removing specific things but this is the official docs and we shouldn't say a prop is mandatory if it is not.

docs/api-reference/carto/carto-layer.md
felixpalmer: OK, I've added it back in marking as optional

borja-munoz (approved):
Looks good to me

PR #7108, opened 2022-06-29
CARTO: Handle minZoom&maxZoom in H3TileLayer

felixpalmer:

#### Change List
- Use `minZoom`&`maxZoom` props in addition to `min/maxresolution` when limiting H3 tile requests

### DIFF
--- a/modules/carto/src/layers/h3-tile-layer.ts
+++ b/modules/carto/src/layers/h3-tile-layer.ts
@@ -67,7 +67,6 @@ export default class H3TileLayer<DataT = any, ExtraPropsT = {}> extends Composit
     const {data, tileJSON} = this.state;
     let minresolution = parseInt(tileJSON.minresolution);
     let maxresolution = parseInt(tileJSON.maxresolution);
-    console.log(minresolution, maxresolution);
 
     // Convert Mercator zooms provided in props into H3 res levels
     // and clip into valid range provided from the tilejson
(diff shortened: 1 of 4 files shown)


modules/carto/src/layers/h3-tile-layer.ts
```diff
@@ -65,15 +65,35 @@ export default class H3TileLayer<DataT = any, ExtraPropsT = {}> extends Composit
...
+    // of a Tileset2D refer to the resolution levels, not the Mercator zooms
     return [
       new SpatialIndexTileLayer(this.props, {
         id: `h3-tile-layer-${this.props.id}`,
         data,
         // @ts-expect-error Tileset2D should be generic over TileIndex
         TilesetClass: H3Tileset2D,
         renderSubLayers,
-        maxZoom
+        minZoom: minresolution,
```
alasarr, line 95:
```suggestion
        // minZoom and maxZoom are in h3 resolution, they will be converted to XYZ zoom levels by H3Tileset2D
        minZoom: minresolution,
```

alasarr (approved):
LGTM, just added a commit suggestion with a comment to make it easier to understand

modules/carto/src/layers/h3-tile-layer.ts
felixpalmer: That's not quite right, H3Tileset2D will treat them as H3 resolutions, it is just the base Tileset2D class is hardcoded to treat zoom and resolution as equal

felixpalmer, line 95:
```suggestion
        // minZoom and maxZoom are H3 resolutions, however we must use this naming as that is what the Tileset2D class expects
        minZoom: minresolution,
```

alasarr: ok, good to know :)

Issue #7064, opened 2022-06-20
Mapbox/Maplibre elevation compatibility


Baseform:
### Description

We're using interleaved terrain rendering with Deck and Mapbox/Maplibre. 
There is a mismatch of elevation on renders - probably due to camera positioning between Deck and the others

### Flavors

- [ ] React
- [ ] Python/Jupyter notebook
- [ ] MapboxLayer
- [ ] GoogleMapsOverlay
- [ ] CartoLayer
- [ ] DeckLayer/DeckRenderer for ArcGIS

### Expected Behavior

I'd expect to see an overlap. WebMercatorViewport docs mention compatibility of default parameters

### Steps to Reproduce

This codepen allows for toggling between both deck and the new Maplibre Terrain 3D versions.
We've also tested with Mapbox with the same results.

[https://codepen.io/diogovit/full/wvyNjwW]

### Environment

- Framework version: deck.gl@8.7.11 (latest)
- Browser: any
- OS: any

Pessimistress:
Can you help me understand how this is going to be used? deck.gl does not support offseting/overlaying layers over a base terrain (it's something we are exploring, but not yet implemented), so even if the camera movement matches that of mapbox/maplibre, the data positions may not be very useful.

With that said, the mapbox module (`MapboxLayer`, `MapboxOverlay`) should definitely make an effort to match the camera matrices. I do not think they support terrain in the base map right now.

On a high level, deck.gl's TerrainLayer works just like other layers - it renders in a self-contained context and does not affect the camera control or any other layers. I agree that this is something we should address, probably by introducing a new [operation](https://github.com/visgl/deck.gl/blob/3a593b3e9e1f40fef477d5a5b0197cfdf4484496/modules/core/src/types/layer-props.ts#L127). Whether deck's default behavior should be identical to that of Mapbox's is open for discussion.

@kylebarron @chrisgervang

Baseform:
@Pessimistress 

We have developed a rich visualization solution for water utility network at [baseform.com](url). This includes mapping geojson/mvt infrastructure (pipes, buildings, valves and others), as well as visualizations of hydraulics (similar to trips layer) and aggregation analytic quantities, from population, to consumer usage, to water quality and water flows. The non-deck based solution is presented here: [https://baseform.com/np4/papps0.html](url) or [https://baseform.com/np4/papps1.html](url)
OSM based vector maps as an alternative to satellite imagery are available as background.

[...]

PR #7114, opened 2022-07-01
[mapbox] Match base map camera when terrain is used

Pessimistress:
For #7064 

Note that this only matches the camera matrices between deck and the base map, so that positions are projected correctly. Any data with z=0 in deck layers will still be rendered at sea level and not align with the terrain surface.

#### Change List
- Upgrade math.gl dependency to include https://github.com/uber-web/math.gl/pull/295
- Calculate `viewState.position` if terrain is used

### DIFF
--- a/modules/mapbox/src/deck-utils.ts
+++ b/modules/mapbox/src/deck-utils.ts
@@ -3,12 +3,18 @@ import type {DeckProps, MapViewState, Layer} from '@deck.gl/core';
 import type MapboxLayer from './mapbox-layer';
 import type {Map} from 'mapbox-gl';
 
+import {lngLatToWorld, unitsPerMeter} from '@math.gl/web-mercator';
+
 type UserData = {
   isExternal: boolean;
   currentViewport?: WebMercatorViewport | null;
   mapboxLayers: Set<MapboxLayer<any>>;
   mapboxVersion: {minor: number; major: number};
 };
+
+// Mercator constants
+const TILE_SIZE = 512;
+const DEGREES_TO_RADIANS = Math.PI / 180;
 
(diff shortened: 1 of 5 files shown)


modules/mapbox/src/deck-utils.ts
```diff
@@ -140,6 +155,53 @@ export function getViewState(map: Map): MapViewState & {
...
+    const cameraToCenterDistanceGround = Math.sqrt(dx * dx + dy * dy);
+
+    const pitchRadians = pitch! * DEGREES_TO_RADIANS;
+    const altitudePixels = 1.5 * height;
+    const scale = (altitudePixels * Math.sin(pitchRadians)) / cameraToCenterDistanceGround;
+    viewState.zoom = Math.log2(scale);
+
+    const cameraZFromSurface = (altitudePixels * Math.cos(pitchRadians)) / scale;
+    const surfaceElevation = cameraZ - cameraZFromSurface;
+    viewState.position = [0, 0, surfaceElevation / unitsPerMeter(latitude)];
```
chrisgervang, line 197: Just trying to get confirmation on my understanding:
So, Deck's MapView will now have support for offsetting camera position from it's geolocation anchor point through `position`.

We're accessing mapbox free camera to calculate a scale value used to determine viewState `position.z` and `zoom` so that the camera can be moved as it traverses over terrain.

If that's the case, that's pretty great. The new position property is useful beyond terrain for viewing a variety of 3D datasets.

Pessimistress: Position should work in production already.

chrisgervang (approved):

PR #7120, opened 2022-07-04
[ts] Improve carto/style typings.

zbigg:

#### Change List
- Improve `carto/style` typings for `color*` functions - improve typings and and use types from `@deck.gl/core`


Clebal (approved):
LGTM

Issue #6923, opened 2022-05-19
[Feat] Use `pydeck` without pulling in `jupyter/ipython` dependencies


anders-kiaer:
### Target Use Case

We use `pydeck` as a backend utility library for transforming input to [`dash-deck`](https://github.com/plotly/dash-deck).

The Dash app (with `pydeck`) is deployed to cloud web apps. Indirectly installing `jupyter/ipython` dependencies (through `pydeck`) is adding extra installation time to `pip install pydeck`, and also adding extra code to app (= larger attack surface).

It will also be relevant for users following in this category mentioned in `README.md`: https://github.com/visgl/deck.gl/blob/edad7cf75b75a137d63ec89d2d644727d3082e25/bindings/pydeck/README.md?plain=1#L57-L61

### Proposal

Moving `jupyter/ipython` related dependencies to `extra` section (https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies). Jupyter/ipython users will probably also have `jupyter/ipython` related dependencies installed prior to installing `pydeck` (or install them separately).

ajduberstein:
Thank you for the suggestion @anders-kiaer. I've put a PR. The change will go out in the next beta of pydeck.

chrisgervang:
This appears to have been released for a while. We've just released pydeck 0.9.2 which transitioned build systems to `uv`. The migration maintained the optional dependencies, so this should continue to work as before. Feel free to reopen if you run into any new issues.

PR #7125, opened 2022-07-05
pydeck: Make Jupyter widget into a Python extra / optional module

ajduberstein:
For #6923

Right now pydeck ships with all the dependencies for the Jupyter extension, which are quite large.

`pip install pydeck` will now install only the portion of pydeck necessary for `.to_html` to work
`pip install pydeck[jupyter]` will install the dependencies for pydeck to work with Jupyter.

#### Background

See #6923

#### Change List
-


bindings/pydeck/setup.py
```diff
@@ -208,13 +208,14 @@ def run(self):
         ],
         extras_require={
             "carto": ["pydeck_carto"],
-            "testing": ["pytest"],
+            "jupyter": [
+                "ipywidgets>=7.0.0",
+                'ipython>=5.8.0;python_version<"3.4"',
+                "traitlets>=4.3.2",
+                'ipykernel>=5.1.2;python_version>="3.4"',
```
kylebarron, line 215: Which of these do we actually depend on? I would've thought we only should require ipywidgets and traitlets, and others would be installed as part of the user's jupyter environment

kylebarron (approved):

kylebarron:
Did you check that this works in an environment without ipywidgets and traitlets?

bindings/pydeck/setup.py
ajduberstein: You're correct, I've cut traitlets and ipwidgets. The `.to_html()` function still works.

kylebarron: Wait what? I was saying the opposite. We directly depend on `ipywidgets` and `traitlets`... so you certainly need those

kylebarron (changes requested):
Sorry my previous review was on my phone. This doesn't work because you'll get an import error if you try to import pydeck without the jupyter deps because we have jupyter deps at the top level.

```
make setup-env
make init
./env/bin/pip uninstall ipykernel traitlets ipywidgets
```

```
> ./env/bin/python
Python 3.8.10 (default, Jul 28 2021, 14:21:59)
[Clang 12.0.5 (clang-1205.0.22.11)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import pydeck
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/kbarron/github/mapping/deck.gl/bindings/pydeck/pydeck/__init__.py", line 7, in <module>
    from .bindings import map_styles, Deck, Layer, LightSettings, View, ViewState  # noqa
  File "/Users/kbarron/github/mapping/deck.gl/bindings/pydeck/pydeck/bindings/__init__.py", line 1, in <module>
    from .deck import Deck  # noqa
  File "/Users/kbarron/github/mapping/deck.gl/bindings/pydeck/pydeck/bindings/deck.py", line 7, in <module>
    from ..widget import DeckGLWidget
  File "/Users/kbarron/github/mapping/deck.gl/bindings/pydeck/pydeck/widget/__init__.py", line 1, in <module>
    from .widget import DeckGLWidget  # noqa
  File "/Users/kbarron/github/mapping/deck.gl/bindings/pydeck/pydeck/widget/widget.py", line 4, in <module>
    from ipywidgets import register, CallbackDispatcher, DOMWidget
ModuleNotFoundError: No module named 'ipywidgets'
```

bindings/pydeck/requirements/requirements.txt
```diff
@@ -1,5 +1,3 @@
 ipykernel>=5.1.2;python_version>="3.4"
-ipywidgets>=7.0.0
-traitlets>=4.3.2
```
kylebarron, line 3: Removing this from `requirements.txt` does nothing for a local (and I assume CI) build of pydeck because `make init` also installs `requirements-dev.txt` and we depend on the `jupyter` metapackage there:
https://github.com/visgl/deck.gl/blob/e275028f7907a14bfca021d5c65de3bd93b79f5c/bindings/pydeck/requirements/requirements-dev.txt#L1

bindings/pydeck/pydeck/bindings/deck.py
```diff
@@ -76,20 +84,26 @@ def __init__(
...
 
         self.description = description
         self.effects = effects
-
         self.map_provider = str(map_provider).lower() if map_provider else None
-        self.deck_widget.map_provider = map_provider
+        self._tooltip = tooltip
+
+        if has_jupyter_extra():
+            from ..widget import DeckGLWidget
```
kylebarron, line 96: As noted above, instead of re-importing again and again, I think it would be clearer (and maybe faster?) to try to import `DeckGLWidget` at the top level, and otherwise set it to `None`

```py
try:
    from ..widget import DeckGLWidget
except ImportError:
    DeckGLWidget = None
```

Then just check
```py
if DeckGLWidget:
    self.deck_widget = DeckGLWidget()
    ...
```

bindings/pydeck/setup.py
```diff
@@ -209,6 +209,8 @@ def run(self):
         extras_require={
             "carto": ["pydeck_carto"],
             "jupyter": [
+                "ipywidgets>=7.0.0",
+                "traitlets>=4.3.2",
                 'ipython>=5.8.0;python_version<"3.4"',
                 'ipykernel>=5.1.2;python_version>="3.4"',
```
kylebarron, line 215: It looks like the only place where we use `ipykernel` is in the tests:
https://github.com/visgl/deck.gl/blob/1049956d723fe7a92c39ea203bc2b5c5fe00cc36/bindings/pydeck/tests/conftest.py#L9

And the only place where we use `ipython` outside of tests is in
https://github.com/visgl/deck.gl/blob/017d960d95160c456204f781dceac41eafd49579/bindings/pydeck/pydeck/io/html.py#L102
and there we have handling so it works whether or not `ipython` exists, right? So do we need to have a runtime dep on that?

kylebarron:
It would be ideal to have a CI check that tests that `pydeck` can be imported without any jupyter dependencies to protect against unintentional regressions in the future that cause it to fail to import.
This would have to be done in a different environment as the rest of the CI testing though, as the existing CI env has all the jupyter deps.

bindings/pydeck/tests/conftest.py
```diff
@@ -1,58 +0,0 @@
...
-# Copyright (c) Jupyter Labs Team
-# Distributed under the terms of the Modified BSD License.
-
-import pytest
-
-from ipykernel.comm import Comm
-from ipywidgets import Widget
-
-
-class MockComm(Comm):
```
kylebarron, line 13: Were we not using this?

bindings/pydeck/tests/io/test_html.py
```diff
@@ -1,8 +1,6 @@
 import html
 import webbrowser
 
-import IPython
```
kylebarron, line 4: We still always have access to `IPython` in tests, right?

kylebarron (approved):
The review comments above are still relevant, but not worth blocking a merge

bindings/pydeck/tests/conftest.py
ajduberstein: We were not

bindings/pydeck/tests/io/test_html.py
ajduberstein: That's correct, it's still available through the dev requirements installed during testing.

Jesus89 (approved):
LGTM

I have a couple of minor improvements:

1. The line in docs `You can run the local tests to verify that the installation worked via ``make test``.` should be located in **Development** instead of **Installing from source**
2. The line in example 2 `from IPython.core.display import display` should be changed to `from IPython.display import display` to remove the warning

@ajduberstein do you want me to go ahead with the release process of 0.8.0b1?

Issue #7143, opened 2022-07-16
[Bug] mapboxApiAccessToken still in the docs


ShaneLee:
### Description

mapboxApiAccessToken has been replaced in favour of mapboxAccessToken

### Flavors

- [X] React
- [ ] Python/Jupyter notebook
- [ ] MapboxLayer
- [ ] GoogleMapsOverlay
- [ ] CartoLayer
- [ ] DeckLayer/DeckRenderer for ArcGIS

### Steps to Reproduce

Just documentation

### Environment

- Framework version: 8.8.4
- Browser: N/A
- OS: N/A

Pessimistress:
There is no mistake in the prop names used in this repo. Please do not mix the API of deck.gl and react-map-gl.

ShaneLee:
Ah okay - I was just confused by this page (https://deck.gl/docs/get-started/using-with-react) which has mapboxApiAccessToken but the examples don't work until it's swapped for mapboxAccessToken.

Pessimistress:
The prop name for access token changed between react-map-gl v6 and v7. You are welcome to open a PR to clarify that in the documentation. The PR you opened was trying to replace API names in the deck.gl source code.

PR #7144, opened 2022-07-16
Replace mapboxApiAccessToken with mapAccessToken

ShaneLee:
For #7143
#### Change List
- Replace mapboxApiAccessToken with mapAccessToken


ShaneLee:
Published as opposed to leaving as draft until the checks are run as I need someone to approve running the checks.

PR #7145, opened 2022-07-16
Replace mapboxApiAccessToken with mapAccessToken

ShaneLee:
For https://github.com/visgl/deck.gl/issues/7143

Change List
- Replace mapboxApiAccessToken with mapAccessToken in the getting started with react documentation.


Pessimistress:
`StaticMap` only exists in react-map-gl v5 and v6, where the token prop is called `mapboxApiAccessToken`. If you wish to use v7 the component is called `Map`.

Pessimistress (approved):
Thank you!

PR #7160, opened 2022-07-21
CARTO: binary data for spatial index layers

felixpalmer:
I have a PoC working with binary data for spatial index datasets (H3 & Quadbin).

I have implemented a pair of conversion functions `spatialjsonToBinary` and `binaryToSpatialjson` [here](https://github.com/visgl/deck.gl/blob/1ed9d66c7486c0c5c55d2586e142d5d441cdfef9/modules/carto/src/layers/schema/spatialjson-utils.ts). These take the existing format and convert it to a format ready to serialized using protocol buffers (as we do for binary data in geospatial tiles).

While the frontend uses the `pbf` library for parsing, I see that on the backend `protobufjs` is being used to convert the data to a buffer, so I have defined a `.proto` file which [describes the schema](https://github.com/visgl/deck.gl/blob/1ed9d66c7486c0c5c55d2586e142d5d441cdfef9/modules/carto/src/layers/schema/carto-spatial-tile.proto) and created a [test](https://github.com/visgl/deck.gl/blob/1ed9d66c7486c0c5c55d2586e142d5d441cdfef9/test/modules/carto/layers/spatialjson.spec.js) which writes and reads using `protobufjs` and the `.proto` file. Unfortunately this library doesn’t handle BigInts automatically, so it is necessary to convert the data before/after the serialization, [see here](https://github.com/visgl/deck.gl/blob/1ed9d66c7486c0c5c55d2586e142d5d441cdfef9/modules/carto/src/layers/schema/carto-spatial-tile.ts#L88-L118). 

If you pull my branch and try running the `test/apps/carto-map` example all the spatial index layers will be using the conversion methods. Currently it is [just hacked in](https://github.com/visgl/deck.gl/blob/1ed9d66c7486c0c5c55d2586e142d5d441cdfef9/modules/carto/src/layers/h3-tile-layer.ts#L34-L53), by taking the data from the server and converting it to a protobuf buffer and back before displaying it.

As a next step I would suggest the backend team try and take my functions and implementing serialization using the `.proto` schema on the backend. Like with geospatial datasets, the backend should only return data in binary format when `TILE_FORMAT` is set to `'binary'`.


modules/carto/src/layers/schema/carto-spatial-tile-loader.ts
```diff
@@ -0,0 +1,57 @@
+import Protobuf from 'pbf';
+import {LoaderOptions, LoaderWithParser} from '@loaders.gl/loader-utils';
+
+import {TILE_FORMATS} from '../../api/maps-api-common';
```
alasarr, line 4: Never used warning

modules/carto/src/layers/schema/carto-spatial-tile.proto
```diff
@@ -0,0 +1,35 @@
+syntax = "proto3";
```
alasarr, line 1: As backend has already implemented it, I think you can remove it

modules/carto/src/layers/schema/carto-spatial-tile-loader.ts
```diff
@@ -0,0 +1,57 @@
...
+  version: '1',
+  id: 'cartoSpatialTile',
+  module: 'carto',
+  extensions: ['pbf'],
+  mimeTypes: ['application/x-protobuf'],
+  category: 'geometry',
+  worker: false,
+  parse: async (arrayBuffer, options) => parseCartoSpatialTile(arrayBuffer, options),
+  parseSync: parseCartoSpatialTile,
+  options: {}
```
alasarr, line 20: Don't you need to add a default to binary?

alasarr (changes requested):
Overall looks great man!

A couple of things we should add:
- The default value of formatTiles should be binary. Right now is json
- I really like your idea of separating the loader into a new file, let's do it for carto-tile too.
- Add the following mimeTypes so we will be able to add it in backend in the future: ` application/vnd.carto-vector-tile` and  `application/vnd.carto-spatial-vector-tile`

modules/carto/src/layers/schema/carto-spatial-tile-loader.ts
alasarr: @thedae ☝️  remember to not modify the defaults.

modules/carto/src/layers/carto-tile-layer.ts
```diff
@@ -67,7 +67,10 @@ const CartoTileLoader: LoaderWithParser = {
   id: 'cartoTile',
   module: 'carto',
   extensions: ['pbf'],
-  mimeTypes: ['application/x-protobuf'],
+  mimeTypes: [
+    'application/vnd.carto-vector-tile',
```
alasarr, line 71: Didn't you suggest `application/vnd.carto-tile`? I'm fine with this by the way

alasarr (changes requested):

docs/api-reference/carto/carto-layer.md
```diff
@@ -132,6 +132,8 @@ Name of the connection registered in the CARTO workspace.
 
 ##### `formatTiles` (String, optional)
 
+- Default: `TILE_FORMATS.BINARY`
```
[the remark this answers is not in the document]
felixpalmer: As per discussion, I've reverted this commit

test/apps/carto-dynamic-tile/app.js
```diff
@@ -7,65 +7,74 @@ import {CartoLayer, FORMATS, MAP_TYPES} from '@deck.gl/carto';
...
+const INITIAL_VIEW_STATE = {longitude: 8, latitude: 47, zoom: 6};
 const COUNTRIES =
   'https://d2ad6b4ur7yvpq.cloudfront.net/naturalearth-3.3.0/ne_50m_admin_0_scale_rank.geojson';
 
 // Skip CDN
 // const apiBaseUrl = 'https://direct-gcp-us-east1.api.carto.com';
 // PROD US GCP
-const apiBaseUrl = 'https://gcp-us-east1.api.carto.com';
+// const apiBaseUrl = 'https://gcp-us-east1.api.carto.com';
+const apiBaseUrl = 'https://gcp-us-east1-06.dev.api.carto.com';
```
alasarr, line 19: Update to prod

alasarr (approved):

modules/carto/src/layers/carto-tile-layer.ts
felixpalmer: It feels more specific and aligned with `carto-spatial-tile`

felixpalmer:
> CartoTileLoader is still inside `carto-tile-layer`. Do you want to address that in a separate PR?

Yes, I want to do that separately as it isn't adding new function, but merely refactoring/cleaning up

PR #7161, opened 2022-07-21
[mapbox] docs use getFillColor instead of getColor

chrisgervang:
Deprecation warning when following docs: 
```
getColor` is deprecated and will be removed in a later version. Use `getFillColor/getLineColor` instead
```

### DIFF
--- a/docs/api-reference/mapbox/mapbox-overlay.md
+++ b/docs/api-reference/mapbox/mapbox-overlay.md
@@ -26,7 +26,7 @@ const overlay = new MapboxOverlay({
       ],
       getPosition: d => d.position,
       getRadius: d => d.size,
-      getColor: [255, 0, 0]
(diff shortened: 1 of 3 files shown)


chrisgervang:
@Pessimistress general comment, these docs aren't obvious to find. What do you think of linking to them from https://deck.gl/docs/get-started/using-with-react#adding-a-base-map?

Pessimistress (approved):

Pessimistress:
I think the new structure is a little confusing.

[This overview page](https://deck.gl/docs/get-started/using-with-map) should be moved to the get-started section.

[The React page](https://deck.gl/docs/get-started/using-with-react#adding-a-base-map) should link to the [using with Mapbox page](https://deck.gl/docs/developer-guide/base-maps/using-with-mapbox#react-map-gl)

PR #7163, opened 2022-07-21
[docs] `fitBounds` requires width and height

chrisgervang:
Without this clarification `fitBounds` doesn't behave correctly because the default value for `width` and `height` is `1`.

### DIFF
--- a/docs/api-reference/core/web-mercator-viewport.md
+++ b/docs/api-reference/core/web-mercator-viewport.md
@@ -134,7 +134,9 @@ Parameters:
 * `bounds` (Array) - Bounding box in `[[longitude, latitude], [longitude, latitude]]`.
 * `opts` (Object) - See additional options in [@math.gl/web-mercator](https://math.gl/modules/web-mercator/docs/api-reference/web-mercator-utils#fitboundsopts)
   + `width` (Number) - If not supplied, will use the current width of the viewport (default `1`)
-  + `height` (Number) - Required unless set on viewport.
+  + `height` (Number) - If not supplied, will use the current height of the viewport (default `1`)
+  + `minExtent` (Number) - In degrees, 0.01 would be about 1000 meters
+  + `maxZoom` (Number) - Max zoom level
   + `padding` (Number) - The amount of padding in pixels to add to the given bounds.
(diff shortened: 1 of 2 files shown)


docs/api-reference/core/web-mercator-viewport.md
```diff
@@ -127,12 +127,14 @@ Returns:
...
-Returns a new viewport that fit around the given bounding box. Only supports non-perspective mode.
+Returns a new viewport that fit around the given bounding box. Viewport `width` and `height` must be either set or provided as options. Only supports non-perspective mode.
 
 Parameters:
 
 * `bounds` (Array) - Bounding box in `[[longitude, latitude], [longitude, latitude]]`.
-* `opts` (Object)
+* `opts` (Object) - See additional options in [@math.gl/web-mercator](https://math.gl/modules/web-mercator/docs/api-reference/web-mercator-utils#fitboundsopts)
+  + `width` (Number) - Required unless set on viewport.
+  + `height` (Number) - Required unless set on viewport.
```
Pessimistress, line 137: Maybe also add typing to the method definition here? https://github.com/visgl/deck.gl/blob/master/modules/core/src/viewports/web-mercator-viewport.ts#L303

docs/api-reference/core/web-mercator-viewport.md
```diff
@@ -127,12 +127,14 @@ Returns:
...
 
-Returns a new viewport that fit around the given bounding box. Only supports non-perspective mode.
+Returns a new viewport that fit around the given bounding box. Viewport `width` and `height` must be either set or provided as options. Only supports non-perspective mode.
 
 Parameters:
 
 * `bounds` (Array) - Bounding box in `[[longitude, latitude], [longitude, latitude]]`.
-* `opts` (Object)
+* `opts` (Object) - See additional options in [@math.gl/web-mercator](https://math.gl/modules/web-mercator/docs/api-reference/web-mercator-utils#fitboundsopts)
+  + `width` (Number) - Required unless set on viewport.
```
Pessimistress, line 136:
```suggestion
  + `width` (Number) - If not supplied, will use the current width of the viewport (default `1`)
```

PR #7177, opened 2022-07-31
Restrict h3-js version in scripting examples

Pessimistress:
h3-js has published a new major version which contains breaking changes.

@nrabinowitz Unrelated but `4.0.0-rc1` should not be tagged `latest`?

#### Change List
- Documentations of H3 layers
- In-doc examples of H3 layers

### DIFF
--- a/docs/api-reference/geo-layers/h3-cluster-layer.md
+++ b/docs/api-reference/geo-layers/h3-cluster-layer.md
@@ -73,7 +73,7 @@ new H3ClusterLayer({});
 To use pre-bundled scripts:
 
 ```html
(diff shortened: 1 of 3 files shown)


nrabinowitz (approved):
I was assuming the `-rc1` version would be handled correctly by `yarn`/`npm`, but didn't consider unpinned `unpkg` versions. I'm not sure what I can reasonably do here - I think the `latest` tag is applied automatically in `npm`, and even if it wasn't correct for the release candidate it would be correct soon for the `v4.0.0` release. Please let me know if you have recommendations here.

kylebarron (approved):

PR #7188, opened 2022-08-09
CARTO: Add headers prop to CartoLayer

felixpalmer:
#### Background
Adds support for passing arbitrary headers to the map instantiation request

#### Change List
- `fetchLayerData` accepts a `headers` parameter which is applied to the map instantiation request
- New `CartoLayer` `headers` prop
- Tests & doc


alasarr (approved):

alasarr:
Just realized we need to add it to fetchMap, the headers should be sent to fetchMap, and all the fetchLayerData requests from the map

Issue #7128, opened 2022-07-08
[Bug] maplibre Interleaved layer coved by 3d building layer


longxiutmm:
### Description

I use MapboxLayer 、ArcLayer and maplibre 2.1.9, the 3d building layer always cover the arcLayer
![image](https://user-images.githubusercontent.com/19338062/177907379-c976b83b-381d-451e-a01a-66832accbbfb.png)

### Flavors

- [ ] React
- [ ] Python/Jupyter notebook
- [X] MapboxLayer
- [ ] GoogleMapsOverlay
- [ ] CartoLayer
- [ ] DeckLayer/DeckRenderer for ArcGIS

### Steps to Reproduce

blow is my code.

```html

<html>
<head>
    <title>Interleaving deck.gl with Mapbox Layers</title>
    <script src="https://unpkg.com/deck.gl@^8.8.0/dist.min.js"></script>

    <script src='https://unpkg.com/maplibre-gl@2.1.9/dist/maplibre-gl.js'></script>
    <link href='https://unpkg.com/maplibre-gl@2.1.9/dist/maplibre-gl.css' rel='stylesheet' />

</head>

<body style="margin:0">


</body>

<script type="text/javascript">

    const {MapboxLayer, ScatterplotLayer, ArcLayer} = deck;

    var map = new maplibregl.Map({
        container: document.body,
        style: 'https://api.maptiler.com/maps/streets/style.json?key=get_your_own_OpIi9ZULNHzrESv6T2vL', // stylesheet location
        center: [-122.4, 37.79],
        zoom: 15,
        pitch: 60
    });


    map.on('load', () => {
        const firstLabelLayerId = map.getStyle().layers.find(layer => layer.type === 'symbol').id;


        map.setPaintProperty('building-3d','fill-extrusion-opacity',0.8);


        map.addLayer(new MapboxLayer({
            id: 'deckgl-circle',
            type: ScatterplotLayer,
            data: [
                {position: [-122.402, 37.79], color: [255, 0, 0], radius: 1000}
            ],
            getPosition: d => d.position,
            getFillColor: d => d.color,
            getRadius: d => d.radius,
            opacity: 0.3
        }), firstLabelLayerId);

        map.addLayer(new MapboxLayer({
            id: 'deckgl-arc',
            type: ArcLayer,
            data: [
                {source: [-122.3998664, 37.7883697], target: [ -122.41460563510304,  37.79512692863844]}
            ],
            getSourcePosition: d => d.source,
[...]

PR #7193, opened 2022-08-10
Fix maplibre-gl interleaving

Pessimistress:
For #7128

maplibre 2.0 does not expose `map.version`, see issue at https://github.com/maplibre/maplibre-gl-js/issues/1426

I'm proposing to drop compatibility with mapbox-gl v0.50-1.2, which are more than 3 years old (mapbox v1's latest is 1.13 and maplibre v1's latest is 1.15). These versions are currently identified by the absence of `map.version`.

#### Change List
- Always use the most recent near plane value (technically a breaking change, for mapbox-gl <=1.2)

### DIFF
--- a/modules/mapbox/src/deck-utils.ts
+++ b/modules/mapbox/src/deck-utils.ts
@@ -9,7 +9,7 @@ type UserData = {
   isExternal: boolean;
   currentViewport?: WebMercatorViewport | null;
   mapboxLayers: Set<MapboxLayer<any>>;
-  mapboxVersion: {minor: number; major: number};
+  // mapboxVersion: {minor: number; major: number};
 };
 
 // Mercator constants
@@ -85,7 +85,7 @@ export function getDeckInstance({
   }
 
   (deckInstance.userData as UserData).mapboxLayers = new Set();
-  (deckInstance.userData as UserData).mapboxVersion = getMapboxVersion(map);
+  // (deckInstance.userData as UserData).mapboxVersion = getMapboxVersion(map);
   map.__deck = deckInstance;
   map.on('render', () => {
     if (deckInstance.isInitialized) afterRender(deckInstance, map);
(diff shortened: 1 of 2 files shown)


modules/mapbox/src/deck-utils.ts
```diff
@@ -9,7 +9,7 @@ type UserData = {
   isExternal: boolean;
   currentViewport?: WebMercatorViewport | null;
   mapboxLayers: Set<MapboxLayer<any>>;
-  mapboxVersion: {minor: number; major: number};
+  // mapboxVersion: {minor: number; major: number};
```
Pessimistress, line 12: Version parsing is no longer used, keeping the code in case another breaking change is introduced in the base map libraries in the future.

chrisgervang (approved):
How will we document this?

modules/mapbox/src/deck-utils.ts
```diff
@@ -208,47 +208,32 @@ function centerCameraOnTerrain(map: Map, viewState: MapViewState) {
...
-  );
+  return new WebMercatorViewport({
+    id: 'mapbox',
+    x: 0,
+    y: 0,
+    width: deck.width,
+    height: deck.height,
+    ...getViewState(map),
+    nearZMultiplier: useMapboxProjection
+      ? // match mapbox's projection matrix
```
chrisgervang, line 232: Can we say matching mapbox 1.3.0+?

modules/mapbox/src/deck-utils.ts
```diff
@@ -83,7 +83,7 @@ export function getDeckInstance({
   }
 
   (deckInstance.userData as UserData).mapboxLayers = new Set();
-  (deckInstance.userData as UserData).mapboxVersion = getMapboxVersion(map);
+  // (deckInstance.userData as UserData).mapboxVersion = getMapboxVersion(map);
```
chrisgervang, line 86: Can you add a warning if we detect version <= 1.2.0?

Pessimistress: We are unable to detect version <=1.2.0 because they do not expose `version`.

Pessimistress: The gist of this bug is that when `map.version` is undefined, it is either mapbox <= 1.2.0 or maplibre >= 2.0.0. Before this PR we assume it is mapbox. After this PR we assume it is maplibre.

Issue #7203, opened 2022-08-16
[Bug] Geojson with only z change doesn't render filled


Lyeeedar:
### Description

When giving a geojson layer a polygon with only Z changes it only renders a stroke at the top and bottom, but no filled polygon interior. Giving the geojson layer lineBillboard: true renders the full outline, but still no interior.

See pic for the issue. Red is the geojson, yellow with lineBillboard, blue when adding 0.00001 to the points at the upper line (so they arent perfectly vertical anymore).

![image](https://user-images.githubusercontent.com/1607501/184928041-297a40a7-c1b0-426a-8e67-f20918401268.png)

### Flavors

- [X] React
- [ ] Python/Jupyter notebook
- [ ] MapboxLayer
- [ ] GoogleMapsOverlay
- [ ] CartoLayer
- [ ] DeckLayer/DeckRenderer for ArcGIS

### Steps to Reproduce

https://codepen.io/lyeeedar/pen/MWVzQXX

### Environment

- Framework version: deck.gl@8.7.11
- Browser: Firefox 
- OS: Windows 10

Pessimistress:
There isn't a quick fix for this. We tesselate polygons using `earcut` which only handles 2D input. Technically the GeoJSON specification is also 2D. The z component is only respected when projection is performed. Adding a small shift to some points may be the easiest workaround for now. If you are only rendering vertical rectangles in your layer, you could also consider using `SimpleMeshLayer`.

We could use a section in the `PolygonLayer`/`GeoJsonLayer` docs to explain this limitation.

Lyeeedar:
Sadly I'm currently rendering a large amount of arbitrary geojson with z, and just noticed this when some faces weren't rendering. I'd like to avoid triangulating myself to make geometries (I assume that SimpleMeshLayer is only going to handle triangles).

In theory would we maybe be able to have an optional parameter on geojson (and presumably polygon layer) that uses a 3d variant of the earcut algorithm (e.g. if earcut produces no triangles flip 2 of the axis, cut, then flip back. Or maybe something cleverer)? I'm more than happy to try to take a look if you think thats something worth persuing. Would you be able to point me to where in the code it does the earcutting?

Pessimistress:
https://github.com/visgl/deck.gl/blob/6708cd139680863d17335600066257a2d16c2c68/modules/layers/src/solid-polygon-layer/polygon.ts#L305-L321

[...]

PR #7205, opened 2022-08-18
[feat] Add a _full3d prop that uses the plane with the maximum area to tesselate

Lyeeedar:
This should fix rendering of full 3d data (data that changes only on the z plane). See https://github.com/visgl/deck.gl/issues/7203

As per the suggested solution to the issue https://github.com/mapbox/earcut/issues/21#issuecomment-98119437 this should calculate the plane with the maximum area, then use that plane to tesselate on.

I've also bumped the earcut version to the latest as it was a few behind, and according to the patch notes there have been a lot of performance improvements made.

### DIFF
--- a/modules/layers/src/geojson-layer/geojson-layer.ts
+++ b/modules/layers/src/geojson-layer/geojson-layer.ts
@@ -198,6 +198,13 @@ type _GeoJsonLayer3DProps<DataT> = {
   wireframe?: boolean;
 
   /**
+   * (Experimental) This prop is only effective with `XYZ` data.
+   * When true, polygon tesselation will be performed on the plane with the largest area, instead of the xy plane.
+   * @default false
+   */
+  _full3d?: boolean;
+
+  /**
    * Elevation valur or accessor.
    *
    * Only used if `extruded: true`.
@@ -300,6 +307,7 @@ const defaultProps: DefaultProps<GeoJsonLayerProps> = {
   filled: true,
   extruded: false,
   wireframe: false,
+  _full3d: false,
   iconAtlas: {type: 'object', value: null},
   iconMapping: {type: 'object', value: {}},
   getIcon: {type: 'accessor', value: f => f.properties.icon},
--- a/modules/layers/src/solid-polygon-layer/polygon.ts
+++ b/modules/layers/src/solid-polygon-layer/polygon.ts
@@ -332,6 +332,8 @@ export function getSurfaceIndices(
 
   let positions = getPositions(polygon);
 
+  const is3d = full3d && positionSize === 3;
+
   if (preproject) {
     // When tesselating lnglat coordinates, project them to the common space for accuracy
     const n = positions.length;
@@ -344,7 +346,7 @@ export function getSurfaceIndices(
       p[0] = positions[i];
       p[1] = positions[i + 1];
 
-      if (full3d && positionSize === 3) {
+      if (is3d) {
         p[2] = positions[i + 2];
       }
 
@@ -353,13 +355,13 @@ export function getSurfaceIndices(
       positions[i] = xy[0];
       positions[i + 1] = xy[1];
 
-      if (full3d && positionSize === 3) {
+      if (is3d) {
         positions[i + 2] = xy[2];
       }
     }
   }
 
-  if (full3d && positionSize === 3) {
+  if (is3d) {
     // calculate plane with largest area
     const xyArea = getPlaneArea(positions, 0, 1);
     const xzArea = getPlaneArea(positions, 0, 2);
--- a/modules/layers/src/solid-polygon-layer/polygon.ts
+++ b/modules/layers/src/solid-polygon-layer/polygon.ts
@@ -331,22 +331,8 @@ export function getSurfaceIndices(
   }
 
   let positions = getPositions(polygon);
+  let positionsCloned = false;
 
-  if (preproject) {
-    // When tesselating lnglat coordinates, project them to the common space for accuracy
-    const n = positions.length;
-    // Clone the array
-    positions = positions.slice();
-    const p: number[] = [];
-    for (let i = 0; i < n; i += positionSize) {
-      p[0] = positions[i];
-      p[1] = positions[i + 1];
-      const xy = preproject(p);
-      positions[i] = xy[0];
-      positions[i + 1] = xy[1];
-    }
-  }
-  
   if (full3d && positionSize === 3) {
     // calculate plane with largest area
     const xyArea = getPlaneArea(positions, 0, 1);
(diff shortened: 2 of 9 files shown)


Lyeeedar:
I have been unable to test it in my actual use case as I'm not sure how to build a package I can drop in instead of the published deck.gl.

Additionally, trying to run 'yarn start-local' from the examples (e.g. get-started/react/basic) just gives an error.
'$ webpack-dev-server --env.local --progress --hot --open
/bin/sh: 1: webpack-dev-server: not found
error Command failed with exit code 127.'

I'm using WSL, and followed all the instructions (I think) in the contributing guide for windows + linux. I can run the tests, just not the actual examples.

Pessimistress:
Thank you for tackling this! I can pull and test it on my side.

A few issues (not necessarily addressed in this PR) remaining:

- Currently `preproject` only applies to the XY coordinates, so z is way out of proportion (in web Mercator about 78,000x along the equator).
- If the base plane is non-XY, should the extrusion axis switch too?

modules/layers/src/geojson-layer/geojson-layer.ts
```diff
@@ -197,6 +197,13 @@ type _GeoJsonLayer3DProps<DataT> = {
    */
   wireframe?: boolean;
 
+  /**
+   * (Experimental) This prop is only effective with `XYZ` data.
+   * When true, polygon tesselation will be performed on the plane with the largest area, instead of the xy plane.
+   * @default false
+   */
+  _full3d?: boolean;
```
Pessimistress, line 205: Add to the documentation?

Lyeeedar:
> 

Maybe it'd be better to do the axis swapping before we preproject in that case, lngLatToWorld doesn't support projecting the z. I guess doing that would still be correct enough.

By extrusion axis do you mean the 'extrude' mode of the layer? I haven't looked into that in any detail as I haven't needed to use it yet.

modules/layers/src/geojson-layer/geojson-layer.ts
Lyeeedar: Have taken a stab at it in the latest commit, take a look and say if you think it needs more fleshing out

Pessimistress:
Permutation needs to happen after preprojection. The Web Mercator world space (degrees/degrees/meters) is non-linear, and the area calculation will not work until you convert it to the common space (uniform cartesian). You can read more about this in https://deck.gl/docs/developer-guide/coordinate-systems#concepts.

The `preproject` function is specified here:

https://github.com/visgl/deck.gl/blob/e923ad9261b366d223c537f6641fef85524c4498/modules/layers/src/solid-polygon-layer/solid-polygon-layer.ts#L168-L169

If in "full 3D" mode, we can use `viewport.projectPosition` (3D) instead of `viewport.projectFlat` (2D), and trade some performance for accuracy.

Pessimistress:
We can deal with `extruded: true` in a follow up PR. It likely will involve some changes to the shader. I'm happy to look into it after we land this fix.

Pessimistress:
@kylebarron @felixpalmer @ibgreen Thoughts about the prop name?

Lyeeedar:
> Permutation needs to happen after preprojection. The Web Mercator world space (degrees/degrees/meters) is non-linear, and the area calculation will not work until you convert it to the common space (uniform cartesian). You can read more about this in https://deck.gl/docs/developer-guide/coordinate-systems#concepts.
> 
> The `preproject` function is specified here:
> 
> https://github.com/visgl/deck.gl/blob/e923ad9261b366d223c537f6641fef85524c4498/modules/layers/src/solid-polygon-layer/solid-polygon-layer.ts#L168-L169
> 
> If in "full 3D" mode, we can use `viewport.projectPosition` (3D) instead of `viewport.projectFlat` (2D), and trade some performance for accuracy.

That you for this, I'll look into making these adjustments on monday

modules/layers/src/solid-polygon-layer/solid-polygon-layer.ts
```diff
@@ -156,19 +164,31 @@ export default class SolidPolygonLayer<DataT = any, ExtraPropsT = {}> extends La
...
         // Lnglat coordinates are usually projected non-linearly, which affects tesselation results
         // Provide a preproject function if the coordinates are in lnglat
-        preproject:
-          coordinateSystem === COORDINATE_SYSTEM.LNGLAT && viewport.projectFlat.bind(viewport),
+        preproject,
         fp64: this.use64bitPositions(),
-        IndexType: !gl || hasFeatures(gl, FEATURES.ELEMENT_INDEX_UINT32) ? Uint32Array : Uint16Array
+        IndexType:
+          !gl || hasFeatures(gl, FEATURES.ELEMENT_INDEX_UINT32) ? Uint32Array : Uint16Array,
+        full3d: _full3d
```
Pessimistress, line 191: You don't need this since it's always set in `updateState`

modules/layers/src/solid-polygon-layer/polygon.ts
```diff
@@ -305,15 +335,55 @@ export function getSurfaceIndices(
...
+
     // Clone the array
     positions = positions.slice();
+
     const p: number[] = [];
     for (let i = 0; i < n; i += positionSize) {
       p[0] = positions[i];
       p[1] = positions[i + 1];
+
+      if (full3d) {
```
Pessimistress, line 347: `positionSize === 3` should be required here too?

Lyeeedar: PositionSize is calculated in the tesselator, but the preproject function is passed in from the solid-polygon-layer which doesnt have access to this so always sets the preproject to the xyz variant when full3d is true. So in here when full3d is true we are always going to have the xyz variant of the preproject function.
Ofc if the projectPosition class works with xy and xyz data, then we need to add the positionSize===3 here too, but I assumed it didn't as the argument didn't seem to suggest it.

Lyeeedar: Ignore me, I just looked and I see it supports both. Shall change.

modules/layers/src/solid-polygon-layer/polygon.ts
```diff
@@ -305,15 +335,55 @@ export function getSurfaceIndices(
...
+
     // Clone the array
     positions = positions.slice();
+
     const p: number[] = [];
     for (let i = 0; i < n; i += positionSize) {
       p[0] = positions[i];
       p[1] = positions[i + 1];
+
+      if (full3d && positionSize === 3) {
```
Pessimistress, line 347: Can you take this check out of the for loop and make a new variable?

Pessimistress (approved):
Thank you!

PR #7208, opened 2022-08-21
Fix controller bug when view props change

Pessimistress:
For #7156

`Controller.makeViewport` currently assumes that the associated `view` never changes.

#### Change List
- Always use the latest view to calculate view states


chrisgervang (approved):

Pessimistress:
Published v8.8.9.

Issue #7211, opened 2022-08-23
[Bug] Text-Layer SDF settings not working as expected (text outlines)


DanielBoa:
### Description

When trying to tweak the Text-Layer SDF settings to achieve a nicely rendered outline I am unable to do so.

![Screenshot 2022-08-22 at 09 44 41](https://user-images.githubusercontent.com/1169866/185882031-31bdc4ec-fdea-4ea9-9736-ac3ef0956aa6.png)
_Note the clipping of the outline._

After digging into the implementation a bit I think there are a few problems contributing to this:

1. The `FontAtlasManager` generates the atlas mapping in the same manner regardless if SDF mode is being used or not, this results in the gradient part of the SDF being "clipped".
    - Part of the issue is that the width of the character's texture is also used as its horizontal advance when laying out the text. 

![image](https://user-images.githubusercontent.com/1169866/186162250-5293d0dd-dce2-4904-9b78-0054ceeea1d4.png)
_An example of a generated atlas, I've added blue bounds from the generated mapping to illustrate how the texture is being sliced. You can see that gradient around some characters are clipped._

2. Increasing the buffer size should in theory help this, however as mentioned above, as the texture width is also used as the character advance the mapping width does not change as the buffer does.
3. As the buffer increases the characters begin to move vertically downwards in the mapping, their calculated bounding box doesn't account for this resulting in them being clipped.
4. I cannot seem to affect the width of the outline using the `outlineWidth` property, this just seems to turn the outline on or off.

I have actually fixed these issues in a branch and plan on opening a PR for review shortly.

### Flavors

- [ ] React
- [ ] Python/Jupyter notebook
- [ ] MapboxLayer
- [ ] GoogleMapsOverlay
- [ ] CartoLayer
- [ ] DeckLayer/DeckRenderer for ArcGIS

### Expected Behavior

1. Tweaking the SDF settings should allow the outline of text to be rendered correctly; without clipping.
2. `outlineWidth` should actually affect the width of the rendered outline.

### Steps to Reproduce

```
fontSettings: {
  sdf: true,
  fontSize: 64,
  buffer: 4,
  radius: 12,
  cutoff: 0.25,
  smoothing: 0.1,
  // I tried tweaking these parameters a fair amount with no success
},
```

I used observable to investigate the generation of the atlas – this is where the image above came from – you can find that at the link below:
https://observablehq.com/d/60b1a4333859aa11

### Environment

- Framework version: 8.8.0-beta.5
[...]

PR #7212, opened 2022-08-23
[FIX][HELP WANTED] Text-Layer: Improve SDF Font Rendering (Text Outlines)

DanielBoa:
For #7211 

Screenshot 2022-08-22 at 09 40 27
_How the same label with the same settings – well I eyeballed the colour – as shown in the referenced issue now renders._

#### Behaviour changed by this PR
1. `sdf` mode can render text outlines without clipping.
2. `buffer` property no longer causes text to be offset vertically.
3. `outlineWidth` scales width of the rendered outline from 0-10.

#### Change List
- Bump `@mapbox/tiny-sdf` from `^1.1.0=>^2.0.5`
- In the `FontAtlasManager` there is now two distinct paths taken when generating the atlas and mapping, either:
  1. One for the "regular" atlas – this is largely the same as before with the SDF stuff removed.
  2. One for the SDF mode – this now uses the updated `@mapbox/tiny-sdf` package to also produce the atlas/mapping, in the mapping it now stores more information per character that can be used to correctly layout the text whilst allowing for the whole of the rendered character in the atlas to be used as the texture for the mutli-icon layer.
- In the draw method of the multi icon layer [there is now two render passes per character when SDF mode is on](https://github.com/DanielBoa/deck.gl/blob/963bb0e61a780bbfb2749b2cb338124cc419e336/modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer.ts#L121-L131), this is to ensure that the text outline is never rendered above a previous characters "centre". **This is obviously not ideal, see below for comment.**

#### Help wanted
1. As mentioned above the fact that two render passes are made when using SDF mode isn't ideal, my hope is that this is possible using blend modes or something but I just lack the `webgl`/`luma.gl` knowledge and was hoping someone could point me in the right direction.
2. The updated `@mapbox/tiny-sdf` is now exported using ESM module syntax and has `"type": "module"` in it's `package.json`, this is causing the `ocular-dev-tools` to throw (`ERR_REQUIRE_ESM`) which means I haven't been able to run the repo test suite. Curiously the package seems to work fine when running `examples/layer-browser` which is where I worked on these changes. Any guidance here would also be appreciated.


Pessimistress:
Thank you for submitting a fix!

If I understand the clipping issue correctly, there are 3 frames of reference:

image

Layout bounds (red) is the real bounding box of the character.
Render bounds (green) is what *should* be processed by the fragment shader.
Packing bounds (blue) adds extra bleeding to prevent interference/overlap between characters.

Right now `fontSettings.buffer` controls the size difference between red and blue. (It does not seem to be applied correctly on the y axis in SDF mode, as you have observed)
Green is always the same as red, which is ok in non-SDF, but causes the clipping in SDF mode.

image

I think an easy fix is to just add another constant (either a `fontSettings` option or a value calculated from `radius`) to control the render bounds. 

Regarding adding another draw call for `outline` to blend correctly: I do think it's a fair trade off for better rendering results.

modules/layers/src/text-layer/font-atlas-manager.ts
```diff
@@ -188,26 +197,98 @@ export default class FontAtlasManager {
...
+    const rows = Math.ceil(characterSet.size / columns);
+    const requiredCanvasHeight = nextPowOfTwo(rows * cellSize);
+
+    if (canvas.height !== requiredCanvasHeight) {
+      resizeCanvas(ctx, canvas.width, requiredCanvasHeight);
+    }
+
+    let x = 0;
+    let y = 0;
+    for (const char of characterSet) {
```
Pessimistress, line 248: You are basically duplicating the `buildMappings` function to handle dynamic character width returned by `SDF.draw`. What advantages does this provide? I see that this new code still assumes that `width` and `height` returned by `tinySDF.draw` is contained in the `cellSize` square, is that true? If so, can we modify `buildMapping` to always use `cellSize` as the packing width in SDF mode?

modules/layers/src/text-layer/font-atlas-manager.ts
```diff
@@ -188,26 +197,98 @@ export default class FontAtlasManager {
...
+      ctx.putImageData(imageData, x, y);
+
+      mapping[char] = {
+        x,
+        y,
+        // use glyphAdvance over glyphWidth due to whitespace characters
+        width: glyphAdvance,
+        textureHeight: cellSize,
+        textureWidth: width,
+        textureOffsetY: fontSize - glyphTop
```
Pessimistress, line 262: I find the names `texture*` confusing (since everything is in a texture). Going back to my diagram, `x` `y` `width` and `height` should describe the green box, and we should only need two more fields `layoutWidth` and `layoutHeight` to describe the red box.

DanielBoa: I've removed the `texture*` prefix so there is now; `x`, `y`, `width`, `height`, `layoutWidth`, & `layoutOffsetY`.

The reason I've used `layoutOffsetY` is because the vertical alignment of the font is no longer baked into the SDF atlas as with the regular atlas, instead the characters are rendered from the very top of their allocated "cell" and the vertical alignment is fixed when positioning the icons instead. The reason I did this was because aligning them within their cell would often cause dangling letters like "y" to have their gradient clipped.

How the atlas looks without the alignment baked in. The green outline being the "render bounds" & the red being the "layout" bounds.
image

Here's how it looks with the alignment being baked in, notice the "y" and a few other get their gradient clipped.
image

Not baking in the alignment would also allow for each character to have a far smaller texture height due to not requiring the whitespace, however when I try this for some reason it completely breaks the positioning of the character icons, I've not figured out why. Also having varying height character icons would also require them to have their scaled individually.

examples/layer-browser/webpack.config.js
```diff
@@ -20,7 +20,8 @@ const CONFIG = {
     alias: {
       // From mapbox-gl-js README. Required for non-browserify bundlers (e.g. webpack):
       'mapbox-gl$': resolve('./node_modules/mapbox-gl/dist/mapbox-gl.js'),
-      '@math.gl/web-mercator': resolve('../../node_modules/@math.gl/web-mercator')
+      '@math.gl/web-mercator': resolve('../../node_modules/@math.gl/web-mercator'),
+      '@mapbox/tiny-sdf': resolve('../../node_modules/@mapbox/tiny-sdf')
```
Pessimistress, line 24: I think you can drop this now that the dependency conflict is fixed

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer.ts
```diff
@@ -102,18 +103,30 @@ export default class MultiIconLayer<DataT, ExtraPropsT = {}> extends IconLayer<
...
+
     super.draw(params);
+
+    // draw text without outline on top to ensure a thick outline won't occlude other characters
+    if (sdf && outlineWidth) {
+      const {iconManager} = this.state;
+      const iconsTexture = iconManager.getTexture();
+
+      if (iconsTexture) {
+        this.state.model.setUniforms({outlineBuffer: 0}).draw();
```
Pessimistress, line 127: This uniform used to be `DEFAULT_BUFFER` when `outlineWidth: 0`. Is the change intentional?

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer.ts
```diff
@@ -102,18 +103,30 @@ export default class MultiIconLayer<DataT, ExtraPropsT = {}> extends IconLayer<
   draw(params) {
     const {sdf, smoothing, outlineWidth} = this.props;
     const {outlineColor} = this.state;
+    const outlineBuffer = outlineWidth ? DEFAULT_BUFFER * (1 - outlineWidth) : -1;
```
Pessimistress, line 106: As discussed, this value needs an upper limit to avoid visual artifacts.

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer.ts
```diff
@@ -125,8 +138,11 @@ export default class MultiIconLayer<DataT, ExtraPropsT = {}> extends IconLayer<
   }
 
   getInstanceIconFrame(icons: string): number[] {
-    return icons
-      ? Array.from(icons).flatMap(icon => super.getInstanceIconFrame(icon))
-      : EMPTY_ARRAY;
+    if (!icons) return EMPTY_ARRAY;
```
Pessimistress, line 141: This change is now unnecessary

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer.ts
DanielBoa: No, the intention is just to draw the text sans outline, I'll use `DEFAULT_BUFFER` instead

modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer.ts
DanielBoa: I've reverted to using the original method of calculating `outlineBuffer` and it appears to work well, i.e. no artefacts and large `outlineWidth` values result in an outline clamped at the upper limit.

examples/layer-browser/webpack.config.js
DanielBoa: Without this line the `layer-browser` example doesn't seem to work for me as it ends up using `@mapbox/tiny-sdf` version `1.2.5` instead of `2.0.5`.

`yarn why @mapbox/tiny-sdf` shows it's because of a conflict with `react-map-gl`.

```
=> Found "@mapbox/tiny-sdf@1.2.5"
info Reasons this module exists
   - "react-map-gl#mapbox-gl" depends on it
   - Hoisted from "react-map-gl#mapbox-gl#@mapbox#tiny-sdf"
```

PR #7235, opened 2022-09-05
Added validation function for carto layers in pydeck-carto

tuxskar:
Added a function to catch the common issues expected before trying to build the html

Using this function the errors are clearer and narrow the issues with descriptive explanations

For instance, a wrong credential token or a wrong column selection would end up in an error on the front end when rendering the HTML on the browser but the errors would be quite odd such as unauthorized instead of "wrong token" or "invalid types.String parameter"


bindings/pydeck-carto/pydeck_carto/carto_auth/layer_validator.py
```diff
@@ -0,0 +1,64 @@
...
+    if len(token_parts) != 3:
+        raise ValueError(f'Access token with wrong format')
+    else:
+        for part, part_label in zip(token_parts, ['header', 'payload', 'signature']):
+            if not part:
+                raise ValueError(f'Missing "{part_label}" of the token')
+
+    if layer.connection != CartoConnection.CARTO_DW:
+        available_connections = carto_auth.connections_list()
+        if layer.connection != CartoConnection.CARTO_DW and \
```
Jesus89, line 47: Duplicated `if layer.connection != CartoConnection.CARTO_DW`

bindings/pydeck/pydeck/types/string.py
```diff
@@ -25,3 +25,6 @@ def __eq__(self, other):
 
     def __repr__(self):
         return self.value
+
+    def __hash__(self):
```
Jesus89, line 29: Where is this used?

bindings/pydeck-carto/pydeck_carto/carto_auth/layer_validator.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+def is_valid_carto_layer(layer: Layer, carto_auth: CartoAuth):
+    if layer.type != 'CartoLayer':
+        raise ValueError('The layer type must be "CartoLayer"')
+
+    validate_attr(layer, 'credentials', dict)
+    validate_attr(layer, 'data', str)
+
+    required_strings = ['type_', 'connection']
```
Jesus89, line 23: Why the loop instead of

```
validate_attr(layer, 'type_')
validate_attr(layer, 'connection')
```

bindings/pydeck-carto/pydeck_carto/carto_auth/layer_validator.py
```diff
@@ -0,0 +1,64 @@
...
+        validate_attr(layer, attr)
+
+    if layer.type_ not in [MapType.QUERY, MapType.TABLE, MapType.TILESET]:
+        raise ValueError(f'Invalid layer type selected')
+
+    for attr in ('accessToken', 'apiVersion', 'apiBaseUrl'):
+        if attr not in layer.credentials:
+            raise ValueError(f'Missing {attr} from layer.credentials')
+
+    if 'carto.com' not in layer.credentials['apiBaseUrl']:
```
Jesus89, line 34: Remove this check. CARTO Self-hosted can have a different apiBaseUrl without carto.com.

bindings/pydeck-carto/pydeck_carto/carto_auth/layer_validator.py
```diff
@@ -0,0 +1,64 @@
...
+
+    if layer.type_ == MapType.QUERY:
+        cnt_params_expected = layer.data.count('?')
+        if cnt_params_expected > 0 and cnt_params_expected != len(layer.query_parameters):
+            raise ValueError(
+                f'Mismatch query_parameters and parameters expected on the query')
+
+    if hasattr(layer, 'geo_column') and layer.geo_column is not None:
+        validate_attr(layer, 'geo_column')
+        if layer.geo_column not in {GeoColumnType.H3, GeoColumnType.QUADBIN}:
```
Jesus89, line 61: Remove this check. Attribute `geo_column` can have any value.

bindings/pydeck-carto/pydeck_carto/carto_auth/auth.py
```diff
@@ -215,3 +212,16 @@ def get_carto_dw_client(self):
 
         project_id, cw_token = self.get_carto_dw_credentials()
         return GClient(project_id, credentials=GCredentials(cw_token))
+
+    def connections_list(self) -> list:
```
Jesus89, line 216: Rename to `list_connections(self)`.

bindings/pydeck-carto/pydeck_carto/carto_auth/auth.py
```diff
@@ -215,3 +212,16 @@ def get_carto_dw_client(self):
...
+        """Returns the list of available connections using this credential"""
+        url = f'{self.api_base_url}/v3/connections'
+        headers = self._api_headers()
+
+        response = requests.get(url, headers=headers)
+        connections_response = response.json()
+        return [conn.get('name') for conn in connections_response]
+
+    def is_valid_connection(self, connection_id):
```
Jesus89, line 225: Remove this. Not used in the validation

bindings/pydeck-carto/examples/carto_layer_quadbin_tileset.py
```diff
@@ -22,6 +22,8 @@
     pickable=True,
 )
 
+
```
Jesus89, line 25: Missing assert is_valid_carto_layer?

Jesus89 (changes requested):
Added some comments regarding wrong validations and minor rename.

bindings/pydeck-carto/pydeck_carto/carto_auth/layer_validator.py
tuxskar: To group all the fields that require `types.String` type as parameters, but maybe as it is only 2 it makes sense to follow the same pattern as the previous validations, I'll change it

bindings/pydeck-carto/pydeck_carto/carto_auth/layer_validator.py
tuxskar: Ok

bindings/pydeck/pydeck/types/string.py
tuxskar: When checking if a value is in a list of potential values it is using the hash of the object like in this check:
```
    if layer.type_ not in [MapType.QUERY, MapType.TABLE, MapType.TILESET]:
        raise ValueError(f'Invalid layer type selected')
```
If `__hash__` is not implemented, the check will raise an error

bindings/pydeck-carto/examples/carto_layer_quadbin_tileset.py
tuxskar: yes, added

bindings/pydeck/pydeck/types/string.py
Jesus89: This requires a new release of pydeck, so I would comment on that code until the new release is available.

Jesus89 (approved):

Issue #7204, opened 2022-08-17
MaskExtension not working in FirstPersonView


Radik24:
### Description

Hi, I would like to use MaskExtension for filtering data in FirstPersonView, but it's throwing errors

### Flavors

- [ ] React
- [ ] Python/Jupyter notebook
- [ ] MapboxLayer
- [ ] GoogleMapsOverlay
- [ ] CartoLayer
- [ ] DeckLayer/DeckRenderer for ArcGIS

### Expected Behavior

Should be working same as in MapView

### Steps to Reproduce

Codepen: https://codepen.io/radik24/pen/NWYeKbR?editors=1111

You can comment/uncomment FirstPersonView/MapView to see the behavior

### Environment

- Framework version: latest
- Browser: Chrome 104.0.5112.79
- OS: macOS 12.5

### Logs

![image](https://user-images.githubusercontent.com/23722745/185115240-d6d3d3e0-bc17-4e7c-a85f-cb33a964f7f1.png)

PR #7240, opened 2022-09-06
Support MaskExtension in FirstPersonView

felixpalmer:
For https://github.com/visgl/deck.gl/issues/7204

#### Change List
- Use `WebMercatorViewport` when rendering mask in `FirstPersonView`
- Add test app showing usage
- Fix other mask test app (maskEnabled button was broken)


felixpalmer:
New test app:

Screenshot 2022-09-06 at 14 24 33

Pessimistress (approved):

modules/core/src/effects/mask/utils.ts
```diff
@@ -93,7 +93,7 @@ export function getMaskViewport({
   width -= padding * 2;
   height -= padding * 2;
 
-  if (viewport instanceof WebMercatorViewport) {
+  if (viewport.isGeospatial) {
```
Pessimistress, line 96: `isGeospatial` is also true for `GlobeView` and I suspect that this won't work for non-mercator projections

felixpalmer: It doesn't, although it isn't far off - the rendered mask is stretched. I've added a warning and updated the docs to make clear that `GlobeView` currently doesn't support masking in order to land this PR

PR #7249, opened 2022-09-08
Add styling functions in pydeck

Jesus89:
This PR contains the required changes in pydeck, and pydeck-carto to support functions from custom libraries and implement carto styling functions in pydeck-carto.

**pydeck**
- Add types.Function

**pydeck-carto**
- Add color_bins, color_categories, color_continuous functions


alasarr (changes requested):
Nice job here, super excited about having the possibility in pydeck for styling functions. 

I think would be better to split this PR in two, as it's addressing different purposes. One for the export of the execute function and another for the python implementation

modules/json/src/helpers/execute-function.js
```diff
@@ -0,0 +1,17 @@
+// This attempts to execute a function
+export function executeFunction(targetFunction, props, configuration) {
+  // Find the function
+  const Function = configuration.functions[targetFunction];
+
+  // Check that the function is in the configuration.
+  if (!Function) {
+    const {log} = configuration; // eslint-disable-line
+    const stringProps = JSON.stringify(props, null, 0).slice(0, 40);
```
felixpalmer, line 9: Move into if(log) block

modules/jupyter-widget/src/playground/create-deck.js
```diff
@@ -25,6 +25,18 @@ function extractClasses(library = {}) {
   return classesDict;
 }
 
+function extractFunctions(library = {}) {
```
felixpalmer, line 28: This is very similar to `extractClasses`. Perhaps create a shared method a supply the filter as a parameter?

modules/json/src/helpers/execute-function.js
```diff
@@ -0,0 +1,17 @@
+// This attempts to execute a function
+export function executeFunction(targetFunction, props, configuration) {
+  // Find the function
+  const Function = configuration.functions[targetFunction];
```
felixpalmer, line 4: Nitpick, not a fan of `Function` as a variable name, I'd stick with `matchedFunction`

felixpalmer (approved):
Looking good! Just some minor comments

Jesus89:
OK. I'll split the PR in two.

Jesus89:
@alasarr @felixpalmer, modules code split in this PR: https://github.com/visgl/deck.gl/pull/7253.

Jesus89:
@alasarr this PR contains only the pydeck, pydeck-carto code.

PR #7256, opened 2022-09-09
Changes extension on PR (#7255)

tuxskar:
Added some tests
Simplified how to add styles on the notify_error function
Removed most of the is_valid_carto_layer from the examples

### DIFF
--- a/bindings/pydeck-carto/examples/carto_layer_geo_query_param.py
+++ b/bindings/pydeck-carto/examples/carto_layer_geo_query_param.py
@@ -24,8 +24,6 @@ layer = pdk.Layer(
     pickable=True,
 )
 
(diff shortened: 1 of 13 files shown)


bindings/pydeck-carto/examples/carto_layer_geo_query_param.py
```diff
@@ -24,8 +24,6 @@
     pickable=True,
 )
 
-assert is_valid_carto_layer(layer, carto_auth)
```
Jesus89, line 27: Could you remove also the import?

tuxskar: sure, on it

Jesus89 (approved):

PR #7257, opened 2022-09-09
CARTO: switch to using quadbin library

felixpalmer:
#### Background

To better allow developers we have extracted the quadbin logic to a new library: https://github.com/CartoDB/quadbin-js

#### Change List
- Replace internal functions with versions from `quadbin` package
- Switch to using bigint to represent the quadbin index in memory
- Test updates

There is a change to the internal format so that we pass around a BigInt representation rather than a hex string. Basically changing 

```
type QuadbinTileIndex = {i: string};
```

to 

```
type QuadbinTileIndex = {i: bigint};
```

This will help us avoid things like this:

```
bigIntToHex(cellToParent(hexToBigInt(i)))
```

It will help performance as we won't needlessly do these conversions.


padawannn (changes requested):
I cannot approve the PR because this change breaks C4R, check [this line](https://github.com/CartoDB/carto-react/blob/master/packages/react-core/src/filters/tileFeaturesSpatialIndex.js#L76)

To merge this PR we would need to coordinate a new release of c4r and builder

padawannn (approved):
I've seen the changes in C4R, thank you for the PR

VictorVelarde:
We have the other PR in C4R, ready to coordinate here.

PR #7259, opened 2022-09-09
Add docs for pydeck-carto

tuxskar:
Added docs in Read the docs format


Jesus89:
Note: update with master when this PR is merged: https://github.com/visgl/deck.gl/pull/7255

felixpalmer:
Not my area, but from a cursory glance it appears that this is duplicating many files from the `pydeck` binding, e.g. the `scripts` appear to be the same. Would it not be possible to avoid this duplication somehow? E.g. move common files to a higher level?

Jesus89:
We have implemented a simplified version of the docs in this PR: https://github.com/visgl/deck.gl/pull/7277. We can close this one.

PR #7279, opened 2022-09-18
Add pydeck-carto in test workflow

Jesus89:
#### Change List
- Include `bindings/pydeck-carto` in the `test` workflow along with the `bindings/pydeck` one.


.github/workflows/test.yml
```diff
@@ -55,7 +55,14 @@ jobs:
...
-      - name: Run tests
+      - name: Run pydeck tests
         run: |
           cd bindings/pydeck
           make test
+
+      - name: Run pydeck-carto tests
+        run: |
+          cd bindings/pydeck-carto
+          make init
```
felixpalmer, line 66: I would put the `make init` in the "Install dependencies" section

.github/workflows/test.yml
```diff
@@ -55,7 +55,14 @@ jobs:
...
+          make lint
```
felixpalmer, line 67: Introduce `make ci` which executes `lint` & `test`? Makes it consistent with the JS and clearer from the Makefile what CI will run

felixpalmer (changes requested):

felixpalmer (approved):

PR #7287, opened 2022-09-21
CARTO: Pass access token in load options

felixpalmer:
Instead of passing the access token in the URL use a header. Enables a shared cache as tile URLs no longer need to have the access token hardcoded


alasarr (approved):
Looks good to me, but don't merge it yet, I'd like to do more acceptance tests

alasarr (approved):

PR #7311, opened 2022-10-04
carto/fetchMap: support custom markers with original colors

zbigg:
#### Change List
- carto: fetchMap supports markers with original, not-masked colors

### DIFF
--- a/modules/carto/src/api/parseMap.ts
+++ b/modules/carto/src/api/parseMap.ts
@@ -254,11 +254,12 @@ function createChannelProps(
     const maxIconSize = getMaxMarkerSize(visConfig, visualChannels);
     const {getPointRadius, getFillColor} = result;
     result.pointType = 'icon';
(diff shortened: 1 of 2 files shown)


modules/carto/src/api/parseMap.ts
```diff
@@ -254,11 +254,12 @@ function createChannelProps(
     const maxIconSize = getMaxMarkerSize(visConfig, visualChannels);
     const {getPointRadius, getFillColor} = result;
     result.pointType = 'icon';
+    const useMaskedIcons = visConfig.filled;
```
VictorVelarde, line 257: How about just destructuring here both required pieces of info...

```suggestion
    const { filled: useMaskedIcons, customMarkersUrl: fallbackUrl } = visConfig;
```

zbigg: Suggestion applied, thanks

alasarr (approved):

Issue #7316, opened 2022-10-06
[Bug] Inconsistent line height in WebMercatorView


4nthonylin:
### Description

When drawing lines with a non-zero height, the height of the lines seems to be fixed to a single height above the map plane. However, it seems to depend on the projected height of first point.

<img width="409" alt="image" src="https://user-images.githubusercontent.com/4007580/194247033-1d2d01f4-bc73-46e4-9a65-3bd1dd8f708e.png">

Line layer with following data:
```
data: [
  // red line
  {
    start: [-120, 60, 550000],
    end: [-40, 75, 550000]
  },
  // green line
  {
    start: [-40, 75, 550000],
    end: [-120, 60, 550000]
  }
],
```

Note that the red line and green line have different "projected" heights on the viewport even though they have the same altitude. I'm still quite new to the code base but couldn't figure out what it could be. Guessing it's related to the line-layer vertex shader and is not properly interpolating the line segments with the commonspace height.

Will continue digging around but wondering if anyone else with better glsl knowledge can shed some insights. 

### Flavors

- [ ] React
- [ ] Python/Jupyter notebook
- [ ] MapboxLayer
- [ ] GoogleMapsOverlay
- [ ] CartoLayer
- [ ] DeckLayer/DeckRenderer for ArcGIS

### Expected Behavior

Expect the lines to be interpolated between the projected height and the order of the start/end should not impact the rendering.

Although thinking it might be cleaner to add a new mode to project the height consistently. A constant altitude over the map plane instead of projecting the altitude to be farther away. 

### Steps to Reproduce

https://codepen.io/4nthonylin/pen/mdLGbEG

### Environment

- Framework version: https://unpkg.com/deck.gl@latest/dist.min.js
- Browser: Edge Version 106.0.1370.34
- OS: Windows/MacOS/Linux

4nthonylin:
Just wanted to add that globe view has the correct behavior.
<img width="507" alt="image" src="https://user-images.githubusercontent.com/4007580/194249328-da9f5fd2-bed2-40e6-a746-3908545afe02.png">
Only see one line since it overlaps the first line.

Pessimistress:
Fixed in v8.8.13.

4nthonylin:
Thanks for the quick fix!

4nthonylin:
Confirmed codepen now works
[...]

PR #7318, opened 2022-10-06
Fix altitude projection dependence on geometry.worldPosition

Pessimistress:
For #7316

This was introduced by #6117 to GPU-based web-mercator projection. `project_size` automatically infers the latitude from the `geometry` struct. For some layers (arc, line, path) `geometry.worldPosition` is set to the start position of the segment, but used when projecting the altitude at both the start and the end.

#### Change List
- Explicitly set reference latitude in `project_position`

### DIFF
--- a/modules/core/src/shaderlib/project/project.glsl.ts
+++ b/modules/core/src/shaderlib/project/project.glsl.ts
@@ -62,6 +62,11 @@ const float EARTH_RADIUS = 6370972.0; // meters
 const float GLOBE_RADIUS = 256.0;
 
 // returns an adjustment factor for uCommonUnitsPerMeter
+float project_size_at_latitude(float lat) {
+  float y = clamp(lat, -89.9, 89.9);
+  return 1.0 / cos(radians(y));
+}
+
 float project_size() {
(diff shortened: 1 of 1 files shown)


modules/core/src/shaderlib/project/project.glsl.ts
```diff
@@ -62,6 +62,11 @@ const float EARTH_RADIUS = 6370972.0; // meters
 const float GLOBE_RADIUS = 256.0;
 
 // returns an adjustment factor for uCommonUnitsPerMeter
+float project_size_at_latitude(float lat) {
+  float y = clamp(lat, -89.9, 89.9);
+  return 1.0 / cos(radians(y));
```
felixpalmer, line 67: Due to the symmetry you could do `1.0 / max(cos(radians(lat)), 0.001745328365898)`. I have a memory that `max` is more performant than `clamp` but I can't find the source

felixpalmer (approved):

modules/core/src/shaderlib/project/project.glsl.ts
Pessimistress: I'll make a separate PR for this. Keeping this one clean for a patch.

PR #7325, opened 2022-10-13
Fix pydeck render in Google Colab

### LINKED ISSUE #7280: [Bug] Pydeck error in Google Colab
### Description

When rendering a pydeck map in Google Colab, the map is rendered, but then it raises the following error:

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
[/usr/local/lib/python3.7/dist-packages/IPython/core/formatters.py](https://localhost:8080/#) in __call__(self, obj)
    343             method = get_real_method(obj, self.print_method)
    344             if method is not None:
--> 345                 return method()
    346             return None
    347         else:

[/usr/local/lib/python3.7/dist-packages/pydeck/bindings/deck.py](https://localhost:8080/#) in _repr_html_(self)
    221         # doesn't actually need the HTML packaging in iframe_with_srcdoc,
    222         # so we just take the HTML.data part
--> 223         return self.to_html(notebook_display=True).data

AttributeError: 'str' object has no attribute 'data'
```

This only happens when rendering the Deck object, not with the `to_html()` function, but in this case, it returns the HTML code, which I think it's not convenient.

I think this bug comes in two steps:
- Return html_str in colab: https://github.com/visgl/deck.gl/pull/5657/files#diff-4da9d85cf3adb9c98acd2c74a17bf1563587afe99e1ea93e2155c054901fe9eeR156
- Use to_html().data as repr_str: https://github.com/visgl/deck.gl/pull/5486/files#diff-bfa722fd9c40db85f3622f615422be4e4234fc86f427251ffa429612f2a44d03R195

IMO, the fix is just to remove [this line](https://github.com/visgl/deck.gl/blob/master/bindings/pydeck/pydeck/io/html.py#L156), but would like to contrast with @ajduberstein. 

### Flavors

- [ ] React
- [X] Python/Jupyter notebook
- [ ] MapboxLayer
- [ ] GoogleMapsOverlay
- [ ] CartoLayer
- [ ] DeckLayer/DeckRenderer for ArcGIS

### Expected Behavior

Render pydeck maps in *Google Colab* without the error message, or the HTML output.

### Steps to Reproduce

Execute this code in a *Google Colab* notebook:

```py
import pydeck as pdk

UK_ACCIDENTS_DATA = 'https://raw.githubusercontent.com/visgl/deck.gl-data/master/examples/3d-heatmap/heatmap-data.csv'

layer = pdk.Layer(
    'HexagonLayer',  # `type` positional argument is here
    UK_ACCIDENTS_DATA,
    get_position=['lng', 'lat'],
    auto_highlight=True,
    elevation_scale=50,
    pickable=True,
    elevation_range=[0, 3000],
    extruded=True,
    coverage=1)

# Set the viewport location
view_state = pdk.ViewState(
    longitude=-1.415,
    latitude=52.2323,
    zoom=6,
    min_zoom=5,
    max_zoom=15,
    pitch=40.5,
    bearing=-27.36)

# Combined all of it and render a viewport
pdk.Deck(layers=[layer], initial_view_state=view_state)
```

### Environment

- Framework version: deck.gl@8.8.11
- Browser: Chrome 104.0
- OS: Ubuntu 20.04.4 LTS

Jesus89:
Closes #7280

### DIFF
--- a/bindings/pydeck/pydeck/io/html.py
+++ b/bindings/pydeck/pydeck/io/html.py
@@ -157,6 +157,7 @@ def deck_to_html(
 
     if in_google_colab:
         render_for_colab(html_str, iframe_height)
+        return None
     elif not filename and as_string:
         return html_str
     elif notebook_display:
--- a/bindings/pydeck/pydeck/io/html.py
+++ b/bindings/pydeck/pydeck/io/html.py
@@ -155,9 +155,8 @@ def deck_to_html(
     if notebook_display is None:
         notebook_display = in_jupyter()
 
-    if notebook_display and in_google_colab:
+    if in_google_colab:
         render_for_colab(html_str, iframe_height)
-        return html_str
     elif not filename and as_string:
         return html_str
     elif notebook_display:
--- a/bindings/pydeck/pydeck/bindings/deck.py
+++ b/bindings/pydeck/pydeck/bindings/deck.py
@@ -230,3 +230,5 @@ class Deck(JSONMixin):
         html = self.to_html(notebook_display=True)
         if hasattr(html, 'data'):
             return html.data
+        else:
+            return ''
--- a/bindings/pydeck/pydeck/bindings/deck.py
+++ b/bindings/pydeck/pydeck/bindings/deck.py
@@ -228,7 +228,4 @@ class Deck(JSONMixin):
         # doesn't actually need the HTML packaging in iframe_with_srcdoc,
         # so we just take the HTML.data part
         html = self.to_html(notebook_display=True)
-        if hasattr(html, "data"):
-            return html.data
-        else:
-            return ""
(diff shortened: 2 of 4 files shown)


bindings/pydeck/pydeck/io/html.py
```diff
@@ -155,9 +155,8 @@ def deck_to_html(
     if notebook_display is None:
         notebook_display = in_jupyter()
 
-    if notebook_display and in_google_colab:
+    if in_google_colab:
         render_for_colab(html_str, iframe_height)
```
felixpalmer, line 159: Nit: I would explicitly `return None` here to make the intention clear

felixpalmer (approved):

ajduberstein (approved):

PR #7332, opened 2022-10-14
CARTO: fetchMap disable depthTest for point layers

felixpalmer:
#### Background

To match behavior in Kepler, `depthTest: false` needs to be set on point layers

#### Change List
- disable depthTest for point layers
- test updates

### DIFF
--- a/modules/carto/src/api/layer-map.ts
+++ b/modules/carto/src/api/layer-map.ts
@@ -160,7 +160,7 @@ export function getLayer(
       Layer: GeoJsonLayer,
       propMap: {
         columns: {
-          altitude: x => ({parameters: {depthTest: x?.fieldIdx > -1}})
+          altitude: x => ({parameters: {depthTest: x !== null}})
         },
         visConfig: {outline: 'stroked'}
       }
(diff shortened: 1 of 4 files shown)


modules/carto/src/api/layer-map.ts
```diff
@@ -153,7 +158,12 @@ export function getLayer(
   > = {
     point: {
       Layer: GeoJsonLayer,
-      propMap: {visConfig: {outline: 'stroked'}}
+      propMap: {
+        columns: {
+          altitude: x => ({parameters: {depthTest: x?.fieldIdx > -1}})
```
zbigg, line 163: It looks like we not only _disable_ depthTest in most cases, but enable for some very specific ones. Why not adding test for this scenario?

Second, i can only generate maps with `columns.altitude` with type `string`:
```
{
    "lat": "latitude",
    "lng": "longitude",
    "altitude": "year"
}
```

Please check if this condition is correct.

felixpalmer: Changed to test for "null" and I've updated the tests to explicitly test for the `depthTest: true` case

zbigg (approved):
LGTM

Issue #7370, opened 2022-11-02
[Bug] FillStyleExtension does not render non square pattern correctly


juneidy:
### Description

The pattern seems to be shrinked to fit a square. See:
![image](https://user-images.githubusercontent.com/13762553/199422805-eaae2222-1c82-4c2c-83aa-583c7a6391fe.png)

Pattern dimension: `{ width: 32, height: 19 }`

### Flavors

- [ ] React
- [ ] Python/Jupyter notebook
- [ ] MapboxLayer
- [ ] GoogleMapsOverlay
- [ ] CartoLayer
- [ ] DeckLayer/DeckRenderer for ArcGIS

### Expected Behavior

What the pattern should look like:
![image](https://user-images.githubusercontent.com/13762553/199422861-0b45a6f5-62b3-42f6-bff4-15344aa8e5a7.png)

### Steps to Reproduce

Apply this image as pattern:
![image](https://user-images.githubusercontent.com/13762553/199423290-82d16c58-5e46-46d4-8ebf-512dac7414c2.png)

### Environment

- Framework version: deck.gl@8.8.10
- Browser: Chromium 107.0
- OS: Linux 5.19

Pessimistress:
Thanks for reporting. Fixed in 8.8.17.

PR #7373, opened 2022-11-03
Fix FillStyleExtension scaling

Pessimistress:
For #7370

#### Change List
- Use 2D scaling


kylebarron (approved):

kylebarron:
@Jesus89 it looks like the new `carto-auth` version broke deck.gl tests?

Issue #7374, opened 2022-11-04
[Feat] CollideExtension


felixpalmer:
### Target Use Case

When a layer(s) is rendered with many dense features, the features will overlap, which results in a cluttered visualization. 
<img width="736" alt="Screenshot 2022-11-03 at 14 01 13" src="https://user-images.githubusercontent.com/453755/199727274-92fe05a4-bd3e-408a-8906-d613fc83044e.png">

This is especially a problem with text labels, but applies in general with other layers like the `ScatterplotLayer`: 

<img width="736" alt="Screenshot 2022-11-03 at 14 01 22" src="https://user-images.githubusercontent.com/453755/199727372-0296d2a8-2b22-4ca7-a343-09addcd2638d.png">

There is a solution for the label case in [this demo](https://github.com/visgl/deck.gl/tree/master/examples/website/tagmap), however it is implemented on the CPU and doesn't support rotating or tilting the camera. https://github.com/visgl/deck.gl/issues/6417 is also relevant for background.

Inspired by the approach taken with the `MaskExtension`, the `CollideExtension` would allow a layer to be rendered with a new `operation`, `OPERATION.COLLIDE`, so that overlapping features would be hidden, with only the frontmost features being visible.

### Proposal

To perform the collision detection, a `Layer` is created with the `OPERATION.COLLIDE` and then also rendered normally. This allows the collision hit areas to be defined independently. This is similar to a layer can be drawn with `OPERATION.MASK` and then read by other layers. The difference is that there is no analogue of `maskId` - that it, there is only one global collision render target.

## API

```javascript
const pointsProps = {
  data: PLACES, 
  pointType: 'circle', 
  getText: f => f.properties.name, 
  getTextSize: 18
};

const layers = [
  new GeoJsonLayer({
    id: 'hit-areas',
    operation: OPERATION.COLLIDE,
    ...pointsProps,
    getPointRadius: 2 * pointsProps.getPointRadius // Enlarge point to increase hit area
  }),
  new GeoJsonLayer({
    id: 'points',
    extensions: [new CollideExtension()],
    collideEnabled: true, // Defaults to true
    ...pointsProps
  })
];
```

## Technical details

[...]

PR #7375, opened 2022-11-04
CollideExtension

felixpalmer:
PoC for https://github.com/visgl/deck.gl/issues/7374 - please review new API proposal there

#### Change List
- Add new `CollideExtension` & `CollideEffect`
- New supporting `CollidePass` 
- Support for sorting by priority via `getCollidePriority`
- Test app demonstrating usage for point, text and tile layers


modules/extensions/src/collide/shader-module.ts
```diff
@@ -0,0 +1,130 @@
...
+float collide_isInBounds(vec2 texCoords, vec3 pickingColor) {
+  if (!collide_enabled) {
+    return 1.0;
+  }
+
+  float O = 0.0;
+  vec2 dx = 2.0 * vec2(1.0 / 2048.0, 0.0);
+  vec2 dy = dx.yx;
+
+  // Visibility test
```
Pessimistress, line 40: - Use for loops?
- Is it necessary to sample so many pixels? Add some comments?

felixpalmer: Switched to for-loop. I'm not 100% happy with this solution, as the object flicker when the view is tilted as the aliasing produces slightly different results between frame. A better solution might be to add some form of leaky bucket to smooth these high-frequency changes out over a few frames, but I'm not going to attempt that in this PR

modules/core/src/lib/effect.ts
```diff
@@ -11,7 +11,9 @@ export type PostRenderOptions = LayersPassRenderOptions & {
 export interface Effect {
   id: string;
   props: any;
+  useInCollide?: boolean;
```
Pessimistress, line 14: I would prefer the core to not have any knowledge of this extension. If this needs to be hardcoded anyways, hard code it in the CollideEffect, not in the core.

modules/core/src/lib/layer.ts
```diff
@@ -468,8 +468,13 @@ export default abstract class Layer<PropsT = {}> extends Component<PropsT & Requ
 
     const {props, oldProps} = params;
     const neededPickingBuffer =
-      Number.isInteger(oldProps.highlightedObjectIndex) || oldProps.pickable;
-    const needPickingBuffer = Number.isInteger(props.highlightedObjectIndex) || props.pickable;
+      Number.isInteger(oldProps.highlightedObjectIndex) ||
```
Pessimistress, line 471: Again there shouldn't be any knowledge of a particular extension in the core classes. Can we add a method to the LayerExtension interface (e.g. `needsPickingBuffer()`) instead?

modules/core/src/types/layer-props.ts
```diff
@@ -1,4 +1,4 @@
-import type {CoordinateSystem} from '../lib/constants';
+import type {CoordinateSystem, OPERATION} from '../lib/constants';
```
Pessimistress, line 1: Moving away from OPERATION, see discussion on #7604

modules/core/src/lib/effect.ts
```diff
@@ -11,7 +11,9 @@ export type PostRenderOptions = LayersPassRenderOptions & {
...
   useInPicking?: boolean;
+  didRender?: boolean;
```
Pessimistress, line 16: This seems weirdly specific. What if the Effect does not render anything? Maybe something like `get isEnabled(): boolean` more clearly reflects that it is a state of the instance?

modules/extensions/src/collide/collide.ts
```diff
@@ -0,0 +1,98 @@
...
+    this: Layer<CollideExtensionProps>,
+    params: UpdateParameters<Layer<CollideExtensionProps>>,
+    extension: this
+  ) {
+    const attributeManager = this.getAttributeManager();
+    if (!attributeManager) return;
+    const hasAttribute = attributeManager.getAttributes()['collidePriorities'];
+
+    if (Boolean(this.props.getCollidePriority) && !hasAttribute) {
+      attributeManager.add({
```
Pessimistress, line 84: Move this to `initializeState` now that we fixed default props for extensions? A constant attribute will have virtually no cost.

modules/extensions/src/collide/shader-module.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  // Visibility test, sample area of 5x5 pixels in order to fade in/out.
+  // Due to the locality, the lookups will be cached
+  // This reduces the flicker present when objects are shown/hidden
+  const int N = 2;
+  float accumulator = 0.0;
+  vec2 step = vec2(1.0 / project_uViewportSize);
+
+  const float floatN = float(N);
+  vec2 delta = -floatN * step;
```
Pessimistress, line 41: How's the perf here?

modules/extensions/src/collide/collide-effect.ts
```diff
@@ -0,0 +1,254 @@
...
+    gl: WebGLRenderingContext,
+    {effects: allEffects, layers, layerFilter, viewports, onViewportActive, views}: PreRenderOptions
+  ): void {
+    if (!this.dummyCollideMap) {
+      this.dummyCollideMap = new Texture2D(gl, {width: 1, height: 1});
+    }
+
+    const collideLayers = layers.filter(
+      // @ts-ignore
+      ({props: {visible, extensions, collideGroup}}) =>
```
Pessimistress, line 41: It will be easier to just add a `collideEnabled` prop and default to `true` following the other extensions' convention.

modules/extensions/src/collide/collide-effect.ts
```diff
@@ -0,0 +1,254 @@
...
+        visible &&
+        extensions.find(e => e.constructor.extensionName === 'CollideExtension') &&
+        collideGroup
+    ) as Layer<CollideExtensionProps>[];
+    if (collideLayers.length === 0) {
+      this.channels = {};
+      return;
+    }
+
+    const effects = allEffects?.filter(({useInCollide}) => useInCollide);
```
Pessimistress, line 51: In this case... could just filter by `useInPicking` here?

modules/extensions/src/collide/collide-effect.ts
```diff
@@ -0,0 +1,254 @@
...
+    // Collect layers to render
+    const channels = this._groupByCollideGroup(gl, collideLayers);
+
+    const viewport = viewports[0];
+    const viewportChanged =
+      !this.lastViewport || !this.lastViewport.equals(viewport) || otherEffectRendered;
+
+    // Resize framebuffers to match canvas
+    for (const collideGroup in channels) {
+      const collidePass = this.collidePasses[collideGroup];
```
Pessimistress, line 65: Nit: I know I set a bad precedence in `MaskPass`, but a pass shouldn't really own any resources...

modules/extensions/src/collide/collide-effect.ts
```diff
@@ -0,0 +1,254 @@
...
+    }
+    this.channels = {};
+    for (const collidePass of Object.values(this.collidePasses)) {
+      collidePass.delete();
+    }
+    this.collidePasses = {};
+    this.lastViewport = undefined;
+  }
+
+  // Debug show FBO contents on screen
```
Pessimistress, line 201: Shall we add this as a common utility somewhere?

modules/extensions/src/collide/collide-effect.ts
```diff
@@ -0,0 +1,254 @@
...
+      return;
+    }
+
+    const renderInfoUpdated =
+      // If render info is new
+      renderInfo === oldRenderInfo ||
+      // If sublayers have changed
+      oldRenderInfo.layers.length !== renderInfo.layers.length ||
+      // If a sublayer's positions have been updated, the cached bounds will change shallowly
+      renderInfo.layerBounds.some((b, i) => b !== oldRenderInfo.layerBounds[i]);
```
Pessimistress, line 111: `equals` from `@math.gl/core` will deal with small precision changes

modules/core/src/lib/effect-manager.ts
```diff
@@ -16,9 +17,10 @@ export default class EffectManager {
     this._setEffects([]);
   }
 
-  addDefaultEffect(effect: Effect) {
-    if (!this._defaultEffects.find(e => e.constructor === effect.constructor)) {
-      this._defaultEffects.push(effect);
+  addDefaultEffect(effect: Effect, slot: number) {
```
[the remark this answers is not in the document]
felixpalmer: Are there really many people out there writing custom effects? 

My understanding is that the purpose of the `deck._addDefaultEffect` API is to be a internal way to allow extensions to register a default effect without having to bloat the core. Having a full dependency graph as well as an order seems overkill. As you mentioned elsewhere, we could rethink the Effect API in the future. As for a stopgap, how about spacing the internal effects so that an external effect can always be placed inbetween

```js
const EffectOrder = {
  MaskEffect: 10,
  CollideEffect: 20
}
```

modules/core/src/lib/effect.ts
felixpalmer: The whole point is to detect if something was rendered. In this case if the mask map is updated, as this requires the collide map to be redrawn. 

Another option here would be to add some state to the mask map (e.g. `MaskEffect.maskMap.userData.version`) which `MaskPass.render` would increment, but I prefer to keep it at the effect level.

modules/extensions/src/collide/shader-module.ts
felixpalmer: Good, I don't see any difference between doing a single lookup and sampling over an area and I think having the fading makes a big difference to the UX

modules/core/src/lib/effect-manager.ts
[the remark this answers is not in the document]
Pessimistress: Ok. Can we at least make it so that effects are sorted by this number, instead of overwritten?

modules/core/src/lib/effect.ts
Pessimistress: My worry is that this field is essentially a dirty flag, but there is no mechanism to clear it. We have like 5 different effects now, and we shouldn't add a new flag then leave it up to each effect class to manage.

May I propose the following change:

- `Effect.preRender(options)` can return an arbitrary value
- The render pipeline assembles these return values and pass it downstream, i.e. `options.preRenderStats["mask-effect"] = effect.preRender(options)`
- `CollideEffect.preRender` can then look at `options.preRenderStats` to see the preRender result of its dependencies

Would that solve your problem?

modules/core/src/lib/layer.ts
```diff
@@ -468,8 +468,13 @@ export default abstract class Layer<PropsT = {}> extends Component<PropsT & Requ
...
+      oldProps.pickable ||
+      oldProps.extensions?.some(e => e.needsPickingBuffer());
+    const needPickingBuffer =
+      Number.isInteger(props.highlightedObjectIndex) ||
+      props.pickable ||
+      props.extensions?.some(e => e.needsPickingBuffer());
```
Pessimistress, line 477: `extensions` is always an array (it is used elsewhere without a null check)

modules/core/src/lib/layer.ts
```diff
@@ -468,8 +468,13 @@ export default abstract class Layer<PropsT = {}> extends Component<PropsT & Requ
 
     const {props, oldProps} = params;
     const neededPickingBuffer =
-      Number.isInteger(oldProps.highlightedObjectIndex) || oldProps.pickable;
-    const needPickingBuffer = Number.isInteger(props.highlightedObjectIndex) || props.pickable;
+      Number.isInteger(oldProps.highlightedObjectIndex) ||
+      oldProps.pickable ||
+      oldProps.extensions?.some(e => e.needsPickingBuffer());
```
Pessimistress, line 473: Not blocking, but I don't think this will work - if `extension.needsPickingBuffer()` accesses layer props, it will always end up reading the current value, not the previous value. `needsPickingBuffer` may have to be cached in `internalState`.

modules/extensions/src/collide/collide.ts
```diff
@@ -0,0 +1,92 @@
...
+  getShaders(this: Layer<CollideExtensionProps>): any {
+    return {modules: [collide]};
+  }
+
+  /* eslint-disable camelcase */
+  draw(this: Layer<CollideExtensionProps>, {uniforms, context, moduleParameters}: any) {
+    const {collideEnabled, collideGroup} = this.props;
+    const {drawToCollideMap, collideFBOs = {}} = moduleParameters;
+    const collideGroups = Object.keys(collideFBOs);
+    const enabled = collideEnabled && collideGroup && collideGroups.includes(collideGroup);
```
Pessimistress, line 53: Is it necessary to check `collideGroup` here? It looks like `preRender` ensures that only layers belonging to the same group are drawn.

Pessimistress: You may want to check `collideEnabled` in `preRender` too since that affects whether the collide map needs rerender.

modules/extensions/src/collide/collide.ts
```diff
@@ -0,0 +1,92 @@
...
+    uniforms.collide_enabled = enabled;
+
+    if (drawToCollideMap) {
```
Pessimistress, line 56: I recommend moving this logic to `shaderModule.getUniforms`. The MaskExtension was calculating some of the uniforms in `draw` because it needed the layer's projection methods.

modules/extensions/src/collide/collide-effect.ts
```diff
@@ -0,0 +1,252 @@
...
+      oldRenderInfo.layers.length !== renderInfo.layers.length ||
+      // If a sublayer's bounds have been updated
+      renderInfo.layerBounds.some((b, i) => !equals(b, oldRenderInfo.layerBounds[i]));
+
+    this.channels[collideGroup] = renderInfo;
+
+    if (renderInfoUpdated || viewportChanged) {
+      this.lastViewport = viewport;
+      const collideFBO = this.collideFBOs[collideGroup];
+      const collidePass = this.collidePasses[collideGroup];
```
Pessimistress, line 135: Is there any reason `CollidePass` cannot be reused across multiple channels?

modules/extensions/src/collide/collide-effect.ts
```diff
@@ -0,0 +1,252 @@
...
+        this.destroyPass(collideGroup);
+      }
+    }
+
+    return channelMap;
+  }
+
+  getModuleParameters(): {collideFBOs: Record<string, Framebuffer>; dummyCollideMap: Texture2D} {
+    const {collideFBOs, dummyCollideMap} = this;
+    return {collideFBOs, dummyCollideMap};
```
Pessimistress, line 200: Just pass a single FBO by `collideFBO[layer.props.collideGroup]`. The less the layer knows about the internals of the effect, the better.

modules/extensions/src/collide/collide-effect.ts
felixpalmer: Good point, just using one now

modules/extensions/src/collide/collide.ts
```diff
@@ -0,0 +1,92 @@
...
+    }
+  }
+
+  initializeState(this: Layer<CollideExtensionProps>, context: LayerContext, extension: this) {
+    if (!extension.isEnabled(this)) {
+      return;
+    }
+    this.context.deck?._addDefaultEffect(new CollideEffect());
+    const attributeManager = this.getAttributeManager();
+    attributeManager!.add({
```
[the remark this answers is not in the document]
felixpalmer: The `isEnabled` check will handle this

modules/extensions/src/collide/collide.ts
felixpalmer: It is being checked https://github.com/visgl/deck.gl/pull/7375/files#diff-79baa7569c66db2edb07b9e7d33a5567b1a23ca5fc90b7864c59a0d7114389aeR59

modules/extensions/src/collide/collide-pass.ts
```diff
@@ -0,0 +1,45 @@
...
+        colorMask,
+        depthTest: true,
+        depthRange: [0, 1]
+      },
+      () => this.render({...options, target, pass: 'collide'})
+    );
+  }
+
+  shouldDrawLayer(layer) {
+    return (
```
Pessimistress, line 29: I still don't get why you need this check. In what scenario will this check fail?

- CollideEffect already filters the layers list by checking `collideEnabled` so no disabled  layers could be passed into this step. 
- We don't look for a particular extension anywhere else, the existence of this prop is good enough indication.

Pessimistress (approved):

modules/extensions/src/collide/collide.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    if (drawToCollideMap) {
+      // Override any props with those defined in collideTestProps
+      // @ts-ignore
+      this.props = this.clone(this.props.collideTestProps).props;
+    }
+  }
+
+  initializeState(this: Layer<CollideExtensionProps>, context: LayerContext, extension: this) {
+    if (!extension.isEnabled(this)) {
```
Pessimistress, line 63: This seems unnecessary. If `isEnabled` is only called in `initializeState` then just move the one-line check in here?

modules/extensions/src/collide/collide-pass.ts
```diff
@@ -0,0 +1,45 @@
...
+    const padding = 1;
+
+    return withParameters(
+      gl,
+      {
+        scissorTest: true,
+        scissor: [padding, padding, target.width - 2 * padding, target.height - 2 * padding],
+        clearColor: [0, 0, 0, 0],
+        blend: false,
+        colorMask,
```
Pessimistress, line 20: Remove, it's already the default

modules/extensions/src/collide/collide-pass.ts
felixpalmer: Removed `shouldDrawLayer` - I think you're right

PR #7383, opened 2022-11-07
CARTO: Support JSON format in QuadbinLayer

felixpalmer:
Regression introduced by https://github.com/visgl/deck.gl/pull/7257

#### Change List
- Convert hexidecimal ids when reading JSON data
- Test update


ibgreen (commented):
Can we add an actual test case that triggers this?

ibgreen (approved):

Issue #7184, opened 2022-08-06
Mask Extension does not seem to work with HeatmapLayer [Bug]


andresramosp:
### Description

I took the official example for HeatmapLayer (https://github.com/visgl/deck.gl/tree/8.8-release/examples/website/heatmap), and just changed the dataset and added a new GeoJsonLayer with a polygon to mask the heatmap. 

This image shows the GeoJsonLayer and the HeatmapLayer without applying the mask extension yet:

![image](https://user-images.githubusercontent.com/44118551/183243799-ea285fcb-97cf-4979-b215-4aa5341e09dc.png)

Then, when applying the mask extension, the first rendering just shows the heatmap without mask:, 

![image](https://user-images.githubusercontent.com/44118551/183243830-dcfc3a85-160c-4a16-811b-53b626cf59ae.png)

and when I zoom the map, the heatmap dissapears and I got the following warning on console:

[.WebGL-00002132058C4E00] GL_INVALID_OPERATION: Feedback loop formed between Framebuffer and active Texture.

If I replace the HeatmapLayer by a ScatterplotLayer, all works as expected:

### Flavors

- [X] React
- [ ] Python/Jupyter notebook
- [ ] MapboxLayer
- [ ] GoogleMapsOverlay
- [ ] CartoLayer
- [ ] DeckLayer/DeckRenderer for ArcGIS

### Expected Behavior

The heatmap clipped by the polygon mask.

### Steps to Reproduce

https://github.com/andresramosp/DeckGL_Heatmap_Masked

### Environment

- Framework version: 8.8.0
- Browser: Chrome
- OS: Windows

### Logs

[.WebGL-00002132058C4E00] GL_INVALID_OPERATION: Feedback loop formed between Framebuffer and active Texture.

sliterok:
I have same problem with ScenegraphLayer. Is it expected behavior?

felixpalmer:
@sliterok I don't think it should be related. Could you make a separate issue with a jsfiddle demonstrating your issue?

andresramosp:
Hi Felix! Any estimate for the fix for Mask on HeatmapLayer? Did you figure out the problem?

PR #7397, opened 2022-11-10
Support MaskExtension with HeatmapLayer

felixpalmer:
For https://github.com/visgl/deck.gl/issues/7184

#### Background

https://user-images.githubusercontent.com/453755/201100833-4ca2fcc3-0f99-4aaf-aef5-c96201ba7dd1.mov

Due to the aggregation the mask extension wasn't working with the `HeatmapLayer`. This PoC works well, but it doesn't fit cleanly with the deck architecture. @Pessimistress could you give some input on:

- How to cleanly access the `EffectManager` in `AggregationLayer.getModuleParameters`
- How best to get the `weightsTransform` to execute *after* the mask is rendered. I couldn't find a way without adding the `postEffect` hook

#### Change List
- Provide correct uniforms to shaders in `weightsTransform` & `TriangleLayer`
- Update shaders to work with extension

#### TODO
- Agree on how to cleanly integrate
- Revert changes to test app
- Render tests

### DIFF
--- a/modules/aggregation-layers/src/aggregation-layer.ts
+++ b/modules/aggregation-layers/src/aggregation-layer.ts
@@ -53,7 +53,7 @@ export default abstract class AggregationLayer<
   updateState(opts: UpdateParameters<this>) {
     super.updateState(opts);
     const {changeFlags} = opts;
-    if (changeFlags.extensionsChanged) {
+    if (changeFlags.extensionsChanged || changeFlags.propsChanged) {
       const shaders = this.getShaders({});
       if (shaders && shaders.defines) {
         shaders.defines.NON_INSTANCED_MODEL = 1;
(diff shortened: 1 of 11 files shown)


Pessimistress (changes requested):
Layers should not have any knowledge of a particular extension. I believe that the transform (which generates the heatmap texture) is already created with the shader of the MaskExtension. The issue is that when the layer updates its module settings, parameters are only sent to the model that draws to screen, bypassing the transform.

andresramosp:
Hello, this is still in progress? any estimation of release? Thanks.

andresramosp:
Hello, sorry for asking again, is there any forecast to release this fix? In my company we need to mask the heatmap layer at port level very accurately and it would be nice to know if we can count on it or look for another solution. Thanks!

kv9898:
I'm also wondering... This is so useful

Issue #7243, opened 2022-09-06
[Bug] map query more than 2048 chars doesn't send geoColumn


jantolg:
### Description

Using a CartoLayer with geomColumn from spatial index (h3/quadbin) including a source type MAP_TYPES.QUERY and a SQL sentence greather than 2048 characters, internally it does a request from POST to maps api, but this request not include geoColumn, aggregationExp nor aggregationResLevel parameter

![image](https://user-images.githubusercontent.com/17105273/188721070-51dc53d8-32aa-40c0-b244-cd64ff6985e7.png)

### Flavors

- [ ] React
- [ ] Python/Jupyter notebook
- [ ] MapboxLayer
- [ ] GoogleMapsOverlay
- [X] CartoLayer
- [ ] DeckLayer/DeckRenderer for ArcGIS

### Expected Behavior

The same behaviour like a GET request

### Steps to Reproduce

Create a CartoLayer with a source as sql query greather than 2048 chars and include a geomColumn diferent as default (geom)

### Environment

- Framework version:
- Browser:
- OS:

jantolg:
cc: @Josmorsot @VictorVelarde

VictorVelarde:
Thx Janto, we'll address it asap

PR #7404, opened 2022-11-11
CARTO: Include aggregation and geoColumn in POST

felixpalmer:
For https://github.com/visgl/deck.gl/issues/7243

#### Change List
- Include aggregation and geoColumn in POST
- Test


zbigg (approved):
LGTM

Issue #7439, opened 2022-11-21
[Feat] Support inverted mask in MaskExtension


felixpalmer:
### Target Use Case

When reading a mask layer it would be useful to be able to invert the mask operation, i.e. to show the objects that would normally be hidden and vice-versa.

https://user-images.githubusercontent.com/453755/203010776-54920172-4b89-46a9-9ec5-d7fb068d6a6a.mov

### Proposal

Add a new Boolean prop to the `MaskExtension`, `maskInverted`, which inverts the masking operation on the rendered layer. This allows the same source mask to be used for both the normal and inverted masking operation.

```javascript
new ScatterplotLayer({
  id: 'sources',
  data: sources,
  extensions: [new MaskExtension()],
  maskId: 'mask',
  maskInverted: true
})
```

PR #7440, opened 2022-11-21
Add maskInverted prop to MaskExtension

felixpalmer:
https://github.com/visgl/deck.gl/issues/7439

#### Change List
- Add `maskInverted` prop to `MaskExtension`
- Update test app
- Update tests

### DIFF
--- a/docs/api-reference/extensions/mask-extension.md
+++ b/docs/api-reference/extensions/mask-extension.md
@@ -89,6 +89,10 @@ Id of the layer that defines the mask. The mask layer must use the prop `operati
 
 ![maskByInstance](https://raw.githubusercontent.com/visgl/deck.gl-data/master/images/docs/mask-by-instance.png)
 
+##### `maskInverted` (Boolean, optional)
+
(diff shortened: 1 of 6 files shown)


docs/api-reference/extensions/mask-extension.md
```diff
@@ -89,6 +89,10 @@ Id of the layer that defines the mask. The mask layer must use the prop `operati
 
 ![maskByInstance](https://raw.githubusercontent.com/visgl/deck.gl-data/master/images/docs/mask-by-instance.png)
 
+##### `maskInverted` (Boolean, optional)
+
+When `maskInverted` is true the result of the masking operation is inverted. Inversion is applied when reading the mask, thus is it possible to use the same mask normally as well as using the inverse on different layers. Defaults to `false`.
```
Pessimistress, line 94:
```suggestion
When `maskInverted` is true the result of the masking operation is inverted. Inversion is applied when reading the mask, thus it is possible to use the same mask normally on some layers and inverted on other layers. Defaults to `false`.
```

Pessimistress (approved):

PR #7441, opened 2022-11-22
Fix BitmapLayer translucency

Pessimistress:
For #7427

#### Background

When the `transparentColor` prop was added for an Uber internal use case (I believe back during v6), it changed the default behavior of `BitmapLayer` with `opacity < 1`. When a layer is drawn into the screen buffer, our default blending settings assume that it is NOT premultiplied. The way `BitmapLayer` currently handles the blending between `transparentColor` and the image pixel, it outputs a premultiplied color. For example, if a pixel from `image` has value `[a, a, a, a]`, with the default `transparentColor` prop, the fragment shader outputs `[a*a, a*a, a*a, a]`. This leads to overly muted output as alpha is technically applied twice.

I created a Codepen to showcase the issue in 8.8: https://codepen.io/Pessimistress/pen/JjZpREP?editors=0010. 

#### Change List
- BitmapLayer blends `transparentColor` and pixel color more naturally
- Clarify the purpose of `transparentColor` in documentation
- Update render test reference image


chrisgervang (approved):
Premultiplication strikes again.

felixpalmer (approved):

PR #7445, opened 2022-11-22
add type for viewState

arthurgailes:
Add type definition for viewState, which causes type warnings when setting the state of a predefined object, even when that object has correct parameters

Used "any" type for the following properties: normalize, position, transitionEasing,  transitionInterpolator; which should probably be defined.

TS Warning Example:
```
Argument of type 'Record<string, any>' is not assignable to parameter of type 'SetStateAction<{ longitude: number; latitude: number; zoom: number; pitch: number; minZoom: number; }>'.
  Type 'Record<string, any>' is missing the following properties from type '{ longitude: number; latitude: number; zoom: number; pitch: number; minZoom: number; }': longitude, latitude, zoom, pitch, minZoomts(2345)
```

### DIFF
--- a/modules/core/src/controllers/controller.ts
+++ b/modules/core/src/controllers/controller.ts
@@ -110,14 +110,34 @@ export type InteractionState = {
   isZooming?: boolean;
 }
 
+export type ViewState = {
+  zoom: number;
+  longitude: number;
+  latitude: number;
+  bearing?: number;
+  pitch?: number;
+  transitionDuration?: number;
+  width?: number;
+  height?: number;
+  altitude?: number;
+  maxPitch?: number;
+  maxZoom?: number;
+  minPitch?: number;
+  minZoom?: number;
+  normalize?: any;
+  position?: number[];
+  transitionEasing?: any;
+  transitionInterpolator?: any;
+}
+
 /** Parameters passed to the onViewStateChange callback */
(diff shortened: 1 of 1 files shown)


modules/core/src/controllers/controller.ts
```diff
@@ -110,14 +110,34 @@ export type InteractionState = {
...
+  position?: number[];
+  transitionEasing?: any;
+  transitionInterpolator?: any;
+}
+
 /** Parameters passed to the onViewStateChange callback */
 export type ViewStateChangeParameters = {
   /** The next view state, either from user input or transition */
-  viewState: Record<string, any>;
+  viewState: ViewState;
```
felixpalmer, line 136: This is assuming a `WebMercatorViewport`, we also have other view types, e.g. https://deck.gl/docs/api-reference/core/orthographic-view#view-state

Adding types here is a good idea, but rather than duplicating the type here, it is better to import the types used elsewhere

felixpalmer (changes requested):

Radik24:
I'm not sure right now what is returned onViewStateChange, but what about multiple views? I often use multiple views and it requires something like:

```
type: ViewState {
  [key: string]: SomeType
}
```

chrisgervang:
Closed by #8663

PR #7465, opened 2022-12-02
Use default loader to load icons

Pessimistress:
For https://github.com/visgl/deck.gl/discussions/7449
The discrepancy comes from loading images with and without premultiplication.

The default ImageLoader is registered with options: https://github.com/visgl/deck.gl/blob/691e3f5bce548aacbf00a16e65c87fae5501b4f2/modules/core/src/lib/init.ts#L56-L60

This is used for loading the `iconAtlas` prop.

When using auto packing, the `IconManager` calls `load` directly with `ImageLoader`, in which case the default options are not used.

#### Change List
- Use default loader to load individual icons

### DIFF
--- a/modules/layers/src/icon-layer/icon-manager.ts
+++ b/modules/layers/src/icon-layer/icon-manager.ts
@@ -422,7 +422,7 @@ export default class IconManager {
 
     for (const icon of icons) {
       this._pendingCount++;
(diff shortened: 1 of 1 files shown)


ibgreen (approved):
Highlights the problems with designing global registration systems...

Issue #7474, opened 2022-12-05
[Bug] Lines visible between cells in QuabinLayer


felixpalmer:
### Description

Thin lines are visible between cells in the `QuadbinLayer`

<img width="528" alt="Screenshot 2022-12-05 at 12 29 31" src="https://user-images.githubusercontent.com/453755/205626738-8aeb97ac-129d-41a2-915a-6a2b94403407.png">

### Flavors

- [ ] React
- [ ] Python/Jupyter notebook
- [ ] MapboxLayer
- [ ] GoogleMapsOverlay
- [ ] CartoLayer
- [ ] DeckLayer/DeckRenderer for ArcGIS

### Expected Behavior

No lines between adjacent cells

### Steps to Reproduce

First example in `test/apps/carto-map` has this issue

### Environment

- Framework version: deck.gl@8.8.20
- Browser: all
- OS: all

PR #7475, opened 2022-12-05
To avoid z-fighting reduce polygon footprint when extruding

felixpalmer:
For https://github.com/visgl/deck.gl/issues/7474

#### Change List
- Only reduce cell coverage when extruding


zbigg (approved):
Look good

Issue #7458, opened 2022-11-30
[Bug] Overwritten parameters in MapboxOverlay


chrisgervang:
### Description

`MapboxOverlay` makes no guarantee for when `parameters` will be set. If the user sets some `parameters` of their own, they'll either overwrite what is set by the overlay or the overlay will eventually overwrite what the user set. The overlay waits to fully initialize and sets the parameters until the first time there are layers to draw (when `layers.length > 0`). 

I'm not sure there's a way for the user decide how these internal props will mix with their props - `map.on('load')` is insufficient since the first deck layer may not be present by that time, and the internal instance is tucked under many private layers.

This issue affects the deck instance created within `MapboxOverlay` when `interleaved: true`, and may affect `interleaved: false` as well. The alpha doesn't blend correctly without these parameters, so this issue leads to visual artifacts. 
https://github.com/visgl/deck.gl/blob/1342a961ba9ce320ddd0eebe1cd53a021c3c8382/modules/mapbox/src/deck-utils.ts#L46-L54

For context, the call order is:
- After `basemap.map.addControl(deck);`, onAdd is called, then  https://github.com/visgl/deck.gl/blob/1342a961ba9ce320ddd0eebe1cd53a021c3c8382/modules/mapbox/src/mapbox-overlay.ts#L91-L92
- Which calls `resolveLayers` and adds new layers. Note `deck` is defined, so it's "external". https://github.com/visgl/deck.gl/blob/1342a961ba9ce320ddd0eebe1cd53a021c3c8382/modules/mapbox/src/resolve-layers.ts#L48
- Once the layer is added, onAdd is called https://github.com/visgl/deck.gl/blob/1342a961ba9ce320ddd0eebe1cd53a021c3c8382/modules/mapbox/src/mapbox-layer.ts#L35-L37 
- Then `getDeckInstance`. I don't believe `map.__deck` is defined before this point.
https://github.com/visgl/deck.gl/blob/1342a961ba9ce320ddd0eebe1cd53a021c3c8382/modules/mapbox/src/deck-utils.ts#L29-L31
- And finally the parameter props are set https://github.com/visgl/deck.gl/blob/ac418c1da58ee505293be0babcf5fae543867c11/modules/mapbox/src/deck-utils.ts#L77

I see two issues in this.
[...]

PR #7477, opened 2022-12-05
fix(mapbox) merge deck props when interleaving

chrisgervang:
For #7458 

#### Change List
- Set interleaving deck props in MapboxOverlay when control is added instead of waiting until a layer is added.
- Synchronize Mapbox overlay props with props from the overlay's deck instance when interleaving.

### DIFF
--- a/modules/mapbox/src/deck-utils.ts
+++ b/modules/mapbox/src/deck-utils.ts
@@ -30,33 +30,23 @@ export function getDeckInstance({
     return map.__deck;
   }
 
+  // Only initialize certain props once per context
   const customRender = deck?.props._customRender;
-  const useDevicePixels =
-    deck && 'useDevicePixels' in deck.props ? deck.props.useDevicePixels : true;
 
-  const deckProps: DeckProps = {
-    useDevicePixels,
-    _customRender: () => {
-      map.triggerRepaint();
-      // customRender may be subscribed by DeckGL React component to update child props
-      // make sure it is still called
-      // Hack - do not pass a redraw reason here to prevent the React component from clearing the context
-      // Rerender will be triggered by MapboxLayer's render()
-      customRender?.('');
-    },
-    // TODO: import these defaults from a single source of truth
-    parameters: {
-      depthMask: true,
-      depthTest: true,
-      blend: true,
-      blendFunc: [gl.SRC_ALPHA, gl.ONE_MINUS_SRC_ALPHA, gl.ONE, gl.ONE_MINUS_SRC_ALPHA],
-      polygonOffsetFill: true,
-      depthFunc: gl.LEQUAL,
-      blendEquation: gl.FUNC_ADD,
-      ...(deck && deck.props.parameters)
-    },
-    views: (deck && deck.props.views) || [new MapView({id: 'mapbox'})]
-  };
+  const deckProps = getDeckProps({
+    gl,
+    currProps: {
+      ...deck?.props,
+      _customRender: () => {
+        map.triggerRepaint();
+        // customRender may be subscribed by DeckGL React component to update child props
+        // make sure it is still called
+        // Hack - do not pass a redraw reason here to prevent the React component from clearing the context
+        // Rerender will be triggered by MapboxLayer's render()
+        customRender?.('');
+      }
+    }
+  });
 
   let deckInstance: Deck;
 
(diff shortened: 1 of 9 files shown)


modules/mapbox/src/deck-utils.ts
```diff
@@ -33,7 +33,7 @@ export function getDeckInstance({
   const customRender = deck?.props._customRender;
 
   const deckProps: DeckProps = {
-    useDevicePixels: true,
+    useDevicePixels: (deck && deck.props.useDevicePixels) || true,
```
Pessimistress, line 36: This expression will always be true

modules/mapbox/src/deck-utils.ts
```diff
@@ -44,6 +44,7 @@ export function getDeckInstance({
     },
     // TODO: import these defaults from a single source of truth
     parameters: {
+      ...(deck && deck.props.parameters),
```
Pessimistress, line 47: user defined parameters should override defaults

modules/mapbox/src/mapbox-overlay.ts
```diff
@@ -95,6 +95,14 @@ export default class MapboxOverlay implements IControl {
       gl: map.painter.context.gl
     });
 
+    this._deck = getDeckInstance({
+      map,
+      // @ts-ignore non-public map property
+      gl: map.painter.context.gl,
+      deck: this._deck
```
Pessimistress, line 102:
```suggestion
      deck: new Deck({
        ...this._props,
        // @ts-ignore non-public map property
        gl: map.painter.context.gl
      });
```

modules/mapbox/src/mapbox-overlay.ts
```diff
@@ -95,6 +95,14 @@ export default class MapboxOverlay implements IControl {
...
+    });
+    this.setProps(this._deck.props);
```
Pessimistress, line 104: Is this still necessary?

yarn.lock
```diff
@@ -4604,9 +4604,9 @@ camelcase@^5.0.0, camelcase@^5.3.1:
   integrity sha512-L28STB170nwWS63UjtlEOE3dldQApaJXZkOI1uMFfzf3rRuPegHaHesyee+YxQ+W6SvRDQV6UrdOdRiR153wJg==
 
 caniuse-lite@^1.0.30001254:
-  version "1.0.30001374"
-  resolved "https://registry.npmjs.org/caniuse-lite/-/caniuse-lite-1.0.30001374.tgz"
-  integrity sha512-mWvzatRx3w+j5wx/mpFN5v5twlPrabG8NqX2c6e45LCpymdoGqNvRkRutFUqpRTXKFQFNQJasvK0YT7suW6/Hw==
+  version "1.0.30001436"
```
Pessimistress, line 4607: Can you revert this? Not related to this PR.

modules/mapbox/src/mapbox-overlay.ts
chrisgervang: If the user changes `parameters` at some point later on, they'll be removing the necessary ones set by `getDeckInstance`. How should they access the necessary parameters to keep them intact?

I'm not sure this change addresses the problem - it only synchronizes this._props and this._deck.props, but it'd still be up to the user to spread these internal props into their parameters before calling setProps.

modules/mapbox/src/deck-utils.ts
chrisgervang: This also wouldn't won't work since `useDevicePixels` can be validly be set to `false`. Do you have a preferred way to check if the object is set (just _not_ undefined)?

chrisgervang: `'useDevicePixels' in deck.props` should do it.

modules/mapbox/src/mapbox-overlay.ts
chrisgervang: I split the prop merging out and call it in `getDeckInstance` and the overlay's `setProps`. I do not attempt to sync `overlay._props` and `overlay._deck.props` anymore, since it shouldn't be necessary because most default props are reapplied whenever a user changes them.

modules/mapbox/src/deck-utils.ts
```diff
@@ -30,30 +30,23 @@ export function getDeckInstance({
...
-      depthFunc: gl.LEQUAL,
-      blendEquation: gl.FUNC_ADD
-    },
-    views: (deck && deck.props.views) || [new MapView({id: 'mapbox'})]
-  };
+  const deckProps = getDeckProps({
+    gl,
+    currProps: {
+      ...deck?.props,
+      _customRender: () => {
```
chrisgervang, line 40: I choose to not reapply the `_customRender` prop since it would stack infinitely. Seems to only need to be set once anyways.

modules/mapbox/src/deck-utils.ts
```diff
@@ -94,6 +87,29 @@ export function getDeckInstance({
   return deckInstance;
 }
 
+export function getDeckProps({gl, currProps}: {gl: WebGLRenderingContext; currProps: DeckProps}) {
```
chrisgervang, line 90: We could add `@luma.gl/constants` to the deps instead of an arg. Preference?

modules/mapbox/src/mapbox-overlay.ts
```diff
@@ -71,10 +77,15 @@ export default class MapboxOverlay implements IControl {
     });
     this._container = container;
 
-    this._deck = new Deck({
-      ...this._props,
-      parent: container,
-      viewState: getViewState(map)
+    this._deck = getDeckInstance({
```
chrisgervang, line 80: These props also need to set in the overlaid use case in the event `layers: []`

modules/mapbox/src/deck-utils.ts
Pessimistress: Always import constants from `@luma.gl/constants`.

modules/mapbox/src/mapbox-overlay.ts
Pessimistress: Overlaid mode should not call `getDeckInstance`. There is no context sharing and deck canvas is not dependent on Mapbox behaviors.

Pessimistress (changes requested):

modules/mapbox/src/deck-utils.ts
```diff
@@ -94,6 +85,29 @@ export function getDeckInstance({
   return deckInstance;
 }
 
+export function getDeckProps(currProps: DeckProps) {
```
Pessimistress, line 88: Nit: call it `getInterleavedProps` for clarity?

Pessimistress (approved):

PR #7493, opened 2022-12-08
Update loading-data.md

harderthan:
For the doc

#### Change List
- wrote the importing component to help beginners understand.


ibgreen (approved):

Pessimistress (commented):
Why do we need this? What exactly is hard to understand? Every other example on this page also reference some layer, why only update this one?

felixpalmer:
> Why do we need this? What exactly is hard to understand? Every other example on this page also reference some layer, why only update this one?

Agreed, this change just makes the docs more verbose

Issue #7509, opened 2022-12-12
[Bug] DataFilterExtension very poor performance 


AnthonyWhitaker:
### Description

Extremely poor performance (3-4 FPS) when using the data-filter extension on a PolygonLayer where the geometry *would* render hundreds of overlapping fullscreen polygons. Even if the majority, or all, of the polygons are filtered - the fragment shader must run countless times only to `discard` all of the fragments.

### Flavors

- [X] React
- [ ] Python/Jupyter notebook
- [ ] MapboxLayer
- [ ] GoogleMapsOverlay
- [ ] CartoLayer
- [ ] DeckLayer/DeckRenderer for ArcGIS

### Expected Behavior

Improved performance.

### Steps to Reproduce

Create a geojson with 100s of whole world quads and attempt to render using the data-filter.

### Environment

- Framework version: deck.gl@8.8.6
- Browser: Chrome 108.0.5358.94 (arm64)
- OS: Mac Ventura 13.0.1 (Apple M1 Max) 64 GB RAM

AnthonyWhitaker:
To avoid this, we move the vertex shaders' output beyond clipspace and to a single degenerate point at (-2, -2, -2). With only this four-line change, we saw the framerate jump back to 120 FPS for the same geometry.

Pessimistress:
Thank you. Fix published in v8.8.21.

PR #7510, opened 2022-12-12
Improve filtered data rendering performance

AnthonyWhitaker:
Moves filtered data geometry off-screen to avoid FP ops

For #7509 

#### Background
Poor performance for a common use case in our application lead to a slight change in the data-filter extension - resulting in the frame rate jumping from 3-4 FPS to 120 FPS.

#### Change List
- Tweaked the data-filter extension's shader to move the geometry outside of clip space resulting in significantly reduced fragment costs.


modules/extensions/src/data-filter/shader-module.ts
```diff
@@ -146,6 +146,12 @@ const inject = {
     #endif
   `,
 
+  'vs:#main-end': `
+    if (dataFilter_value == 0.0) {
+      gl_Position = vec4(-2, -2, -2, 1);
```
Pessimistress, line 151: `vec4(0.)` should suffice?

Pessimistress, line 151:
```suggestion
      gl_Position = vec4(0.);
```

Pessimistress (approved):

modules/extensions/src/data-filter/shader-module.ts
AnthonyWhitaker: The goal was to move all geometry outside of clip space, so anything outside of -1 -> 1 on any axis. I'm not a fan of the magic -2 values. Definitely open to suggestions. Any ideas?

Issue #7496, opened 2022-12-09
[Bug] Accessors added by extensions not diffed correctly


felixpalmer:
### Description

When the `_propTypes` object [is created](https://github.com/visgl/deck.gl/blob/c1d3921798720280a9a4c99e298bdadc0d1d6ef8/modules/core/src/lifecycle/create-props.ts#L97) for a Layer the props added by any extensions are not considered. 

Later in the render cycle [when layers are diffed](https://github.com/visgl/deck.gl/blob/master/modules/core/src/lifecycle/props.ts#L144-L164), any accessors that are added by an extension are considered to be changing shallowly as the propType for the extension accessor is not present on `Layer._propTypes`.

The impact is that state updates happen needlessly on every frame.

### Flavors

- [ ] React
- [ ] Python/Jupyter notebook
- [ ] MapboxLayer
- [ ] GoogleMapsOverlay
- [ ] CartoLayer
- [ ] DeckLayer/DeckRenderer for ArcGIS

### Expected Behavior

An accessor added by an extension behaves the same as a standard accessor, in that it invokes the [default accessor equality function](https://github.com/visgl/deck.gl/blob/master/modules/core/src/lifecycle/prop-types.ts#L131-L136).

### Steps to Reproduce

Add the following test to `test/modules/core/lib/layer-extension.spec.js`:

```javascript
// Modify MockExtension.updateState to include:
MockExtension.changeFlags = updateParams.changeFlags;

test('LayerExtension#props diff', t => {
  const extension = new MockExtension({assert: t.ok});

  MockExtension.resetStats();

  testLayer({
    Layer: ScatterplotLayer,
    testCases: [
      {
        props: {
          id: 'test-layer',
          data: [
            {coordinates: [-122, 38], value: 1},
            {coordinates: [-123, 37], value: 2}
          ],
          getPosition: d => d.coordinates,
          ext_getValue: d => d.value,
          extensions: [extension]
        },
        onAfterUpdate: ({layer}) => {
          t.is(MockExtension.initializeCalled, 1, 'initializeState called');
          t.is(MockExtension.updateCalled, 1, 'updateState called');
          t.is(MockExtension.finalizeCalled, 0, 'finalizeState not called');
        }
      },
      {
        updateProps: {
          getPosition: d => [12, 34]
        },
        onAfterUpdate: () => {
          t.is(MockExtension.updateCalled, 1, 'updateState not called');
          t.is(
            MockExtension.changeFlags.propsChanged,
[...]

PR #7513, opened 2022-12-13
Handle defaultProps of extensions

Pessimistress:
For #7496

Before this change, we parse the static `defaultProps` object once for each layer class, and then merge it recursively with parent class defaults once for each layer class as well. The result is stored in e.g. `ScatterplotLayer["_mergedDefaultProps"]`.

After this change, we parse the static `defaultProps` object once for each layer class and layer extension class, and merge them recursively with parent class defaults for each layer class + extension class list combination. The result is stored in e.g. `ScatterplotLayer["_mergedDefaultProps:DataFilterExtension:FillStyleExtension"]`.

This approach ensures that `ExtensionClass.defaultProps` behave the same as `LayerClass.defaultProps` when used together.

My main concern about merging this is the perf impact. Here are the resulting `createProp` benchmarks:

| Test | Before (iterations/s) | After (iterations/s) |
|---|---|---|
| empty | 3.20M | 3.20M |
| with props | 117K  | 118K |
| props with extensions | 117K  | 104K |

#### Change List
- Update `createProps` to consider extensions
- Unit tests
- Bench tests

### DIFF
--- a/modules/core/src/lifecycle/create-props.ts
+++ b/modules/core/src/lifecycle/create-props.ts
@@ -95,15 +95,24 @@ function createPropsPrototypeAndTypes(
   const componentDefaultProps = getOwnProperty(componentClass, 'defaultProps') || {};
   const componentPropDefs = parsePropTypes(componentDefaultProps);
 
-  // Merged default props object
-  let defaultProps: any = {...parentDefaultProps, ...componentPropDefs.defaultProps};
-  // Merged prop types object
-  const propTypes = {...parentDefaultProps?.[PROP_TYPES_SYMBOL], ...componentPropDefs.propTypes};
-  // Merged deprecation list
-  const deprecatedProps = {
-    ...parentDefaultProps?.[DEPRECATED_PROPS_SYMBOL],
-    ...componentPropDefs.deprecatedProps
-  };
+  // Merged default props object. Order: parent, self, extensions
+  const defaultProps: any = Object.assign(
+    Object.create(null),
+    parentDefaultProps,
+    componentPropDefs.defaultProps
+  );
+  // Merged prop type definitions. Order: parent, self, extensions
+  const propTypes = Object.assign(
+    Object.create(null),
+    parentDefaultProps?.[PROP_TYPES_SYMBOL],
+    componentPropDefs.propTypes
+  );
+  // Merged deprecation list. Order: parent, self, extensions
+  const deprecatedProps = Object.assign(
+    Object.create(null),
+    parentDefaultProps?.[DEPRECATED_PROPS_SYMBOL],
+    componentPropDefs.deprecatedProps
+  );
 
   if (extensions?.length) {
     for (const extension of extensions) {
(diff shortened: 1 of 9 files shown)


modules/core/src/lifecycle/create-props.ts
```diff
@@ -46,22 +57,40 @@ export function createProps<T>(
...
+
 // Return precalculated defaultProps and propType objects if available
 // build them if needed
-function getPropsPrototype(componentClass) {
-  const defaultProps = getOwnProperty(componentClass, '_mergedDefaultProps');
+function getPropsPrototype(componentClass, extensions?: any[]) {
+  // A string that uniquely identifies the extensions involved
+  let cacheKey = MergedDefaultPropsCacheKey;
+  if (extensions) {
+    for (const extension of extensions) {
```
felixpalmer, line 68: Sort the extensions so that `_mergedDefaultProps:ExtensionA:ExtensionB` doesn't duplicate `_mergedDefaultProps:ExtensionB:ExtensionA`?

modules/core/src/lifecycle/create-props.ts
```diff
@@ -46,22 +57,40 @@ export function createProps<T>(
...
+        cacheKey += `:${ExtensionClass.extensionName || ExtensionClass.name}`;
+      }
+    }
+  }
+
+  const defaultProps = getOwnProperty(componentClass, cacheKey);
   if (!defaultProps) {
-    createPropsPrototypeAndTypes(componentClass);
-    return componentClass._mergedDefaultProps;
+    return (componentClass[cacheKey] = createPropsPrototypeAndTypes(
```
felixpalmer, line 78: Rather than using a compound cache key that includes all the extensions, would it not be possible to store the defaultProps of the extension on the Extension class and then merge them here? Are you worried about the GC overhead of doing this on every frame?

modules/core/src/lifecycle/create-props.ts
Pessimistress: - Sort is going to have a much bigger perf impact. The key generation could be invoked indefinitely and every frequently, while caching only happen once or twice (due to duplicates). It is also unlikely that the user keep changing the order of extensions upon rerender.
- Technically speaking the order does matter, if the extensions define overlapping props.

modules/core/src/lifecycle/create-props.ts
Pessimistress: How is caching on the Extension class different from caching on the Layer class? Don't you still need to differentiate the layer with which the extension is used?

modules/core/src/lifecycle/create-props.ts
felixpalmer: Good point regarding overlapping props

modules/core/src/lifecycle/create-props.ts
felixpalmer: I'm thinking about how large the cache would grow. As more extensions are added over time the same layer with different extensions would have the defaultProps duplicated. But perhaps this is needlessly worrying about a tiny bit of memory

felixpalmer (approved):

PR #7515, opened 2022-12-13
EffectManager improvements

Pessimistress:
#### Background

This PR is preparation for the addition of `CollisionExtension` and `TerrainExtension`. Also addresses some legacy perf issues with the `EffectManager`.

#### Change List
- Add experimental `_addDefaultEffect` API to `Deck` for extensions to register custom effects. This allows us to add optional global effects without bloating the core bundle.
- `MaskEffect` is moved from core to the extensions module. It is only attached if a layer is rendered with the `MaskEffect` extension.
- Add optional `setProps` method to the `Effect` interface. This allows the `EffectManager` to avoid destroying and recreating WebGL resources when an effect's option changes.
- Prior to this PR `EffectManager` wipes the resources from all effects if the `effects` prop changes. After this PR, old and new effect instances are matched by id, and only the necessary updates are applied.
- Unit tests


modules/extensions/src/mask/mask.ts
```diff
@@ -31,6 +32,10 @@ export default class MaskExtension extends LayerExtension {
   static defaultProps = defaultProps;
   static extensionName = 'MaskExtension';
 
+  initializeState(this: Layer<MaskExtensionProps>) {
+    this.context.deck?._addDefaultEffect(new MaskEffect());
```
felixpalmer, line 36: As above, do we want to let this silently succeed if there is no deck instance?

felixpalmer (approved):

modules/extensions/src/mask/mask.ts
Pessimistress: There is no deck instance when running test utilities.

Issue #7555, opened 2023-01-04
[Bug] Typescript definition does not allow SimpleMeshLayerProps to have wireframe: true


thpica:
### Description

When creating a `SimpleMeshLayer`, the typescript compilation fails if the `wireframe` prop is set to `true`.

It fails with the following error:
```console
test.ts:4:3 - error TS2322: Type 'true' is not assignable to type 'false'.

4   wireframe: true,
    ~~~~~~~~~

  node_modules/@deck.gl/mesh-layers/typed/simple-mesh-layer/simple-mesh-layer.d.ts:70:5
    70     wireframe?: false;
           ~~~~~~~~~
    The expected type comes from property 'wireframe' which is declared here on type 'SimpleMeshLayerProps<any>'
```

### Flavors

- [X] React
- [ ] Python/Jupyter notebook
- [ ] MapboxLayer
- [ ] GoogleMapsOverlay
- [ ] CartoLayer
- [ ] DeckLayer/DeckRenderer for ArcGIS

### Expected Behavior

Setting wireframe to true should be accepted as it is a valid value.

### Steps to Reproduce

```typescript
import { SimpleMeshLayerProps } from "@deck.gl/mesh-layers/typed";

const props: SimpleMeshLayerProps = {
    wireframe: true
};
```

### Environment

- Framework version: deck.gl@8.8.20
- Typescript: 4.9.4

Pessimistress:
Thank you for submitting a fix. It is included in v8.8.22.

PR #7556, opened 2023-01-04
Fix SimpleMeshLayerProps.wireframe and SimpleMeshLayerProps._instanced type

thpica:
For #7555

#### Change List
- Fix SimpleMeshLayerProps wireframe type

### DIFF
--- a/modules/mesh-layers/src/simple-mesh-layer/simple-mesh-layer.ts
+++ b/modules/mesh-layers/src/simple-mesh-layer/simple-mesh-layer.ts
@@ -145,12 +145,12 @@ type _SimpleMeshLayerProps<DataT> = {
    * setting _instanced to false interpreted mesh positions as lnglat deltas.
    * @default true
    */
-  _instanced?: true; // TODO - formalize API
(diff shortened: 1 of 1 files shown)


modules/mesh-layers/src/simple-mesh-layer/simple-mesh-layer.ts
```diff
@@ -150,7 +150,7 @@ type _SimpleMeshLayerProps<DataT> = {
    * Whether to render the mesh in wireframe mode.
    * @default false
    */
-  wireframe?: false;
+  wireframe?: boolean;
```
felixpalmer, line 153: Apply the same fix to `_instanced`?

felixpalmer (changes requested):

modules/mesh-layers/src/simple-mesh-layer/simple-mesh-layer.ts
thpica: Done

Pessimistress (approved):

PR #7585, opened 2023-01-18
fixed MapboxOverlay props typing

YonatanHanan:


YonatanHanan (commented):
Hi,

I fixed this little typing :)

felixpalmer:
You've reformated the whole file. Please use the prettier config or run `yarn lint`

Pessimistress:
What is the reason for this change? `interleaved` is only available in the constructor, and not in `setProps`.

YonatanHanan:
> What is the reason for this change? `interleaved` is only available in the constructor, and not in `setProps`.

yes, but when you use this component in react you have to copy the `interleaved` property as if it is part of the props.
this is my code:
```TS
import { useControl } from 'react-map-gl';
import { MapboxOverlay } from '@deck.gl/mapbox/typed';
import type { MapboxOverlayProps } from '@deck.gl/mapbox/typed';

type MBDeckGLOverlayProps = MapboxOverlayProps & {
    interleaved?: boolean;
};

export const MBDeckGLOverlay = (props: MBDeckGLOverlayProps) => {
    const overlay = useControl<MapboxOverlay>(() => new MapboxOverlay(props));
    overlay.setProps(props);
    return null;
};
```

instead of 
```TS
import { useControl } from 'react-map-gl';
import { MapboxOverlay } from '@deck.gl/mapbox/typed';
import type { MapboxOverlayProps } from '@deck.gl/mapbox/typed';

export const MBDeckGLOverlay = (props: MBDeckGLOverlayProps) => {
    const overlay = useControl<MapboxOverlay>(() => new MapboxOverlay(props));
    overlay.setProps(props);
    return null;
};
```

Pessimistress (approved):

PR #7596, opened 2023-01-23
Export CursorState

timnyborg:
Simple type export for end-users


Pessimistress:
Don't you need it exported at `src/index.ts`?

Issue #4872, opened 2020-08-17
Ripple effect for selected point


UttamRahane:
Is there any way to add ripple effect for selected point in ScatterplotLayer/GeoJsonLayer/IconLayer? I'm trying to highlight selected object with some animation effects. like [https://docs.mapbox.com/mapbox-gl-js/example/add-image-animated](url)

UttamRahane:
![image](https://user-images.githubusercontent.com/21005702/90478697-2b17d380-e14b-11ea-87a9-254fab354cae.png)

UttamRahane:
Perfect, Thank you so much @Pessimistress for the help.

Nitikaa:
@UttamRahane  I wanted to add ripple effect on lineLayer,  the same thing I'm facing like its re-rendering all the layers (with large data). can you brief on iTime approach ?

Nitikaa:
@Pessimistress I wanted to add ripple effect on lineLayer, the same thing I'm facing like its re-rendering all the layers (with large data). can you brief on iTime draw() approach ?

PR #7601, opened 2023-01-26
Add docs to clarify layer updates in React

kevinschaich:
For #7598

#### Change List
- Added docs to clarify layer instantiation updates & in React, and discourage use of `layer.clone(...)`

### DIFF
--- a/docs/api-reference/core/layer.md
+++ b/docs/api-reference/core/layer.md
@@ -597,6 +597,9 @@ Create a copy of this layer, optionally change some prop values.
 Arguments:
 - `overrideProps` (Object, optional) - layer props to update.
 
+Remarks:
+- When using React, instead of calling `layer.clone(...)` you should construct a new layer instance on every render. Instead of the above, use `const updatedLayer = new ScatterPlotlayer({...props, ...overrideProps})` each time (both initialization and re-rendering).
+
 ##### `setState`
 
 Used to update the layers [`state`](/docs/api-reference/core/layer.md#state) object. Calling this method will also cause the layer to rerender.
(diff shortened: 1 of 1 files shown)


docs/api-reference/core/layer.md
```diff
@@ -597,6 +597,13 @@ Create a copy of this layer, optionally change some prop values.
 Arguments:
 - `overrideProps` (Object, optional) - layer props to update.
 
+Remarks:
+- When using React, instead of calling `layer.clone(...)` you should construct a new layer instance on every render. Instead of the above, use the following each time (both initialization and re-rendering):
```
felixpalmer, line 601: I don't see why this is necessary, we have a [whole section on React](https://deck.gl/docs/get-started/using-with-react) with example usage and on this page it is mentioned that: 

```
    Layer methods are designed to support the creation of new layers or layer sub-classing and are NOT intended to be called by applications.
```

kevinschaich:
@felixpalmer it was unclear to me because I was following another example in vanilla JS (https://github.com/visgl/deck.gl/issues/4872, [codepen](https://codepen.io/Pessimistress/pen/jOqVKVB?editors=0010)) where `layer.clone(...)` works as expected, but has no effect in react.

Feel free to close it if you don't think it's helpful, just trying to save someone else the headache in the future

Pessimistress:
@kevinschaich your code in that discussion did not work because you were using the wrong prop values, not because you were using React. `clone` works anywhere, just not necessary in React.

kevinschaich:
Got it - thanks guys, sincerely appreciate your help.

PR #7603, opened 2023-01-27
move comments into proper places

fatality1922:
For #

#### Change List
-


ibgreen:
@fatality1922 Thanks for the PR, but why do you want the documents below the declarations? Is there some specific tool or IDE you use that expects that? When the comments precede the declarations, vscode parses them and displays them when hovering over fields of that type, which is great since many vis.gl contributors use vscode.

fatality1922:
> @fatality1922 Thanks for the PR, but why do you want the documents below the declarations? Is there some specific tool or IDE you use that expects that? When the comments precede the declarations, vscode parses them and displays them when hovering over fields of that type, which is great since many vis.gl contributors use vscode.

hmm, so I'm not sure right now. I did correction because I had problem exactly with that. 
bug

As you see info is for sizeScale, but comment is clearly for transformMatrix

ibgreen:
That is strange. Maybe you have to restart and update vscode and if that doesn't work, google a bit...

Screen Shot 2023-01-27 at 2 52 04 PM

PR #7610, opened 2023-01-31
DECKGL_FILTER_GL_POSITION hook consistency

Pessimistress:
For #7195

`TerrainExtension` rewrites gl_Position in the `DECKGL_FILTER_GL_POSITION` hook based on the common position. This breaks layers like `IconLayer`, `TextLayer`, `ScatterplotLayer`, `PathLayer` in billboard mode, where multiple vertices are offset from the same clipspace anchor.

`DECKGL_FILTER_GL_POSITION` is currently called after the last assignment of `gl_Position`.

This PR moves it to before any clipspace offset is applied, i.e. where `gl_Position` corresponds to `geometry.position`.

#### Change List
- Invoke `DECKGL_FILTER_GL_POSITION` hook before screen space offset is applied

### DIFF
--- a/modules/layers/src/arc-layer/arc-layer-vertex.glsl.ts
+++ b/modules/layers/src/arc-layer/arc-layer-vertex.glsl.ts
@@ -246,8 +246,8 @@ void main(void) {
     getExtrusionOffset((next.xy - curr.xy) * indexDir, positions.y, widthPixels),
     0.0);
   DECKGL_FILTER_SIZE(offset, geometry);
+  DECKGL_FILTER_GL_POSITION(curr, geometry);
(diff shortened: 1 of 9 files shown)


ibgreen (approved):
Do deck.gl have any docs for the shader hooks? If so add something about this requirement? If not, create a minimal page?

felixpalmer (approved):

Pessimistress:
> Do deck.gl have any docs for the shader hooks?

Yes, https://deck.gl/docs/developer-guide/custom-layers/writing-shaders#vsdeckgl_filter_gl_position

igorDykhta (approved):

PR #7633, opened 2023-02-09
CollideExtension docs & render tests

felixpalmer:
For https://github.com/visgl/deck.gl/issues/7374

#### Change List
- `CollideExtension` docs
- Render tests


docs/api-reference/extensions/collide-extension.md
```diff
@@ -0,0 +1,96 @@
...
+
+Collision group this layer belongs to. If it is not set, the 'default' collision group is used. Two (or more) layers that share the same `collideGroup` will be considered together when calculating collisions.
+
+##### `collideTestProps` (Object, optional)
+
+Props to override when computing collisions. A common use case is to increase the size of the features when computing collisions to provide greater spacing between visible features. For the `ScatterplotLayer` this would be done by:
+
+```js
+collideTestProps: {radiusScale: 2}
+`
```
Pessimistress, line 77: Code block not closed

docs/api-reference/extensions/collide-extension.md
```diff
@@ -0,0 +1,96 @@
...
+
+When added to a layer via the `extensions` prop, the `CollideExtension` adds the following properties to the layer:
+
+##### `collideEnabled` (Boolean, optional)
+
+Enable/disable collisions. If collisions are disabled, all objects are rendered. Defaults to `true`.
+
+##### `collideGroup` (string, optional)
+
+Collision group this layer belongs to. If it is not set, the 'default' collision group is used. Two (or more) layers that share the same `collideGroup` will be considered together when calculating collisions.
```
Pessimistress, line 69: I think a little more explanation is needed here. Maybe an example use case. My understanding is that you are only supposed to combine multiple layers into a single group if they use exactly the same data.

Pessimistress (approved):

PR #7641, opened 2023-02-15
Update example dependencies

Pessimistress:
For #7635 

For all examples and test apps where it applies:

#### Change List
- Removed `tsc` from `build` script. The examples do not currently contain their own tsconfig.json and `tsc` just fails.
- Updated `react-map-gl` to v7 and use `maplibre-gl` as map provider in all except the following examples:
  + `get-started/react/mapbox` - depends on `NavigationControl` which is no longer supported in v7
  + `layer-browser` - depends on `NavigationControl` which is no longer supported in v7
  + `safegraph` (the mapbox integration example on the [website](https://deck.gl/examples/mapbox/)) - for testing against the latest version of mapbox-gl
- Updated `react` and `react-dom` to v18

### DIFF
--- a/test/apps/mapbox-layers/package.json
+++ b/test/apps/mapbox-layers/package.json
@@ -5,9 +5,10 @@
   },
   "dependencies": {
     "deck.gl": "^8.4.0",
-    "react": "^16.3.0",
-    "react-dom": "^16.3.0",
-    "react-map-gl": "^5.0.0"
+    "mapbox-gl": "^2.0.0",
+    "react": "^18.0.0",
+    "react-dom": "^18.0.0",
+    "react-map-gl": "^7.0.0"
   },
   "devDependencies": {
     "vite": "^4.0.0"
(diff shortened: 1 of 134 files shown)


wipfli:
Happy to see they you are switch to MapLibre GL JS in the examples. Cool to have you in the community!

test/apps/mapbox-layers/package.json
```diff
@@ -5,9 +5,10 @@
   },
   "dependencies": {
     "deck.gl": "^8.4.0",
-    "react": "^16.3.0",
-    "react-dom": "^16.3.0",
-    "react-map-gl": "^5.0.0"
+    "mapbox-gl": "^2.0.0",
```
chrisgervang, line 8: Should we also have a maplibre example? The packages are diverging overtime and it's a lot of effort to switch this example back and forth between them at the moment in testing.

chrisgervang (approved):

test/apps/mapbox-layers/package.json
Pessimistress: Yes I agree. I will open another PR to consolidate test apps.

PR #7642, opened 2023-02-15
Better support for maplibre in scripting interface

Pessimistress:
For #7635 

It's possible to use maplibre with the current bundle, but the user is required to pass `map: maplibregl` to the `DeckGL` constructor. This PR make it possible to use maplibre without this manual step.

#### Change List
- Auto detect `maplibregl` global object
- Documentation
- Move scripting examples to use maplibre-gl

### DIFF
--- a/modules/core/bundle/deckgl.js
+++ b/modules/core/bundle/deckgl.js
@@ -56,7 +56,7 @@ export default class DeckGL extends Deck {
 
     const viewState = props.viewState || props.initialViewState;
     const isMap = Number.isFinite(viewState && viewState.latitude);
-    const {map = window.mapboxgl} = props;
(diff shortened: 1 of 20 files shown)


felixpalmer (approved):
I can't see the code change where the autodetection happens, could you point me to it @Pessimistress ?

modules/core/bundle/deckgl.js
```diff
@@ -56,7 +56,7 @@ export default class DeckGL extends Deck {
 
     const viewState = props.viewState || props.initialViewState;
     const isMap = Number.isFinite(viewState && viewState.latitude);
-    const {map = window.mapboxgl} = props;
+    const {map = window.mapboxgl || window.maplibregl} = props;
```
Pessimistress, line 59: @felixpalmer

PR #7643, opened 2023-02-15
CARTO: Increase GET limit to 8192

felixpalmer:
POST requests create issues with the API, increase the size limit for GET request to allow larger queries


borja-munoz (approved):
Looks good to me!

PR #7646, opened 2023-02-15
Fix jupyter widget tests in dist mode

Pessimistress:
See background in #7645

#### Change List
- Upgrade ocular-dev-tools to include https://github.com/uber-web/ocular/pull/418
- Remove cross-module import with relative path
- Fix unit test that fails in dist mode


chrisgervang (approved):
That's a lot clearer, thanks - I was wondering how new modules were added to the jupyter bundle.

ajduberstein (approved):

Issue #7532, opened 2022-12-20
[Bug] MapboxOverlay adds map move listener that is not removed when instance is finalized 


AdamRatcliffe-TomTom:
### Description

In deck-utils `getDeckInstance` a `move` listener is added but when `finalize` is called on the overlay the listener is not removed. In my use case the overlay may be removed before the end of the map lifetime, after the overlay is removed moving the map will result in an error being thrown in `Deck.setProps`:

````
deck.ts:456 Uncaught TypeError: Cannot read properties of null (reading 'setProps')
    at Deck.setProps (deck.ts:456:1)
    at onMapMove (deck-utils.ts:229:1)
    at i.<anonymous> (deck-utils.ts:66:1)
    at It.fire (maps.min.js:1:1)
    at gi._fireEvent (maps.min.js:1:1)
    at gi._fireEvents (maps.min.js:1:1)
    at gi._updateMapTransform (maps.min.js:1:1)
    at gi._applyChanges (maps.min.js:1:1)
    at Object.callback (maps.min.js:1:1)
    at _i.run (maps.min.js:1:1)
    at i._render (maps.min.js:1:1)
    at maps.min.js:1:1
````

### Flavors

- [ ] React
- [ ] Python/Jupyter notebook
- [X] MapboxLayer
- [ ] GoogleMapsOverlay
- [ ] CartoLayer
- [ ] DeckLayer/DeckRenderer for ArcGIS

### Expected Behavior

Map `move` listener is removed when Deck instance is finalized.

### Steps to Reproduce

Add a `MapboxOverlay` to the Mapbox GL map. Call `finalize` on the overlay instance. Move the map to trigger the error.

### Environment

- Framework version: 8.8.20
- Browser: Chrome 108.0.5359.124
- OS: Mac OS  12.6

YonatanHanan:
I think i get the same error when calling `map.jumpto` have you found a solution?

AdamRatcliffe-TomTom:
@YonatanHanan unfortunately I have not.

YonatanHanan:
So using the documentation https://deck.gl/docs/api-reference/mapbox/mapbox-overlay#using-with-react-map-gl did not work for me when trying to `jumpto` but when I added the `MapboxOverlay` as a control manually it did work.
Could that mean I have a different problem?

AdamRatcliffe-TomTom:
[...]

PR #7649, opened 2023-02-16
fix(mapbox) only move viewState once deck is loaded and only until deck is finalized

chrisgervang:
For #7532 

#### Change List
- comments to clarify current `getDeckInstance` use cases (React vs pure JS is no longer relevant) 
- only watch for map move events after deck is initialized
- protect deck object from mapbox move events after finalization


modules/mapbox/src/deck-utils.ts
```diff
@@ -60,7 +60,7 @@ export function getDeckInstance({
     });
     // If using the WebGLContext created by deck (React use case), we use deck's viewState to drive the map.
     // Otherwise (pure JS use case), we use the map's viewState to drive deck.
-    map.on('move', () => onMapMove(deckInstance, map));
+    map.on('move', () => deckInstance.isInitialized && onMapMove(deckInstance, map));
```
chrisgervang, line 63: Alternatively, we could try to call `map.off`, but it's stateful requiring the original function be passed in.

chrisgervang: Is there an event deck fires when it is finalized? We could utilize that here since `overlay.finalize()` calls `deck.finalize()`, and we have access to the deck instance here.

Pessimistress: I think the most common way to trigger this is by removing MapboxOverlay? In which case you can just call something from the overlay's `onRemove`.

If you want to address the more generic case, I guess you can make the callback deregister itself after the first time it's called with deck finalized.

chrisgervang: Yeah, `onRemove` in the overlay is the natural place to call something. https://github.com/visgl/deck.gl/pull/7649/commits/2f7fde03324e55aaf0c644daf270894fcb42490e shows what that could look like in the overlay, except `MapboxLayer` needs a one-time `map.on('move')` applied so I can't simply delete it in `getDeckInstance`.

chrisgervang: I'm leaning towards preferring the generic solution where the callback deregisters itself... I see we don't have a `deck.isFinalized` and that many of the signals may either mean yet-to-be-initialized or finalized (https://github.com/visgl/deck.gl/blob/ef8ea132f694aa1d47b1f2e3a262c388534eb6ba/modules/core/src/lib/deck.ts#L484-L485). Do you know what we can use to detect that deck is finalized?

modules/mapbox/src/deck-utils.ts
```diff
@@ -60,7 +60,15 @@ export function getDeckInstance({
     });
     // If using the WebGLContext created by deck (React use case), we use deck's viewState to drive the map.
     // Otherwise (pure JS use case), we use the map's viewState to drive deck.
-    map.on('move', () => onMapMove(deckInstance, map));
+    const _handleMapMove = () => {
+      // deregister self if called after deck is finalized.
+      if (!deckInstance?.layerManager) {
```
chrisgervang, line 65: This works, however I'd prefer to implement a `deck.isFinalized` since `layerManager` is a protected variable.

modules/mapbox/src/mapbox-overlay.ts
```diff
@@ -98,6 +98,7 @@ export default class MapboxOverlay implements IControl {
       })
     });
 
+    map.on('move', this._handleMapMove);
```
Pessimistress: Doesn't this register a second move handler, on top of the one existing in `getDeckInstance`?

modules/mapbox/src/deck-utils.ts
Pessimistress: `deck.isInitialized` basically does this. However it may be `false` in two cases:
- After Deck is constructed and before context become available
- After Deck is finalized

chrisgervang: Can I add an `isFinalized` that is only true when deck has been finalized?

modules/mapbox/src/mapbox-overlay.ts
chrisgervang: Correct

modules/mapbox/src/deck-utils.ts
Pessimistress: I think we need to discuss how `isInitialized` should behave if we add more tags... Alternatively you can do this

```js
if (deck.isInitialized) {
  watchMapMove(map, deck);
} else {
  deckProps.onLoad = () => watchMapMove(map, deck)
}
```

leo-abandon:
I've the same issue

Pessimistress (approved):

PR #7651, opened 2023-02-16
IconManager preserve aspect ratio when auto packing

Pessimistress:
#### Background

This issue has been raised in the Slack channel:

When using the `IconLayer` with auto packing, the application may not know the dimensions of an icon before loading it. Providing the wrong width/height will cause the rendered icon to be stretched.

After this PR, the `IconManager` preserves the aspect ratio of the incoming icon image while resizing.

This is technically a breaking change with no way of reverting to the old behavior. I would argue that it is the expected behavior, though.

Other caveats:
- The packing algorithm is still run before any icon is loaded using the user-provided width/height, so there may end up being some empty pixels in the texture when all is loaded. However, this way we only have to resize the texture once.
- If an icon image's aspect ratio is different from the pre-defined dimension, the pixels are centered in the provided bounding box. The default `anchorX`/`anchorY` (center of bounding box) will not be affected, but other values may appear off.

#### Change List
- IconManager preserve aspect ratio when auto packing
- Unit test

### DIFF
--- a/test/modules/layers/icon-manager.spec.js
+++ b/test/modules/layers/icon-manager.spec.js
@@ -1,6 +1,8 @@
 import test from 'tape';
 import IconManager, {buildMapping, getDiffIcons} from '@deck.gl/layers/icon-layer/icon-manager';
 import {gl} from '@deck.gl/test-utils';
+import {isBrowser} from '@probe.gl/env';
+import {loadImage} from 'canvas';
 
 const DATA = [
   {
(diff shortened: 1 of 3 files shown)


test/modules/layers/icon-manager.spec.js
```diff
@@ -241,3 +243,56 @@ test('IconManager#events', t => {
...
+  };
+
+  iconManager.setProps({
+    autoPacking: true,
+    loadOptions: isBrowser() ? {} : [nodeImageLoader]
+  });
+  iconManager.packIcons(
+    [
+      {id: 'no-resize', width: 16, height: 16},
+      {id: 'down-size', width: 12, height: 12},
```
chrisgervang, line 290: Can they be up-sized as well?

chrisgervang (approved):

test/modules/layers/icon-manager.spec.js
Pessimistress: Yes

PR #7652, opened 2023-02-16
TypeScript: fix some TS erros

Pessimistress:
We previously set `PropsT` default to `{}` because `any` defeats type checking on `layer.props`. However, this leads to the type error `Layer<PropsT> does not extend Layer<{}>` when we use `this` where `Layer` is expected.

See [playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAMghiCAnACkg9mAzlAvFAbwCgpSoATOYOALijgDsQBtAXQG4SzM4BjAS1B0GAVwC2AI2ScyUAG78s-CQBsIdCenRrGnAL5EiAeiNReKuFhzxESADwBRAB7AkcNJiwAVPPSYA+QmNTc0trBGRHFzcPbB98Aj1A4lCrWAj7Z1d3DDjfEQYAawZ0AHcGZK5SMFysOiyY2p8AMnTbWKxOKrN0BixXEV5gdCQAChrPeuiczxa25A6ASiDZWWAAC0UAOgnsX13O7oNu4Ah+0YtbOhtkReuMldXSEygAZXX0ERVyKGQMJCgowK5AgADN+AwID9doturJLsgdrUdvxeIU4KpoPhQXAVFgIF0ns9TO9Pt9fkh-oDShgGABzKC7KCgSCwolQBFIJGeLY8ASgXwAcghclx-B+opUIgggsJRJepK+PykUHQhUBN3sHS8gUU9HmSDZRKQEGAIiQDGZm0OsgMBiAA)

#### Change List
- `ExtraPropsT` defaults to `unknown` instead of `{}`

### DIFF
--- a/modules/aggregation-layers/src/aggregation-layer.ts
+++ b/modules/aggregation-layers/src/aggregation-layer.ts
@@ -31,7 +31,7 @@ import {filterProps} from './utils/prop-utils';
 export type AggregationLayerProps<DataT = any> = CompositeLayerProps<DataT>;
 
 export default abstract class AggregationLayer<
-  ExtraPropsT = {}
+  ExtraPropsT = unknown
 > extends CompositeLayer<ExtraPropsT> {
   static layerName = 'AggregationLayer';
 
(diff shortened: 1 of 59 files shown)


ibgreen (commented):
To make sure the inheritance is understood by the compiler, I think the normal technique is to do something like:
```
Layer<PropsT extends {} = {}>
```

ibgreen (approved):

PR #7654, opened 2023-02-16
Consolidate common utils in extensions

Pessimistress:
The test apps seem to be working ok. All tests pass. I'd appreciate if you could give it another manual test.

#### Change List
- Consolidate common utils between MaskExtension and TerrainExtension


modules/extensions/src/mask/mask-effect.ts
```diff
@@ -2,13 +2,13 @@ import {Layer, Viewport, Effect, PreRenderOptions, CoordinateSystem, log} from '
 import {Texture2D} from '@luma.gl/core';
 import {equals} from '@math.gl/core';
 import MaskPass from './mask-pass';
-import {getMaskBounds, getMaskViewport, MaskBounds} from './utils';
```
felixpalmer, line 5: I'm seeing blending artifacts when running `test/apps/collision/app.jsx`

Screenshot 2023-02-17 at 11 02 45

felixpalmer (changes requested):

felixpalmer (approved):

PR #7660, opened 2023-02-17
Add embedded demo to new extension docs

Pessimistress:
For #7635 

https://codepen.io/vis-gl/pen/oNPXXzm

image

@felixpalmer There is a bug here - the collide map is not rerendered after the icon layer is loaded, you'll need to move around to get it to update.

https://codepen.io/vis-gl/pen/VwGLLeR

image

#### Change List
- CollideExtension doc
- TerrainExtension doc

### DIFF
--- a/docs/api-reference/extensions/terrain-extension.md
+++ b/docs/api-reference/extensions/terrain-extension.md
@@ -9,8 +9,14 @@ To use this extension, first define a terrain source with the prop `operation: '
 
 For each layer that should be fitted to the terrain surface, add the `TerrainExtension` to its `extensions` prop.
 
-<!-- TODO: Codepen demo -->
 <div style={{position:'relative',height:450}}></div>
+<div style={{position:'absolute',transform:'translateY(-450px)',paddingLeft:'inherit',paddingRight:'inherit',left:0,right:0}}>
+  <iframe height="450" style={{width: '100%'}} scrolling="no" title="deck.gl TerrainExtension" src="https://codepen.io/vis-gl/embed/VwGLLeR?height=450&theme-id=light&default-tab=result" frameborder="no" loading="lazy" allowtransparency="true" allowfullscreen="true">
+    See the Pen <a href='https://codepen.io/vis-gl/pen/VwGLLeR'>deck.gl TerrainExtension</a> by vis.gl
+    (<a href='https://codepen.io/vis-gl'>@vis-gl</a>) on <a href='https://codepen.io'>CodePen</a>.
+  </iframe>
+</div>
+
 
 ```js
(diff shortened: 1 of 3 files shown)


docs/api-reference/extensions/collide-extension.md
```diff
@@ -5,8 +5,14 @@ The `CollideExtension` allows layers to hide features which overlap with other f
 
 To use this extension on a layer, add the `CollideExtension` to the layer's `extensions` prop.
 
-<!-- TODO: Codepen demo -->
 <div style={{position:'relative',height:450}}></div>
+<div style={{position:'absolute',transform:'translateY(-450px)',paddingLeft:'inherit',paddingRight:'inherit',left:0,right:0}}>
```
felixpalmer, line 9: Nice demo, I think we could make it prettier though. Add a point layer for the nodes?

```js
const pointLayer = new ScatterplotLayer({
    id: 'points',
    data: Object.values(nodes),
    getPosition: node => [node.x, node.y],
    getRadius: getIconSize,
    radiusScale: 0.05,
    radiusUnits: 'pixels',
    lineWidthUnits: 'pixels',
    stroked: true,
    getFillColor: [222, 222, 222],
    getLineColor: [100, 100, 100]
  });
```
Screenshot 2023-02-23 at 10 11 17

docs/api-reference/extensions/terrain-extension.md
```diff
@@ -9,8 +9,14 @@ To use this extension, first define a terrain source with the prop `operation: '
 
 For each layer that should be fitted to the terrain surface, add the `TerrainExtension` to its `extensions` prop.
 
-<!-- TODO: Codepen demo -->
 <div style={{position:'relative',height:450}}></div>
+<div style={{position:'absolute',transform:'translateY(-450px)',paddingLeft:'inherit',paddingRight:'inherit',left:0,right:0}}>
```
felixpalmer, line 13: Why not use aerial imagery like we do on the https://deck.gl/examples/terrain-layer/? By adjusting the fragment shader we could even make it look more "winter-like" (basically make it black and white)

felixpalmer (approved):

docs/api-reference/extensions/terrain-extension.md
Pessimistress: Is there a free aerial imagery dataset? I don't want to include a token in a codepen unless we have to.

chrisgervang: https://gist.github.com/chrisgervang/9fc4aa7691c633eefaf2887d720f7cd1

PR #7667, opened 2023-02-20
docs: Add section on binary data format in GeoJsonLayer

felixpalmer:
People have struggled to understand the binary data format used in GeoJsonLayer. This section gives an overview, similarly as we do in the [TextLayer docs](https://deck.gl/docs/api-reference/layers/text-layer#use-binary-attributes)


Pessimistress (commented):
Link to this part from the MVTLayer's doc?

docs/api-reference/layers/geojson-layer.md
```diff
@@ -354,6 +354,156 @@ The GeoJsonLayer renders the following sublayers:
...
+  }
+  pathIndices: {value: Uint16Array([0, 4, ...]), size: 1}
+  featureIds: {value: Uint16Array([0, 0, 0, 0, 1, 1, ...]), size: 1}
+  fields: [{id: 123}, {id: 456}]
+}
+```
+
+#### Polygons
+
+Polygons are an extension of the idea introduced with lines, but instead of `pathIndices` the `polygonIndicies` array specifies where each polygon starts inside the `positions` array. Because pygons can have holes, the offsets for the outer and inner rings are stored separately in the `primitivePolygonIndices` array. A polygon that has an outer ring consisting of 60 vertices and a hole with 40 vertices is represented as:
```
kylebarron, line 478:
```suggestion
Polygons are an extension of the idea introduced with lines, but instead of `pathIndices` the `polygonIndicies` array specifies where each polygon starts inside the `positions` array. Because polygons can have holes, the offsets for the outer and inner rings are stored separately in the `primitivePolygonIndices` array. A polygon that has an outer ring consisting of 60 vertices and a hole with 40 vertices is represented as:
```

kylebarron: It's good to note that in our custom format, every index value points into the _positions_ array, while other offset formats like Arrow have each index pointing into the next index array (i.e. polygon indices would ring indices would point into positions)

docs/api-reference/layers/geojson-layer.md
```diff
@@ -354,6 +354,156 @@ The GeoJsonLayer renders the following sublayers:
...
+polygons: {
+  positions: {value: Float32Array([x0, y0, ...]), size: 2},
+  polygonIndices: {value: Uint16Array([0, 100, ...]), size: 1}
+  primitivePolygonIndices: {value: Uint16Array([0, 60, 100, ...]), size: 1}
+}
+```
+
+#### Global feature ids
+
+Because the `features` array in the GeoJSON con contain a mix of different geometry types, in order to represent this ordering each of the `points`, `lines` and `polygons` objects contains a `globalFeatureIds` array, which contains the per-vertex indices into the original GeoJSON `features' array.
```
kylebarron, line 490:
```suggestion
Because the `features` array in the GeoJSON can contain a mix of different geometry types, in order to represent this ordering each of the `points`, `lines` and `polygons` objects contains a `globalFeatureIds` array, which contains the per-vertex indices into the original GeoJSON `features' array.
```

kylebarron (approved):

PR #7668, opened 2023-02-20
Avoid picking in overlays before deck is initialized

JannikGM:
#### Background

I've originally reported a bug on Slack: https://deckgl.slack.com/archives/C34AC5MSQ/p1676629109034669
The picking functions were used by the mapbox event handlers before deck was fully initialized (which will be fixed by #7723).

Also, while looking into it, I noticed that the explicit picking support provided by the overlays also isn't guarded against incorrect use before deck is initialized.
Because the user of the overlay doesn't have direct deck access either, I proposed to forward `isInitialized`.

I've also added an assert in case `this._deck` doesn't exist yet to the picking functions and made comments more consistent.

#### Change List
- Add `isInitialized` to Mapbox- and GoogleMapsOverlay
- Add types to GoogleMapsOverlay picker
- Consistent picker comments in Mapbox- and GoogleMapsOverlay

---

**None of this was tested**
I'm not sure how to run automated tests for the deck.gl codebase. `yarn bootstrap` fails with gyp not able to find `python` (which should be `python3` on this macOS machine). `yarn test` failed during linter step in untouched code.
All my deck.gl projects use another private library which has hardcoded dependency on a specific deck.gl version, so manual testing would be cumbersome.

### DIFF
--- a/modules/google-maps/src/google-maps-overlay.ts
+++ b/modules/google-maps/src/google-maps-overlay.ts
@@ -7,7 +7,7 @@ import {
   getViewPropsFromOverlay,
   getViewPropsFromCoordinateTransformer
 } from './utils';
-import {Deck} from '@deck.gl/core';
+import {assert, Deck} from '@deck.gl/core';
 
 import type {DeckProps} from '@deck.gl/core';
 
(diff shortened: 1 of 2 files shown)


modules/google-maps/src/google-maps-overlay.ts
```diff
@@ -43,6 +43,11 @@ export default class GoogleMapsOverlay {
 
   /* Public API */
 
+  /** Equivalent of `deck.isInitialized`. */
+  get isInitialized(): boolean {
```
felixpalmer, line 47: It would be nice if the user wouldn't have to worry about this, could we not instead have `pickObject` and friends return null when deck is uninitialized?

Pessimistress: I want to note that the next planned release is v9.0, so we can make breaking changes if it makes more sense.

PR #7671, opened 2023-02-21
CARTO: POST body consistency with GET parameters

felixpalmer:
#### Background

Some of the parameters present in the GET requests were missing in the POST body.

#### Change List
- Encode `geo_column` as `geo_column` rather than `geoColumn`
- Add `v` parameter
- Share code to avoid diverging in future


padawannn (approved):
LGFM

PR #7672, opened 2023-02-21
chore: Update Ubuntu for CI

felixpalmer:
#### Background

Ubuntu 18 will be soon deprecated, CI is failing as a result today due to a brownout: https://github.com/actions/runner-images/issues/6002

#### Change List
- Bump CI tests to ubuntu 20


modules/layers/src/text-layer/multi-icon-layer/multi-icon-layer-fragment.glsl.ts
```diff
@@ -28,7 +28,7 @@ uniform sampler2D iconsTexture;
 uniform float gamma;
 uniform bool sdf;
 uniform float alphaCutoff;
-uniform float buffer;
+uniform float sdfBuffer;
```
Pessimistress, line 31: headless-gl does not compile if a uniform is named `buffer`. Seems to trigger some internal reserved word.

Pessimistress (approved):

PR #7679, opened 2023-02-22
Test app CollisionFilterExtension + TextLayer alignment

felixpalmer:
For https://github.com/visgl/deck.gl/issues/7635

#### Background

Because the anchor points for labels can outside the rendered bounds of a label when alignment/baseline are not centered it is not trivial to get good behavior with the `CollisionFilterExtension`. This test app demonstrates a technique to make the collision filter work in this case.

https://user-images.githubusercontent.com/453755/220633955-e6518da4-cc3b-4376-a12d-0331c6623138.mov

#### Change List
- New test app

### DIFF
--- a/test/apps/text-collision/app.jsx
+++ b/test/apps/text-collision/app.jsx
@@ -1,0 +1,218 @@
+/* global fetch */
+import React, {useCallback, useState, useMemo} from 'react';
+import {render} from 'react-dom';
+import {StaticMap} from 'react-map-gl';
+import DeckGL from '@deck.gl/react';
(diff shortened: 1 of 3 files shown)


Pessimistress:
Can we try make one test app for each feature? There already are three mask extension test apps. It's hard to keep track of what needs to be tested when I make a change.

felixpalmer:
> Can we try make one test app for each feature? There already are three mask extension test apps. It's hard to keep track of what needs to be tested when I make a change.

I will combine the two collision test apps

PR #7680, opened 2023-02-22
Rerender collision map when isLoaded flag changes

felixpalmer:
For https://github.com/visgl/deck.gl/issues/7635 (fixes bug reported in https://github.com/visgl/deck.gl/pull/7660)

#### Change List
- Compare `isLoaded` value between renders in `CollisionFilterEffect`
- test


modules/extensions/src/collide/collide-effect.ts
```diff
@@ -171,15 +178,12 @@ export default class CollideEffect implements Effect {
...
-          collideGroup,
-          layers: [],
-          layerBounds: []
-        };
+        channelInfo = {collideGroup, layers: [], layerBounds: [], layerIsLoadedFlags: []};
         channelMap[collideGroup] = channelInfo;
       }
       channelInfo.layers.push(layer);
       channelInfo.layerBounds.push(layer.getBounds());
+      channelInfo.layerIsLoadedFlags.push(layer.isLoaded);
```
Pessimistress, line 186: I think a single flag for "all layers are loaded" should suffice.

Pessimistress (approved):

PR #7681, opened 2023-02-22
Mask Extension example for website

felixpalmer:
For https://github.com/visgl/deck.gl/issues/7635

Screenshot 2023-02-24 at 16 02 15

#### Background

@Pessimistress New example for https://deck.gl/examples. The idea is to make something that shows how the masking from one layer can be dynamic and thus performantly perform a spatial join on the client. As the planes fly over cities they are highlighted by an invisible mask layer with a longer trail than used for the visualization.

#### Change List
- New example app
- Update to website files


Pessimistress:
@felixpalmer I pushed a refactor that removed some of the complexity that was copied over from the globe example.

Pessimistress (approved):

PR #7693, opened 2023-02-24
docs: add warning about updating viewport

jamesscottbrown:
This might seem obvious, but an explicit statement might have helped me to avoid a bug I've just spent a while debugging.

(I was using a deck.gl MapboxOverlay, and updating the viewport with`deckOverlay.setProps({viewState })` - this worked, but interacting with the map would snap back to the previous viewport. The fix was to update the viewport on the mapbox map instead).


Pessimistress (approved):
Thank you!

PR #7699, opened 2023-02-27
Refactor globe view example

Pessimistress:
Bring in some of the refactor from #7681 

#### Change List
- Move the grouping and culling of flights into a composite layer. This added complexity makes the React component hard to read.
- Move the processing of timestamps in with the data loading logic. It is specific to this dataset (we try to show several disjoint dates in one smooth animation) and should not be lumped in with the rendering side of things.


chrisgervang (approved):
Looks cleaner. I notice we have a variety of custom layers implemented throughout our examples, but they're not super discoverable. Perhaps they'd be useful to readers of the [Writing Custom Layers](https://deck.gl/docs/developer-guide/custom-layers) articles?

PR #7702, opened 2023-02-28
docs: correct division sign to multiplication

jamesscottbrown:
I *think* this is a typo in the docs.

I think a multiplication sign would mean that the map is scaled such that whatever circle of latitude is being viewed corresponds to 512 common units, which seems reasonable.

A multiplication sign would also be consistent with how OSM map tlles work - the OSM wiki states [1]:

> The horizontal distance represented by each square tile, measured along the parallel at a given latitude, is given by:
> $S_{tile} = C ∙ cos(latitude) / 2^{zoomlevel}$

I initially noticed this because points on a scatteplot layer were about bigger than expected, and the error seemed to be roughly a factor of $1/cos^2(\phi)$,

[1]: https://wiki.openstreetmap.org/wiki/Zoom_levels

### DIFF
--- a/docs/developer-guide/coordinate-systems.md
+++ b/docs/developer-guide/coordinate-systems.md
@@ -132,7 +132,7 @@ The most performant way to transform the dimensions for each object is to use an
 
 The conversion between meter sizes and common sizes depend on the projection mode:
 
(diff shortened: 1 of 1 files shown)


Pessimistress (approved):
Thank you!

PR #7709, opened 2023-03-02
Fix docusaurus website dark mode

Pessimistress:
For #7635 

#### Change List
- Use theme variables in styled components
- Switch map style based on theme

### DIFF
--- a/website/src/doc-demos/demo-base.js
+++ b/website/src/doc-demos/demo-base.js
@@ -2,7 +2,7 @@ import React from 'react';
 import DeckGL from '@deck.gl/react';
 import {Map} from 'react-map-gl';
 import styled from 'styled-components';
-
+import { useColorMode } from '@docusaurus/theme-common'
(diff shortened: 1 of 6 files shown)


website/src/doc-demos/demo-base.js
```diff
@@ -83,22 +83,27 @@ function evalObject(source, globals, output) {
...
+    return text && {
+      text,
+      style: TOOLTIP_STYLE
     };
+  };
+
+  const layerProps = evalObject(props, imports);
+
+  function Demo() {
+    const { isDarkTheme } = useColorMode();
```
chrisgervang, line 101: This seems to be deprecated.
Screen Shot 2023-03-03 at 5 10 58 PM

chrisgervang (approved):
Switching the map style is a nice touch. Looks so nice I think we should do it for the home page hero, and other examples as well!

PR #7712, opened 2023-03-02
Collisions filter example

menusal:
For [#7635](https://github.com/visgl/deck.gl/issues/7635)

@Pessimistress New example for [https://deck.gl/examples](https://deck.gl/examples). The idea is to make something that shows how the new Collision Filter Extension shows collisions filters in GeoJSON and Text layers.

#### Background
image

https://user-images.githubusercontent.com/5357085/224036932-cbec269c-ec32-4faf-aa8f-ada88b2ddfff.mp4

#### Change List
- new example app


cayetanobv:
> For [#7635](https://github.com/visgl/deck.gl/issues/7635)
> 
> @Pessimistress New example for https://deck.gl/examples. The idea is to make something that shows how the new Collision Filter Extension shows collisions filters in GeoJSON and Text layers.
> 
> #### Background
> image
> 
> #### Change List
> * new example app

@menusal Could you change the image? 
image

examples/website/collision-filter/app.jsx
```diff
@@ -0,0 +1,102 @@
+import React from 'react';
+import {createRoot} from 'react-dom/client';
+import {Map} from 'react-map-gl';
+import maplibregl from 'maplibre-gl';
+import DeckGL from '@deck.gl/react';
+import {GeoJsonLayer, TextLayer} from '@deck.gl/layers';
+import data from './data/road_accidents.json';
```
cayetanobv, line 7: We need to remember to change data sources to [deck.gl-data](https://github.com/visgl/deck.gl-data)

cayetanobv (changes requested):

examples/website/collision-filter/app.jsx
cayetanobv: Added in this PR: https://github.com/visgl/deck.gl-data/pull/31

cayetanobv:
> I am not convinced that this example is the best showcase for the extension. We already have a [TextLayer example](https://pessimistress.github.io/deck.gl/examples/text-layer) that shows collision among text labels. The case for the scatterplot is relatively weak. Usually, the desired way to visualize overlapping points at low zoom levels is to show heatmap or clusters instead of simply hiding some, i.e. by aggregation, not reduction. The [HexagonLayer example](https://deck.gl/examples/hexagon-layer/) presents exactly the same dataset, and this new example in my opinion offers neither a more appealing representation nor a different insight.

Hi @Pessimistress , I agree we can find another better example as a showcase for this extension, but the case we are presenting is not weak. We are using ordinal data, which is categorical data: road accidents by severity (with three ordered categories). For this reason, I don't think we can replace it properly with a heatmap. Clustering doesn't work totally fine in this case either. Actually, ordered data is pretty hard to visualize on a map when you have a lot of data. The key here is that using this Extension we can find a valid way, although not perfect, to show where the importance of the variable is distributed: the fatality of accidents, not the number of accidents. Obviously, you need to play with the radius to find a good equilibrium at different zoom levels. On the other hand, the example you are referring (HexagonLayer) uses another dataset of the same data source and is showing a quantitative variable: the number of accidents by cell. This means that our use case is adding new insights. Finally, the labels layer is included only as a complement to see where the variable is distributed; the text layer example you mention is better to show the effect of the extension on labels.

alasarr:
Ey guys! Agree with @Pessimistress here, the example is good "thematic" example, but not the first example to understand how the extension works. 

I'd use this extension to add text labels and icons over a line or polygon. I'd work on that direction

alasarr:
In fact, we developed this feature, to improve the capabilities of labeling, the example should go in that direction because many users will use it for that purpose

alasarr:
thanks! I appreciate!

examples/website/collision-filter/app.jsx
```diff
@@ -0,0 +1,99 @@
...
+  sizeScale = 4,
+  collisionEnabled = true,
+  routeName = 'US-101',
+}) {
+  const roadName = routeName.split('-');
+  const filteredRoads =
+    routeName === ALL_ROUTES
+      ? roads
+      : roads.features.filter(f => f.properties.number === roadName[1]);
+  const filteredLabels =
```
Pessimistress, line 25: - Labels should be generated from the road geometry, not loaded from a GeoJSON
- The angles do not look correct

website/src/examples/collision-filter-extension.js
```diff
@@ -0,0 +1,126 @@
...
+
+import {makeExample} from '../components';
+
+class CollisionDemo extends Component {
+  static title = 'California Collision Test';
+
+  static code = `${GITHUB_TREE}/examples/website/collision-filter`;
+
+  static parameters = {
+    routeName: {
```
Pessimistress, line 14: I don't think this filter adds much value. Just show everything.

examples/website/collision-filter/README.md
```diff
@@ -0,0 +1,27 @@
...
+npm install
+# or
+yarn
+# bundle and serve the app with vite
+npm start
+```
+
+### Data sources
+
+#### Road accidents
```
Pessimistress, line 18: Data credit needs to be updated

menusal: @Pessimistress `WIP` - I'll inform you about the following review.

examples/website/collision-filter/app.jsx
menusal: Fixing angle
@cayetanobv `Labels should be generated from the road geometry, not loaded from a GeoJSON`

website/src/examples/collision-filter-extension.js
menusal: wip

examples/website/collision-filter/app.jsx
Pessimistress: FYI codepen to align label with path: https://codepen.io/Pessimistress/pen/OJgmXba?editors=0010

examples/website/collision-filter/README.md
cayetanobv: Data credits updated

examples/website/collision-filter/app.jsx
```diff
@@ -0,0 +1,149 @@
+import React from 'react';
+import {createRoot} from 'react-dom/client';
+import {Map} from 'react-map-gl';
+import maplibregl from 'maplibre-gl';
+import DeckGL from '@deck.gl/react';
+import {GeoJsonLayer, TextLayer} from '@deck.gl/layers';
+import roads from './data/ne_10_roads_filtered_usa_california.json';
```
alasarr, line 7: Why don't we use a bigger dataset @cayetanobv? An example with only 20 roads is not very impressive. We can use a bigger area or use another country, like Spain for instance.

cayetanobv: Finally, Mexico is the winner. It works perfectly.

website/src/examples/collision-filter-extension.js
```diff
@@ -0,0 +1,57 @@
...
+      value: true
+    }
+  };
+
+  static mapStyle = MAPBOX_STYLES.DARK;
+
+  static renderInfo(meta) {
+    return (
+      <div>
+        <p>Principals roads from Mexico country</p>
```
Pessimistress, line 33:
```suggestion
        <p>Principal roads from Mexico</p>
```

examples/website/collision-filter/app.jsx
```diff
@@ -0,0 +1,158 @@
...
+const LINE_COLOR = [0, 173, 230];
+
+function calculateLabels(data, pointSpacing) {
+  const routes = data.features.filter(d => d.geometry.type !== 'Point');
+
+  // Add points along the lines
+  const filteredLabels = routes.map(d => {
+    const lineLength = Math.floor(turf.lineDistance(d.geometry));
+
+    const result = {
```
Pessimistress, line 22: There is no reason to create new GeoJSON features if you are not even using a GeoJsonLayer. Just return the generated info with a link back to the source feature.

```ts
{
  label: string;
  anchor: [number, number];
  angle: number;
  priority: number;
  road: Feature;
}[]
```

examples/website/collision-filter/package.json
```diff
@@ -0,0 +1,24 @@
...
+  "scripts": {
+    "start": "vite --open",
+    "start-local": "vite --config ../../vite.config.local.mjs",
+    "build": "vite build"
+  },
+  "dependencies": {
+    "deck.gl": "8.9.0-beta.3",
+    "react": "^18.0.0",
+    "react-dom": "^18.0.0",
+    "d3-scale": "^2.0.0",
```
Pessimistress, line 15: Not used?

examples/website/collision-filter/app.jsx
```diff
@@ -0,0 +1,137 @@
...
+
+    const result = [];
+
+    function addPoint(lineString, dAlong, priority) {
+      let offset = 1;
+      if (dAlong > 0.5 * lineLength) offset *= -1;
+      const feature = turf.along(lineString, dAlong);
+      const nextFeature = turf.along(lineString, dAlong + offset);
+      const {coordinates} = feature.geometry;
+      const next = turf.point(nextFeature.geometry.coordinates).geometry.coordinates;
```
Pessimistress, line 32:
```suggestion
      const next = nextFeature.geometry.coordinates;
```

examples/website/collision-filter/app.jsx
```diff
@@ -0,0 +1,137 @@
...
+const initialViewState = {longitude: -100, latitude: 24, zoom: 5, minZoom: 5, maxZoom: 12};
+
+function calculateLabels(data, pointSpacing) {
+  const routes = data.features.filter(d => d.geometry.type !== 'Point');
+
+  // Add points along the lines
+  const filteredLabels = routes.map(d => {
+    const lineLength = Math.floor(turf.lineDistance(d.geometry));
+
+    const result = [];
```
Pessimistress, line 24: Move `result` declaration out of the loop and skip the `reduce` step in the end:

```js
const result = [];

const addPoint = ...

for (const feature of routes) {
  // generate points
}

return result;
```

Pessimistress (approved):

PR #7723, opened 2023-03-06
Fix bug with picking in overlay mouse handlers before deck is initialized

JannikGM:
This is branched off from #7668 to fix the existing bugs, but without the controversial changes to the API (which remain in #7668).

#### Background

I've originally reported a bug on Slack: https://deckgl.slack.com/archives/C34AC5MSQ/p1676629109034669
The picking functions were used by the mapbox event handlers before deck was fully initialized.

This should hopefully land in v8.x before any breaking changes (in deck.gl v9.0).

#### Change List
- Avoid picking in GoogleMapsOverlay before deck is initialized
- Avoid picking in MapboxOverlay before deck is initialized

---

**None of this was tested**
I'm not sure how to run automated tests for the deck.gl codebase. `yarn bootstrap` fails with gyp not able to find `python` (which should be `python3` on this macOS machine). `yarn test` failed during linter step in untouched code.
All my deck.gl projects use another private library which has hardcoded dependency on a specific deck.gl version, so manual testing would be cumbersome.

### DIFF
--- a/modules/google-maps/src/utils.ts
+++ b/modules/google-maps/src/utils.ts
@@ -286,6 +286,10 @@ function getEventPixel(event, deck: Deck): {x: number; y: number} {
 
 // Triggers picking on a mouse event
 function handleMouseEvent(deck: Deck, type: string, event) {
+  if (!deck.isInitialized) {
+    return;
+  }
+
   const mockEvent: Record<string, any> = {
     type,
     offsetCenter: getEventPixel(event, deck),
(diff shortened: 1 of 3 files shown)


Pessimistress:
@JannikGM You need to comment out these tests for CI to pass:

https://github.com/visgl/deck.gl/blob/574caf1e43a184b3822e1e44d45f3912588592e4/test/modules/google-maps/google-maps-overlay.spec.js#L202-L208

Deck will never be initialized in this test with the mock map API.

Pessimistress:
@JannikGM literally commenting them out is fine. We will look into improving the mock Map class later to restore these tests.

Issue #7708, opened 2023-03-02
[Bug] `IconLayer` fails shallow comparison of `textureParameters` prop, because it's not deep-compared


JannikGM:
### Description

`IconLayer` with `textureParameters` will fail a shallow comparison when naively passing an object.
This is in contast to the `Layer` prop `parameters` (which accepts similar values) which does a deep-compare instead.

I've originally brought this up on Slack in https://deckgl.slack.com/archives/C34AC5MSQ/p1677684985788469:

> Shouldn't `textureParameters` on `IconLayer` behave similar to `parameters` on base `Layer` (~ deep compare: `{type: 'object', value: {}, optional: true, compare: true, depth: 1},`)?
>
> Currently, having
> ```
> textureParameters: {
>   [GL.TEXTURE_MIN_FILTER]: GL.NEAREST,
>   [GL.TEXTURE_MAG_FILTER]: GL.LINEAR
> }
>  ```
> will cause shallow comparison failure which triggers an update to the layer.

### Flavors

- [ ] React
- [ ] Python/Jupyter notebook
- [ ] MapboxLayer
- [ ] GoogleMapsOverlay
- [ ] CartoLayer
- [ ] DeckLayer/DeckRenderer for ArcGIS

### Expected Behavior

There should be a deep-comparison, no failed comparison should be logged at log level 3.

- `parameters` on `Layer` has an explicit prop-type: https://github.com/visgl/deck.gl/blob/2cab2c75ca48ecfbddb94515241a32368a99af2d/modules/core/src/lib/layer.ts#L160
- `textureParameters`on `IconLayer`  **does not**  have an explicit prop-type: https://github.com/visgl/deck.gl/blob/2cab2c75ca48ecfbddb94515241a32368a99af2d/modules/layers/src/icon-layer/icon-layer.ts#L131-L132

### Steps to Reproduce

/

### Environment

- Framework version: All
- Browser: /
- OS: /

Pessimistress:
Related to https://github.com/visgl/deck.gl/issues/7506

@felixpalmer Now I think more about this, I do not believe deep comparison is the right thing to do. Prop changes in `loadOptions` and `textureParameters` are NOT handled by the layers. The documentation even states that they are only used when resources are loaded. I think what we need is a "ignore" flag in the prop type.

felixpalmer:
@Pessimistress, agreed - it also saves us needlessly comparing potentially deep parameters objects

PR #7726, opened 2023-03-07
Prop types API improvements

Pessimistress:
For #7708

This PR should NOT introduce any breaking changes to the prop types API (compared to v8.8).

#### Change List
- Consistent behavior cross prop types array, object, function: use `ignore` to disable comparison, `compare` to set depth of comparison
- Some layers props definitions
- Documentation
- Unit tests


docs/developer-guide/custom-layers/prop-types.md
```diff
@@ -181,14 +201,14 @@ A function.
 
 - Options:
   + `optional` (boolean, optional) - accept `null` or `undefined`. Default `false`.
-  + `compare` (boolean, optional) - compare strictly during prop comparison. Default `true`.
+  + `ignore` (boolean, optional) - if `true`, prop comparison is disabled. Default `true`.
 - Default `validate`: value is a function
 - Default `equal`: `true` if `compare: false`, otherwise strict equal
```
felixpalmer, line 206: Comment for `equal` doesn't make sense with `compare` option removed

modules/core/src/lifecycle/prop-types.ts
```diff
@@ -54,15 +54,29 @@ type ColorPropType = BasePropType<Color | null> & {
 type ArrayPropType<T = any[]> = BasePropType<T> & {
   type: 'array';
   optional?: boolean;
-  compare?: boolean;
+  /** Ignore change in the prop value.
+   * @default false
+   */
+  ignore?: boolean;
+  /** Deep-compare two prop values. Only used if `ignore: false`.
+   * When a number is supplied, used as the depth of depp-comparison. 0 is equivalent to shallow comparison, -1 is infinite depth
```
felixpalmer, line 62:
```suggestion
   * When a number is supplied, used as the depth of deep-comparison. 0 is equivalent to shallow comparison, -1 is infinite depth
```

modules/core/src/lifecycle/prop-types.ts
```diff
@@ -74,8 +88,16 @@ type ImagePropType = BasePropType<Texture | null> & {
...
   type: 'object';
   optional?: boolean;
-  compare?: boolean;
-  depth?: number;
+  /** Ignore change in the prop value.
+   * @default false
+   */
+  ignore?: boolean;
+  /** Deep-compare two prop values. Only used if `ignore: false`.
+   * When a number is supplied, used as the depth of depp-comparison. 0 is equivalent to shallow comparison, -1 is infinite depth
```
felixpalmer, line 96:
```suggestion
   * When a number is supplied, used as the depth of deep-comparison. 0 is equivalent to shallow comparison, -1 is infinite depth
```

felixpalmer (approved):

docs/developer-guide/custom-layers/prop-types.md
```diff
@@ -161,6 +162,25 @@ MyLayerClass.defaultProps = {
 }
 ```
 
+##### `object` {#array}
+
+An arbitrary object.
+
+- Options:
+  + `optional` (boolean, optional) - accept `null` or `undefined`. Default `false`.
+  + `ignore` (boolean, optional) - if `true`, prop comparison is disabled. Default `false`.
```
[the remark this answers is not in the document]
Pessimistress: Some prop changes are already ignored. For example, changing `onHover` does not trigger a layer update, it will only be applied next time an hover event occurs.

Similarly, layers never handled changes in `loaders` and `loadOptions`. They are only used when data URL changes.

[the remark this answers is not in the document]
Pessimistress: > it's impossible to change this prop after it has been set once.

They can be changed, just that the change alone does not trigger any layer action (update or redraw) because they are "passive". Their values will be respected when other situations arise that do require a layer action.

PR #7734, opened 2023-03-09
docs(mapbox) discoverability for integration examples

chrisgervang:
#### Change List
- update standalone version
- use maplibre
- example links

### DIFF
--- a/docs/api-reference/mapbox/overview.md
+++ b/docs/api-reference/mapbox/overview.md
@@ -13,8 +13,9 @@ This module makes it easy to use deck.gl as native layers and controls in the Ma
 ### Include the Standalone Bundle
 
 ```html
-<script src="https://unpkg.com/deck.gl@^8.1.0/dist.min.js"></script>
-<script src='https://api.tiles.mapbox.com/mapbox-gl-js/v1.10.0/mapbox-gl.js'></script>
+<script src="https://unpkg.com/deck.gl@^8.9.0/dist.min.js"></script>
+<script src='https://unpkg.com/maplibre-gl@latest/dist/maplibre-gl.js'></script>
+<link href='https://unpkg.com/maplibre-gl@latest/dist/maplibre-gl.css' rel='stylesheet' />
 <script type="text/javascript">
   const {MapboxOverlay} = deck;
 </script>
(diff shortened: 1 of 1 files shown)


docs/api-reference/mapbox/overview.md
```diff
@@ -30,6 +31,17 @@ npm install @deck.gl/mapbox
 import {MapboxOverlay} from '@deck.gl/mapbox';
 ```
 
+## Examples
+
+[test/apps/mapbox-integration](https://github.com/visgl/deck.gl/tree/8.9-release/test/apps/mapbox-integration)
```
chrisgervang, line 36: Link to master or release?

Pessimistress: We should not link to test/apps from docs. These are not maintained with the mindset that they will be used by external users. If you think our existing examples are insufficient you can add more to examples/get-started

chrisgervang: Ok, that makes sense. I'd like to maintain two public examples for the mapbox integration using interleaved MapboxOverlay with maplibre in react and pure-js since it's a common usage when someone reaches for @deck.gl/mapbox.

docs/api-reference/mapbox/overview.md
```diff
@@ -13,8 +13,9 @@ This module makes it easy to use deck.gl as native layers and controls in the Ma
 ### Include the Standalone Bundle
 
 ```html
-<script src="https://unpkg.com/deck.gl@^8.1.0/dist.min.js"></script>
-<script src='https://api.tiles.mapbox.com/mapbox-gl-js/v1.10.0/mapbox-gl.js'></script>
+<script src="https://unpkg.com/deck.gl@^8.9.0/dist.min.js"></script>
+<script src='https://unpkg.com/maplibre-gl@latest/dist/maplibre-gl.js'></script>
+<link href='https://unpkg.com/maplibre-gl@latest/dist/maplibre-gl.css' rel='stylesheet' />
```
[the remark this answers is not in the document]
chrisgervang: You bring up some good points to consider. We're currently focused on supporting both libraries with the same integration module since they're effectively drop-in replacements for most use cases.

Doing a hard split would add maintenance overhead to consider as well, since there are many things to test and most of the code would remain the same between the modules.

I agree showing both library options is probably more useful at this stage, as well as tracking the differences and limitations common to both, and unique to each.

chrisgervang:
I'm going to close this one. A lot of the feedback has been incorporated into v9, namely we've added separate examples for mapbox and maplibre (more in #8541) that cover different environments (pure js, react, scripting) and rendering modes (overlaid and interleaved).

PR #7750, opened 2023-03-13
Replace proj4 with lightweight alternative

Pessimistress:
Saves ~78k in bundle size

#### Change List
- Move proj4 to dev dependencies (for testing)
- Alternative projection util
- Tests


ibgreen (approved):
Neat. Wonder if we should push it upstream to math.gl

kylebarron (approved):
Agree that proj4 has a huge overhead and shouldn't be a default dependency

PR #7757, opened 2023-03-15
TextLayer.fontSettings propType compare depth 1

felixpalmer:
#### Background

`TextLayer.fontSettings` propType doesn't include a compare depth leading to unnecessary state updates


modules/layers/src/text-layer/text-layer.ts
```diff
@@ -218,7 +218,7 @@ const defaultProps: DefaultProps<TextLayerProps> = {
   lineHeight: DEFAULT_LINE_HEIGHT,
   outlineWidth: {type: 'number', value: 0, min: 0},
   outlineColor: {type: 'color', value: DEFAULT_COLOR},
-  fontSettings: {},
+  fontSettings: {type: 'object', value: {}, optional: true, compare: 1},
```
Pessimistress, line 221: `optional` means it can be null, I don't think you should set that.

Pessimistress (approved):

PR #7762, opened 2023-03-15
Fix bug when using default value in attributes with size greater than 1

gabeboning:
I was experiencing a bug that traced back to the attribute auto updater using data left over in reallocated buffers. This happens when an attribute's size is larger than 1 and the accessor returns `undefined`:

In the auto updater, the relevant block is https://github.com/visgl/deck.gl/blob/8.9-release/modules/core/src/lib/attribute/attribute.ts#L404-L410. When `objectValue` is undefined, calling `attribute._normalizeValue` ends up in https://github.com/visgl/deck.gl/blob/8.9-release/modules/core/src/lib/attribute/data-column.ts#L511-L514 -- which only puts `defaultValue[0]` into `out`. Thus when the auto updater calls [`fillArrray`](https://github.com/visgl/deck.gl/blob/master/modules/core/src/lib/attribute/attribute.ts#L405), it will only copy over a single value per vertex (`out` is plumbed through to where `fillArray` determines how many values to populate in `target`: https://github.com/visgl/deck.gl/blob/8.8-release/modules/core/src/utils/flatten.ts#L61 ). Since only one value per vertex is being populated, there are many values in the array left with whatever was in them last time they were handed back to the allocator!

### DIFF
--- a/modules/core/src/lib/attribute/data-column.ts
+++ b/modules/core/src/lib/attribute/data-column.ts
@@ -509,7 +509,10 @@ export default class DataColumn<Options, State> implements IShaderAttribute {
       return out;
     }
     if (!value) {
-      out[start] = defaultValue[0];
+      let i = size;
+      while (--i >= 0) {
+        out[start + i] = defaultValue[i];
+      }
       return out;
     }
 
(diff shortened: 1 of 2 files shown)


Pessimistress (approved):
I think the original intention was that the non-numeric accessors should never return `undefined`, but the implementation definitely leads to some confusing behaviors. Since we are not throwing when encountering `undefined` I agree that this is the right thing to do.

Issue #7776, opened 2023-03-21
[Bug] Artifact when using CollisionFilterExtension with text outline


felixpalmer:
### Description

When creating a `TextLayer` with the `CollisionFilterExtension` the sdf outline drawing can in some cases create an ugly artifact across a quarter of the view

https://user-images.githubusercontent.com/453755/226573412-7e621255-7b31-40a4-b7b5-9a965aa7cbc2.mov

### Flavors

- [ ] React
- [ ] Python/Jupyter notebook
- [ ] MapboxLayer
- [ ] GoogleMapsOverlay
- [ ] CartoLayer
- [ ] DeckLayer/DeckRenderer for ArcGIS

### Steps to Reproduce

NOTE: you need to resize to window to observe this bug, it doesn't happen at all resolutions

```js
const {DeckGL, TextLayer, CollisionFilterExtension} = deck;

const layer = new TextLayer({
  id: 'TextLayer',
  data: 'https://raw.githubusercontent.com/visgl/deck.gl-data/master/website/bart-stations.json',
  fontSettings: {sdf: true},
  getAlignmentBaseline: 'center',
  getColor: [255, 128, 0],
  getPosition: d => d.coordinates,
  getText: d => d.name,
  getSize: 12,
  getTextAnchor: 'middle',
  outlineColor: [255, 0, 0, 255],
  outlineWidth: 5,
  extensions: [new CollisionFilterExtension()],
});
new DeckGL({
  mapStyle: 'https://basemaps.cartocdn.com/gl/positron-gl-style/style.json',
  initialViewState: {
    longitude: -122.4,
    latitude: 37.74,
    zoom: 11,
    maxZoom: 20,
    pitch: 30,
    bearing: 0
  },
  controller: true,
  layers: [layer]
});
  
```

### Environment

- Framework version:
- Browser:
- OS:

Pessimistress:
Fixed included in v8.9.4

PR #7777, opened 2023-03-21
CollisionFilterExtension: Discard outside clip space

felixpalmer:
For https://github.com/visgl/deck.gl/issues/7776

#### Background

Using `vec4(0.)` to discard a vertex seems to have precision issues, safer to have a point that is clearly outside (e.g. where `w !== 0`.

#### Change List
- Discard vertex by placing outside clip space

### DIFF
--- a/modules/extensions/src/collision-filter/shader-module.ts
+++ b/modules/extensions/src/collision-filter/shader-module.ts
@@ -72,7 +72,8 @@ const inject = {
     vec2 collision_texCoords = collision_getCoords(collision_common_position);
     collision_fade = collision_isVisible(collision_texCoords, geometry.pickingColor / 255.0);
     if (collision_fade < 0.0001) {
-      position = vec4(0.);
+      // Position outside clip space bounds to discard
+      position = vec4(0.0, 0.0, 2.0, 1.0);
     }
   }
   `,


Pessimistress:
Interesting. Have you seen any documentation / explanation of this from an external source? If you search for `gl_Position = vec4(0.);` in the repo we use it in quite a few places.

felixpalmer:
This [comment](https://community.khronos.org/t/some-clarification-about-clipping/74041/7) is saying that `vec4(0.)` is not in fact clipped. So our code dropping vertices is collapsing triangles rather than moving them out of clip space. What is not clear to me is where this vertex is positioned, e.g. if we draw a line from `vec4(-1.0, -1.0, 0.0, 1.0)` to `vec4(0.)`

With the clip coordinate being `x/w, y/w, z/w` choosing `vec4(0.)` seems like a risky choice if the goal is to entirely discard a triangle, when `vec4(0.0, 0.0, 2.0, 1.0)` achieves the same thing. 

For completeness, here are all the locations where we do this in shaders

```
modules/layers/src/line-layer/line-layer-vertex.glsl.ts
83:      gl_Position = vec4(0.);

modules/layers/src/solid-polygon-layer/solid-polygon-layer-vertex-main.glsl.ts
56:    gl_Position = vec4(0.);

modules/extensions/src/data-filter/shader-module.ts
151:      gl_Position = vec4(0.);

```

I would hesitate to change the first two as the discarded vertex is part of a larger shape and could introduce artifacts of its own. I did see if I could reproduce the bug with the `DataFilterExtension` but I could not.

Pessimistress (approved):

Issue #7783, opened 2023-03-22
[Bug] pydeck widget tests fail


ajduberstein:
### Description

One of the unpinned dependencies in pydeck seems to be causing issues with the widget-specific tests. See #7781.

### Flavors

- [ ] React
- [X] Python/Jupyter notebook
- [ ] MapboxLayer
- [ ] GoogleMapsOverlay
- [ ] CartoLayer
- [ ] DeckLayer/DeckRenderer for ArcGIS

### Steps to Reproduce

`make test` or `make ci` e.g.  https://github.com/visgl/deck.gl/actions/runs/4482393204/jobs/7880901714

### Environment

- Framework version: pydeck 0.8.4b

donmccurdy:
As the tests have been disabled in #7784, I've changed the type of this issue from bug to chore and hope to look at fixing the test as part of the pydeck v9 release.

PR #7784, opened 2023-03-22
Skip widget tests temporarily

ajduberstein:
For #7783

#### Background

CI tests started failing: https://github.com/visgl/deck.gl/actions/runs/4482393204/jobs/7880901714 Error seems to be related to an ipywidgets dependency but this needs more investigation.

#### Change List
- Skips failing tests


bindings/pydeck/pydeck/_version.py
```diff
@@ -1 +1 @@
-__version__ = "0.8.0"
+__version__ = "0.8.0b4"
```
Jesus89, line 1: Why 0.8.0b4? I think it should be 0.8.1 in case a bump is needed here

Jesus89 (approved):

PR #7798, opened 2023-03-28
CARTO: data loading performance improvements

felixpalmer:
#### Background

The parsing of protobuffers is slower than it needs to be, due to the implementation iterating over large arrays of fixed size data, when it can just map the whole array into memory in some cases.

#### Change List
- Implement direct memory copy versions of `readPackedDouble` and `readPackedFixed64`
- Return the correct data structure from `TileReader` to avoid second pass to `unpackProperties`
- Speed up picking with spatial index layers by exploiting the tile/data heirarchy

### DIFF
--- a/modules/carto/src/layers/schema/carto-spatial-tile-loader.ts
+++ b/modules/carto/src/layers/schema/carto-spatial-tile-loader.ts
@@ -1,7 +1,7 @@
 import {LoaderOptions, LoaderWithParser} from '@loaders.gl/loader-utils';
 
 import {Tile, TileReader} from './carto-spatial-tile';
-import {parsePbf, unpackProperties} from './tile-loader-utils';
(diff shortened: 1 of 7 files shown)


ibgreen (commented):
Don't mean to get involved in carto module reviews, but since the PR mentions "performance improvements" I got curious... Would any of these 100x optimizations make sense for loaders.gl or are they strictly carto specific?

felixpalmer:
@ibgreen it is possible, yes. Whenever pbf iterates over single values and decodes them separately the same trick could be used. It doesn't work if the values as encoded as variable length though.

zbigg (approved):
LGTM

TIL: that we webplatform just use native endianess, but assuming that 99% of the world works in [little-endian mode](https://stackoverflow.com/questions/7869752/javascript-typed-arrays-and-endianness), we're safe

PR #7814, opened 2023-04-04
Honor changing view controller type.

hkfb:
#### Background
When replacing existing Views with changed Controller types, if the new Views happen to have the same ids, then the new Controller types are not honored, and instead the old, possibly incompatible Controllers are used instead.

#### Change List
- Create new Controller if type has changed

### DIFF
--- a/modules/core/src/lib/view-manager.ts
+++ b/modules/core/src/lib/view-manager.ts
@@ -326,8 +326,9 @@ export default class ViewManager {
         height: viewport.height
       };
 
-      // TODO - check if view / controller type has changed, and replace the controller
-      if (!controller) {
+      // Create controller if not already existing or if the type of the
+      // controller has changed.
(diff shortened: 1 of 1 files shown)


felixpalmer (changes requested):
This doesn't check if the view has changed, as the original TODO mentions. Could you also add that? Please also add some tests that demonstrate that this works

hkfb:
> This doesn't check if the view has changed, as the original TODO mentions. Could you also add that? Please also add some tests that demonstrate that this works

I think it's the view / controller combination that is the intention of the TODO comment. The change of the View is handled elsewhere.

I added  a test that demonstrates the issue. It fails without the changes in this PR.

hkfb:
@felixpalmer would you mind taking another look?

felixpalmer (approved):

PR #7817, opened 2023-04-06
Fix depth picking from GeoJSON layer for binary features

gabeboning:
Depth picking from binary features does not work when the features are not all the same type (i.e. not all polygons). This is because the global feature index -- the index in the GeoJSON equivalent -- does not necessarily match the index of the feature/datum in each of the point/polygon/line sublayers.

This PR adds a `_getSublayerPickingIndex` method to allow composite layers to translate picked ids from a global id into the correct one (or multiple, to correctly handle MultiPolygon/MultiLineString cases where a single global id has multiple local ids) to be used for clearing the picking buffer.

It doesn't feel like the cleanest approach, but I do like that it keeps the knowledge about indices in the layer that would know (the composite parent). Happy to attempt it some other way if anyone has a proposed better approach; though this does seem to work reliably.


modules/core/src/lib/layer.ts
```diff
@@ -746,7 +746,24 @@ export default abstract class Layer<PropsT extends {} = {}> extends Component<
   /** (Internal) Sets the picking color at the specified index to null picking color. Used for multi-depth picking.
      This method may be overriden by layer implementations */
   disablePickingIndex(objectIndex: number): void {
-    this._disablePickingIndex(objectIndex);
+    // When rendered by a composite layer, the objectIndex may be indexed by all data in the composite layer, but this
+    // layer only renders a subset. So we allow the parent layer to translate the index into the correct one for this layer.
+    if (this.parent) {
+      const transformedIndex = this.parent._getSublayerPickingIndex(objectIndex);
```
Pessimistress, line 752: I am not a fan of this solution. It adds yet another hidden API and complicates the interaction between a composite layer and its sub layers.

The issue you are trying to address here is that `objectIndex` does not map to the index into `props.data`. The problem exists because picking colors in the binary GeoJSON case is recalculated and passed down to the sub layers via `data.attribute.instancePickingColors`. You can resolve the index transform locally by searching within this value:

```js
disablePickingIndex(objectIndex: number) {
  const {pickingColors, instancePickingColors} = this.getAttributeManager().attributes;
  const colors = pickingColors || instancePickingColors;
  const {attributes, length} = this.props.data;
  const externalColorAttribute = colors && attributes && attributes[colors.id];
  if (externalColorAttribute) {
    const values = externalColorAttribute.value;
    const objectColor = this.encodePickingColor(objectIndex);
    for (let index = 0; index < length; index ++) {
      const i = colors.getVertexOffset(index);
      if (values[i] === objectColor[0] &&
        values[i + 1] === objectColor[1] &&
        values[i + 2] === objectColor[2]) {
          objectIndex = index;
          break;
      }
    }
  }
  this._disablePickingIndex(objectIndex);
}

Comment (gabeboning) [reply modules/core/src/lib/layer.ts:752]: would you prefer to put that logic here (in layer.ts) or distribute it between the path and solid polygon layers (akin to the current handling in those layers for when they are used in the geojson layer: https://github.com/visgl/deck.gl/blob/master/modules/layers/src/path-layer/path-layer.ts#L292-L300)

Comment (Pessimistress) [reply modules/core/src/lib/layer.ts:752]: It can go into layer.ts.

Comment (gabeboning) [reply modules/core/src/lib/layer.ts]: Updated to search for matching values (can't exit early from the search loop -- that doesn't work for multi polygons/lines) when there is a picking color attribute defined

File: modules/core/src/lib/layer.ts [hunk 2]
```diff
@@ -745,8 +747,35 @@ export default abstract class Layer<PropsT extends {} = {}> extends Component<
...
   /** (Internal) Sets the picking color at the specified index to null picking color. Used for multi-depth picking.
      This method may be overriden by layer implementations */
-  disablePickingIndex(objectIndex: number): void {
-    this._disablePickingIndex(objectIndex);
+  disablePickingIndex(objectIndex: number) {
+    // @ts-ignore (TS2531) this method is only called internally with attributeManager defined
+    const {pickingColors, instancePickingColors} = this.getAttributeManager().attributes;
+
+    const colors = pickingColors || instancePickingColors;
+    const binaryData = this._getBinaryData();
```
Pessimistress, line 755: A new method seems excessive? You can get the correct type by

```ts
disablePickingIndex(objectIndex: number) {
	const data = this.props.data as LayerData<any>;
	if (!('attributes' in data)) {
	  this._disablePickingIndex(objectIndex);
	  return;
	}
	// carry on check external attributes
```

modules/core/src/lib/layer.ts
```diff
@@ -48,10 +48,12 @@ import type Attribute from './attribute/attribute';
 import type {Model} from '@luma.gl/engine';
 import type {PickingInfo, GetPickingInfoParams} from './picking/pick-info';
 import type Viewport from '../viewports/viewport';
-import type {NumericArray} from '../types/types';
+import type {NumericArray, TypedArray} from '../types/types';
```
Pessimistress, line 51:
```suggestion
import type {NumericArray} from '../types/types';
```

modules/core/src/lib/layer.ts
```diff
@@ -48,10 +48,12 @@ import type Attribute from './attribute/attribute';
...
 import type {DefaultProps} from '../lifecycle/prop-types';
-import type {LayerProps} from '../types/layer-props';
+import type {LayerData, LayerProps} from '../types/layer-props';
 import type {LayerContext} from './layer-manager';
+import type {Buffer} from '@luma.gl/webgl';
```
Pessimistress, line 55:
```suggestion
```

modules/core/src/lib/layer.ts
```diff
@@ -48,10 +48,12 @@ import type Attribute from './attribute/attribute';
...
+import {BinaryAttribute} from './attribute/attribute';
```
Pessimistress, line 56:
```suggestion
```

modules/core/src/lib/layer.ts
```diff
@@ -745,8 +747,38 @@ export default abstract class Layer<PropsT extends {} = {}> extends Component<
...
+    const {pickingColors, instancePickingColors} = this.getAttributeManager().attributes;
+
+    const colors = pickingColors || instancePickingColors;
+    const data = this.props.data as LayerData<any>;
+    if (!('attributes' in data)) {
+      this._disablePickingIndex(objectIndex);
+      return;
+    }
+
+    const externalColorAttribute = data.attributes?.[colors.id];
```
Pessimistress, line 761:
```suggestion
    const data = this.props.data as LayerData<any>;
    if (!('attributes' in data)) {
      this._disablePickingIndex(objectIndex);
      return;
    }

    const {pickingColors, instancePickingColors} = this.getAttributeManager().attributes;
    const colors = pickingColors || instancePickingColors;
    const externalColorAttribute = colors && data.attributes && data.attributes[colors.id];
```

modules/core/src/lib/layer.ts
```diff
@@ -745,8 +745,37 @@ export default abstract class Layer<PropsT extends {} = {}> extends Component<
...
+      const values = externalColorAttribute.value;
+      const objectColor = this.encodePickingColor(objectIndex);
+      for (let index = 0; index < data.length; index++) {
+        const i = colors.getVertexOffset(index);
+        if (
+          values[i] === objectColor[0] &&
+          values[i + 1] === objectColor[1] &&
+          values[i + 2] === objectColor[2]
+        ) {
+          this._disablePickingIndex(index);
```
Pessimistress, line 773:
```suggestion
          this._disablePickingIndex(index);
          break;
```

modules/core/src/lib/layer.ts
```diff
@@ -50,7 +50,7 @@ import type {PickingInfo, GetPickingInfoParams} from './picking/pick-info';
 import type Viewport from '../viewports/viewport';
 import type {NumericArray} from '../types/types';
 import type {DefaultProps} from '../lifecycle/prop-types';
-import type {LayerProps} from '../types/layer-props';
+import type {LayerData, LayerProps} from '../types/layer-props';
 import type {LayerContext} from './layer-manager';
```
Pessimistress, line 55:
```suggestion
import type {BinaryAttribute} from './attribute/attribute';

```

modules/core/src/lib/layer.ts
```diff
@@ -745,8 +746,35 @@ export default abstract class Layer<PropsT extends {} = {}> extends Component<
...
+      const objectColor = this.encodePickingColor(objectIndex);
+      for (let index = 0; index < data.length; index++) {
+        const i = colors.getVertexOffset(index);
+        if (
+          values[i] === objectColor[0] &&
+          values[i + 1] === objectColor[1] &&
+          values[i + 2] === objectColor[2]
+        ) {
+          this._disablePickingIndex(index);
+          break;
```
gabeboning, line 772: This break will prevent disabling picking colors correctly for multi polygons or multi line strings -- every match needs to disabled, not just the first one.

Pessimistress: I see, sorry about that. Could you kill the line and make sure this PR fixes the original issue before we merge? I think the only thing missing is unit tests, but I can add that in a follow up PR.

gabeboning: Removed + verified that it works as I'd expect now!

PR #7845, opened 2023-04-19
CARTO: Bump quadbin library to 0.1.8

felixpalmer:
Includes fix from https://github.com/CartoDB/quadbin-js/pull/13

### DIFF
--- a/modules/carto/package.json
+++ b/modules/carto/package.json
@@ -46,7 +46,7 @@
     "h3-js": "^3.7.0",
     "moment-timezone": "^0.5.33",
     "pbf": "^3.2.1",
-    "quadbin": "^0.1.2"
+    "quadbin": "^0.1.8"
   },
   "devDependencies": {
     "@loaders.gl/schema": "^3.3.3",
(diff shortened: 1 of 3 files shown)


modules/carto/package.json
```diff
@@ -46,7 +46,7 @@
     "h3-js": "^3.7.0",
     "moment-timezone": "^0.5.33",
     "pbf": "^3.2.1",
-    "quadbin": "^0.1.2"
+    "quadbin": "^0.1.8"
```
VictorVelarde, line 49: LGTM

modules/carto/package.json
```diff
@@ -46,7 +46,7 @@
     "h3-js": "^3.7.0",
```
VictorVelarde, line 46: Have we considered bumping here to major 4.1.x ?. It looks like it manages correctly the previous error when using h3-js from web workers: https://github.com/uber/h3-js/releases

VictorVelarde (approved):
👍🏻

Issue #7735, opened 2023-03-10
SSR Error After Upgrading from 8.8 to 8.9


aemonm:
### Description

After upgrading from 8.8 to 8.9 via `npm update`, I receive the following error when trying to build my NextJS App:

```ERR_REQUIRE_ESM]: require() of ES Module /Users/aemon/Worky/cerulean-ui/node_modules/@mapbox/tiny-sdf/index.js from /Users/aemon/Worky/cerulean-ui/node_modules/@deck.gl/layers/dist/es5/text-layer/font-atlas-manager.js not supported.
Instead change the require of index.js in /Users/aemon/Worky/cerulean-ui/node_modules/@deck.gl/layers/dist/es5/text-layer/font-atlas-manager.js to a dynamic import() which is available in all CommonJS modules.
    at Object.<anonymous> (/Users/aemon/Worky/cerulean-ui/node_modules/@deck.gl/layers/dist/es5/text-layer/font-atlas-manager.js:17:39)
    at Object.<anonymous> (/Users/aemon/Worky/cerulean-ui/node_modules/@deck.gl/layers/dist/es5/text-layer/text-layer.js:32:49)
    at Object.<anonymous> (/Users/aemon/Worky/cerulean-ui/node_modules/@deck.gl/layers/dist/es5/geojson-layer/sub-layer-map.js:16:41)
    at Object.<anonymous> (/Users/aemon/Worky/cerulean-ui/node_modules/@deck.gl/layers/dist/es5/geojson-layer/geojson-layer.js:32:20)
    at Object.<anonymous> (/Users/aemon/Worky/cerulean-ui/node_modules/@deck.gl/layers/dist/es5/index.js:119:44)
    at 9801 (/Users/aemon/Worky/cerulean-ui/.next/server/pages/index.js:1391:18)
    at __webpack_require__ (/Users/aemon/Worky/cerulean-ui/.next/server/webpack-runtime.js:25:42)
    at 5297 (/Users/aemon/Worky/cerulean-ui/.next/server/pages/index.js:134:79)
    at __webpack_require__ (/Users/aemon/Worky/cerulean-ui/.next/server/webpack-runtime.js:25:42)
    at /Users/aemon/Worky/cerulean-ui/.next/server/pages/index.js:1093:79
    at __webpack_require__.a (/Users/aemon/Worky/cerulean-ui/.next/server/webpack-runtime.js:89:13)
[...]

PR #7847, opened 2023-04-21
Add SSR issue to upgrade guide

Pessimistress:
For #7735

#### Change List
- Add SSR issue to upgrade guide


chrisgervang (approved):

chrisgervang:
A note about SSR disabling in getting started guides would be helpful for a good first experience. I've run into this with libs before and it's frustrating, but the workaround is pretty easy.

PR #7855, opened 2023-04-26
CARTO: support long queres for stats requests via POST

felixpalmer:
#### Background

Followup to https://github.com/visgl/deck.gl/pull/7643

#### Change List
- Support switching to POST with long queries against the stats API
- Test for stats request flow


zbigg (approved):
LGTM, 

Not related:
Note i've just realized`_fetchTilestats` doesn't support `queryParameters` and thus will fail in parametrized queries.

Issue #7852, opened 2023-04-25
[Bug] Tile layer issuing excess tile requests when map gets hidden


ilijapuaca:
### Description

When working with tile layer, hiding the entire map by setting the `display` style to `none` issues lots of new tile requests. 
This issue first manifested itself in the Jupyter notebook use case, but reproducing it in a React environment seems relatively easy too (I've pushed a trivial example on how to reproduce it to [this branch](https://github.com/visgl/deck.gl/tree/repro/tile-requests)).

Based on some light research, the problem seems to be that the viewport changes to quarter of its size when the DOM element gets hidden. Unsure whether this stems from some weird browser behavior where it emits updates with mixed device pixel sizes.

### Flavors

- [ ] Script tag
- [X] React
- [X] Python/Jupyter notebook
- [ ] MapboxOverlay
- [ ] GoogleMapsOverlay
- [ ] CartoLayer
- [ ] ArcGIS

### Expected Behavior

No tile requests get issued

### Steps to Reproduce

Start the `examples/website/map-tile` example on [this branch](https://github.com/visgl/deck.gl/tree/repro/tile-requests), open up the network tab and press the `Toggle Map` button

### Environment

- Framework version: 8.8.0, also on latest `master`
- Browser: Chrome 112.0.5615.137
- OS: macOS 13.3.1

Pessimistress:
The issue with this code is that you are only hiding the canvas, not the element that receives pointer events, so the viewport will continue to update if you drag and scroll. To hide the React component you should do something like

```jsx
<DeckGL
  style={{display: showMap ? 'block' : 'none'}}
  ... />
```

ilijapuaca:
I did not check what the `id` property of `DeckGL` gets assigned to under the hood, I assumed it was the top-level container but as you noted it seems to be the canvas.

I have revisited the example per your suggestion and pushed to the same branch, the same issue is still present. To make sure we're on the same page, there is no dragging or scrolling involved that causes the new tiles to be requested, I'm attaching a video just in case

https://user-images.githubusercontent.com/3406718/234706097-a7287998-d273-476c-9a0d-f5d259925855.mov

Pessimistress:
Fixed in v8.9.12

PR #7856, opened 2023-04-26
Handle hidden canvas

Pessimistress:
For #7852

#### Change List
- Allow Deck class set `width` and `height` to 0
- Ignore viewport of size 0 (will skip `updateState` & `draw` of corresponding layers)
- Tests

### DIFF
--- a/modules/core/src/lib/deck.ts
+++ b/modules/core/src/lib/deck.ts
@@ -734,9 +734,9 @@ export default class Deck {
     if (!canvas) {
       return;
     }
-    // Fallback to width/height when clientWidth/clientHeight are 0 or undefined.
-    const newWidth = canvas.clientWidth || canvas.width;
-    const newHeight = canvas.clientHeight || canvas.height;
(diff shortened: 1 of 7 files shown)


ilijapuaca:
I tried rebasing my repro branch on top of this one and giving it a try, the issue still seems to be there.

Per the original report and based on me doing some light investigation, the canvas size doesn't go down to 0 when the element gets hidden, but the width rather becomes half of the original width and same thing with the height, which seemed like some weird device pixel density interaction

Pessimistress:
I am unable to reproduce your issue with this change. Make sure you run `yarn start-local` to test against the source.

ilijapuaca:
You're right, I initially tried running `start-local` but ran into issues that ended up being due to yarn version mismatch/browser content blocker.

I can confirm that no new tile requests get issued when the map is hidden now. Now I just need to figure out how to give this a try in Jupyter, any pointers as to what the easiest way might be?

ilijapuaca (approved):

PR #7859, opened 2023-04-27
Fix typescript errors mapbox module

alasarr:
With the following version of typescript:
```bash
$ npx tsc -v
Version 4.6.4
```
I'm getting these errors:
```
modules/mapbox/src/mapbox-overlay.ts:279:29 - error TS2352: Conversion of type '{ type: string; deltaX?: number | undefined; deltaY?: number | undefined; offsetCenter: { x: number; y: number; }; srcEvent: MapMouseEvent; tapCount?: number | undefined; }' to type 'MjolnirPointerEvent' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.

279         deck._onPointerMove(mockEvent as MjolnirPointerEvent);
```

I've applied a fix in this PR, but I keep it under draft because I don't understand why it happens to me and not in the CI

### DIFF
--- a/modules/mapbox/src/mapbox-overlay.ts
+++ b/modules/mapbox/src/mapbox-overlay.ts
@@ -203,6 +203,7 @@ export default class MapboxOverlay implements IControl {
     }
   };
 
+  // eslint-disable-next-line complexity
   private _handleMouseEvent = (event: MapMouseEvent) => {
     const deck = this._deck;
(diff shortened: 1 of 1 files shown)


alasarr:
It was happening to me because I was using a higher version of Mjolnir in my local environment. This PR should be required if we upgrade the library of Mjolnir in the future. 

cc: @Pessimistress

Issue #7857, opened 2023-04-27
[Bug] Map does not scale with browser zoom when using `interleaved: true`


kochis:
### Description

We're using DeckGL + MapLibre in interleaved mode, and we noticed an issue when the browser is scaled with cmd+ or cmd- . It looks like the Deck layer scales as expected, but the underlying map does not.  This seems to be an issue only when using interleaved, the map & Deck layer scale as expected when interleaved: false.

Have an example here: https://codepen.io/boon21/pen/abRwpPy

### Flavors

- [ ] Script tag
- [ ] React
- [ ] Python/Jupyter notebook
- [X] MapboxOverlay
- [ ] GoogleMapsOverlay
- [ ] CartoLayer
- [ ] ArcGIS

### Expected Behavior

The map & Deck layer should scale at the same rate.

### Steps to Reproduce

CodePen: https://codepen.io/boon21/pen/abRwpPy

Scale the browser's zoom level `cmd+` or `cmd-` and zoom in/out on the visualization. You will see they map and layer are no longer in sync.

### Environment

- Framework version: latest
- Browser: Chrome 112
- OS: MacOS

kochis:
Also noticed, this works as expected on Safari, so the bug just applies to Chrome and Firefox (have not tested Edge)

Pessimistress:
Fixed in v8.9.13.

PR #7867, opened 2023-05-01
Fix MapboxOverlay scale issue when device pixel ratio changes

Pessimistress:
For #7857

Mapbox, Maplibre and deck.gl each treat device pixel ratio change differently. Let the base map do the scaling when using interleaved mode.

#### Change List
- `Deck` class: disable `autoResizeDrawingBuffer` if using an external gl context (AFAIK this is only used when interleaving with Mapbox/Google Maps/ArcGIS)
- Bump luma dependency for https://github.com/visgl/luma.gl/pull/1746


chrisgervang:
Does this change the way to manually control the resolution in interleaved mode?

Pessimistress:
You shouldn't be able to control the resolution in interleaved mode. I'm pretty sure it will break the base map rendering. What is the use case?

modules/mapbox/src/deck-utils.ts
```diff
@@ -112,11 +112,8 @@ export function removeDeckInstance(map: Map & {__deck?: Deck | null}) {
 }
 
 export function getInterleavedProps(currProps: DeckProps) {
-  const useDevicePixels = 'useDevicePixels' in currProps ? currProps.useDevicePixels : true;
-
   const nextProps: DeckProps = {
     ...currProps,
-    useDevicePixels,
```
chrisgervang, line 119: Resolution can currently be controlled in two parts. One, by setting the width/height css style of the react-map-gl HTML container, and two, multiplying width/height by setting `useDevicePixels` prop on deck and `pixelRatio` prop on react-map-gl for maplibre. 

Was this redundant or does this change modify behavior?

Pessimistress: You should be able to just set `pixelRatio` on the base map after this PR.

chrisgervang (approved):

PR #7877, opened 2023-05-09
Disable transition render tests

Pessimistress:
Fix CI failure on master

As far as I can tell, this test is broken in the Chrome v113 headless mode only. The browser test passes without any issues. The `instancePositions` data at each step read correctly, but the rendering outputs do not match the buffer.

Downgrading puppeteer also makes the issue go away.

#### Change List
- Temporarily disable the attribute transition render tests.


felixpalmer (approved):
The puppeteer bump also replaces Chromium with Chrome For Testing which on my machine runs a lot slower

Issue #7888, opened 2023-05-13
[Bug] GoogleMapsOverlay.setMap(null) doesn't hide or remove (with .finalize) the overlay


chriszrc:
### Description

The codesandbox is here:

https://codesandbox.io/s/typescript-deckgl-googlemaps-v8swpy

There's both a button to click and timeout that calls `.setMap(null)` and `.finalize()` but the overlay is not removed unless the window is resized or the map is physically interacted with or [EDIT] modified in some way; I did find if I called `map.setCenter()` with slightly different coords that was enough to trigger the redraw.

### Flavors

- [ ] Script tag
- [ ] React
- [ ] Python/Jupyter notebook
- [ ] MapboxOverlay
- [X] GoogleMapsOverlay
- [ ] CartoLayer
- [ ] ArcGIS

### Expected Behavior

I expected the overlay to be removed from the map

### Steps to Reproduce

https://codesandbox.io/s/typescript-deckgl-googlemaps-v8swpy

### Environment

- Framework version: 8.9.13
- Browser: Chrome Chrome, Firefox, Safari, etc
- OS: Mac

Pessimistress:
Fixed in v8.9.16.

PR #7902, opened 2023-05-17
GoogleMapsOverlay: requestRedraw when removing interleaved overlay

felixpalmer:
For https://github.com/visgl/deck.gl/issues/7888

#### Change List
- requestRedraw when removing interleaved overlay
- test


Pessimistress:
Should this be reported as a bug to Google Maps? I would expect `setMap(null)` to trigger a redraw on the map side.

Pessimistress (approved):

PR #7911, opened 2023-05-23
Extend tooltip style.transform instead of overwriting it

nvn-nil:
#### Background

The tooltip element is shown below the picked object and this is not optimal when your object is in the bottom right of the map. This is not the best in terms of UX and the tooltip can overflow. Since deck doesn't handle tooltip position based on the dimensions of the tooltip element we need a way to pass in our own `style.transform` to setTooltip.

This PR has a minor improvement that uses user-provided `style.transform` without overwriting it at the end of setTooltip function. A similar change has been made to the overwrite of `style.display` because it's also hardcoded. See discussion in #7024 for more context.

#### Change List
- ENH: Extend el.style.transform in setTooltip
- TST: Add test to show getTooltip using user provided 'style.transform'

### DIFF
--- a/modules/core/src/lib/tooltip.ts
+++ b/modules/core/src/lib/tooltip.ts
@@ -62,6 +62,9 @@ export default class Tooltip {
       return;
     }
 
+    el.style.display = 'block';
+    el.style.transform = `translate(${x}px, ${y}px)`;
+
     if (typeof displayInfo === 'string') {
       el.innerText = displayInfo;
     } else if (!displayInfo) {
@@ -81,8 +84,6 @@ export default class Tooltip {
       Object.assign(el.style, displayInfo.style);
     }
     this.isVisible = true;
-    el.style.display = el.style.display || 'block';
-    el.style.transform = el.style.transform || `translate(${x}px, ${y}px)`;
   }
 
   remove(): void {
--- a/modules/core/src/lib/tooltip.ts
+++ b/modules/core/src/lib/tooltip.ts
@@ -62,9 +62,6 @@ export default class Tooltip {
       return;
     }
 
-    el.style.display = 'block';
-    el.style.transform = `translate(${x}px, ${y}px)`;
-
     if (typeof displayInfo === 'string') {
       el.innerText = displayInfo;
     } else if (!displayInfo) {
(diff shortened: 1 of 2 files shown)


modules/core/src/lib/tooltip.ts
```diff
@@ -81,8 +81,8 @@ export default class Tooltip {
       Object.assign(el.style, displayInfo.style);
     }
     this.isVisible = true;
-    el.style.display = 'block';
-    el.style.transform = `translate(${x}px, ${y}px)`;
+    el.style.display = el.style.display || 'block';
+    el.style.transform = el.style.transform || `translate(${x}px, ${y}px)`;
```
Pessimistress, line 85: I don't think this will work. `el.style.transform` will always be defined after the first time it is rendered.

You need to check `displayInfo.style` instead.

nvn-nil: I see. So, the tooltip element doesn't get removed after use. This will indeed use transform and display style from the last time it was set until it is changed by the user explicitly. But this is the behavior with all the style properties.

I made a codepen based on the iconlayer example to show this. https://codepen.io/nvn-nil/pen/eYPxMXP

We'll need to do `el.style = {}` or something in the `!displayInfo` block to prevent this from happening.

I'll use displayInfo for this check here. I can open another issue for the persisting styles or I can fix it here if you think that is not the intended behavior.

nvn-nil: I moved 
```js
    el.style.display = 'block';
    el.style.transform = `translate(${x}px, ${y}px)`;
```
before all the displayInfo blocks since it being at the end is what's overwriting the user set styles.

modules/core/src/lib/tooltip.ts
```diff
@@ -62,6 +62,9 @@ export default class Tooltip {
       return;
     }
 
+    el.style.display = 'block';
+    el.style.transform = `translate(${x}px, ${y}px)`;
```
Pessimistress, line 66: There's no point setting `transform` if the tooltip is going to be hidden.
I would suggest keeping this block in its original place, instead move `Object.assign(el.style, displayInfo.style)` to after these lines.

nvn-nil: I can do that but then once a tooltip is provided style `display: none` by the user it'll persist and no tooltips will be shown afterwards (even when no display style is provided).

I'll update my codepen (in the previous comment) to show this.

nvn-nil: Done

PR #7913, opened 2023-05-24
update h3-js to v4

charlierudolph:
#### Background

As a workaround for my own build chain issue, I would like to upgrade deck.gl's h3-js version. Given its a major version upgrade that mainly improves some functions names seemed like a good thing to do regardless.

Running tests on my machine appeared to start hanging so figured would just let CI handle it based on the limited set of changes. Let me know if there are other specific verification asks you'd like done

#### Change List
- Update h3-js from 3.7 to 4.1
  - Used their migration guide for name changes: https://h3geo.org/docs/library/migrating-3.x (one instance needed to have try / catch around it as an error was now thrown)
  - Converted some `import *` to specific imports where possible


Pessimistress (approved):

Pessimistress:
@charlierudolph Thank you for tackling this!

Given that this is a breaking change (especially for script tag users) we can't release it in a patch. We could publish a new beta release from the master branch (say `8.10.0-alpha.1`) - will that work for you?

charlierudolph:
Yeah that would great. Thanks!

charlierudolph:
Thanks for merging! Do you still plan to do an alpha release with this change or can you give an estimate when a release would include this change?

Pessimistress:
Just published `8.10.0-alpha.1`.

charlierudolph:
Thanks! Alpha released solved my issue (though did appear to also have other breaking changes that needed updates - just some minimal type changes)

Issue #7918, opened 2023-05-26
[Feat] Make canvas instance readonly instead of protected


akre54:
### Target Use Case

Right now there isn't a way to access the canvas instance [as seems to be suggested](https://github.com/visgl/deck.gl/issues/4436#issuecomment-610472868), as Typescript forbids it. It's possible to cast the deck instance as `any` before getting the canvas prop but this way provides a formalized accessor.

### Proposal

Change the type on `canvas` from `protected` to `readonly`.

felixpalmer:
Fixed in 8.9.31

Issue #4436, opened 2020-03-30
Way of saving canvas to image in deck.gl


liokm:
Hi, I want to save the webgl canvas to image by using `toDataURL` or `gl.readPixels` and have some questions:

- In order to read from canvas, seems I need to use `deck.redraw()` to trigger the render and read from framebuffer. ~~Since I am using `<DeckGL>` in React and, according to [the code](https://github.com/uber/deck.gl/blob/master/modules/react/src/deckgl.js#L64) the `this.deck` is not exposed from `<DeckGL>`, how could I achieve this without passing customized deck instance?~~ Use `ref.current.deck.redraw(true)`
- What is your recommendation of the way saving canvas to image, in deck.gl?
Thanks!

kylebarron:
You probably want to use a `ref`, e.g.:
```jsx
        <DeckGL
          ref={ref => {
            this.deck = ref;
          }}
```
Then you can access deck.gl in the class through `this.deck`

liokm:
@kylebarron Thank you, that makes sense. ref works here.

baiqiushi:
Hi, I'm glad I saw this issue. I'm facing the same issue here, wanna save deck.gl scatterLayer result to an image. But I am using plain JS, can you elaborate more on how to do this? Thank you!

My code looks like this:
```html
<div id="deck-gl-canvas"></div>
```

```javascript
$scope.scatterLayer = new deck.DeckGL({
              container: 'deck-gl-canvas',
              initialViewState: {
                  latitude: $scope.map.getCenter().lat,
                  longitude: $scope.map.getCenter().lng,
                  zoom: $scope.map.getZoom() - 1,
                  bearing: 0,
                  pitch: 0
              },
              controller: true,
              layers: []
       });

const deckGLScatterplot = new deck.ScatterplotLayer({
              id: 'deck-gl-scatter',
              data: $scope.points,
              radiusMinPixels: Math.round($scope.pointRadius),
              getPosition: d => [d[1], d[0]],     // returns longitude, latitude, [altitude]
              getRadius: d => $scope.pointRadius,  // returns radius in meters
              getFillColor: d => [0, 0, 255]           // returns R, G, B, [A] in 0-255 range
    });

$scope.scatterLayer.setProps(
          {
                layers: [deckGLScatterplot],
                viewState:
                  {
                    latitude: $scope.map.getCenter().lat,
                    longitude: $scope.map.getCenter().lng,
[...]

PR #7919, opened 2023-05-26
Added a getCanvas helper function on deck

akre54:
For #7918

#### Background
Change typescript annotation to make the Deck instance canvas readonly instead of protected. Right now there isn't a way to access the canvas instance [as seems to be suggested](https://github.com/visgl/deck.gl/issues/4436#issuecomment-610472868), as Typescript forbids it. It's possible to cast the deck instance as `any` before getting the canvas prop but this way provides a formalized accessor.

#### Change List
- Added a `getCanvas` helper function on deck


modules/core/src/lib/deck.ts
```diff
@@ -286,8 +286,8 @@ export default class Deck {
   readonly height: number = 0;
   // Allows attaching arbitrary data to the instance
   readonly userData: Record<string, any> = {};
-
-  protected canvas: HTMLCanvasElement | null = null;
+  readonly canvas: HTMLCanvasElement | null = null;
```
felixpalmer, line 289: This isn't going to work, as per the errors reported. @Pessimistress, what do you think about adding a `get canvas()` getter method?

felixpalmer (changes requested):

modules/core/src/lib/deck.ts
akre54: Good point - I reverted and added a helper similar to mapbox-gl's / maplibre's - also [exposed via react-map-gl](https://github.com/visgl/react-map-gl/blob/2bf7ca08a1782093f070b8f909d459efa7760777/src/types/lib.ts#L81).

Pessimistress:
Add the new method to the documentation page?

Pessimistress (approved):

felixpalmer (approved):

paulleonartcalvo:
Would love to see this merged! :)

PR #7931, opened 2023-06-06
Add tensorflow example

Pessimistress:
#### Background

Adding an experimental example that demonstrates how WebGL resources can be shared between Tensorflow.js and deck.gl.

There are two demos:

- `src/demo-histogram.ts`: Tensorflow performs bin counting on 1M data points, and the output is copied to a Buffer that is passed to a GridCellLayer.
- `src/demo-spectrogram.ts`: Tensorflow performs [short-time Fourier transform](https://en.wikipedia.org/wiki/Short-time_Fourier_transform) with real-time audio data, and the output is displayed as an image by a custom BitmapLayer.

#### Todo

- In the long term, I am interested in creating an official tfjs connector module. This would free us from implementing our own GPGPU API, given that Tensorflow is well-tested, widely adopted, and well maintained.
- Tensorflow also has a WebGPU backend that would be interesting to test with luma.gl v9.

#### Change List
- Add `experimental/tfjs` example


felixpalmer:
Very cool demos. Regarding the long term plan, is there value in not just providing a connector but also leveraging the GPGPU capabilities provided by Tensorflow.js to implement features within deck? For example our GPU-aggregated layers or data filtering?

ibgreen (approved):
Cool examples! 

We could look into offering tf as an alternative to the Transform class in luma. 

The glue code would probably also work best as a luma module.

Issue #7938, opened 2023-06-07
[Bug] Seams visible in Tile3DLayer


felixpalmer:
### Description

Artifacts are visible between 3D tiles when using the `Tile3DLayer`, for example on https://deck.gl/examples/google-3d-tiles.

### Flavors

- [ ] Script tag
- [ ] React
- [ ] Python/Jupyter notebook
- [ ] MapboxOverlay
- [ ] GoogleMapsOverlay
- [ ] CartoLayer
- [ ] ArcGIS

### Expected Behavior

No seams. Compare the output from https://sandcastle.cesium.com/?src=Google%20Photorealistic%203D%20Tiles.html

### Steps to Reproduce

Run the `examples/website/google-3d-tiles` example, and change the background to red to make it more obvious.

<img width="985" alt="Screenshot 2023-06-07 at 15 59 16" src="https://github.com/visgl/deck.gl/assets/453755/990f6e19-a1db-453c-aaca-51f497cbc5e3">

### Environment

- Framework version: 8.9.18
- Browser: Chrome/Firefox
- OS: MacOS 12.0.1

PR #7939, opened 2023-06-07
Google 3D tiles: Enlarge tiles to close seams

felixpalmer:
For https://github.com/visgl/deck.gl/issues/7938

Simple fix (hack) in the example to hide the seams. Not perfect as it causes misalignment at tile boundaries, but these are at least less obvious than the seams.

A better fix long-term fix would be to fix the errors in the projection

Before fix

Screenshot 2023-06-07 at 15 59 16

After fix

Screenshot 2023-06-07 at 15 59 42

#### Change List
- Increase size of 3D tiles by 1% to hide seams

### DIFF
--- a/examples/website/google-3d-tiles/app.jsx
+++ b/examples/website/google-3d-tiles/app.jsx
@@ -2,8 +2,10 @@ import React, {useState} from 'react';
 import {scaleLinear} from 'd3-scale';
 import {createRoot} from 'react-dom/client';
 import DeckGL from '@deck.gl/react';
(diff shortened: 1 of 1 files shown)


chrisgervang:
Is there a way to do this with a prop override if someone wants to use the technique?

Any indication of how other libraries solve this?

felixpalmer:
@chrisgervang the fix is to correct our precision issue, I believe it should be possible, it's just slow-going finding the bug

PR #7949, opened 2023-06-16
CARTO: merge loadOptions passed in prop

felixpalmer:
#### Background

`loadOptions` passed to `CartoLayer` were getting [overwritten](https://github.com/visgl/deck.gl/blob/master/modules/carto/src/layers/carto-layer.ts#L329)

#### Change List
- Merge `loadOptions` with access token
- Test


zbigg (approved):
lgtm

Issue #7956, opened 2023-06-19
[Bug] Controller is constantly re-constructed when using a custom MapController with MapView


aslatemts:
### Description

I have implemented a custom MapController similar to the example in the docs (https://deck.gl/docs/api-reference/core/map-controller#custom-mapcontroller) but with a custom constructor, and using it by passing it to the 'controller' option on the Deck class.

However I am experiencing issues with the controller (unable to zoom, pan, etc.) , as it is constantly being re-constructed. I have found the root cause of the issue, which was introduced in the following change: https://github.com/visgl/deck.gl/pull/7814/files

The MapView returns [MapController as the controller type by default](https://github.com/visgl/deck.gl/blob/8.9-release/modules/core/src/views/map-view.ts#L54), and does not honour any change to the controller setting, and as such the following comparison will always be true when using a controller that is not the default MapController: `controller.constructor !== view.ControllerType` at https://github.com/visgl/deck.gl/pull/7814/files#diff-bb43f5d7d49b58a3361883cb5eebe6d322009e01817ac5877449fa25f417e7ccR331.

I have worked around this issue by implementing a custom MapView that overrides ControllerType like so: 
```
export class MyMapView extends MapView {
  get ControllerType() {
    return MyMapController;
  }
}
```
but it would be nice for the view to honour the controller type that has been configured by default. 

### Flavors

- [ ] Script tag
- [ ] React
- [ ] Python/Jupyter notebook
- [ ] MapboxOverlay
- [ ] GoogleMapsOverlay
- [ ] CartoLayer
- [ ] ArcGIS

### Expected Behavior

The custom MapController (MyMapController) is only constructed once.

### Steps to Reproduce

Set up Deck with a custom MapController as shown in the docs here: https://deck.gl/docs/api-reference/core/map-controller#custom-mapcontroller

Will probably need to override the constructor as well to experience the same effect like so:
```
constructor(opts: any) {
    super(opts);
    // custom logic here
}
```

### Environment

- Framework version: [v8.9.17](https://github.com/visgl/deck.gl/releases/tag/v8.9.17) and higher
- Browser: Chrome
- OS: Windows

Pessimistress:
Fixed in v8.9.20.

PR #7962, opened 2023-06-22
Use controller.type rather than ControllerType

felixpalmer:
For https://github.com/visgl/deck.gl/issues/7956

#### Change List
- Use controller.type rather than ControllerType

### DIFF
--- a/modules/core/src/lib/view-manager.ts
+++ b/modules/core/src/lib/view-manager.ts
@@ -328,7 +328,7 @@ export default class ViewManager {
 
       // Create controller if not already existing or if the type of the
       // controller has changed.
-      if (!controller || controller.constructor !== view.ControllerType) {
+      if (!controller || controller.constructor !== view.controller.type) {
         controller = this._createController(view, resolvedProps);
       }
       if (controller) {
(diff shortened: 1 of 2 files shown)


modules/core/src/views/view.ts
```diff
@@ -53,7 +53,7 @@ export default abstract class View<
 > {
   id: string;
   abstract get ViewportType(): ConstructorOf<Viewport>;
-  abstract get ControllerType(): ConstructorOf<Controller<any>>;
+  protected abstract get ControllerType(): ConstructorOf<Controller<any>>;
```
felixpalmer, line 56: Added to prevent similar bugs in the future. As far as I can tell `ControllerType` is only used as a default value for `this.controller.type`

Pessimistress (approved):

modules/core/src/lib/view-manager.ts
```diff
@@ -328,7 +328,7 @@ export default class ViewManager {
 
       // Create controller if not already existing or if the type of the
       // controller has changed.
-      if (!controller || controller.constructor !== view.ControllerType) {
+      if (!controller || controller.constructor !== view.controller.type) {
```
Pessimistress, line 331: Use `controllerProps` to avoid executing `view.controller` getter again.

PR #7965, opened 2023-06-25
Refactor: Rename selectedCounty setter function to adhere to common c…

Sen-442b:
…onvention

It is a [common convention](https://legacy.reactjs.org/docs/hooks-state.html#declaring-a-state-variable) to initialize the setter function variable name returned from  useState with 'set'

For #

#### Change List
-


Pessimistress:
This is too trivial of a change to warrant a PR.

Issue #7952, opened 2023-06-16
[Bug] LayerExtension has no effect on Tile3DLayer 


RELNO:
### Description

When using LayerExtension with simple frag shader, most layers are impacted, besides Tile3DLayer.
Example below shows `Tile3DLayer`  and `TileLayer`. `LayerExtension` only impacts `TileLayer` (https://codesandbox.io/p/sandbox/small-water-fkrzcm)
**Note: this applies to all Tile3DLayer sources, regardless if its google 3d tiles, archgis, etc.** 

<img width="1218" alt="Screenshot 2023-06-16 at 1 42 04 PM" src="https://github.com/visgl/deck.gl/assets/14296774/504ca0cd-8c88-4ebe-ad06-f6b9d1851427">
<img width="1221" alt="Screenshot 2023-06-16 at 1 42 10 PM" src="https://github.com/visgl/deck.gl/assets/14296774/c2f6de08-db21-4893-9043-4bb51b60bbc2">

Where Extension is:

```
import { LayerExtension } from "@deck.gl/core";

export default class ColorExtension extends LayerExtension {

  getShaders() {
    return {
      inject: {
        "fs:DECKGL_FILTER_COLOR": `
        color = vec4(1,0,0, 1);
            `,
      },
    };
  }
}

```

### Flavors

- [ ] Script tag
- [X] React
- [ ] Python/Jupyter notebook
- [X] MapboxOverlay
- [X] GoogleMapsOverlay
- [ ] CartoLayer
- [X] ArcGIS

### Expected Behavior

LayerExtension frag shader should apply to all layers 

### Steps to Reproduce

add LayerExtension with shader to Tile3DLayer. See here https://codesandbox.io/p/sandbox/small-water-fkrzcm

### Environment

- Framework version: latest, also tested on prev. release.
- Browser: any
- OS: any

RELNO:
@Pessimistress since tile3d is composite of ScenegraphLayer (in this case), I wonder if https://github.com/visgl/deck.gl/pull/5214 might be related.

RELNO:
Further investigation shows that `getShaders` has no impact on extended `Tile3DLayer`:
```
import { Tile3DLayer } from "@deck.gl/geo-layers";

export default class ExtendedTile3DLayer extends Tile3DLayer {
  initializeState(state) {
    super.initializeState();
    console.log("This will be logged");
  }

  getShaders() {
    console.log("This will not be logged")
[...]

PR #7987, opened 2023-07-05
Fix ScenegraphLayer shader injection from extensions

Pessimistress:
For #7952

#### Change List
- `ScenegraphLayer.getShaders` should call `Layer.getShaders` to handle injection properly


chrisgervang (approved):

felixpalmer (approved):

ibgreen (approved):
Fix is fine as it follows the existing pattern, but just a nit / food for thought: Requiring a call to `super` is not a great pattern (brittle in a couple of ways), and looks like we are forcing the need to do so in all layers by putting common (extension) processing in the base class. I feel it would be nicer if the extension related shader processing was moved out of `Layer.getShaders()` and applied after the call to getShaders. As one idea, maybe the external api could be `getLayerShaders()` and `getShaders()` could be declared protected (or vice versa if we don't want to change the external api).

Issue #7875, opened 2023-05-09
[Bug] PathStyleExtension offset affect dashline


vrman:
### Description

i have 3 custom dashed path layer with different offsets , i want those paths with offset have same dashed style as the layer 0 with no offset (center path), but its impossible to make this happen, please check this example :
[https://codepen.io/vrman/pen/mdzxBOx](https://codepen.io/vrman/pen/mdzxBOx)

### Flavors

- [x] Script tag
- [ ] React
- [ ] Python/Jupyter notebook
- [ ] MapboxOverlay
- [ ] GoogleMapsOverlay
- [ ] CartoLayer
- [ ] ArcGIS

### Steps to Reproduce

[https://codepen.io/vrman/pen/mdzxBOx](https://codepen.io/vrman/pen/mdzxBOx)

### Environment

- Framework version: 8.9.12
- Browser: chrome
- OS: windows

mrsiahpoosh:
@Pessimistress Can you please check this bug? 
thanks

vrman:
@Pessimistress Can you please check this bug?
thanks

vrman:
any comment on this issue ?

Pessimistress:
Fixed in v8.9.21.

PR #7990, opened 2023-07-09
PathStyleExtension: fix dash + offset usage

Pessimistress:
For #7875

#### Change List
- Improve uv calculation when using offsets


ibgreen (approved):
Just a rubberstamp, not familiar enough with the shader.

Issue #7936, opened 2023-06-06
[Bug] Missing @babel/runtime dependency


nemanja-tosic:
### Description

While working on a project that uses Yarn PnP i've started getting errors during the build saying `Module not found: Can't resolve '@babel/runtime/helpers/interopRequireDefault'`. Looking into the distributed package, it appears that the babel runtime plugin is used, however the @babel/runtime dependency is missing. Currently, it appears that another package is providing @babel/runtime (added below) but this is potentially problematic because the project is using a dependency it does not declare. For reference, adding the output of `yarn why @babel/runtime`

```
yarn why v1.22.19
[1/4] 🤔  Why do we have the module "@babel/runtime"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "@babel/runtime@7.22.3"
info Reasons this module exists
   - "@probe.gl#env" depends on it
   - Hoisted from "@probe.gl#env#@babel#runtime"
   - Hoisted from "@probe.gl#log#@babel#runtime"
   - Hoisted from "@probe.gl#stats#@babel#runtime"
   - Hoisted from "deck-graph-layers#@luma.gl#core#@babel#runtime"
   - Hoisted from "@deck.gl#core#@luma.gl#webgl#@babel#runtime"
   - Hoisted from "deck-graph-layers#probe.gl#@babel#runtime"
   - Hoisted from "@deck.gl#core#@math.gl#core#@babel#runtime"
   - Hoisted from "@deck.gl#core#@math.gl#sun#@babel#runtime"
   - Hoisted from "@deck.gl#core#@math.gl#web-mercator#@babel#runtime"
   - Hoisted from "@deck.gl#core#@loaders.gl#core#@babel#runtime"
   - Hoisted from "deck-graph-layers#@luma.gl#core#@luma.gl#engine#@babel#runtime"
   - Hoisted from "deck-graph-layers#@luma.gl#core#@luma.gl#gltools#@babel#runtime"
   - Hoisted from "deck-graph-layers#@luma.gl#core#@luma.gl#shadertools#@babel#runtime"
   - Hoisted from "@deck.gl#core#@loaders.gl#core#@probe.gl#log#@babel#runtime"
   - Hoisted from "@deck.gl#core#@loaders.gl#core#@loaders.gl#loader-utils#@babel#runtime"
[...]

PR #7991, opened 2023-07-09
Add @babel/runtime to dependencies

Pessimistress:
For #7936 

`@babel/runtime` is introduced to both the ES5 and ESM entry points by our current babel setting (`@babel/transform-runtime`). It should be listed as a dependency.

We may be able to eliminate the dependency in v9 with something similar to https://github.com/uber-web/probe.gl/pull/237

#### Change List
- Add @babel/runtime to dependencies


ibgreen (approved):
Interesting, I thought we already had these in place, at least I think we do in other frameworks...

Issue #7904, opened 2023-05-19
[Bug] Quadkey layer has visible gaps even when stroke: false


jtmiclat:
### Description

Hi thanks for everyone's work on this awesome library!

We recently tried using the quadkey layer but noticed there were visible lines in the quadkey layers even when stroke:false. We we able to determine that there were gaps between the quadkey polygons. Specifically because of this line

https://github.com/visgl/deck.gl/blob/06f5bd0c3545b40460e3139efab2ca5a70e12cc1/modules/geo-layers/src/quadkey-layer/quadkey-utils.ts#L17-L21

when we replace this line 

```
 [(x + 0.99) / scale, TILE_SIZE - (y + 0.99) / scale]
 ```
  to 
  ```
  [(x + 1) / scale, TILE_SIZE - (y + 1) / scale]
  ```
  
  removes the gap,

### Flavors

- [X] Script tag
- [X] React

### Expected Behavior

When stroke is false or when getLineColor has an alpha quadkeys should be continuous. 

### Steps to Reproduce

note that stroke=false

```javascript
import { Deck } from "@deck.gl/core";
import { QuadkeyLayer } from "@deck.gl/geo-layers";


new Deck({
  mapStyle: "https://basemaps.cartocdn.com/gl/positron-gl-style/style.json",
  initialViewState: {
    longitude: -122.4,
    latitude: 37.74,
    zoom: 11,
    maxZoom: 20,
  },
  controller: true,
  layers: [
    new QuadkeyLayer({
      id: "QuadkeyLayer",
      data: "https://raw.githubusercontent.com/visgl/deck.gl-data/master/website/sf.quadkeys.json",

      /* props from QuadkeyLayer class */

      getQuadkey: (d) => d.quadkey,
      stroke: false,
      getFillColor: [0, 0, 0],
      getLineColor: [255, 0, 0, 0], // invisible red line just to make sure it isnt rendering the line
    }),
  ],
});


```
produces visible gaps 
<img width="1284" alt="Screenshot 2023-05-19 at 7 06 02 PM" src="https://github.com/visgl/deck.gl/assets/30991698/5f5e5f00-3ad8-43df-9935-dbcd4d9e1ebe">

### Environment

- Framework version: 8.9.15 
- Browser: Firefox/Chrome
- OS: MacOS M2

Pessimistress:
Fixed in v8.9.22.

PR #7995, opened 2023-07-11
Remove hairline present on non-stroked QuadkeyLayer

felixpalmer:
For https://github.com/visgl/deck.gl/issues/7904

#### Change List
- Removes 0.99 scaling factor for non-extruded QuadkeyLayer
- Render & test update


modules/geo-layers/src/quadkey-layer/quadkey-layer.ts
```diff
@@ -28,14 +28,16 @@ export default class QuadkeyLayer<DataT = any, ExtraProps extends {} = {}> exten
...
+    const {data, extruded, getQuadkey} = this.props;
+    // To avoid z-fighting reduce polygon footprint when extruding
+    const coverage = extruded ? 0.99 : 1;
 
     return {
       data,
       _normalize: false,
       positionFormat: 'XY',
 
-      getPolygon: (x: DataT, objectInfo) => getQuadkeyPolygon(getQuadkey(x, objectInfo))
```
Pessimistress, line 38: Does this work when you switch between extruded and non-extruded? Needs an update trigger?

Pessimistress:
> // To avoid z-fighting reduce polygon footprint when extruding

Just want to point out that we don't do this in any other geo cell layers.

Pessimistress (approved):

Issue #8013, opened 2023-07-19
[Bug] CesiumIon tilesets from "asset depot" are failing to load


belom88:
### Description

CesiumIon has changed tileset urls for assets from [Asset Depot](https://ion.cesium.com/assetdepot) so they don't load. Private tilesets work as expected.

### Flavors

- [ ] Script tag
- [X] React
- [ ] Python/Jupyter notebook
- [ ] MapboxOverlay
- [ ] GoogleMapsOverlay
- [ ] CartoLayer
- [ ] ArcGIS

### Steps to Reproduce

* Open https://deck.gl/examples/tile-3d-layer
* The tileset is not loaded

### Environment

- Framework version: deck.gl@latest
- Browser: Chrome
- OS: Ubuntu

Pessimistress:
Fixed in v8.9.22.

PR #8014, opened 2023-07-19
fix(geo-layers): Cesium tilesets loading

belom88:
For #8013 
#### Background
The CesiumIONLoader has preload function that is called. This function can return the actual `tileset.json` url.
#### Change List
- take `tileset.json` url from `preload` response


ibgreen (commented):
While you are patching code that is already here, I do feel this logic should happen in loaders.gl - it is rather too complicated to ask an application to have to understand and mess with these details.

Pessimistress:
There is a larger pending task to support `preload` on loaders.gl side - right now this layer is hardcoded to run `props.loaders[0].preload`.

Pessimistress (approved):

PR #8016, opened 2023-07-22
add(modules) widget module

chrisgervang:
For #7946 

#### Background

With the exception of tooltip, the plan is to distribute widgets in their own module. 
#### Change List
- adding widgets module
- documentation

@Pessimistress, should I target master or `9.0-dev`?

### DIFF
--- a/docs/api-reference/widgets/overview.md
+++ b/docs/api-reference/widgets/overview.md
@@ -1,0 +1,35 @@
+# @deck.gl/widgets
+
+Widgets are UI components around the WebGL canvas to offer controls and information for a better user experience.
+
+This module contains the following extensions:
+
+- [FullscreenWidget](./fullscreen-widget.md)
+
+## Installation
+
+### Install from NPM
+
+```bash
+npm install deck.gl
+# or
+npm install @deck.gl/core @deck.gl/widgets
+```
+
+```js
+import {FullscreenWidget} from '@deck.gl/widgets';
+new FullscreenWidget({});
+```
+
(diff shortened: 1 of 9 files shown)


ibgreen (approved):
Given the ongoing WebGL work - If we are working on an 8.10-alpha track (which makes perfect sense especially for modules like this that don't touch and GPU rendering) I recommend we do so on an `8.10-dev` branch. 

`master` needs to move on to 9.0 soon.

docs/api-reference/widgets/overview.md
```diff
@@ -0,0 +1,35 @@
...
+new FullscreenWidget({});
+```
+
+### Include the Standalone Bundle
+
+```html
+<script src="https://unpkg.com/deck.gl@^8.10.0/dist.min.js"></script>
+<!-- or -->
+<script src="https://unpkg.com/@deck.gl/core@^8.10.0/dist.min.js"></script>
+<script src="https://unpkg.com/@deck.gl/widgets@^8.10.0/dist.min.js"></script>
```
Pessimistress, line 30: The next version will be 9.0

Pessimistress (approved):

chrisgervang:
I'm going to merge this as-is since adding FullscreenWidget will be more discussion and this way other modules can get started in parallel.

docs/api-reference/widgets/overview.md
chrisgervang: Opps, missed this. I'll update to 9

PR #8024, opened 2023-07-25
add(widgets) fullscreen widget

chrisgervang:
For #7946 

#### Background
A fullscreen widget adds a button which users can click to make their map full screen. Click it again to exit full screen.

### Design Goals

`themes.ts` provides 4 built-in themes, and the getting started example shows how these can be applied.

```js
import {FullscreenWidget, DarkTheme} from '@deck.gl/widgets';

new Deck({
  ...
  widgets: [new FullscreenWidget({ style: DarkTheme })
})
```
 
With CSS variables it's possible for applications to make customizations, such as:
- Light and dark mode (i.e. [`prefers-color-scheme`](https://css-tricks.com/a-complete-guide-to-dark-mode-on-the-web/) media-query, or prop override)
- Desktop/mobile sizes (i.e. [`max-width`](https://css-tricks.com/a-complete-guide-to-css-media-queries/) media-query, or prop override)
- Total visual customization, such as glass or flat
- Branding, such as coloring, shadows, corners, etc

Examples

### Implementation Goals
- Provide css stylesheet via NPM for applications to include
- Themable with [CSS variables](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties) and class selectors.
- VanillaJS interface with `preact` UI framework internally to minimize complexity and bundle size.
- For icons, prefer inline SVG over heavy icon fonts

### Compatibility
Currently the widget assumes the Fullscreen API is supported without prefixes since it's a mature API at this point. If it's not, then a "pseudo" fullscreen is styled with css.

#### Change List
- Adds FullscreenWidget
- Adds widgets getting started example
- docs


chrisgervang:
Please leave design comments in our [prototype](https://www.figma.com/proto/KGQqLbZZGUINgyyKYPvDvu/Widgets?type=design&node-id=86-2970&t=0VyO2IPaGYfpQGMj-1&scaling=min-zoom&page-id=86%3A2969). Open to suggestions!

chrisgervang:
Learning a bit more about CSS variables. They aren't supported inside media queries directly, but a combination of JS and CSS can be used. See [discussion](http://stackoverflow.com/questions/56300132/ddg#56550819).

It'd be ideal to keep variable names shorter, such as `--background` vs. `--deckgl-widget-background`. Variables can be scoped with selectors such as `.deckgl-widget` (or similar) instead of `:root`. It can be selector an application likely doesn't already use.

Supporting browsers that lack css variable support is straightforward. Just define each style twice with and without a variable, and the cascading will apply the latter "valid" property.

modules/widgets/src/fullscreen-widget.ts
```diff
@@ -0,0 +1,55 @@
...
+  handleClick() {
+    if (!document.fullscreenElement) {
+      this.requestFullscreen();
+    } else {
+      this.exitFullscreen();
+    }
+  }
+
+  async requestFullscreen() {
+    // @ts-ignore canvas is protected. Merge https://github.com/visgl/deck.gl/pull/7919?
```
ibgreen, line 44: To get notified when the problem is solved use expect-error?
```suggestion
    // @ts-expect-error canvas is protected. Merge https://github.com/visgl/deck.gl/pull/7919?
```

examples/get-started/pure-js/basic/app.js
```diff
@@ -57,5 +58,6 @@ new Deck({
       getTargetColor: [200, 0, 80],
       getWidth: 1
     })
-  ]
+  ],
+  _widgets: [new FullscreenWidget()]
```
ibgreen, line 62: Assume you will create a new example for effects.

modules/widgets/src/fullscreen-widget.ts
```diff
@@ -0,0 +1,55 @@
+import type {Deck, Widget} from '@deck.gl/core';
+
+/* global document */
+const defaultStyle: Partial<CSSStyleDeclaration> = {
```
ibgreen, line 4: Any value in making this a static on the class? 
```suggestion
const defaultStyle: Readonly<Partial<CSSStyleDeclaration>> = {
```

examples/get-started/pure-js/basic/app.js
chrisgervang: @Pessimistress suggested I add to the getting started, but I'm not sure if this is what she meant.

Pessimistress: You should add a separate example unless you think it's a good functionality to surface on all the examples. Each get-started flavor (pure-js, react, script) demonstrate exactly the same functionality.

modules/widgets/src/fullscreen-widget.ts
```diff
@@ -0,0 +1,55 @@
...
+    if (!document.fullscreenElement) {
+      this.requestFullscreen();
+    } else {
+      this.exitFullscreen();
+    }
+  }
+
+  async requestFullscreen() {
+    // @ts-expect-error canvas is protected. Merge https://github.com/visgl/deck.gl/pull/7919?
+    await this.deck?.canvas?.parentElement?.requestFullscreen({navigationUI: 'hide'});
```
chrisgervang, line 45: Allow user to specify their own element to fullscreen.

ibgreen: Allow user to override value of navigationUI option?

ibgreen (approved):
Some thoughts: 
- Do we have provisions for declaratively setting props on this component (let's say that the component offered a `hideNavigationUI` prop and the app wanted to change that after creating the component.
- What if the app programmatically wants to active the effect that an interactive widget controls. Let's say that the app wanted to trigger full screen, how would it do that?


- Docs - In documentation mention that this feature is not available on iPhones?
https://caniuse.com/?search=requestFullscreen
- Tests - testing the fullscreen effect may be overkill, but at least add the control to the DOM and and remove it when testing under browser?

chrisgervang:
Thanks for the comments, @ibgreen. I'll take these into account as iteration continues in this PR. In the meantime, let us know if you have comments on #8023 since that's about to merge in.

chrisgervang:
Evaluated [RE:DOM](https://redom.js.org/#jsx) as a lightweight html-in-js solution that is supposed to offer a JSX syntax. JSX works fine without typescript in this isolated [example](https://codesandbox.io/s/redom-jsx-example-vn2jqw) I made, but I'm not figuring out how to get redom JSX to work with typescript in this other [example](https://codesandbox.io/s/redom-tsx-example-rkh37w?file=/src/index.tsx). 

RE:DOM and typescript don't seem to be pretty commonly documented together. And babel seems to have moved on to a new way to building jsx plugins, which RE:DOM doesn't implement.

Resouces
- https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html
- https://babeljs.io/docs/babel-plugin-transform-react-jsx
- https://www.typescriptlang.org/tsconfig#jsxImportSource

----
Setup notes:
`yarn add babel-plugin-transform-redom-jsx babel-plugin-transform-react-jsx --dev -W`
```js
// babel.config.js
...
plugins: [
  "babel-plugin-transform-redom-jsx",
  [
    "transform-react-jsx",
    {
      "pragma": "el"
    }
  ]
]
```
```js
// tsconfig.json
"compilerOptions": {
  "jsx": "preserve",
  "jsxFactory": "el"
  ...
},
```

modules/widgets/src/fullscreen-widget.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  async requestFullscreen() {
+    // @ts-expect-error canvas is protected. Merge https://github.com/visgl/deck.gl/pull/7919?
+    await this.deck?.canvas?.parentElement?.requestFullscreen({navigationUI: 'hide'});
+    this.isFullscreen = true;
+  }
+
+  async exitFullscreen() {
+    await document.exitFullscreen();
+    this.isFullscreen = false;
```
ibgreen, line 77: IF the user presses escape or uses some other operation to exit full screen does this get out of sync?

modules/widgets/src/fullscreen-widget.tsx
```diff
@@ -0,0 +1,81 @@
...
+  }
+}
+
+export default FullscreenWidget;
```
ibgreen, line 81: I recommend named exports rather than default exports. Lets the compiler track the class name all the way.

modules/widgets/package.json
```diff
@@ -30,7 +30,8 @@
     "prepublishOnly": "npm run build-bundle && npm run build-bundle -- --env=dev"
   },
   "dependencies": {
-    "@babel/runtime": "^7.0.0"
+    "@babel/runtime": "^7.0.0",
+    "preact": "^10.17.0"
```
ibgreen, line 34: Just curious, I haven't really being paying attention. Are we using a react style library for our pure-js implementation? And then we will have react wrappers on top?

modules/widgets/src/fullscreen-widget.ts
chrisgervang: Yes, adding a `container?: HTMLElement` prop.

modules/widgets/src/fullscreen-widget.tsx
chrisgervang: Yes, looking into `document.addEventListener('fullscreenchange')` for this.

modules/widgets/package.json
chrisgervang: I'm pretty certain we don't want the complexity or bundle size of react, but I like its api style.

My goal has been to pick something as close to the DOM as we can while also providing some convenience over `document.createElement`. `preact` seems to serve this niche well... It give us the ability to make basic components with JSX, and implements a thin virtual dom only adding 3kb to the bundle.

Re: react wrappers, I think of a widget as a deck component that happens to have UI. In a sense, how the UI is built within it is ideally isolated from other UI on the page so that deck can manage it. Let's say we used react to make UI widgets in `@deckgl/widgets`... I'd still ideally want them to [create their own react root](https://react.dev/learn/add-react-to-an-existing-project#using-react-for-a-part-of-your-existing-page) separate from the rest of application and other widgets. This way all integrations work through the same interface.

ibgreen: Understood and I agree with the principle that widgets should be free to choose implementation as long as they are externally independent. That said, I would not be happy to see a pure-js deck.gl module have react as a dependency in its package.json. I would think twice before using that in a pure-js app.

For a community widget that could be a different story of course.

preact seems like a reasonable choice.

docs/api-reference/widgets/fullscreen-widget.md
```diff
@@ -0,0 +1,68 @@
+# FullscreenWidget
+
+## Variables
```
chrisgervang, line 3: These variables are intended for all core widgets to use. For now, documenting in FullscreenWidget since it's the only usage.

examples/get-started/pure-js/widgets/app.js
```diff
@@ -0,0 +1,85 @@
...
+
+const prefersDarkScheme = window.matchMedia("(prefers-color-scheme: dark)");
+const widgetTheme = prefersDarkScheme.matches ? {
+  // Dark Glass Theme
+  '--button-background': 'rgba(18, 18, 20, 0.75)',
+  '--button-stroke': 'rgba(18, 18, 20, 0.30)',
+  '--button-inner-stroke': '1px solid rgba(18, 18, 20, 0.75)',
+  '--button-shadow': '0px 0px 8px 0px rgba(0, 0, 0, 0.25), 0px 0px 8px 0px rgba(0, 0, 0, 0.1) inset',
+  '--button-backdrop-filter': 'blur(4px)'
+} : {
```
chrisgervang, line 18: Demonstrates how an application can set the theme from JS.

examples/get-started/pure-js/widgets/index.html
```diff
@@ -0,0 +1,21 @@
...
+  <style>
+    body {margin: 0; width: 100vw; height: 100vh; overflow: hidden;}
+    
+    @media screen and (max-width: 600px) {
+      .deckgl-wdgt {
+        /* mobile */
+        --button-size: 48px;
+        --button-corner-radius: 12px;
+      }
+    }
```
chrisgervang, line 15: Demonstrates how an application can set theme from CSS.

modules/widgets/src/stylesheet.css
```diff
@@ -0,0 +1,33 @@
+.deckgl-wdgt {
+  margin: var(--wdgt-margin, 12px);
+}
+
+.deckgl-wdgt-button-border {
+  width: var(--button-size, 28px);
```
chrisgervang, line 6: `@deck.gl/widgets` won't define values for the CSS variables to avoid cascading issues (e.g. what happens when app css is loaded before lib css?). Instead, deck defines a default in each usage of `var()` according to the Light Flat theme.

.eslintrc.js
```diff
@@ -109,6 +109,12 @@ module.exports = getESLintConfig({
           'import/named': 0,
           'no-new': 0
         }
+      },
+      {
+        files: ['modules/widgets/**/*.tsx'],
+        rules: {
+          'react/react-in-jsx-scope': 0
```
ibgreen, line 116: It would be very helpful to add comments here, future maintainers will likely struggle to understand the context. E.g: I am assuming we are not actually using react but using these rules with preact?

babel.config.js
```diff
@@ -13,6 +13,17 @@ module.exports = getBabelConfig({
           patterns: ['**/*.glsl.js', '**/*.glsl.ts']
         }
       ]
+    ],
+    overrides: [
```
ibgreen, line 17: ditto, comments with justification please

examples/get-started/pure-js/widgets/app.js
ibgreen: OK but the default dark and light themes should not be in the app right, they should be provided by the widgets module? Maybe less confusing if the app was setting a unique theme (say a game UI theme or a theme that favors red color etc)

ibgreen (approved):

examples/get-started/pure-js/widgets/app.js
chrisgervang: I'd prefer to keep the widget's "dumb" in the sense that they won't try to automatically apply a theme based on mobile or dark/light OS theme. Reason being, apps will already own their approach on when to apply these responsive styles and there are potentially many edge cases and oddities to cover. 

For instance, apps may have a custom user-override for light/dark, they may want a different size for iPad, etc..

As long as they can apply a theme in their CSS or via JS, they should be good.

modules/widgets/src/stylesheet.css
ibgreen: Yes... Just to clarify, when i said `deck.gl` I just meant to say that the deck core module should not be bothered with CSS registration type stuff (it should be handled the deck widget module and/or the app).

examples/get-started/pure-js/widgets/app.js
ibgreen: Yews but even if the widgets don't do any auto switching of themes,  the app shouldn't have to define the full themes?

I was assuming that the "themes" should be importable from the widget module.

Seeing that an example app has to tediously itemize detailed sets of CSS values does not give me a sense that I get easy theming with this widget library. 

```
import {LightTheme, DarkTheme} from '@deck.gl/widgets';
const widgetTheme = prefersDarkScheme.matches ? LightTheme : DarkTheme;
```

docs/api-reference/widgets/fullscreen-widget.md
```diff
@@ -0,0 +1,68 @@
...
+
+### Size
+
+| Name | Type | Default |
+| ---- | ---- | ------- |
+| `--button-size` | [Dimension](https://developer.mozilla.org/en-US/docs/Web/CSS/dimension) | `28px` |
+| `--button-border-radius` | [Dimension](https://developer.mozilla.org/en-US/docs/Web/CSS/dimension) | `12px` |
+| `--wdgt-margin` | [Dimension](https://developer.mozilla.org/en-US/docs/Web/CSS/dimension) | `12px` |
```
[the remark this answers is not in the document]
chrisgervang: Users have a variety of ways to control the scope of their CSS variable customizations, and these are good additions for the docs.

To scope to the fullscreen widget users can set the `style` prop in the JS widget - that uses a CS in-line style, so it'll cascade down only for that instance. This would mean two fullscreen buttons can be uniquely styled if desired.

To target all fullscreen widgets variables can be set CSS in the `.deckgl-fullscreen` selector.

To target all widgets set variables in `.deckgl-widget`.

examples/get-started/pure-js/widgets/app.js
chrisgervang: Oh, good idea. Then users could import a theme and modify maybe one or two variables as-needed too.

modules/widgets/src/stylesheet.css
chrisgervang: Ah, ok yes. Totally agree. This approach leaves it up to the app to import the CSS stylesheet with whatever approach they use for other sheets. The widget module just provides an unattached style sheet and should come with usage docs similar to [maplibre](https://maplibre.org/maplibre-gl-js/docs/#maplibre-css).

docs/api-reference/widgets/fullscreen-widget.md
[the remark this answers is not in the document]
chrisgervang: As for prefix's, I'd like them to be semantic names that aren't prescriptive of look.

So `widget` prefix should target the widget as a whole, `button` targets whatever is functionally apart of a button, etc. 

An good, non-prescriptive, name for a CSS `backdrop-filter` property would be `button-backdrop-filter`. `button-backdrop-blur` wouldn't be good since `blur` prescribes a look.

modules/widgets/src/fullscreen-widget.tsx
```diff
@@ -0,0 +1,101 @@
+import type {Deck, Widget, WidgetPlacement} from '@deck.gl/core';
+import {h, render} from 'preact';
+
+/* global document */
+const defaultStyle: Readonly<Partial<CSSStyleDeclaration>> = {
```
Pessimistress, line 5: I assume this is no longer used?

docs/api-reference/widgets/fullscreen-widget.md
```diff
@@ -19,50 +19,3 @@
 | `--button-inner-stroke` | [Border](https://developer.mozilla.org/en-US/docs/Web/CSS/border) | `unset` |
 | `--button-shadow` | [Box Shadow](https://developer.mozilla.org/en-US/docs/Web/CSS/box-shadow) | `0px 0px 8px 0px rgba(0, 0, 0, 0.25)` |
 | `--button-backdrop-filter` | [Backdrop Filter](https://developer.mozilla.org/en-US/docs/Web/CSS/backdrop-filter) | `unset` |
-
```
Pessimistress: It would be nice to include an image showing what each theme looks like.

modules/widgets/src/fullscreen-widget.tsx
```diff
@@ -0,0 +1,141 @@
...
+  viewId = null;
+
+  deck?: Deck;
+  element?: HTMLDivElement;
+
+  fullscreen: boolean = false;
+
+  constructor(props: FullscreenWidgetProps) {
+    this.id = props.id || 'fullscreen';
+    this.viewId = props.viewId || null;
```
Pessimistress, line 35: What does it mean to assign the full screen widget to a specific view? My understanding is that it will always full screen the whole canvas.

modules/widgets/src/fullscreen-widget.tsx
```diff
@@ -0,0 +1,141 @@
...
+    props.style = props.style || {};
+    this.props = props;
+  }
+
+  onAdd({deck}: {deck: Deck}): HTMLDivElement {
+    const {style, className} = this.props;
+    const el = document.createElement('div');
+    el.classList.add('deckgl-widget', 'deckgl-widget-fullscreen');
+    if (className) el.classList.add(className);
+    Object.entries(style).map(([key, value]) => el.style.setProperty(key, value));
```
Pessimistress, line 47: This is how Deck applies style: https://github.com/visgl/deck.gl/blob/master/modules/core/src/lib/deck.ts#L734

I have not tested it but the two approaches seem equivalent?
https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/setProperty#alternative_usage

chrisgervang: I started with that, but it doesn't work for css variables.

`Object.assign(el.style, this.props.style);` results in no variables applied to the middle widget.
Screen Shot 2023-09-09 at 3 19 00 PM

`el.style.setProperty` has inline styles:
Screen Shot 2023-09-09 at 3 19 23 PM

I think we need to change how Deck does it so the theme variables will work.

modules/widgets/src/fullscreen-widget.tsx
chrisgervang: I also think this is a little odd, but I understood it as `WidgetManager` offsets position of the widget to be inside the view of choice. Fullscreen behavior will always apply to a whole container, be it `canvas.parentElement` or user provided `container` prop.

examples/get-started/pure-js/widgets/app.js
```diff
@@ -0,0 +1,75 @@
...
+      getTargetPosition: f => f.geometry.coordinates,
+      getSourceColor: [0, 128, 200],
+      getTargetColor: [200, 0, 80],
+      getWidth: 1
+    })
+  ],
+  widgets: [
+    new FullscreenWidget({}),
+    new FullscreenWidget({id: 'themed', style: widgetTheme}), 
+    new FullscreenWidget({id: 'purple', className: 'purple'})
```
chrisgervang, line 73: @Pessimistress I was thinking of move these variations to a test app. I want to maintain coverage for edge cases somewhere, and keep getting started very simple. Thoughts?

.eslintrc.js
chrisgervang: Added "Disable React-style JSX linting since they conflict with Preact JSX."

examples/get-started/pure-js/widgets/app.js
Pessimistress: Yes agreed

modules/widgets/src/fullscreen-widget.tsx
Pessimistress: I think that will only confuse the user more. I suggest you remove this prop for now and add it back if it is needed.

chrisgervang: Agreed.. just cuz we can position it anywhere doesn't mean we should

chrisgervang:
Users may want to replace the icons, so I've moved them to CSS. To replace with your own we'll document:

Icons are replaceable and their colors can be customized with CSS variables. 
- Prepare your icons as SVG urls for a CSS mask-image.
- The color of embedded in your SVG will be ignored, but ensure it isn’t transparent. Set the color with the appropriate CSS variable.
- Replace the icon with CSS variables like `--icon-fullscreen-enter`.

docs/api-reference/widgets/overview.md
chrisgervang, line 1: @Pessimistress, added sections for replacing icons and made other adjustments. Any comments?

chrisgervang: cc @ilyabo, @felixpalmer, @ibgreen

PR #8040, opened 2023-08-03
[chore] clean typescript output before new build

Pessimistress:
#### Change List
- Prevent outdated files from being published


chrisgervang (approved):

ibgreen (approved):
LGTM... I reckon we are not using incremental typescript builds anymore, I had that working for a while in some of the repos.

PR #8042, opened 2023-08-04
Temporarily disable node test and coverage report

Pessimistress:
Allow CI to function on master while 9.0 work is in progress


chrisgervang (approved):
Lgtm, node tests can still run locally and we can bring them back when they're stable

PR #8051, opened 2023-08-10
CARTO v9 API RFC

felixpalmer:


dev-docs/RFCs/v8.0/carto-v9-rfc.md
```diff
@@ -0,0 +1,222 @@
...
+
+At a high level we want to achieve an API which makes it clearer to the user what is going on, and allows for better extensibility. We also want to separate the options used for loading the data from the props which style the layer.
+
+To do so, the interaction with the CARTO API will be done via a set of typed helper classes which aid the user in obtaining the TileJSON needed for a particular layer.
+
+```javascript
+import CartoQuerySource, CartoQuerySourceOptions, CartoTileLayer from '@deck.gl/carto';
+
+// Global options
+setDefaultCredentials({accessToken: 'XXXX'});
```
alasarr, line 57: I think we should deprecated this too. The credentials are storing two things today: accessToken and apiBaseUrl. The apiBaseUrl is not a `credential`...

I propose the following:
- Add accessToken and apiBaseUrl to the source object. apiBaseUrl defaults to the US region.
- Add a setCartoAccessToken and setCartoApiBaseUrl. I add `Carto` because setAccessToken will be super generic when it's used in large applications. I know we could solve as alias, but I think it's better to force the default alias to carto

dev-docs/RFCs/v8.0/carto-v9-rfc.md
```diff
@@ -0,0 +1,222 @@
...
+
+  // ...style props
+});
+```
+
+## Class names
+
+The goal is to have descriptive names for classes and options, even if this means they are more verbose.
+
+| Source class                | Layer                   |
```
alasarr, line 79: What about removing the `tile` suffix? Everything is a tile layer and the idea is to continue like that

dev-docs/RFCs/v8.0/carto-v9-rfc.md
```diff
@@ -0,0 +1,222 @@
...
+| `CartoQuadbinQuerySource`   | `CartoQuadbinTileLayer` |
+| `CartoQuadbinTilesetSource` | `CartoQuadbinTileLayer` |
+| `CartoRasterSource`         | `CartoRasterTileLayer`  |
+
+### Options
+
+Each of the `XXXSource` classes takes an options object as a constructor parameter, all inheriting from the common `CartoSourceOptions` base type.
+
+```javascript
+type CartoSourceOptions = {
```
alasarr, line 97: I would include the path URL of maps. We could potentially change the path parameter of maps api and for widgets we will need the path to maps api.

mapsApiUrlPath = /v3/maps
(future) sqlApiUrlPath = /v3/sql
(future) statsApiUrlPath = /v3/stats

alasarr (changes requested):
Thanks for setting this RFC @felixpalmer!  I have added some comments. 

We also need to define here how fetchLayerData will be affected.

dev-docs/RFCs/v8.0/carto-v9-rfc.md
```diff
@@ -0,0 +1,222 @@
...
+  connectionName: string,
+  credentials: CloudNativeCredentials,
+  clientId?: string,
+  formatTiles?: TileFormat,
+  headers?: Headers,
+  v?: string
```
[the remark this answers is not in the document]
felixpalmer: Just `apiVersion` is more consistent with other naming. I wonder how much value there is allowing configuring this at all. Ideally the version of the client would just work with a single API version. At least I think we should type enforce what this string should be to guide users

dev-docs/RFCs/v8.0/carto-v9-rfc.md
felixpalmer: Yes, it's a pain having it everywhere. I've removed it, but I'm wary of keeping the generic `CartoLayer` name as people will just expect it to continue to work as before. `CartoVectorLayer`?

dev-docs/RFCs/v8.0/carto-v9-rfc.md
felixpalmer: Why just special case those two? In general what we are trying to do is to provide global `Source` configuration. One could argue there is value in allowing global configuration of `headers`, `clientId` etc.

In general I don't love the `setDefaultCredentials` pattern as it means all of the parameters (`accessToken`, `mapsUrl`) need to be typed as optional and we have no way to know at compile time if the user has forgotten to include something we require (for example access token). Also we rely on the user setting some global state which isn't great.

How about this as a pattern?

```javascript
import {SOURCE_DEFAULTS, CartoQuerySource, CartoVectorLayer} from '@deck.gl/carto';

const data = CartoQuerySource({
  ...SOURCE_DEFAULTS,
  accessToken: 'XXX',
  connectionName: 'bigquery',
  sqlQuery: 'select * from carto.tables.example'
});

new CartoVectorLayer({data});
```

In a larger codebase with different sources the user would define their own source:

```javascript
// AuthenticatedQuerySource.js
export default function AuthenticatedQuerySource(options) {
  return CartoQuerySource({...SOURCE_DEFAULTS, accessToken: 'XXX', connectionName: 'bigquery', ...options});
};

// Elsewhere
import AuthenticatedQuerySource from 'AuthenticatedQuerySource';

new CartoVectorLayer({
  data: AuthenticatedQuerySource({sqlQuery: 'select * from carto.tables.example'});
});
```

dev-docs/RFCs/v8.0/carto-v9-rfc.md
```diff
@@ -0,0 +1,183 @@
...
+new CartoVectorLayer({
+  loadOptions: {
+    fetch: {
+      headers: {Authorization: `Bearer ${accessToken}`}
+    }
+  })
+});
+```
+
+This is in line with [deck.gl docs](https://deck.gl/docs/developer-guide/loading-data#example-fetch-data-with-credentials) and also makes it clearer how to set custom headers on the requests.
```
alasarr, line 177: 👍

dev-docs/RFCs/v8.0/carto-v9-rfc.md
alasarr: Yes, what I wanted to propose is something more general rather than setDefaultCredentials. 

I like your approach

alasarr (approved):

padawannn (changes requested):
What happens with the dynamic aggregation? I mean, if you have a point table that you want to aggregate dynamically using Quandbin, must you use a `CartoVectorTableSource` with an aggregation expression to render a `CartoQuadbinTileLayer` layer? I think it's a little confusing

felixpalmer:
> What happens with the dynamic aggregation? I mean, if you have a point table that you want to aggregate dynamically using Quandbin, must you use a `CartoVectorTableSource` with an aggregation expression to render a `CartoQuadbinTileLayer` layer? I think it's a little confusing

@padawannn wouldn't you just use a `CartoQuadbinTableSource` in that case? The `CartoQuadbinTableSource` is just basically a function that takes some parameters, converts them to a URL and fetches the tileJSON

padawannn:
> > What happens with the dynamic aggregation? I mean, if you have a point table that you want to aggregate dynamically using Quandbin, must you use a `CartoVectorTableSource` with an aggregation expression to render a `CartoQuadbinTileLayer` layer? I think it's a little confusing
> 
> @padawannn wouldn't you just use a `CartoQuadbinTableSource` in that case? The `CartoQuadbinTableSource` is just basically a function that takes some parameters, converts them to a URL and fetches the tileJSON

So if I understand well, it would be a `CartoQuadbinTableSource` and we would have to set the `spatialDataColumn` parameter, right? Internally we would call the Maps API with `geoColum=quadbin:geom` if we would set the `spatialDataColumn` as `geom`

@felixpalmer

PR #8055, opened 2023-08-13
Added new numSegments prop in ArcLayer

karsta26:
For #8052

#### Background
Added a new parameter to be able to modify the number of segments used to generate arcs. Previously the number was hardcoded - 50. 
This will allow rendering smooth arcs in case of long paths with `greatCircle` param.
#### Change List
- Added new `numSegments` prop in `ArcLayer`
- Fixed some broken links


test/render/golden-images/great-circle-modified-segments.png
felixpalmer, line 1: The render output looks strange to me, any idea why those lines are becoming thin and forming the square shapes? It's quite different from the `great-circle.png` output

felixpalmer: I think perhaps a better test case is one demonstrating what this PR is for, i.e. `numSegments > 50` showing a zoomed-in portion of a curve that is now smooth, whereas before it had visible segments - like you have in the [Issue](https://github.com/visgl/deck.gl/issues/8052)

karsta26: Regarding square shapes:
This is due to a projection and very long paths that are used in this example. 

To better visualize this: this is the data from the `great-circle` test case shown on a map with countries with `numSegments = 50`.
image

And this is the same data but with `numSegments = 8` as in the new test.
image

As you can see for the very long path there are not enough segments to draw it properly and because of this square shape is visible. 


I agree that perhaps this is not the best way to demonstrate the new parameter.
I have changed the test case to have only one arc. 

With the default value of the parameter, the test fails and produces this image:
image

And with `numSegments = 150` the new golden image:
image

felixpalmer (changes requested):

modules/layers/src/arc-layer/arc-layer.ts
```diff
@@ -52,6 +52,7 @@ const defaultProps: DefaultProps<ArcLayerProps> = {
   getTilt: {type: 'accessor', value: 0},
 
   greatCircle: false,
+  numSegments: {type: 'number', value: 50, min: 0},
```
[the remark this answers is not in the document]
karsta26: Changed to 1

felixpalmer (approved):

Issue #8073, opened 2023-08-22
[Bug] TileLayer.isLoaded leads to a crash


ilyabo:
### Description

Calling `tileLayer.isLoaded()` may lead to a crash when called early before `selectedTiles` is initialized in `tileset`.

### Flavors

- [ ] Script tag
- [X] React
- [ ] Python/Jupyter notebook
- [ ] MapboxOverlay
- [ ] GoogleMapsOverlay
- [ ] CartoLayer
- [ ] ArcGIS

### Expected Behavior

There should be no crash when calling `tileLayer.isLoaded()`.

### Steps to Reproduce

In summary: Create a TileLayer and call `tileLayer.isLoaded()` just after creation.
Has been observed in FSQ Studio.
Will try to come up with a minimal example.

### Environment

- Framework version: 8.9.12
- Browser: Chrome 114
- OS: Mac OS

Pessimistress:
8.9.27

PR #8074, opened 2023-08-22
Fix TileLayer isLoaded crash

ilyabo:
Solves #8073

#### Change List
- Adds a type guard to `isLoaded()` to prevent crash when `tileset.selectedTiles` is `null`


modules/geo-layers/src/tile-layer/tile-layer.ts
```diff
@@ -165,7 +165,7 @@ export default class TileLayer<DataT = any, ExtraPropsT extends {} = {}> extends
   }
 
   get isLoaded(): boolean {
-    return this.state?.tileset?.selectedTiles.every(
+    return this.state?.tileset?.selectedTiles?.every(
```
jwasilgeo, line 168: Question: would it be worth it to use something like `...tileset?.isLoaded`? At a quick glance the potential tradeoffs and subtle differences aren't fully clear to me yet.

https://github.com/visgl/deck.gl/blob/master/modules/geo-layers/src/tileset-2d/tileset-2d.ts#L183-L185

Pessimistress (approved):

PR #8076, opened 2023-08-23
CARTO: Don't hardcode mime-types

felixpalmer:
Followup to https://github.com/visgl/deck.gl/pull/8056

#### Background

The old versions of the CARTO API did not set explicit mime-types for responses. Both the vector and spatial index binary format simply returned `Content-Type: application/x-protobuf` which meant loaders.gl has no way to select a loader automatically. As of v=3.2 the API returns explicit mimeTypes so loaders.gl can automatically select the correct loader without having to be informed explictly via `loadOptions`

#### Change List
- Remove hardcoded mime-types in Layer classes
- Update test mock API to return mime-type for binary vector tile responses


zbigg (approved):
lgtm

Issue #8108, opened 2023-09-12
[Bug] Missing typings for React DeckGL children set to a render callback


mz8i:
### Description

The following simple setup

```tsx
<DeckGL>
    {({viewState, width, height}) => (<>
      <Map viewState={viewState} width={width} height={height} />
    </>)}
  </DeckGL>
```

 produces a TypeScript error:
```
Type '({ viewState, width, height }: { viewState: any; width: any; height: any; }) => Element' is not assignable to type 'ReactNode'.
```

### Flavors

- [ ] Script tag
- [X] React
- [ ] Python/Jupyter notebook
- [ ] MapboxOverlay
- [ ] GoogleMapsOverlay
- [ ] CartoLayer
- [ ] ArcGIS

### Expected Behavior

Passing a render callback as the React component's children is allowed and documented for JavaScript, so should be supported in TypeScript.

### Steps to Reproduce

```tsx
<DeckGL>
    {({viewState, width, height}) => (<>
      <Map viewState={viewState} width={width} height={height} />
    </>)}
  </DeckGL>
```

### Environment

- Framework version: 8.9.26
- Browser: N/A
- OS: N/A

Pessimistress:
Fix published in v8.9.28

PR #8109, opened 2023-09-12
Add typings for React render callback children

mz8i:
For #8108

#### Change List
- add typings for React DeckGL render callback


Pessimistress (approved):
Thank you!

Issue #8060, opened 2023-08-19
[Bug]: Layers are sinking under Mapbox Terrain


belom88:
### Description

I'm trying to use `ScenegraphLayer` with `Mapbox 2` terrain in interleaved mode.
Having the correct z-value a model only seen on high `pitch` value close to 90 degrees. If I tilt the camera to a lower `pitch` value, the layer sinks under terrain.
Example to reproduce: https://stackblitz.com/edit/js-njts6h?file=index.js (Mapbox token required)
<img width="296" alt="image" src="https://github.com/visgl/deck.gl/assets/18549629/d0804909-bd0a-49e6-9d3b-d84b29efb1de">
<img width="276" alt="image" src="https://github.com/visgl/deck.gl/assets/18549629/748d3ed8-edb1-4a55-b094-c809fb9dc4b1">
<img width="295" alt="image" src="https://github.com/visgl/deck.gl/assets/18549629/6a2c6498-0e39-4f11-a505-5fa735e073df">

### Flavors

- [ ] Script tag
- [ ] React
- [ ] Python/Jupyter notebook
- [X] MapboxOverlay
- [ ] GoogleMapsOverlay
- [ ] CartoLayer
- [ ] ArcGIS

### Expected Behavior

The layer position don't change when I tilt a map with terrain.

### Steps to Reproduce

* Create mapbox map with terrain;
* Add a layer created with `MapboxLayer` class

### Environment

- Framework version: 8.9.25
- Browser: Chrome 116.0.5845.96 (Official Build) (64-bit)
- OS: Windows 11

Pessimistress:
My guess is that deck's projection matrix near/far planes do not match those of mapbox when terrain is on.

piergiorgio-roveda:
Hi there,

I've encountered the same error while working with Mapbox and DeckGL for 3D visualization using GeoJSON. Has there been any progress in resolving this issue? I'm eager to get this sorted out as it's affecting my project's visualization. Any updates or suggestions on a workaround would be greatly appreciated. Thanks for your efforts in addressing this matter!

Best regards,
Pj

https://github.com/visgl/deck.gl/assets/74056468/103ec74d-fc5b-4ba9-bdef-a4bcab2f569c

felixpalmer:
@piergiorgio-roveda @belom88 could you ask on the Mapbox project for clarity on how the projection matrix changes when terrain is enabled?

piergiorgio-roveda:
[...]

PR #8111, opened 2023-09-13
[mapbox] Fix interleaving when base map has terrain

Pessimistress:
For #8060

#### Background

The current projection matrix parameters in WebMercatorViewport are designed to match an early release of Mapbox. Between mapbox-gl, maplibre-gl, and their different major versions, the calculation has diverged significantly. This hack extracts near and far plane positions from a private property on the mapbox-gl map.

For the long term, I will open feature requests for the base maps to expose their projection matrix.

#### Change List
- WebMercatorViewport accepts external nearZ/farZ positions
- MapboxLayer uses base map nearZ/farZ if available


Pessimistress:
Mapbox issue: https://github.com/mapbox/mapbox-gl-js/issues/12888
Maplibre issue: https://github.com/maplibre/maplibre-gl-js/issues/3080

felixpalmer (approved):
Good find!

Issue #8091, opened 2023-08-31
[Bug] Interleaved Mapbox overlay does not work with terrain elevation


ssaornil:
### Description

We are using MapboxOverlay in interleaved mode together with the TerrainExtension on the deck layers so they are positioned in z according to the terrain elevation.

The deck layers are not visible at all. I suspect they remain at z zero.

### Flavors

- [ ] Script tag
- [ ] React
- [ ] Python/Jupyter notebook
- [X] MapboxOverlay
- [ ] GoogleMapsOverlay
- [ ] CartoLayer
- [ ] ArcGIS

### Expected Behavior

The deck layers ignore the z attribute and use the terrain elevation instead

### Steps to Reproduce

[Code sandbox](https://codesandbox.io/s/deck-layers-with-3d-927gh8?file=/src/App.tsx)

If you change the "interleaved" flag, it "works" as expected but we need the interleaved mode.

### Environment

- Framework version: deck.gl@8.9.x
- Browser: Chrome 116.0
- OS: Ubuntu 22.04.3 LTS

dsactionengine:
Could be related to https://github.com/visgl/deck.gl/issues/8060

ssaornil:
I updated the code sandbox with [version 8.9.28](https://github.com/visgl/deck.gl/releases/tag/v8.9.28) which fixes the issue mentioned in the previous comment but the behavior is still the same so there must be something else at play here.

ssaornil:
Thanks @Pessimistress.

After updating the code sandbox with [version 8.9.30](https://codesandbox.io/s/deck-layers-with-3d-927gh8) the layers are finally displayed in interleaved mode but they seem to be partially hidden by the terrain depending on the zoom level, pitch and rotation angles. The expectation would be for them to show completely on top of the terrain.

PR #8115, opened 2023-09-14
[mapbox] fix drawing TerrainExtension in interleaved mode

Pessimistress:
For #8091

#### Change List
- Include terrain source when drawing each MapboxLayer

### DIFF
--- a/modules/mapbox/src/deck-utils.ts
+++ b/modules/mapbox/src/deck-utils.ts
@@ -162,7 +162,8 @@ export function drawLayer(deck: Deck, map: Map, layer: MapboxLayer<any>): void {
 
   deck._drawLayers('mapbox-repaint', {
     viewports: [currentViewport],
-    layerFilter: ({layer: deckLayer}) => layer.id === deckLayer.id,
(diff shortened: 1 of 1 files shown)


modules/mapbox/src/deck-utils.ts
```diff
@@ -162,7 +162,8 @@ export function drawLayer(deck: Deck, map: Map, layer: MapboxLayer<any>): void {
 
   deck._drawLayers('mapbox-repaint', {
     viewports: [currentViewport],
-    layerFilter: ({layer: deckLayer}) => layer.id === deckLayer.id,
+    layerFilter: ({layer: deckLayer}) =>
+      layer.id === deckLayer.id || deckLayer.props.operation.includes('terrain'),
```
felixpalmer, line 166: Do we have the same issue with `OPERATION.MASK`?

felixpalmer (approved):

modules/mapbox/src/deck-utils.ts
Pessimistress: I don't think MaskEffect is using `layerFilter` to exclude mask sources.

Issue #8120, opened 2023-09-20
[Bug] deck.gl website compiled with incorrect __VERSION__


felixpalmer:
### Description

The deck.gl website reports `0.0.0` as the deck.gl version. This causes [examples which rely](https://deck.gl/examples/carto) on this compile time constant to be valid to fail.

### Flavors

- [ ] Script tag
- [ ] React
- [ ] Python/Jupyter notebook
- [ ] MapboxOverlay
- [ ] GoogleMapsOverlay
- [ ] CartoLayer
- [ ] ArcGIS

### Expected Behavior

The correct version to be reported.

### Steps to Reproduce

Open the browser console on deck.gl and type `deck.VERSION`

### Environment

- Framework version:
- Browser:
- OS:

felixpalmer:
The cause here is the `website/package.json` which has `version: "0.0.0"`. Should we update this to match `modules/xxx/package.json` and add `website` to packages in `lerna.json`?

PR #8124, opened 2023-09-21
Chore: Include website in lerna config

felixpalmer:
For https://github.com/visgl/deck.gl/issues/8120

#### Change List
- Include `website` directory in lerna.json

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -4,7 +4,8 @@
   "npmClient": "yarn",
   "useWorkspaces": true,
   "packages": [
-    "modules/*"
+    "modules/*",
+    "website"
   ],
   "command": {
     "bootstrap": {}


felixpalmer:
Note, I haven't been able to properly test this without running the publish process

Pessimistress:
This does not look like the correct solution. Is the version constant referenced in the carto module or in the example itself? Why does the website version matter?

felixpalmer:
It is referenced in the carto module, where it should be compiled-in. The example just imports the carto module, so I'm not really clear on where the "0.0.0" deck version gets injected

Pessimistress (approved):

felixpalmer:
@Pessimistress I've been able to test this with `npx lerna version --force-publish --no-git-tag-version --no-push` and it works but I had to add the configuration to `package.json` due to `useWorkspaces` being specified in `lerna.json`


Screenshot 2023-09-22 at 15 43 51

Pessimistress (changes requested):
@felixpalmer Adding it to the root workspace will change the dependency tree (currently website dependencies are installed separately and controlled by `website/yarn.lock`). It could have unexpected consequences.

Alternatively you can add a script to `website/scripts/build.sh` to update `website/package.json` to match `modules/core/package.json`.

Pessimistress (approved):

PR #8130, opened 2023-09-21
Carto: Initial implementation of v9 API

felixpalmer:

#### Background

READ FIRST: Implements [RFC detailed here ](https://github.com/visgl/deck.gl/blob/master/dev-docs/RFCs/v8.0/carto-v9-rfc.md)

Based off 8.9 to ease testing. This PR will not be merged into master until binary data handling is fixed in the `VectorTileLayer`

#### Change List
- Implementation of Source classes for all types of data sources
- Rename `_CartoTileLayer` -> `VectorTileLayer` & export previously private Layers
- Automatic injection of access token in Layers
- Test app to show usage with all data types
- Types for data sources and return values


alasarr:
@padawannn @zbigg @VictorVelarde can you help me with the review of this PR?. I'm not going to be able to check until the end of the week and don't want to slow it down

modules/carto/src/sources/index.ts
```diff
@@ -0,0 +1,33 @@
+export {SOURCE_DEFAULTS} from './common';
+export type {
+  CartoSourceOptionalOptions,
```
VictorVelarde, line 3: CartoSourceOptionalParams / CartoSourceRequiredParams ? to avoid the cacophony?

test/apps/carto-dynamic-tile/app.tsx
```diff
@@ -0,0 +1,142 @@
...
+}
+
+async function fetchLayerData() {
+  const data = await CartoVectorTableSource({...globalOptions, tableName: 'carto-demo-data.demo_tables.chicago_crime_sample', format: 'geojson'});
+  console.log(data.tiles); // <- Typescript error
+  console.log(data.features); // <- type: GeoJSON
+}
+fetchLayerData();
+
+function createVectorLayer(datasource) {
```
VictorVelarde, line 103: Same, we should add the 'use' preffix on any function using react hooks, unless they are indeed React components with JSX

modules/carto/src/sources/index.ts
felixpalmer: Well they are options rather than parameters if you go by the `fn(param1, param2)` vs `fn({option1, option2})` convention. I'm actually not going to export these types in the end, they are mainly there internally to help distinguish between options for which the client has defaults and those that mush be provided by the user

test/apps/carto-dynamic-tile/app.tsx
```diff
@@ -0,0 +1,142 @@
...
+        title="dataset"
+        obj={Object.keys(datasets)}
+        value={dataset}
+        onSelect={setDataset}
+      />
+    </>
+  );
+}
+
+function createH3Layer(datasource) {
```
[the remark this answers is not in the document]
felixpalmer: I've changed the "create" prefix to "use" to make clearer. I'm not sure what the best practice here is but this is just a test app so don't have strong feelings

modules/carto/src/sources/index.ts
```diff
@@ -0,0 +1,32 @@
+export {SOURCE_DEFAULTS} from './common';
+export type {CartoTilejsonResult, GeojsonResult, JsonResult} from './common';
+
+export {CartoH3QuerySource} from './h3-query-source';
```
aaranadev, line 4: @felixpalmer the functions in JS should be in camelCase instead of PascalCase, can you change them? 

You should review all of them.

Thanks

modules/carto/src/sources/h3-query-source.ts
```diff
@@ -0,0 +1,48 @@
...
+  geo_column?: string;
+  q: string;
+  queryParameters?: string;
+};
+
+const CartoH3QuerySource: TilejsonSource<CartoH3QuerySourceOptions> = async function (
+  options: CartoH3QuerySourceOptions
+): Promise<any> {
+  const {
+    aggregationExp = '1 AS value',
```
aaranadev, line 25: @felixpalmer, I would like to know, why this default value 🤔.  Below you are checking if the value exist, it's always true by default and I don't know If this default value can hide possible errors by not passing the parameter from outside

modules/carto/src/sources/utils.ts
```diff
@@ -0,0 +1,63 @@
...
+      response = await fetch(url, {headers});
+    }
+    let json: any;
+    try {
+      json = await response.json();
+    } catch {
+      json = {error: ''};
+    }
+    if (!response.ok) {
+      throw new CartoAPIError(json.error, errorContext, response);
```
aaranadev, line 42: @felixpalmer the error is always `{error: ''}`, do we have a way to know the error?

modules/carto/src/sources/index.ts
felixpalmer: OK, fair enough, I'll rename to `cartoXXXSource` etc. I didn't want to implement these as classes, as they are essentially just wrappers around `fetch()`

modules/carto/src/sources/h3-query-source.ts
felixpalmer: In our current code we provide this as a default, perhaps it is better to remove it to make it clearer to the user they need to provide the expression. After all just aggregating using `1 AS value` isn't very useful.

I'll remove the default value and mark it as required

modules/carto/src/sources/utils.ts
felixpalmer: This is how the current code works. `CartoAPIError` will display an error message which will log out the response error code. The error will only by `''` when no JSON response can be parsed

aaranadev: Ok, thx!

modules/carto/src/sources/index.ts
aaranadev: Great, thx!

modules/carto/src/sources/h3-query-source.ts
aaranadev: I agree with @felixpalmer, the user must provide the expression

PR #8131, opened 2023-09-23
Support css variables in `style` prop

chrisgervang:
For #7946 

#### Background
We aim to offer a convenient method for theming UI widgets using CSS variables through the deck `style` property. However, our current styling approach supports only CSS properties and not variables.

#### Change List
- Use `setProperty` instead of `assign` for setting css style values.

### DIFF
--- a/modules/core/src/lib/deck.ts
+++ b/modules/core/src/lib/deck.ts
@@ -736,7 +736,10 @@ export default class Deck {
       parent.appendChild(canvas);
     }
 
-    Object.assign(canvas.style, props.style);
+    // Set css variables and properties
+    Object.entries(props.style).map(([key, value]) =>
+      (canvas as HTMLCanvasElement).style.setProperty(key, value as string)
+    );
 
     return canvas;
   }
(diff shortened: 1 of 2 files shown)


chrisgervang:
@Pessimistress, the canvas is `null` in our tests. Do we mock a canvas or can you think of some way to test this?

modules/core/src/lib/deck.ts
```diff
@@ -466,6 +466,7 @@ export default class Deck {
     // Merge with existing props
     Object.assign(this.props, props);
 
+    this._setStyle(this.props);
```
Pessimistress, line 469: Move this to a standalone function in utils

modules/core/src/lib/deck.ts
```diff
@@ -736,11 +737,29 @@ export default class Deck {
       parent.appendChild(canvas);
     }
 
-    Object.assign(canvas.style, props.style);
+    // Set css variables and properties
+    Object.entries(props.style).map(([key, value]) =>
```
Pessimistress, line 741: Why not use a for loop?

chrisgervang:
I don't think this approach will with since it applies styles to the deck canvas rather than the container. Closing in favor of a API change in discussion in #8160

PR #8138, opened 2023-09-26
Enable render tests

Pessimistress:
For #7457 

Re-enable render tests for core layers and simple composite layers. Still disabled: binary attribute test cases, GPU aggregation layers, GPU-based effects and extensions

#### Change List
- Upgrade to luma.gl `9.0.0-alpha.36`
- Update WebGL functions to new API name
- Enable some render tests
- Bug fix in constant attribute + shader attribute handling
- Bug fix in `PathStyleExtension`
- Bug fix in `TextBackgroundLayer`


ibgreen (approved):
Great progress. Let's land and make a plan for which of the commented tests we want to unbreak first.

PR #8167, opened 2023-10-04
CARTO: Initial implementation of v9 API

felixpalmer:

Squashed and cherry-picked from 8.9 branch: https://github.com/visgl/deck.gl/pull/8130

#### Change List
- Implementation of Source classes for all types of data sources
- Rename `_CartoTileLayer` -> `VectorTileLayer` & export previously private Layers
- Automatic injection of access token in Layers
- Test app to show usage with all data types
- Types for data sources and return values


padawannn (approved):
LGFM

PR #8168, opened 2023-10-04
CARTO: Support v3.3 spatialDataType & spatialDataColumn

felixpalmer:

#### Change List
- Support v3.3 spatialDataType & spatialDataColumn


zbigg (approved):

modules/carto/src/sources/common.ts
```diff
@@ -39,6 +39,8 @@ export type CartoTilesetSourceOptions = {
   tableName: string;
 };
 
+export type SpatialDataType = 'geometry' | 'h3' | 'quadbin';
```
alasarr, line 42: Should be Geo.

Issue #8125, opened 2023-09-21
[Bug] MaskExtension show interactivity in features filtered


aaranadev:
### Description

That mask extension isn't working properly, the features are filtered but the interactivity is enable, the hover and click are available.

The example is available in the doc official.

https://codepen.io/vis-gl/pen/ExbKoYg

### Flavors

- [ ] Script tag
- [ ] React
- [ ] Python/Jupyter notebook
- [ ] MapboxOverlay
- [ ] GoogleMapsOverlay
- [ ] CartoLayer
- [ ] ArcGIS

### Steps to Reproduce

- Add a new layer with Mask Extension
- Add a mask layer

### Environment

- Framework version: deck.gl@8.9.29
- Browser: Chorme 117
- OS: macOS 13

felixpalmer:
Fixed in 8.9.31

PR #8176, opened 2023-10-06
Include required mask layers when picking

felixpalmer:
For https://github.com/visgl/deck.gl/issues/8125

#### Change List
- Include required mask layers when rendering to picking buffer
- Test

### DIFF
--- a/modules/core/src/lib/deck-picker.ts
+++ b/modules/core/src/lib/deck-picker.ts
@@ -173,9 +173,17 @@ export default class DeckPicker {
     if (this._pickable === false) {
       return null;
     }
-    const pickableLayers = layers.filter(
+    let pickableLayers = layers.filter(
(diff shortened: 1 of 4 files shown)


modules/core/src/lib/deck-picker.ts
```diff
@@ -173,9 +173,17 @@ export default class DeckPicker {
     if (this._pickable === false) {
       return null;
     }
-    const pickableLayers = layers.filter(
+    let pickableLayers = layers.filter(
       layer => this.pickLayersPass.shouldDrawLayer(layer) && !layer.isComposite
     );
+
+    const requiredMasks = new Set(
```
Pessimistress, line 180: The core should not know about an extension.

Pessimistress (changes requested):

PR #8177, opened 2023-10-06
Carto: Support styling layer by color column

Josmorsot:
#### Background

Allow to style a layer using the a color property.

#### Change List
- support for `HexColor column` in `fetchMap`


modules/carto/src/api/layer-map.ts
```diff
@@ -341,15 +342,19 @@ export function getColorValueAccessor({name}, colorAggregation, data: any) {
...
   data: any
 ) {
-  const scale = SCALE_FUNCS[scaleType as any]();
+  let scale = SCALE_FUNCS[scaleType as any]();
   let domain: (string | number)[] = [];
   let scaleColor: string[] = [];
+  let accessorKeys = getAccessorKeys(name, aggregation);
 
-  if (Array.isArray(colorMap)) {
+  if (type === HEX_COLOR_COLUMN) {
```
felixpalmer, line 354: Rather than having the special case here, could you try adding [scaleIdentity](https://github.com/d3/d3-scale#scaleIdentity) to `SCALE_FUNCS` and then using that?

felixpalmer (changes requested):

modules/carto/src/api/layer-map.ts
Josmorsot: Done here: [c53c724](https://github.com/visgl/deck.gl/pull/8177/commits/c53c724c4d87c479746af854840c63874061cc12)

test/apps/carto-map/app.js
```diff
@@ -74,7 +74,8 @@ const examples = [
   '12119dd2-0ddb-4fd2-b48a-15a1b7511e52', // Overlapping empty circles
   '3c892452-3806-4ebf-821b-a76f4562dd0c', // points and lines
   '21cc8261-e626-4778-a78b-76fe8b808214', // markers tilesets
-  '4e8f215f-97b4-4f4e-8b53-465c3908c317' // markers points
+  '4e8f215f-97b4-4f4e-8b53-465c3908c317', // markers points
+  '27de26b4-b94f-4e94-b291-41d1a21d3d02' // HexColumn color
```
VictorVelarde, line 78: In which account is this mapId predefined for testing it later?

Josmorsot: frontend-team

modules/carto/src/api/layer-map.ts
```diff
@@ -380,6 +372,33 @@ export function getColorAccessor(
   return normalizeAccessor(accessor, data);
 }
 
+export function getDomainAndScaleColor(name, scaleType, range, data) {
```
felixpalmer, line 375: Why do you need to introduce this new function? I feel it just makes the code harder to follow

modules/carto/src/api/layer-map.ts
```diff
@@ -339,29 +340,20 @@ export function getColorValueAccessor({name}, colorAggregation, data: any) {
 }
 
 export function getColorAccessor(
-  {name},
+  field: VisualChannelField,
```
felixpalmer, line 343: Just destructure `colorColumn` rather than referencing `field`

felixpalmer (changes requested):

modules/carto/src/api/layer-map.ts
Josmorsot: I don't need but I think that it's more clear than having a long function. This function is just calculating the `domain` and `scaleColor` for each case.

felixpalmer: OK, but can your rename to `calculateLayerScale`, remove the `export` and put these three lines in the helper function:

```
scale.domain(domain);
scale.range(scaleColor);
scale.unknown(UNKNOWN_COLOR);
```

and then just return the `scale`

felixpalmer (changes requested):

felixpalmer (approved):

PR #8178, opened 2023-10-06
[chore] Fix widget module publishing

Pessimistress:
For #7946 #8182

#### Change List
- Fix babel config syntax
- Bump ocular-dev-tools to include https://github.com/uber-web/ocular/pull/435 for publishing the widgets module


chrisgervang (approved):
I thought babel's transform-jsx plugin was sufficient for widgets. Was there some part of the publishing script that needed the full react preset?

Issue #8182, opened 2023-10-07
@deck.gl/widgets v8.10.0-alpha.4 use React.createElement instead of the preact jsxFactory h


amarant:
### Description

Hi,
I was trying the @deck.gl/widgets v8.10.0-alpha.4 and it crashes because the translation from JSX to JS use `React.createElement` instead of the preact version `h`. I don't know why it didn't compile with it because the tsconfig seems to indicate the preact jsxFactory.

notify @chrisgervang who made the PR #8024 

### Flavors

- [ ] Script tag
- [X] React
- [ ] Python/Jupyter notebook
- [ ] MapboxOverlay
- [ ] GoogleMapsOverlay
- [ ] CartoLayer
- [ ] ArcGIS

### Steps to Reproduce

create a simple deck map using :

```
<DeckGL  widgets={[new FullscreenWidget({ id: 'test' })]} />
```

### Environment

- Framework version: v8.10.0-alpha.4
- Browser: Chrome
- OS:

### Logs

```
deck: React is not defined undefined
```

chrisgervang:
How do you install deck.gl in your environment? I'm not expecting our npm bundles to include JSX. Are you building from source?

amarant:
I use the npm bundle, and the esm or es5 distributions don't use jsx, but it uses `React.createElement` instead of the preact function `h`, just look at [Npm in @deck.gl/widgets/dist/esm/fullscreen-widget.js ](https://www.npmjs.com/package/@deck.gl/widgets/v/8.10.0-alpha.4?activeTab=code).

![Screenshot_20231008-110034](https://github.com/visgl/deck.gl/assets/2281471/5ef90096-91e9-4d50-b7fb-b4282794fdd9)

chrisgervang:
Just published `8.10.0-alpha.5` with a fix. Please give it a try.
<img width="430" alt="Screen Shot 2023-10-17 at 4 08 36 PM" src="https://github.com/visgl/deck.gl/assets/2461547/d5e4544e-776d-4456-9ca8-b6747456af0b">

amarant:
Thanks, this part works now, but there is an [error in the stylesheet](https://github.com/visgl/deck.gl/blob/69743353d537ce9e5b797e6cb18d597f1ca159cb/modules/widgets/src/stylesheet.css#L72).

chrisgervang:
Released 8.10.0-alpha.6 with that fixed.

amarant:
This is all fixed now, thanks.

PR #8202, opened 2023-10-13
chore(tests) bump puppeteer version

chrisgervang:
For #8182 

#### Background
`puppeteer` [maps its version to a chromium version](https://pptr.dev/chromium-support). ~~Currently our version is locked to 10.2, which in turn uses [Chromium 93.0.4577.0](https://chromium.googlesource.com/chromium/src.git/+/refs/tags/93.0.4577.0) from mid 2021.~~ Edit: See @Pessimistress's comment, `yarn bootstrap` installs the latest puppeteer and Chrome globally. 

`@proble.gl/test-utils` does not specify which version of puppeteer to use, so I'd like to add a resolution for a consistent test setup across environments.

#### Change List
- update `puppeteer` and add resolution
- ~~in turn, updates to [Chrome for Testing](https://docs.google.com/document/d/1XJvxyqAQjhPfJ0rX84PjfXXb5sBx3m8DXzMxR0ipQNs/edit#heading=h.xzptrog8pyxf) 117.0.5938.149~~ Edit: global `puppeteer` already did this.
- bump ocular to use new headless chrome mode.

#### Testing
- Confirmed browser tests pass on `master` and `8.10-release` when render tests are disabled.


Pessimistress:
The dev setup does not download a local copy of chromium. There is a global version of puppeteer installed with `yarn global add puppeteer` which installs the latest version (at least in CI):

https://github.com/uber-web/ocular/blob/master/modules/dev-tools/scripts/bootstrap.sh#L7

During test, `PUPPETEER_EXECUTABLE_PATH` is pointed to this installation:

https://github.com/uber-web/ocular/blob/master/modules/dev-tools/scripts/test.sh#L23

So our tests have always been using the latest version of chromium, but controlled with an old version of puppeteer. Otherwise you wouldn't see any warning messages.

chrisgervang:
Ah, thanks for writing this up. The environment is modified by a number locations. I'd prefer to use `devDependecies` for puppeteer instead of a global. But if there's a reason it is the way it is, I could also add a "resolutions_note" to package.json to explain the global location.

Pessimistress:
It is set up like this because all of the vis.gl repos use puppeteer for testing. Use a global installation outside these repos:

- avoids downloading hundreds of MBs of dependencies for every repo
- makes sure we always run against the latest chromium in CI without manual upgrade

Pessimistress (approved):

PR #8214, opened 2023-10-20
CARTO: Enhancements to v9 API

felixpalmer:

#### Change List
- [Implement request cache](https://github.com/visgl/deck.gl/commit/1e48c5aab4156a02b7009b7ffd122cb054103af6)
- [Remove carto function prefix](https://github.com/visgl/deck.gl/commit/2d9b55e5d2347ac73c912412e70c6daee1e6a8e2)
- [Update demo app & remove useMemo](https://github.com/visgl/deck.gl/commit/716d20c9b406c6e00b415a493acd86a26a464627)


modules/carto/src/sources/utils.ts
```diff
@@ -3,6 +3,7 @@ import {encodeParameter, MapType} from '../api/maps-api-common';
 import {DEFAULT_HEADERS, DEFAULT_PARAMETERS, MAX_GET_LENGTH} from './common';
 import {buildMapsUrlFromBase} from '../config';
 
+const REQUEST_CACHE = new Map();
```
padawannn, line 6: I'm a bit worried about this cache. For example, in Builder when we want to refresh data we make a new request with `Cache-Control: no-cache` and with this REQUEST_CACHE It wouldn't work right?

padawannn (changes requested):

modules/carto/src/sources/utils.ts
padawannn: I mean, we could refresh data multiple times with `Cache-Control: no-cache` so it always would be the same request

test/apps/carto-dynamic-tile/app.tsx
```diff
@@ -69,17 +69,14 @@ function Root() {
 function useH3Layer(datasource) {
   const {getFillColor, source, aggregationExp, columns, spatialDataColumn, sqlQuery, tableName} =
     datasource;
-  // useMemo to avoid a map instantiation on every re-render
```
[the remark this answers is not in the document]
felixpalmer: The whole idea is that now calling this function repeatedly will not hit the server, but instead just return the tileJSON that was already fetched. The worry was that many users would forget to implement the `useMemo` pattern (or similar in vanilla JS) and would fetch the data on every frame

modules/carto/src/sources/utils.ts
felixpalmer: Currently the way `CartoLayer` works [is the same](https://github.com/visgl/deck.gl/blob/master/modules/carto/src/layers/carto-layer.ts#L231-L241). If you have a `CartoLayer` and don't change any props it won't refetch the data. 

I think it is reasonable that the client doesn't issue another request for exactly the same URL&headers, if you need to force a refresh I would suggest that in Builder you change one of the headers, [If-Modified-Since](https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching#if-modified-since) seems appropriate in this case

padawannn: OK, in any case, it going to be a breaking change in Builder and we need to validate with infra the [If-Modified-Since](https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching#if-modified-since) approach

padawannn (approved):

PR #8222, opened 2023-10-23
Use node 18

chrisgervang:
#### Background
Node 18.18.2 is now LTS. Discussion started in #8220 

#### Change List
- Use node 18 in github actions and volta


ibgreen (approved):

birkskyum:
Node is moving pretty fast - v20 is the active LTS now.
Screenshot 2023-10-24 at 23 37 02

ibgreen:
yeah. A bit too fast sometimes :) not all npm modules are able to catch up so while being on the latest sounds appealing one can run into some unwelcome surprises. 

As an example we are just able to get loaders.gl upgraded to 18&20 now in the 4.0 release. That was a **lot** of work. hopefully it will be much less going forward.

PR #8242, opened 2023-10-31
[website] Fix ghcn website example

felixpalmer:
For https://github.com/visgl/deck.gl/issues/8193

#### Change List
- Convert example to Typescript
- Use `MeshAttributes` for `borderMesh`
- Fix `wireframe` rendering in `SimpleMeshLayer`


ibgreen (approved):
Nit: What on earth is `ghcn`? Do we have to have such a cryptic name? We are pretty strict with asking people to type out variable names - should we not apply at least a minimal similar standard for directories?

felixpalmer:
@ibgreen it refers to the dataset, `Global Historical Climatology Network`. I'd be in favour of renaming `examples/website
/ghcn` > `examples/website/orthographic-view` as it more useful to see what the example demonstrates rather than what it contains

PR #8268, opened 2023-11-08
Use getStride() in bufferLayoutEqual() comparison

felixpalmer:
For https://github.com/visgl/deck.gl/issues/8219

#### Background

`ColumnLayer` was not rendering correctly, a regression first introduced in https://github.com/visgl/deck.gl/pull/8153 by adding a [layoutChanged()](https://github.com/visgl/deck.gl/pull/8153/files#diff-6865938bd9e33227cb4501cefc20abe93abb4098905de00a0653cd1a80750544R641)
 check to the core `Layer` class. In the case of the `ColumnLayer` this was evaluating to `true` becuase the underlying comparison in [bufferLayoutEqual()](https://github.com/visgl/deck.gl/pull/8153/files#diff-11742470930e1b6ff2ef1c13c6102c1a9d18f0cbd76ef0bdb1655b68573b8b55R69) was detecting a change in the `instancePickingColors`, specifically the `stride` prop missing in the old accessor and set to `4` in the new accessor.

`Data-Column` uses the `getStride()` function to calculate the stride when it is missing and this evaulates to 4 for the old accessor also, thus the update triggered by `bufferLayoutEqual()` returning true is unnecessary. 

This fix passes the `getStride()` function into `bufferLayoutEqual()` to avoid such false positives. While this fixes the `ColumnLayer`, it is somewhat a happy accident, and I'm still not clear on why the layer doesn't render without this change

#### Change List
- Use `getStride()` in `bufferLayoutEqual()`

### DIFF
--- a/modules/core/src/lib/attribute/gl-utils.ts
+++ b/modules/core/src/lib/attribute/gl-utils.ts
@@ -62,11 +62,15 @@ export function getBufferAttributeLayout(
   };
 }
 
+export function getStride(accessor: DataColumnSettings<any>): number {
+  return accessor.stride || accessor.size * accessor.bytesPerElement;
+}
+
 export function bufferLayoutEqual(accessor1: BufferAccessor, accessor2: BufferAccessor) {
   return (
     (accessor1.type ?? GL.FLOAT) === (accessor2.type ?? GL.FLOAT) &&
     accessor1.size === accessor2.size &&
-    accessor1.stride === accessor2.stride &&
+    getStride(accessor1) === getStride(accessor2) &&
     (accessor1.offset || 0) === (accessor2.offset || 0)
   );
 }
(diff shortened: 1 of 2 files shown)


modules/core/src/lib/attribute/gl-utils.ts
```diff
@@ -62,11 +62,15 @@ export function getBufferAttributeLayout(
...
+export function getStride(accessor: DataColumnSettings<any>): number {
+  return accessor.stride || accessor.size * accessor.bytesPerElement;
+}
+
 export function bufferLayoutEqual(accessor1: BufferAccessor, accessor2: BufferAccessor) {
   return (
     (accessor1.type ?? GL.FLOAT) === (accessor2.type ?? GL.FLOAT) &&
     accessor1.size === accessor2.size &&
-    accessor1.stride === accessor2.stride &&
+    getStride(accessor1) === getStride(accessor2) &&
```
Pessimistress, line 73: This does not look correct to me. `BufferAccessor` does not contain a `bytesPerElement`. The right solution could be always passing `Required<BufferAccessor>` into this function.

Pessimistress (changes requested):

modules/core/src/lib/attribute/gl-utils.ts
felixpalmer: Updated types to match what is being used in `attribute/attribute.ts`

Pessimistress (approved):

PR #8272, opened 2023-11-09
Use getCanvas() in GoogleMapsOverlay

felixpalmer:
Followup to https://github.com/visgl/deck.gl/pull/7919

#### Change List
- Also use `getCanvas()` in `GoogleMapsOverlay`
- Defensive checks for when canvas isn't available


akre54 (approved):
Nice!

chrisgervang (approved):

CptCompiler:
@chrisgervang thank you for all your great work. 
I just had a little issue with my packages. 
I'm not an expert, but shouldn't the peerDependency in @deck.gl/google-maps be updated to this new version? (8.9.33)
My deck.gl core package was still version 8.9.25 and I got the error: this._deck.getCanvas is not a function

thank you!

Issue #8260, opened 2023-11-06
Tile3DLayer `this.state` throws undefined during deconstruction


chrisgervang:
https://github.com/visgl/deck.gl/blob/148b3b04a065a70f4d92a079b095e2194b979fda/modules/geo-layers/src/tile-3d-layer/tile-3d-layer.ts#L107-L110

`this.state` isn't defined at all times in a 3d tile layer's lifecycle, and so deconstruction throws when that is the case.

I've encountered this while calling `layer.isLoaded` in a maplibre `onRender` callback while deck is interleaved.

Other layers, like `MVTLayer`, implement `isLoaded` to protect for this, and in my testing this approach appears to fix the issue. 
https://github.com/visgl/deck.gl/blob/3e6d295630af2672c5e0792c6ae7595e6e1822ae/modules/geo-layers/src/mvt-layer/mvt-layer.ts#L127

Could we do something similar?
```js
return Boolean(this.state && this.state.tileset3d && this.state.tileset3d.isLoaded() && super.isLoaded);
```

PR #8279, opened 2023-11-12
Imorove Tile3DLayer.isLoaded

chrisgervang:
For #8260

#### Change List
- Add protections to Tile3DLayer.isLoaded

### DIFF
--- a/modules/geo-layers/src/tile-3d-layer/tile-3d-layer.ts
+++ b/modules/geo-layers/src/tile-3d-layer/tile-3d-layer.ts
@@ -105,8 +105,12 @@ export default class Tile3DLayer<DataT = any, ExtraPropsT extends {} = {}> exten
   }
 
   get isLoaded(): boolean {
-    const {tileset3d} = this.state;
-    return tileset3d !== null && tileset3d.isLoaded();
+    return Boolean(
+      this.state &&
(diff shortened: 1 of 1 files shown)


chrisgervang:
Should I open a second PR for v9 since that file is "disabled"?

modules/geo-layers/src/tile-3d-layer/tile-3d-layer.ts
```diff
@@ -105,8 +105,12 @@ export default class Tile3DLayer<DataT = any, ExtraPropsT extends {} = {}> exten
   }
 
   get isLoaded(): boolean {
-    const {tileset3d} = this.state;
-    return tileset3d !== null && tileset3d.isLoaded();
+    return Boolean(
```
felixpalmer, line 108:
```suggestion
    return Boolean(this.state?.tileset3d?.isLoaded() && super.isLoaded)
```

Like we do in `TileLayer`

felixpalmer (changes requested):

felixpalmer:
> Should I open a second PR for v9 since that file is "disabled"?

That would be great

chrisgervang:
@felixpalmer good to go?

felixpalmer (approved):

PR #8281, opened 2023-11-14
[Test]: GeoJsonLayer applies polygon tessellation on binary geometry format

lixun910:
Test: if GeoJsonLayer can apply polygon tessellation on binary geometry format using `polygonIndices` and `primitivePolygonIndices`, and the function`triangulate()` in module/carto

### DIFF
--- a/modules/carto/src/index.ts
+++ b/modules/carto/src/index.ts
@@ -5,6 +5,7 @@ export {default as H3TileLayer} from './layers/h3-tile-layer';
 export {default as _PointLabelLayer} from './layers/point-label-layer';
 export {default as QuadbinTileLayer} from './layers/quadbin-tile-layer';
 export {default as RasterTileLayer} from './layers/raster-tile-layer';
+export {triangulate} from './layers/schema/carto-vector-tile-loader';
 export {default as BASEMAP} from './basemap';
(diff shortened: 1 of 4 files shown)


lixun910:
For issue: https://github.com/visgl/loaders.gl/pull/2785. What do you think about this kind of fix, @ibgreen? so loaders.gl/arrow will no need to do earcut polygons. Thanks!

felixpalmer:
@lixun910 this doesn't make sense, `GeoJsonLayer` already uses `earcut` for triangulation, as does the `triangulate()` function

lixun910:
Thanks for your reply, @felixpalmer! I can see that `GeoJsonLayer` uses earcut for triangulation on GeoJsonFeature, but not for the binary geometries (polygons) with `polygonIndices` and `primitivePolygonIndicies` (see createLayerPropsFromBinary() in geojson-layer.ts). We are using the GeoJsonLayer with binary geometries to render GeoArrow format. Maybe I missed something, can you point to me how the triangulate() function was applied to the binary geometries? Thanks again!

felixpalmer:
`SolidPolygonLayer` performs triangulation on the data in `PolygonTesselator._updateIndices()`, is this not working for you with binary data?

PR #8269, opened 2023-11-08
CARTO: Add example for boundaries

felixpalmer:
For https://github.com/visgl/deck.gl/issues/8148

#### Change List

- Add `CartoPropertiesTileLoader` to load binary property tile data
- Enhance `VectorTileLayer` to support fetching & merging of property data
- Add `cartoBoundaryQuerySource` & `cartoBoundaryTableSource`
- Update test app with query & table boundary examples

PR #8286, opened 2023-11-15
CARTO: Boundaries handle empty 204 responses

felixpalmer:
Followup to https://github.com/visgl/deck.gl/pull/8269

#### Change List
- Handle empty 204 responses
- Log instead of throw on duplicate geoid

### DIFF
--- a/modules/carto/src/layers/utils.ts
+++ b/modules/carto/src/layers/utils.ts
@@ -1,3 +1,4 @@
+import {log} from '@deck.gl/core';
 import {Tile as PropertiesTile} from './schema/carto-properties-tile';
 import {Tile as VectorTile} from './schema/carto-tile';
 import {_deepEqual as deepEqual} from '@deck.gl/core';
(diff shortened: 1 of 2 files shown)


alasarr (approved):

modules/carto/src/layers/utils.ts
```diff
@@ -19,9 +20,10 @@ export function mergeBoundaryData(geometry: VectorTile, properties: PropertiesTi
   const mapping = {};
   for (const {geoid, ...rest} of properties.properties) {
     if (geoid in mapping) {
-      throw new Error(`Duplicate geoid key in mapping: ${geoid}`);
+      log.warn(`Duplicate geoid key in boundary mapping, using first occurance`)();
```
alasarr, line 23: Could you log the geoid too?

alasarr, line 23:
```suggestion
      log.warn(`Duplicate geoid [${geoid}] key in boundary mapping, using first occurance`)();
```

felixpalmer: I'm deliberately not doing that as it will then potentially log a huge number of messages, but I can add it if you think that is OK

alasarr: You're right

PR #8287, opened 2023-11-15
Use canvas rather than htmlCanvas in deck.spec

felixpalmer:
#### Background

`yarn test` fails when run locally due to `htmlCanvas` not being defined

#### Change List
- [Use canvas rather than htmlCanvas in deck.spec](https://github.com/visgl/deck.gl/commit/973de77c1873546c08135ca2f76a0cc22846cad6)


felixpalmer:
Yes it is, but it may be null if `canvas` is set to an offscreen canvas: https://github.com/visgl/luma.gl/blob/master/modules/core/src/adapter/canvas-context.ts#L111-L119

Issue #8266, opened 2023-11-08
[Bug] OrbitView example broken in v9


felixpalmer:
### Description

The [OrbitView](https://felixpalmer.github.io/deck.gl/examples/plot) example is broken in v9. 

This issue is not so much down to a problem with `OrbitView` itself, but that the example uses 3 custom layers which have yet to be ported to the v9 API. These layers will need to be converted to Typescript and fixed following the approach used in (e.g.) the core layers

### Flavors

- [ ] Script tag
- [ ] React
- [ ] Python/Jupyter notebook
- [ ] MapboxOverlay
- [ ] GoogleMapsOverlay
- [ ] CartoLayer
- [ ] ArcGIS

### Steps to Reproduce

https://felixpalmer.github.io/deck.gl/examples/plot or run `examples/website/plot` on `master` branch

### Environment

- Framework version: 9
- Browser:
- OS:

donmccurdy:
@felixpalmer - Do you (or others) have a preference about converting the entire example to TypeScript, or just the `-layer.js` files, or something else? I do find the type-checking helpful, just want to make sure the example is still approachable. 

Enabling `"allowJs": true` mode in `tsconfig.json` might be an alternative if we want the examples to use plain JavaScript, but to still have compile-time errors about incorrect usage.

felixpalmer:
I think it makes sense to convert the layer files, as they are complex and would benefit from the type checking. Just switching to Typescript should help the migration to v9 as many errors will be flagged

donmccurdy:
Sounds good to me -- opened https://github.com/visgl/deck.gl/pull/8289.

PR #8289, opened 2023-11-15
[website] Migrate 'plot' example to deck.gl v9

donmccurdy:
- Fixes https://github.com/visgl/deck.gl/issues/8266

Updates `examples/website/plot` to deck.gl v9. The migration was easier done with type-checking, and as discussed in #8266 I've left the layer files in TypeScript. I'm also happy to revert the types if we prefer that examples be plain JavaScript.

There were a few things in the PR I wasn't sure about, and I'll add comments inline to call those out.


examples/website/plot/plot-layer/axes-layer.ts
```diff
@@ -222,29 +236,29 @@ export default class AxesLayer extends Layer {
...
         attributes: {
-          positions: new Float32Array(gridPositions),
-          normals: new Float32Array(gridNormals)
+          positions: {value: new Float32Array(gridPositions), size: 3},
+          normals: {value: new Float32Array(gridNormals), size: 3}
         },
         vertexCount: gridPositions.length / 3
-      }),
-      isInstanced: true
+      })
```
donmccurdy, line 251: This works, and types say `isInstanced` no longer exists, but is some indication of instancing expected at this point? `state.instanceCount` is not readily available but could be derived if necessary.

examples/website/plot/plot-layer/axes-layer.ts
```diff
@@ -222,29 +236,29 @@ export default class AxesLayer extends Layer {
       0, -1, 0, 0, -1, 0
     ];
 
-    const grids = new Model(gl, {
+    const grids = new Model(gl.device, {
       id: `${this.props.id}-grids`,
       vs: gridVertex,
       fs: fragmentShader,
+      bufferLayout: this.getAttributeManager().getBufferLayouts(),
```
donmccurdy, line 243: Note: When `bufferLayout` is omitted, type checking still passes and the layer silently displays nothing. To make updates to v9 a bit easier, I wonder if the types could be more strict about this?

examples/website/plot/plot-layer/axes-layer.ts
```diff
@@ -271,20 +285,20 @@ export default class AxesLayer extends Layer {
...
       geometry: new Geometry({
         topology: 'triangle-list',
         attributes: {
           indices: new Uint16Array(labelIndices),
-          positions: new Float32Array(labelPositions),
-          texCoords: {size: 2, value: new Float32Array(labelTexCoords)},
-          normals: new Float32Array(labelNormals)
+          positions: {value: new Float32Array(labelPositions), size: 3},
+          texCoords: {value: new Float32Array(labelTexCoords), size: 2},
+          normals: {value: new Float32Array(labelNormals), size: 3}
```
donmccurdy, line 299: Note: Size now appears to be required – if that isn't an intentional change in v9, I can look into the cause. Requiring it seems fine to me, too.

examples/website/plot/plot-layer/surface-layer.ts
```diff
@@ -109,20 +126,20 @@ export default class SurfaceLayer extends Layer {
...
+    return [(xIndex / (uCount - 1)) * 255, (yIndex / (vCount - 1)) * 255, 1] as $TODO;
   }
 
-  decodePickingColor([r, g, b]) {
+  decodePickingColor([r, g, b]: Color): number {
     if (b === 0) {
       return -1;
     }
-    return [r / 255, g / 255];
+    return [r / 255, g / 255] as $TODO;
```
donmccurdy, line 142: Probably not a huge deal, but TypeScript complains that these types defer from the base class.

examples/website/plot/plot-layer/surface-layer.ts
```diff
@@ -56,48 +83,38 @@ export default class SurfaceLayer extends Layer {
...
 
-  getModel(gl) {
+  getModel() {
     // 3d surface
-    return new Model(gl, {
+    return new Model(this.context.device as $TODO, {
       id: `${this.props.id}-surface`,
       vs: surfaceVertex,
       fs: fragmentShader,
       modules: [picking],
```
donmccurdy, line 106: FYI, I see three warnings in the JS console:

```
luma.gl: Model(PlotLayer-surface-surface): Ignoring buffer "instancePickingColors" for unknown attribute "instancePickingColors"

luma.gl: Model(PlotLayer-axes-grids): Ignoring buffer "instancePickingColors" for unknown attribute "instancePickingColors"

luma.gl: Model(PlotLayer-axes-labels): Ignoring buffer "instancePickingColors" for unknown attribute "instancePickingColors"
```

It doesn't look like they're directly caused by changes in the PR, and I'm not sure what creates the buffer, so I've left these warnings alone for now.

examples/website/plot/tsconfig.json
```diff
@@ -0,0 +1,5 @@
+{
+    "extends": "../../../tsconfig.json",
+    "include": ["./**/*.ts"],
+    "compilerOptions": { "strict": false }
```
donmccurdy, line 4: I opted for `strict: false` here because it makes incremental conversion easier. There were a couple tricky type issues, possibly specific to the 'examples' environment, which didn't seem urgent to deal with quite yet. For example, this line in surface-layer...

```js
return new Model(this.context.device, { ... });
```

... gives ...

```
Argument of type 'import("/Users/donmccurdy/Projects/deck.gl/node_modules/@luma.gl/core/dist/adapter/device").Device' is not assignable to parameter of type 'import("/Users/donmccurdy/Projects/deck.gl/examples/website/plot/node_modules/@luma.gl/core/dist/adapter/device").Device'.
  Property 'getCanvasContext' is missing in type 'import("/Users/donmccurdy/Projects/deck.gl/node_modules/@luma.gl/core/dist/adapter/device").Device' but required in type 'import("/Users/donmccurdy/Projects/deck.gl/examples/website/plot/node_modules/@luma.gl/core/dist/adapter/device").Device'.ts(2345)
device.d.ts(133, 5): 'getCanvasContext' is declared here.
```

With this combination of `strict: false` and some `as $TODO`, the type checker is happy with the PR.

examples/website/plot/plot-layer/axes-layer.ts
```diff
@@ -169,13 +179,17 @@ export default class AxesLayer extends Layer {
...
-        .setUniforms(Object.assign({}, uniforms, baseUniforms, labelTexture))
-        .draw();
+      modelsByName.grids.setUniforms(Object.assign({}, uniforms, baseUniforms));
+      modelsByName.labels.setBindings({labelTexture});
+      modelsByName.labels.setUniforms(
+        Object.assign({}, uniforms, baseUniforms, labelTextureUniforms)
+      );
+
+      modelsByName.grids.draw(this.context.renderPass as $TODO);
+      modelsByName.labels.draw(this.context.renderPass as $TODO);
```
donmccurdy, line 192: `renderPass.id` is increasing each time we draw — is that expected, or could it indicate I've done something incorrectly here?

examples/website/plot/plot-layer/axes-layer.ts
felixpalmer: luma will [autocreate the buffer layout](https://github.com/visgl/luma.gl/blob/master/modules/core/src/adapter/resources/render-pipeline.ts#L31-L34) for any attributes passed (so here just what is defined in the `geometry`, so I can see the logic in `bufferLayout` being optional. However, this pattern of adding more attributes via `bufferLayout` seems to be very common (at least in deck) and so maybe it would be good to make the type required, so that the user has to explicitly pass in `[]` to signal no additional buffers. @ibgreen thoughts?

In terms of failing silently, luma does complain if it cannot match a buffer, e.g.:

```
Model(PlotLayer-surface-surface): Ignoring buffer "instancePickingColors" for unknown attribute "instancePickingColors"
```

examples/website/plot/plot-layer/axes-layer.ts
felixpalmer: That is intentional, due to WebGPU being more strict about attribute dimensions (e.g. you can't pass 2D array for a `vec3 attribute`)

examples/website/plot/plot-layer/axes-layer.ts
felixpalmer: This is happening elsewhere, e.g. in `ScatterplotLayer` and I think it is by design. Could you switch to the `8.9-release` branch and check the behavior there?

examples/website/plot/plot-layer/surface-layer.ts
```diff
@@ -1,24 +1,42 @@
...
   vCount: 100,
   lightStrength: 0.1
 };
 
+/** All props supported by SurfaceLayer. */
+export type SurfaceLayerProps<DataT = unknown> = _SurfaceLayerProps<DataT> & LayerProps;
+
+type _SurfaceLayerProps<DataT> = {
+  data: LayerDataSource<DataT>;
+  getPosition?: $TODO;
```
felixpalmer, line 30: On my branch, I've typed some of these already: https://github.com/visgl/deck.gl/blob/felix/surface-layer-v9/examples/website/plot/plot-layer/surface-layer.ts#L29-L30

examples/website/plot/plot-layer/surface-layer.ts
felixpalmer: I think we can ignore these for this PR

examples/website/plot/plot-layer/surface-layer.ts
felixpalmer: I would remove this method entirely, and rely on the `Layer` implementation. Then, modify `getPickingInfo` to reconstruct `[r, g, b]` from the `index`

examples/website/plot/tsconfig.json
felixpalmer: Try and resolve the remaining issues so we don't need this. I assume you're running the example using `yarn start` in the example directory. Also test running the website (`cd deck.gl/website && yarn && yarn start`). It is currently failing

examples/website/plot/plot-layer/axes-layer.ts
felixpalmer: @ibgreen is this safe to remove? I see we still have it in `ScatterplotLayer` and other layers

felixpalmer (changes requested):

examples/website/plot/plot-layer/surface-layer.ts
donmccurdy: The issue I ran into here was that the types from the branch require `// @ts-ignore` when using the accessors, and I slightly preferred a `$TODO` at the top over `// @ts-ignore` each time they're used.

I'll take another look and see if I can figure out the correct strict types, though.

examples/website/plot/plot-layer/axes-layer.ts
```diff
@@ -104,15 +116,7 @@ export default class AxesLayer extends Layer {
...
-    this.setState(
-      Object.assign(
-        {
-          numInstances: 0,
-          labels: null
-        },
-        this._getModels(gl)
-      )
-    );
+    this.setState(Object.assign({instanceCount: 0}, this._getModels(gl)));
```
donmccurdy, line 119: I'll rename this back to `numInstances` - hadn't realized that `Layer.ts` relies on this naming.

donmccurdy:
Moving some inline conversations into main thread, as they've gotten detached in a rebase.

> Try and resolve the remaining issues so we don't need this. I assume you're running the example using yarn start in the example directory. 

I've added additional type checks so that `strict: false` is no longer necessary.

As far as I can tell we do need a `tsconfig.json` file — without it TypeScript raises many errors that seem to indicate it's type-checking against the wrong version, or multiple versions, of deck.gl and its dependencies.

> Also test running the website (cd deck.gl/website && yarn && yarn start). It is currently failing.

Fixed by adding `allowDeclareFields` to `deck.gl/website/babel.config.js`, so that we can extend the base type declaration of `Layer#state`.


> I would remove this method entirely, and rely on the Layer implementation...

I agree this would be better, but the direct conversion is not working as I would've expected. Perhaps I've missed something about why these methods were added. For now I've just fixed the types to match the base class so we don't need `$TODO` here, and left the overridden methods alone.

***

If there are other concerns with the types, perhaps I should remove TypeScript from the PR for now? It might be easier to deal with the TS issues in a PR that isn't also reworking the code for v9.

examples/website/plot/plot-layer/axes-layer.ts
donmccurdy: It doesn't look like `this.context.renderPass` exists in 8.9, not sure if there's some comparable thing to check for? 

Checking the ArcLayer example on `master` branch — the render pass ID increments in the same way. So, if this is an issue at all, it doesn't appear to be specific to this PR.

felixpalmer:
I think it is best to not modify the typescript config with `allowDeclareFields` in this PR. If using declare fields is a good solution, we should adopt it in the whole codebase when dealing with `state!`. If you think this is better than what we have now, could you create an PR which migrates all the layers to using declare so others can comment? It might be better to first create an Issue to save time though. For background, I found this [explanation useful](https://stackoverflow.com/questions/75563171/what-is-the-practical-use-of-type-only-field-declarations).

I can't figure out why your layer definition of state leads to an error when `ScatterplotLayer` and others seem to be doing the same thing and there is no error though.

I think it would be a shame to remove the TypeScript as you've put in the effort and the types are helpful. In order to land this PR:

- Remove the modifications to tsconfig
- Where possible replace $TODO with the correct type, where not use `any`
- Fix the issue with `state!`, using `ts-ignore` if you can't find a solution

examples/website/plot/plot-layer/surface-layer.ts
donmccurdy: Got this in a state that TS is happy with, without `// @ts-ignore` or `$TODO`. There might be a less verbose way to do it, but I couldn't find other examples, and probably best to just be verbose and strict in the meantime.

donmccurdy:
> - Remove the modifications to tsconfig
> - Where possible replace $TODO with the correct type, where not use any
> - Fix the issue with state!, using ts-ignore if you can't find a solution

✅ Done!

felixpalmer (approved):

PR #8290, opened 2023-11-15
Audit Layer#isLoaded implementations

chrisgervang:
#### Background
Building on top of #8260, I've audited many of the `isLoaded` functions for consistency and quality. Open to feedback or recommendations.

In our docs we say a layer is considered loaded if all asynchronous assets are loaded. Assets aren't formally defined as far as I'm aware, but I interpret them as anything required to load before the layer can be drawn.

I've reviewed each layer to ensure the following types of assets are settled:
- async props (ensure `super.isLoaded`)
- tiles in a tileset
- 3d parsing in mesh layers (Particularly useful for `Tile3DLayer` where `scenegraphs` are constantly loading) 

I've also looked for redundant checks and style inconsistencies (using chained optionals and casting to `Boolean` as needed).

#### Change List
- Wait for `scenegraph` to parse in `ScenegraphLayer`
- Wait for `model` to parse in `SimpleMeshLayer`
- Use `Tileset2D#isLoaded` and `super.isLoaded` on `TileLayer`
- Move cached `Layer#isLoaded` check into `Tileset2D#isLoaded`
- Removed redundant check in MVTLayer#isLoaded
- Optional chain in WMSLayer#isLoaded

### DIFF
--- a/modules/geo-layers/src/tile-layer/tile-layer.ts
+++ b/modules/geo-layers/src/tile-layer/tile-layer.ts
@@ -165,8 +165,12 @@ export default class TileLayer<DataT = any, ExtraPropsT extends {} = {}> extends
   }
 
   get isLoaded(): boolean {
-    return this.state?.tileset?.selectedTiles?.every(
-      tile => tile.isLoaded && tile.layers && tile.layers.every(layer => layer.isLoaded)
+    return Boolean(
+      this.state?.tileset?.isLoaded &&
+        this.state?.tileset?.selectedTiles?.every(tile =>
+          tile.layers?.every(layer => layer.isLoaded)
+        ) &&
+        super.isLoaded
     );
   }
 
--- a/modules/geo-layers/src/tile-layer/tile-layer.ts
+++ b/modules/geo-layers/src/tile-layer/tile-layer.ts
@@ -165,13 +165,7 @@ export default class TileLayer<DataT = any, ExtraPropsT extends {} = {}> extends
   }
 
   get isLoaded(): boolean {
-    return Boolean(
-      this.state?.tileset?.isLoaded &&
-        this.state?.tileset?.selectedTiles?.every(tile =>
-          tile.layers?.every(layer => layer.isLoaded)
-        ) &&
-        super.isLoaded
-    );
+    return Boolean(this.state?.tileset?.isLoaded && super.isLoaded);
   }
 
   shouldUpdateState({changeFlags}): boolean {
--- a/modules/geo-layers/src/mvt-layer/mvt-layer.ts
+++ b/modules/geo-layers/src/mvt-layer/mvt-layer.ts
@@ -114,7 +114,7 @@ export default class MVTLayer<ExtraProps extends {} = {}> extends TileLayer<
   }
 
   get isLoaded(): boolean {
-    return this.state && this.state.data && this.state.tileset && super.isLoaded;
+    return this.state?.data && super.isLoaded;
   }
 
   updateState({props, oldProps, context, changeFlags}: UpdateParameters<this>) {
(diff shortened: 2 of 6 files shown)


felixpalmer (approved):

chrisgervang:
Any idea why I'm getting this error on many tests?
Screen Shot 2023-11-16 at 3 36 54 PM

modules/geo-layers/src/terrain-layer/terrain-layer.ts
```diff
@@ -343,6 +343,10 @@ export default class TerrainLayer<ExtraPropsT extends {} = {}> extends Composite
       );
     }
 
+    if (!this.state.terrain) {
+      return null;
+    }
```
chrisgervang, line 348: The `TerrainLayer#Empty Props` test was hanging. Maybe because it was creating a `SimpleMeshLayer` with an undefined mesh, and that is considered "not loaded"? `terrain` state is actually optional, but the type assumed it was always be defined.. I'm not sure why we'd create a layer when there isn't a anything to show

felixpalmer: `isTiled` will only be true when we have a non-Array form of `elevationData`. When `elevationData` is not defined, we have `isTiled = false` but `this.state.terrain` as `null` (due to `this.loadTerrain()`. In this case it does make sense to not render any sublayers, but I would say it is clearer to instead add the following to the start of `renderLayers()` instead:

```js
if (!elevationData) {
  return null;
}
```

If this check isn't triggered we can be sure that we will be can render a valid layer(s)

felixpalmer (changes requested):

modules/geo-layers/src/terrain-layer/terrain-layer.ts
chrisgervang: Ok, I've pushed that up. I'm chasing some other failing tests related to tiles

modules/geo-layers/src/tile-layer/tile-layer.ts
```diff
@@ -165,9 +165,7 @@ export default class TileLayer<DataT = any, ExtraPropsT extends {} = {}> extends
   }
 
   get isLoaded(): boolean {
-    return this.state?.tileset?.selectedTiles?.every(
-      tile => tile.isLoaded && tile.layers && tile.layers.every(layer => layer.isLoaded)
```
chrisgervang, line 169: An MVTLayer test is failing in the picking module when layer cache checking is pushed into Tileset2D. 

For background, the TileLayer [caches sub layers](https://github.com/visgl/deck.gl/pull/8290/files#diff-8624e1d990265f2e01a5dce4cf739482733edb375fdd8a0df5c5c7df8c30f16fL342) rendered by each tile in the Tileset2D `tiles[...].tile.layers` property. 

Tileset2D#isLoaded exists but only checks if tile data is loaded, whereas TileLayer#isLoaded checks if these layers as well as if the tile data is loaded. Oddly, TileLayer doesn't use Tileset2D#isLoaded. This seemed like an odd choice for TileLayer - why doesn't TileLayer just use Tileset2D's isLoaded?

There's an edge case too. TileLayer is "loaded" if all layers rendered on a tile are present and loaded. I'm not sure what the design intention is for `tile.layers`. It starts null, then eventually it's defined and contains layers.

Alternatively, can we remove the Tileset2D#isLoaded method and just rely on the layer to implement it? I can't find usages outside of tests.

chrisgervang: I'm moving these changes into #8335 since they're blocking other fixes.

felixpalmer (approved):

ibgreen (approved):

PR #8293, opened 2023-11-16
Expose types at the package roots

alasarr:
In v9 we're going to expose types at package roots, so we can import as follows:

```
```

The old version `import {Deck} from '@deck.gl/core/typed'` is deprecated with this PR.


Pessimistress:
Add a line to upgrade guide?

Pessimistress (approved):

alasarr:
>Add a line to upgrade guide?

I will open another PR with this and modifying this page: https://deck.gl/docs/get-started/using-with-typescript

PR #8295, opened 2023-11-20
[chore] Bump peerDependencies to 9.0.0

felixpalmer:
#### Background

Installing the v9 alpha package currently gives warnings of the form:

```
warning " > @deck.gl/aggregation-layers@9.0.0-alpha.3" has incorrect peer dependency "@deck.gl/core@^8.0.0".
```

#### Change List
- Bump all v8 peerDependencies to v9

### DIFF
--- a/modules/aggregation-layers/package.json
+++ b/modules/aggregation-layers/package.json
@@ -39,8 +39,8 @@
     "d3-hexbin": "^0.2.1"
   },
   "peerDependencies": {
-    "@deck.gl/core": "^9.0.0",
-    "@deck.gl/layers": "^9.0.0",
+    "@deck.gl/core": "^9.0.0-alpha",
+    "@deck.gl/layers": "^9.0.0-alpha",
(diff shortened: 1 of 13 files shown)


alasarr (approved):

Pessimistress:
If you want to get rid of the warning, peer dep version should be `^9.0.0-alpha`?

modules/aggregation-layers/package.json
```diff
@@ -39,8 +39,8 @@
     "d3-hexbin": "^0.2.1"
   },
   "peerDependencies": {
-    "@deck.gl/core": "^8.0.0",
-    "@deck.gl/layers": "^8.0.0",
+    "@deck.gl/core": "^9.0.0",
```
alasarr, line 42: Why don't we link to alpha? `^9.0.0-alpha.'`

alasarr (changes requested):
I think if we don't like to alpha it won't work

alasarr (approved):

PR #8296, opened 2023-11-20
[CARTO] User boundaries support

alasarr:
With this PR, users will be able to create their own custom boundaries and join with their data. 

Users will provide a tilesetTable instead of a boundaryId. `boundaryTilesetSource` has been deprecated as a regular `vectorTilesetSource` can be used.


modules/carto/src/sources/boundary-query-source.ts
```diff
@@ -3,26 +3,26 @@ import {baseSource} from './base-source';
...
-  sqlQuery: string;
+  propertiesSqlQuery: string;
   queryParameters?: QueryParameters;
 };
 type UrlParameters = {
-  boundaryId: string;
+  tilesetTableName: string;
   matchingColumn: string;
-  sqlQuery: string;
+  sqlQuery: string; // TODO rename to propertiesSqlQuery
```
felixpalmer, line 14: Why is this TODO? You've changed the`boundaryId` to `tilesetTableName` so API changes are required anyway

felixpalmer (changes requested):

felixpalmer (approved):

PR #8301, opened 2023-11-21
CARTO: Don't render demo in docusaurus build

felixpalmer:
For https://github.com/visgl/deck.gl/issues/8193

#### Background

Use of `fetch` breaks the website build

#### Change List
- Only render demo in browser


donmccurdy (approved):
LGTM based on Docusaurus documentation for `<BrowserOnly/>`.

I'm missing a few steps in my local repo setup to be able to reproduce the original issue, but I'll follow up about that offline.

Issue #8148, opened 2023-09-29
CARTO v9 API tracker


felixpalmer:
Tracker for work needed to land the [new API](https://github.com/visgl/deck.gl/blob/master/dev-docs/RFCs/v8.0/carto-v9-rfc.md) for interacting with the CARTO backend.

### Tasks

- [x] RFC https://github.com/visgl/deck.gl/pull/8051
- [x] Initial implementation https://github.com/visgl/deck.gl/pull/8130 against 8.9
- [ ] Test & review internally
- [x] Fix binary data (to make vector layers work) https://github.com/visgl/deck.gl/pull/8153
- [x] Initial implementaion against `master` https://github.com/visgl/deck.gl/pull/8167
- [x] Update website example https://github.com/visgl/deck.gl/pull/8191
- [x] Update playground example https://github.com/visgl/deck.gl/pull/8192
- [ ] Update pydeck
- [x] Remove CartoLayer https://github.com/visgl/deck.gl/pull/8217
- [x] Cleanup maps-v3-api https://github.com/visgl/deck.gl/pull/8224
- [x] `fetchMap`support https://github.com/visgl/deck.gl/pull/8224
- [ ] Error checking to catch runtime errors (for JS users who won't get TS hints)
- [ ] Documentation https://github.com/visgl/deck.gl/pull/8228
- [ ] Fix tests

aaranadev:
About the RFC #8051, I have a doubt about _[React considerations](https://github.com/visgl/deck.gl/issues/8148#react-considerations)_. 

Now the access_token is required into new source functions, so the access_token need to add as dependency in `useMemo`, I would like to see how the workflow is affected if access_token changes or if you add it to the dependency array.

For instance: 
 - the access_token expires
 - the access_token is changed by the user

As far as I know, Maps API instantiation needs to be renewed.

cc: @menusal @VictorVelarde

felixpalmer:
> Now the access_token is required into new source functions, so the access_token need to add as dependency in `useMemo`, I would like to see how the workflow is affected if access_token changes or if you add it to the dependency array.

In the case of a changing access token, it would be included in the `useMemo` dependency Array, which as you say will then trigger a new instantiation

aaranadev:
> > Now the access_token is required into new source functions, so the access_token need to add as dependency in `useMemo`, I would like to see how the workflow is affected if access_token changes or if you add it to the dependency array.
> 
[...]

PR #8308, opened 2023-11-22
CARTO: Remove mapsUrl

felixpalmer:
For https://github.com/visgl/deck.gl/issues/8148

#### Change List
- Remove `mapsUrl` from options
- Consolidate all URL building into `endpoints.ts`
- Remove broken dataset from test app (regression from https://github.com/visgl/deck.gl/pull/8296)

### DIFF
--- a/modules/carto/src/api/endpoints.ts
+++ b/modules/carto/src/api/endpoints.ts
@@ -2,13 +2,17 @@ import {MapType} from './types';
 
 export type V3Endpoint = 'maps' | 'stats' | 'sql';
 
-function buildUrlFromBase(apiBaseUrl: string, endpoint: V3Endpoint) {
-  let suffix = `/v3/${endpoint}`;
-  if (apiBaseUrl.endsWith('/')) {
-    suffix = suffix.substring(1);
-  }
+function joinPath(...args: string[]): string {
+  return args.map(part => (part.endsWith('/') ? part.slice(0, -1) : part)).join('/');
+}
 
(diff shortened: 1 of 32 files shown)


modules/carto/src/api/endpoints.ts
```diff
@@ -11,28 +11,58 @@ function buildUrlFromBase(apiBaseUrl: string, endpoint: V3Endpoint) {
...
-export function buildMapsUrlFromBase(apiBaseUrl: string): string {
-  return buildUrlFromBase(apiBaseUrl, 'maps');
+export function buildPublicMapUrl({
+  apiBaseUrl,
+  cartoMapId
+}: {
+  apiBaseUrl: string;
+  cartoMapId: string;
+}): string {
+  return `${buildUrlFromBase(apiBaseUrl, 'maps')}/public/${cartoMapId}`;
```
donmccurdy, line 21: Optional: Maybe the name of this function, or structure of this query building, could be clarified a little? In my mind a "URL" usually has any of {protocol, domain, path, query, fragment}, and so it isn't necessarily safe to append new paths at the end of a URL. Or in this case, I couldn't tell it was safe until reading the function implementation.

To me any of these would be easier to read...

```js
// (a)
return joinPath(apiBaseUrl, 'v3', 'maps', 'public', cartoMapId);

// (b)
return joinPath(apiBaseUrl, `/v3/maps/public/${cartoMapId}`);
```

... where `joinPath` takes care of missing or redundant `/` separators.

donmccurdy (approved):

PR #8328, opened 2023-11-30
CARTO: added filter parameter to fetchLayerData

padawannn:
#### Background
We have added a filter parameter to be able to filter sources using a widget language.
#### Change List
- Added filter parameter to fetchLayerData

### DIFF
--- a/modules/carto/src/api/maps-api-common.ts
+++ b/modules/carto/src/api/maps-api-common.ts
@@ -94,3 +94,19 @@ export type NamedQueryParameter = Record<string, QueryParameterValue>;
 export type PositionalQueryParameter = QueryParameterValue[];
 
 export type QueryParameters = NamedQueryParameter | PositionalQueryParameter;
+
+export interface Filters {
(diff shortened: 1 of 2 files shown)


modules/carto/src/api/maps-api-common.ts
```diff
@@ -94,3 +94,19 @@ export type NamedQueryParameter = Record<string, QueryParameterValue>;
...
+export enum FilterTypes {
+  In = 'in',
+  Between = 'between', // [a, b] both are included
+  ClosedOpen = 'closed_open', // [a, b) a is included, b is not
+  Time = 'time',
+  StringSearch = 'stringSearch'
+}
+
+export interface FilterValues {
+  values: string[] | number[] | number[][];
```
felixpalmer, line 111: I assume not all filter types can take all filter values, i.e. `stringSearch` isn't going to take `number[][]`. If this could be expressed in the types, that would be great

felixpalmer (approved):

PR #8335, opened 2023-12-01
Check if tile layers are loaded within tile.

chrisgervang:
#### Change List
- Use `Tileset2D#isLoaded` and `super.isLoaded` on `TileLayer`
- Move cached `Layer#isLoaded` check into `Tile2DHeader#isLoaded`

### DIFF
--- a/modules/geo-layers/src/tileset-2d/tile-2d-header.ts
+++ b/modules/geo-layers/src/tileset-2d/tile-2d-header.ts
@@ -78,7 +78,9 @@ export class Tile2DHeader<DataT = any> {
   }
 
   get isLoaded(): boolean {
-    return this._isLoaded && !this._needsReload;
+    return (
+      this._isLoaded && !this._needsReload && Boolean(this.layers?.every(layer => layer.isLoaded))
+    );
   }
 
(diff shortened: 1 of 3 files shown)


modules/geo-layers/src/tileset-2d/tile-2d-header.ts
```diff
@@ -78,7 +78,9 @@ export class Tile2DHeader<DataT = any> {
   }
 
   get isLoaded(): boolean {
-    return this._isLoaded && !this._needsReload;
+    return (
+      this._isLoaded && !this._needsReload && Boolean(this.layers?.every(layer => layer.isLoaded))
```
chrisgervang, line 82: This assumes a tile isn't loaded until it has rendered layers.

If a tile can be considered loaded without rendering any layers, the logic could look something like

```
if (!this._isLoaded) return false;
if (this._needsReload) return false;
if (tile.layers) return tile.layers.every(layer => layer.isLoaded);
return true;
```

felixpalmer:
@chrisgervang open to discussion, but for the reasons above I don't think we should make this change

PR #8345, opened 2023-12-07
Remove use of deprecated BufferWithAccessor

donmccurdy:
Applies changes required for removing BufferWithAccessor from luma.gl v9.

- https://github.com/visgl/luma.gl/pull/1864.

We'll have to do something about sync `getData`, but I'm leaving that for another PR.

### DIFF
--- a/modules/core/src/lib/attribute/attribute-transition-utils.ts
+++ b/modules/core/src/lib/attribute/attribute-transition-utils.ts
@@ -4,6 +4,7 @@ import {padArray} from '../../utils/array-utils';
 import {NumericArray, TypedArray} from '../../types/types';
 import Attribute from './attribute';
 import type {BufferAccessor} from './data-column';
+import { BufferWithAccessor } from '@luma.gl/webgl';
 
 export interface TransitionSettings {
   type: string;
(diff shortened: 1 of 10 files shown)


modules/core/src/lib/attribute/attribute-transition-utils.ts
```diff
@@ -167,21 +166,26 @@ export function padBuffer({
     ? (i, chunk) => getData(toData, chunk)
     : (i, chunk) => getData(toData.subarray(i + byteOffset, i + byteOffset + size), chunk);
 
-  const bufferWithAccessor = buffer as BufferWithAccessor;
-  const source = bufferWithAccessor.getData({length: fromLength});
-  const data = new Float32Array(toLength);
+  // TODO(donmccurdy): Might need a helper function here.
+  const sourceData = buffer.getData();
```
felixpalmer, line 170: What has happened to `fromLength`?

donmccurdy: Short answer: Oversight, and I've restored it for now, thanks!

Long answer: I think this is being used in combination with dynamic resizing of buffers, which is something we're going to want to avoid in v9. I'm also not sure whether we're keeping a synchronous method for reading buffers, since that doesn't appear to be supported in WebGPU, but switching to an async method here is a larger change than I wanted to try in this PR.

felixpalmer (approved):

donmccurdy:
Merging. I will also look into removing reliance on the `getData()` function (which is part of the Buffer interface now, but only implemented by BufferWithAccessor). But some of the key uses of that function are related to transform feedback, which is not currently in a state we can test here on `master`, so I will probably hold off on those changes until I can test transform feedback in the DeckGL `master` branch.

Issue #8302, opened 2023-11-21
[Bug] Google Maps Overlay not working with vector maps


alasarr:
### Description

When using a vector basemap + GoogleMapsOverlay the basemap only renders the labels:

<img width="1270" alt="image" src="https://github.com/visgl/deck.gl/assets/1161870/e91e5fe3-5e84-462b-a6c1-88e9e34be489">

### Flavors

- [ ] Script tag
- [ ] React
- [ ] Python/Jupyter notebook
- [ ] MapboxOverlay
- [X] GoogleMapsOverlay
- [ ] CartoLayer
- [ ] ArcGIS

### Expected Behavior

Google Maps basemap should render properly with all the layers

### Steps to Reproduce

The following stackblitz example demonstrates the issue: 
https://stackblitz.com/github/CartoDB/deck.gl-examples/tree/master/google-basemap?file=index.ts

With 8.9 it works:
https://stackblitz.com/github/CartoDB/deck.gl-examples/tree/alasarr/google-basemap-8.9/google-basemap?file=index.ts

### Environment

- Framework version: deck.gl@9.0.0
- Browser: Chrome 119
- OS: MacOS Monterey

PR #8351, opened 2023-12-12
[v9] Fix GoogleMapsOverlay by not clearing canvas

zbigg:
For #8302

#### Change List
- [x] Fix LayerPass to obey `clearCanvas: false` and not clear framebuffer
- [x] configure devicePixelRatio in device.canvasContext

### DIFF
--- a/modules/core/src/lib/deck.ts
+++ b/modules/core/src/lib/deck.ts
@@ -1024,6 +1024,7 @@ export default class Deck {
 
     setGLParameters(device, this.props.parameters);
 
+    this.device?.canvasContext?.resize({useDevicePixels: this.props.useDevicePixels});
     this.props.onBeforeRender({device, gl});
 
     const opts = {
--- a/modules/core/src/lib/deck.ts
+++ b/modules/core/src/lib/deck.ts
@@ -1024,7 +1024,6 @@ export default class Deck {
 
     setGLParameters(device, this.props.parameters);
 
-    this.device?.canvasContext?.resize({useDevicePixels: this.props.useDevicePixels});
     this.props.onBeforeRender({device, gl});
 
     const opts = {
--- a/modules/core/src/passes/layers-pass.ts
+++ b/modules/core/src/passes/layers-pass.ts
@@ -66,7 +66,7 @@ export default class LayersPass extends Pass {
         viewport: [0, 0, width, height]
       },
       // clear depth and color buffers, restoring transparency
-      clearColor: options.clearColor || (options.clearCanvas ? [0, 0, 0, 0] : undefined),
+      clearColor: options.clearColor ?? (options.clearCanvas ? [0, 0, 0, 0] : false),
       clearDepth: options.clearCanvas ? 1 : undefined
     });
 
@@ -85,8 +85,7 @@ export default class LayersPass extends Pass {
       viewports,
       views,
       onViewportActive,
-      clearStack = true,
-      clearCanvas = true
+      clearStack = true
     } = options;
     options.pass = options.pass || 'unknown';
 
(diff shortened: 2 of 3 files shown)


modules/core/src/passes/layers-pass.ts
```diff
@@ -85,8 +85,7 @@ export default class LayersPass extends Pass {
       viewports,
       views,
       onViewportActive,
-      clearStack = true,
-      clearCanvas = true
+      clearStack = true
```
zbigg, line 88: clearCanvas was unused in this fn

modules/core/src/lib/deck.ts
```diff
@@ -1024,6 +1024,7 @@ export default class Deck {
 
     setGLParameters(device, this.props.parameters);
 
+    this.device?.canvasContext?.resize({useDevicePixels: this.props.useDevicePixels});
```
zbigg, line 1027: Not sure if this is correct place, but we must do it in one place or another, when we're rendering to canvas we're not "managing" by ourselves.

modules/google-maps/src/google-maps-overlay.ts
```diff
@@ -247,14 +247,6 @@ export default class GoogleMapsOverlay {
...
       deck.needsRedraw({clearRedrawFlags: true});
 
-      // Workaround for bug in Google maps where viewport state is wrong
-      // TODO remove once fixed
-      setGLParameters(gl, {
-        viewport: [0, 0, gl.canvas.width, gl.canvas.height],
-        scissor: [0, 0, gl.canvas.width, gl.canvas.height],
-        stencilFunc: [gl.ALWAYS, 0, 255, gl.ALWAYS, 0, 255]
-      });
-
```
zbigg, line 257: these are managed by beginRenderPass now i recon ... and i think they are not needed now

felixpalmer: Resizing the window results in the overlay being the wrong aspect. Could you check if this is due to removing this code?
Screenshot 2023-12-13 at 13 32 58

modules/core/src/lib/deck.ts
zbigg: After some research, i think the core problem is that luma.gl `CanvasContext.cssToDeviceRatio()` returns invalid value if no-one called `resize({useDevicePixels :true})` or `setDevicePixelRatio`  and it's not called by animation loop for external contexts.

I think best fix would be to fix `cssToDeviceRatio` so it works even if user didn't call `resize`.

felixpalmer: The `useDevicePixels` is being passed down to luma in `_createAnimationLoop`, which should be enough. I also note that there we have a TODO regarding `autoResizeDrawingBuffer` which may be related to my other comment:

https://github.com/visgl/deck.gl/blob/master/modules/core/src/lib/deck.ts#L803-L804

felixpalmer (changes requested):

modules/core/src/lib/deck.ts
ibgreen: The CSS pixel API on the CanvasContext is a bit confusing, and I think it can be a source of subtle errors. 

To create that API, I basically just took a bunch of global CSS pixel helper functions (that had been hacked on over the years) in luma.gl v8 and made them into members of the `CanvasContext`. 

While they fit well there, I feel these functions as written and documented right now are not easy to understand and there are some pitfalls with how drawing buffers and canvases relate in WebGL that require special handling so one has to be careful when refactoring them.

The existing test cases were also too difficult to port as they relied on hacky mocking of WebGL context so we lost that safety net. Such subtle functions without test cases is a bad situation.

For the call in question, I wonder if `useDevicePixels` wouldn't be better as a prop on the CanvasContext rather than a parameter to `resize()`. It kind of looks confusing to not send in width and height but to have to remember to send in useDevicePixels.

felixpalmer (approved):

PR #8379, opened 2024-01-03
chore(luma): Upgrade to luma.gl v9 alpha-50

donmccurdy:
Upgrades to luma.gl v9 alpha.50, and fixes build errors.

PR #8392, opened 2024-01-09
fix(core): Update AttributeTransitionUtils for Luma v9

donmccurdy:
In #8379 we updated to luma.gl v9 alpha-50, and had to disable some tests to keep the updates managable and avoid blocking other PRs. With this PR I'm revisiting attribute transitions, aiming to get those working again, or at least ready for a beta release.

This requires a deckgl-internal helper, `getBufferData`, to provide synchronous reads from WebGL buffers. That isn't future proof, in a WebGPU world, but to reduce moving parts I think we need the workaround for now.

With changes in this PR, attribute transition manager tests pass again. The app `test/apps/attribute-transitions` runs and displays output on screen (it previously failed). The transitions are not visually correct (see below), with the reasons for that still unknown. As this is a strict improvement over transitions throwing errors, I think it can be merged without solving the remaining issues yet.

https://github.com/visgl/deck.gl/assets/1848368/45b2643e-f2d4-44c0-a323-e97c873f3135


donmccurdy:
Current dependencies on/for this PR:
* `master`
  * **PR #8392** <a href="https://app.graphite.dev/github/pr/visgl/deck.gl/8392?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/visgl/deck.gl/8392?utm_source=stack-comment).

modules/core/src/lib/attribute/attribute-transition-manager.ts
```diff
@@ -148,11 +148,9 @@ export default class AttributeTransitionManager {
     // TODO: when switching transition types, make sure to carry over the attribute's
     // previous buffers, currentLength, startIndices, etc, to be used as the starting point
     // for the next transition
-    let needsUpdate = !transition || transition.type !== settings.type;
-    needsUpdate ||=
-      attribute.buffer.byteLength > transition?.attributeInTransition.buffer.byteLength;
```
felixpalmer, line 153: Are you removing this because the buffer length cannot change?

modules/core/src/lib/attribute/attribute-transition-utils.ts
```diff
@@ -184,7 +188,44 @@ export function padBuffer({
...
+    buffer = buffer.device.createBuffer({byteLength: target.byteLength + byteOffset});
   }
   buffer.write(target, byteOffset);
+  return buffer;
+}
+
+/** @deprecated TODO(v9.1): Buffer reads should be asynchronous and avoid accessing GL context. */
+export function getBufferData(
+  buffer: Buffer,
+  Ctor: TypedArrayConstructor,
```
felixpalmer, line 200:
```suggestion
  TypedArray: TypedArrayConstructor,
```

modules/core/src/lib/attribute/attribute-transition-utils.ts
```diff
@@ -184,7 +188,44 @@ export function padBuffer({
...
+
+  // Use GL.COPY_READ_BUFFER to avoid disturbing other targets and locking type
+  _buffer.gl.bindBuffer(GL.COPY_READ_BUFFER, _buffer.handle);
+  _buffer.gl2.getBufferSubData(GL.COPY_READ_BUFFER, byteOffset, dstArray, dstOffset, dstLength);
+  _buffer.gl.bindBuffer(GL.COPY_READ_BUFFER, null);
+
+  return dstArray;
+}
+
+export function getAttributeVertexFormat(size: 1 | 2 | 3 | 4): LumaVertexFormat {
```
felixpalmer, line 219: This only supports `float32`, use [getVertexFormatFromAttribute](https://github.com/visgl/luma.gl/blob/master/modules/core/src/adapter/type-utils/vertex-format-from-attribute.ts#L58) instead?

modules/core/src/transitions/gpu-interpolation-transition.ts
```diff
@@ -87,18 +89,29 @@ export default class GPUInterpolationTransition implements GPUTransition {
...
-    // };
+    const padBufferOpts = {
+      numInstances,
+      attribute,
+      fromLength: this.currentLength,
+      fromStartIndices: this.currentStartIndices,
+      getData: transitionSettings.enter
+    };
+
+    buffers.forEach((buffer, index) => {
```
felixpalmer, line 100: Generally prefer for loops to avoid creating functions

felixpalmer (changes requested):

modules/core/src/lib/attribute/attribute-transition-utils.ts
donmccurdy: It looks like attribute transitions support `float`, `vec2`, `vec3`, and `vec4` types, and to call `getVertexFormatFromAttribute` we'd need to allocate an empty Float32Array... the type of `attribute.value` is actually Float64Array:

https://github.com/visgl/deck.gl/blob/c56c50363ab5e970fc8d4539685d6093aeafb374/modules/core/src/transitions/gpu-interpolation-transition.ts#L120-L122

I've cleaned this up a bit and added JSDoc on the function, but I think it needs to roughly do what it is doing now – let me know if there are other concerns!

modules/core/src/lib/attribute/attribute-transition-manager.ts
donmccurdy: I added this line in https://github.com/visgl/deck.gl/pull/8392 thinking that we'd need to re-create the Transition if attribute length changes, but that turned out to be a mistake. When `transition.start()` runs it calls `padBuffer`. Previously that would resize the buffer, now — since buffers have fixed size — it returns a replacement buffer. By keeping the Transition we avoid overwriting older elements in the buffer, which (at least according to the unit tests) is required.

felixpalmer (approved):

PR #8405, opened 2024-01-12
[v9] Align picking module PickingProps with luma

felixpalmer:
Followup to https://github.com/visgl/deck.gl/pull/8334. Requires https://github.com/visgl/luma.gl/pull/1915

#### Change List
- Use same property names as luma


modules/aggregation-layers/src/contour-layer/contour-layer.ts
```diff
@@ -351,7 +351,8 @@ export default class ContourLayer<
     const {count} = this.state.weights;
     let {aggregationData} = count;
     if (!aggregationData) {
-      aggregationData = count.aggregationBuffer!.getData() as Float32Array;
+      // @ts-ignore
+      aggregationData = count.aggregationBuffer!.readSyncWebGL2() as Float32Array;
```
donmccurdy, line 355: Flagging in case this bites someone (possibly me) later — return type is Uint8Array, not Float32Array. See https://github.com/visgl/deck.gl/pull/8425.

donmccurdy (approved):

PR #8412, opened 2024-01-16
carto(v8): fetchLayerData to support columns prop for query sources

zbigg:
#### Change List
- carto(v8): fetchLayerData to support columns prop for query sources


felixpalmer (approved):
Change looks fine, just needs test against API as discussed

